PR #56, opened 2023-09-13
Check for valid metadata

Eric-Arellano:
In the closed source repository that consumes Qiskit/docs, we have this same check for valid file metadata. It is useful to run it here too so that we fail more eagerly.

Like the internal code, this uses TypeScript. It makes the code more readable and better aligned with the internal code.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,5 +14,7 @@ jobs:
       - name: Install Node.js dependencies
         run: npm ci
 
+      - name: File metadata
+        run: npm run check:metadata
       - name: Spellcheck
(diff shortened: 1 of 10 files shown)


.github/workflows/main.yml
```diff
@@ -14,5 +14,7 @@ jobs:
       - name: Install Node.js dependencies
         run: npm ci
 
+      - name: File metadata
+        run: npm run check:metadata
```
Eric-Arellano, line 18: CI has dedicated steps for each check, rather than simply running `npm run check`, because it makes the output more structured.

axelhzf (approved):

Issue #9, opened 2023-09-06
Migrate Sphinx -> MDX pipeline for open source


Eric-Arellano:
This is currently in closed source, to pull in the API docs for qiskit, qiskit-ibm-runtime, and qiskit-ibm-provider. We're proposing open sourcing it.

At least in a follow up, we should consider running the pipeline on artifacts from CI, rather than the actual published websites. That will allow us to consume specific commits, such as pulling in typo fixes more quickly.

> on artifacts from CI

Local builds for Qiskit take too long to run locally.

PR #61, opened 2023-09-28
Add Jest and Sphinx parsing library

Eric-Arellano:
## Background

First part of https://github.com/Qiskit/documentation/issues/9. The docs infrastructure expects MDX files, but `qiskit`, `qiskit-ibm-provider`, and `qiskit-ibm-runtime` use Sphinx to generate their API references from source code. So, @axelhzf created a pipeline to convert Sphinx output to MDX.

We are moving that pipeline to live in open source since the API references are also open source.

## What this PR does

This adds the `lib/sphinx` code and its tests. The code is copied over without change, other than adding copyright headers. The PR also sets up Jest.

A follow up PR will add the actual parsing script. I split up the PRs since this adds non-trivial infrastructure like Jest, so to have a smaller PR to review.

This code will be deleted from the internal repository once https://github.com/Qiskit/documentation/issues/9 is complete and the internal repository is consuming this one.

## Why Jest?

We use Playwright in https://github.com/Qiskit/qiskit_sphinx_theme, so I considered using that too here for consistency.

But I stuck with the internal repo's decision to use Jest because Playwright doesn't have an equivalent to `toMatchInlineSnapshot`. I think Playwright is awesome, but we're never going to have Playwright integration tests in this repository since we're not testing the docs site itself in this repo, i.e. we don't need to spin up a server. Jest is simpler for our unit tests.


tsconfig.json
```diff
@@ -1,15 +1,14 @@
 {
     "compilerOptions": {
-      "target": "es5",
+      "target": "es2021",
```
Eric-Arellano, line 3: Without this, I had issues with TypeScript saying `string.replaceAll()` doesn't work.

tsconfig.json
```diff
@@ -1,15 +1,14 @@
...
+      "module": "ESNext",
       "allowJs": false,
       "strict": true,
       "noEmit": true,
       "esModuleInterop": true,
       "moduleResolution": "node",
-      "resolveJsonModule": true,
-      "isolatedModules": true,
       "incremental": true,
-      "downlevelIteration": true
```
Eric-Arellano, line 12: These were removed recently from the closed source config I had originally copied this file from.

tsconfig.json
```diff
@@ -1,15 +1,14 @@
...
+      "skipLibCheck": true
```
Eric-Arellano, line 11: I couldn't get this error working:

```
node_modules/micromark-extension-mdxjs/index.d.ts:7:14 - error TS2307: Cannot find module './node_modules/micromark-extension-mdx-expression/dev/lib/syntax.js' or its corresponding type declarations.

7     | import('./node_modules/micromark-extension-mdx-expression/dev/lib/syntax.js').Options
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in node_modules/micromark-extension-mdxjs/index.d.ts:7
```

Using a declaration file didn't work. Tomorrow, I'm going to try to get a minimal reproduction and file an issue upstream.

scripts/lib/stringUtils.ts
```diff
@@ -0,0 +1,38 @@
+// This code is a Qiskit project.
```
Eric-Arellano, line 1: While this file is a copy from the internal repo, I removed functions not used by this API pipeline.

Eric-Arellano (commented):
The code in `scripts/lib` can be skipped over since it's a copy-and-paste, other than adding copyright headers.

axelhzf (approved):
🚀 

When the sync to the internal repository is working, we can probably delete this code from the internal copy and keep it only here.

Issue #6, opened 2023-09-06
Consider auto-formatting MDX files


Eric-Arellano:
I think it would be worth adding Prettier for our MD/MDX files :) But I'll confirm with the content team that they're okay with it first.

Eric-Arellano:
The team has been fine without having this. I suspect it would increase friction too much that it's annoying to need to run, even if we automated it.

Still, I personally would be a +1 for this. I only don't feel strongly enough to argue for it. If someone else wants to take this on, I'd vote yes.

PR #76, opened 2023-10-04
Autoformat TypeScript with Prettier

Eric-Arellano:
Relates to https://github.com/Qiskit/documentation/issues/6, which is about docs. This PR only adds Prettier to our support code. We might want to extend Prettier to docs in a follow up.

Having a consistent format is convenient and saves us time so that humans don't have to manually move things around.

We have a non-trivial amount of TypeScript tooling, so this is worth adding imo.


Eric-Arellano:
First commit needs close review. Second commit is generated.

frankharkins (approved):
Sorry I missed this earlier.

Looks good, thanks!

PR #80, opened 2023-10-05
Migrate support page from closed source

Eric-Arellano:
This was oversight. I didn't realize we needed to migrate this page too.

The page is simply copy-and-pasted.


abbycross (approved):
👍

PR #132, opened 2023-10-09
Set up a staging site

Eric-Arellano:
The content team requested this as a useful tool to show how the docs are rendering on `main`, before the content gets synced to close source and pushed to production.

This uses the same IBM Cloud Engine code we use for PR previews. It doesn't use GitHub Pages because we need to run the server with Docker, whereas GH Pages expects a simple static website.

The staging link should be stable at https://qiskit-docs-preview-staging.1799mxdls7qz.us-south.codeengine.appdomain.cloud.


.github/workflows/staging.yml
```diff
@@ -0,0 +1,36 @@
...
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# This Action builds the docs on `main` so that we can have a staging build of the open
+# source site. It uses IBM Cloud to build the Dockerfile at the root of the repository.
+#
+# Note that there is no cleanup job, unlike preview-docs.yml. The workflow will replace
+# the old deploy with a new one.
```
Eric-Arellano, line 17: I confirmed this makes sense with Va.

javabster:
can we update the navbar tag to say "staging" instead of "local"?

javabster (approved):
LGTM 🚀

PR #136, opened 2023-10-10
Edits to issue templates

abbycross:
From @javabster -- improvements to issue templates:

### content bug or inaccuracy

- [x] can we add a checkbox selection like we had in the old one? something like this:

```
- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] it's just wrong
- [ ] other
```

- [x] description formatting is a bit weird with the slash (/). Can you update to say "Report typos, bugs, out-of-date content, broken links etc."

- [x] can you have it automatically add the "bug" label when the issue is opened?

### UI issue

- [x] typo in the "please describe the UI problem" description, there is an extra ) at the end

- [x] also can you have it automatically add a label, might need to make a new one called "UI feedback" or something

### Request new content
- [x] could you rephrase the last question in the form to emphasise that it is optional? i imagine some people might just want to add a suggestion but dont want to be further involved at all, they just want to submit a request and leave. That currently doesn't have an option in the dropdown and the q is mandatory. You could rephrase the question to "if this new content request is accepted, do you want to write the content?" and then have all the dropdown options including a "no" option
- [x]  can you automatically add the "new" label and also a "content request" label (i think the second will need to be created
### API refs
add the links to the 1 line description of the issue
- [x] https://docs.quantum-computing.ibm.com/api/qiskit
- [x] https://docs.quantum-computing.ibm.com/api/qiskit-ibm-runtime/runtime_service

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-content.yaml
+++ b/.github/ISSUE_TEMPLATE/new-content.yaml
@@ -35,7 +35,7 @@ body:
         - I will write a draft of the proposed content
         - I can help the team by providing enough information to write the material
         - I only want to review the material when it's finished
-        - No
+        - "no"
       default: 0
     validations:
       required: true
--- a/.github/ISSUE_TEMPLATE/new-content.yaml
+++ b/.github/ISSUE_TEMPLATE/new-content.yaml
@@ -1,6 +1,6 @@
 name: ✨ Request new content
 description: Make a request to the Content Team to add a new section in the docs.
-labels: ["NEW CONTENT REQUEST ✨"]
+labels: ["content request", "new"]
 assignees:
   - abbycross
   - beckykd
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/new-content.yaml
```diff
@@ -1,6 +1,6 @@
 name: ✨ Request new content
 description: Make a request to the Content Team to add a new section in the docs.
-labels: ["NEW CONTENT REQUEST ✨"]
+labels: ["content request", "new"]
```
javabster, line 3:
```suggestion
labels: ["content request", "new ✨"]
```

javabster (approved):
LGTM 🚀 

we should check that the automatic labelling action is working properly once this merges

.github/ISSUE_TEMPLATE/new-content.yaml
abbycross: Aha, you've already done so

abbycross:
Confirmed that the automatic labels are working now 🥳

PR #159, opened 2023-10-12
Fix bad internal links to optimize section

Eric-Arellano:
We missed these on accident, in the absence of the link checker.


abbycross (approved):
Many thanks

PR #173, opened 2023-10-16
Add script to check links

### LINKED ISSUE #4: Set up link checker
We should make sure that internal links are valid for the state of the docs in the PR, i.e. HEAD. That is, if we reorganize HTML pages, that's fine as long as we update the right links.

External links should be valid no matter what.
```[tasklist]
### Tasks
```

frankharkins:
This PR adds a link-checking script to verify internal links work. It should be easy to extend this to check external links, and anchors within internal links. All comments welcome.

### Details

Uses `markdown-link-extractor`, which is used by `markdown-link-check` (one of the proposed tools).

***

First part of #4


scripts/commands/checkLinks.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    for (let filePath of this.resolve()) {
+      if (existsSync(filePath)) {
+        return true
+      }
+    }
+
+    console.log(`❌ ${this.origin}: Could not find link '${this.value}'`)
+    return false
```
Eric-Arellano, line 94: Instead of logging directly in this function, it might be a cleaner user interface to collect all the failing links in the `main` function or `checkLinksInFile`, and then bundle them together as a single console error, where you list all the failures.

This is a lot of repeated warnings right now, which can be noisy.

See an example: https://github.com/Qiskit/documentation/blob/a1a72fe18a8fed759b733965ef0fc6440e303ebc/scripts/commands/checkMetadata.ts#L55-L65

You'd have this function return something like structured data or a string representing what the failure is. Return `undefined` if it's okay. Return type is `string?`.

scripts/commands/checkLinks.ts
```diff
@@ -0,0 +1,132 @@
...
+
+function checkLinksInFile(filePath: string): boolean {
+  const source = readFileSync(filePath, {encoding: 'utf8'})
+  const markdown = (path.extname(filePath) === '.ipynb') ? markdownFromNotebook(source) : source
+  const links = markdownLinkExtractor(source).links.map((x: string) => new Link(x, filePath))
+
+  let allGood = true
+  for (let link of links) {
+    allGood = link.check() && allGood
+  }
```
Eric-Arellano, line 116: `return links.all(link => link.check())`

If you take the above suggestion to have `check()` return the error or `undefineed`, rather than directly logging it, then this would be `return links.map(link => link.check()).filter(Boolean)`, where `.filter(Boolean)` gets rid of the undefined.

scripts/commands/checkLinks.ts
```diff
@@ -0,0 +1,132 @@
...
+// This code is licensed under the Apache License, Version 2.0. You may
+// obtain a copy of this license in the LICENSE file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+import { globby } from "globby";
+const { existsSync, readFileSync } = require('fs');
```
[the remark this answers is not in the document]
Eric-Arellano: > Allows us to run multiple futures at once using Promise.all(). It will allow us to check every link in concurrency, which is going to be a lot faster.

That's not going to be relevant if you take my below recommendation to pre-read all the valid files, since they will be synchronous operations. That's fine though - it will be really fast.

It's still good to use async here for consistent style.

Eric-Arellano (commented):
Thank you! This looks great 🙌 

We'll also want this hooked up to `package.json` as `check:links`, and to add it to GitHub Actions so we error on invalid links. Include instructions in the README too.

scripts/commands/checkLinks.ts
frankharkins: With `links.every`, the script will exit on the first broken link it finds, whereas the current loop prints _all_ broken links. Is exiting early ok?

Eric-Arellano: Ah, okay, never mind. We definitely do want to collect every single link. The `.map().filter()` suggestion would do that properly though, if you end up returning the error rather than logging directly.

scripts/commands/checkLinks.ts
```diff
@@ -0,0 +1,140 @@
...
+  for (let cell of JSON.parse(source).cells) {
+    if (cell.source === 'markdown') {
+      markdown += cell.source
+    }
+  }
+  return markdown
+}
+
+function checkLinksInFile(filePath: string, filePaths: string[]): boolean {
+  const source = readFileSync(filePath, {encoding: 'utf8'})
```
Eric-Arellano, line 117: We'll still want this to be an async read, which will require marking `async function checkLinksInFile`. This line will become `const source = await readFile(...)`

Then, lines 130-133 will become:

```typescript
const results = await Promise.all(filePaths.map(fp => checkLinksInFile(fp, filePaths))
if (results.some(x => !x)) {
}
```

That will parallelize things.

scripts/commands/checkLinks.ts
```diff
@@ -0,0 +1,140 @@
...
+
+import { globby } from "globby";
+import { existsSync, readFileSync } from 'fs';
+import path from 'node:path';
+import markdownLinkExtractor from 'markdown-link-extractor';
+
+const DOCS_ROOT = "./docs"
+const CONTENT_FILE_EXTENSIONS = [".md", ".mdx", ".ipynb"]
+
+class Link {
```
Eric-Arellano, line 21: It would be super valuable to have some unit tests for `Link.resolve()` and `Link.check()`. They don't have any side effects anymore and they're now pure functions, so it should be easy to test. 

To do so, you'd add a file `scripts/commands/linkChecker.ts` or something like that. Copy this code over. Then import it from your commands file. And add a test file similar to `dedupeIds.test.ts`.

`npm test` to ensure the test works.

scripts/commands/checkLinks.ts
```diff
@@ -0,0 +1,140 @@
...
+      return true
+    }
+
+    const possiblePaths = this.resolve()
+    for (let filePath of possiblePaths) {
+      if (filePathCache.includes(filePath)) {
+        return true
+      }
+    }
+    // Check disk for files not in cache (images etc.)
```
Eric-Arellano, line 94: Ah, I was thinking the images should be in the cache too. The only files that we expect you to use locally are images from `public/images` and other `docs/` files. Otherwise it's an error.

Eric-Arellano (commented):
I like how this is progressing! The `Link` class is quite nice.

Eric-Arellano (approved):
We're going to merge this now to help out the content team. Frank will make a follow up PR to make some of the suggested refactors

PR #176, opened 2023-10-16
[TO DISCUSS] Hide the backend.run vs Qiskit Runtime comparison

Eric-Arellano:
From @javabster 

> We should remove /compare
> It's not useful enough to warrant it's own page, and I'm still trying to figure out if theres another page we could integrate it into
> for now, can you give it the __ treatment? so the file is still there but it's not deployed?


jyu00:
I agree it doesn't belong in `Migration guides`, but why do you want to hide it before finding a new home? Since the information is still useful and will be lost this way.

Eric-Arellano:
See https://github.com/Qiskit/documentation/pull/177 for my proposal for now before this Wednesday deadline, since it sounds like we need to move Migration Guides to the API dropdown by then. Then we can find a better home for this file after the release.

Eric-Arellano:
We decided to instead consolidate `compare` into the Runtime migration guides. @beckykd is working on that in a separate PR.

Issue #186, opened 2023-10-17
URL audit


abbycross:
URLs should be reflective of the content while, as best as possible, taking into consideration future planned content.

Also, use hyphens rather than underscores.

Rename the following files:

(Nothing in Start)

Build:

- [x] build/feature-table to build/qasm-feature-table

Verify:

- [x] verify/noise to verify/noisy-simulators
- [x] verify/simulators to verify/cloud-based-simulators (FYI the title of this file is changing in a [WIP PR](https://github.com/Qiskit/documentation/pull/184))

Run:

- [x] run/cost to run/manage-cost
- [x] run/error-mitigation to run/configure-error-mitigation
- [x] run/error-suppression to run/configure-error-suppression
- [x] run/estimate-job to run/estimate-job-run-time
- [x] run/hardware to run/system-information
- [x] run/monitor to run/monitor-a-job
- [x] run/queue to run/fair-share-queue
- [ ] run/run-sessions to run/run-jobs-in-a-session

beckykd:
Ah - for the future, can we avoid changing file names?  I link to these topics from Cloud

abbycross:
That reminds me that we also should check for broken links on ibm.com, and we don't have access to their files. It will require some thorough clicking around on the quantum pages.

Eric-Arellano:
> for the future, can we avoid changing file names? I link to these topics from Cloud

Ah, sorry we broke things. Fwiw, the idea with this one-time rename was to fix the URLs to be stable before yesterday's release. The new names are meant to be more future-proof so that we don't run into issues in the future.

PR #189, opened 2023-10-17
Improve URLs to be more specific

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/186.

We want to rename these URLs before an upcoming release and before the new docs have more traffic. The URL names are meant to be more feature proof.

I found what links needed updating by searching with rigprep with the pattern `/file-name`.


abbycross:
@Eric-Arellano were you searching within all docs, so that lab and admin are also included in the searching?

Eric-Arellano:
Good catch, @abbycross! Thank you! I appreciate your attention to detail 🙇

PR #192, opened 2023-10-17
consistency with backend vs system (and fix some links)

abbycross:
In the Get backend information page, replace most instances of system with backend (except where it's in the code).


javabster (commented):
you can change the code snippets as well to use e.g. `backend.properties().backend_name` instead of `system.properties.backend()`

javabster:
are you adding an intro sentence to this as well, or adding it in a follow up PR?

abbycross:
> you can change the code snippets as well to use e.g. `backend.properties().backend_name` instead of `system.properties.backend()`

ah so Qiskit already accepts either? How smart Qiskit is! 😆

javabster:
> ah so Qiskit already accepts either? How smart Qiskit is! 😆

Its more that the user gets to call it whatever it wants haha

take this code example, that you're probably familiar with:

```
from qiskit_ibm_runtime import QiskitRuntimeService, Estimator 
backend = service.get_backend("ibmq_qasm_simulator")
```
I can then call `backend.properties()` like this doc page says and get the config info I need

But I could also do this:
```
from qiskit_ibm_runtime import QiskitRuntimeService, Estimator 
penguin = service.get_backend("ibmq_qasm_simulator")
```
and then call `penguin.properties()` and get the same info, I chose to assign the backend object to the word `penguin`. I could also choose to assign it to the word `backend` or `system` or whatever I want

abbycross:
> are you adding an intro sentence to this as well, or adding it in a follow up PR?

I slacked you about this too, but I don't trust myself to dash off a reasonable intro before our imminent release and will open up a new issue to do so afterward! But if it's literally just a sentence, "we use backend and system interchangeably", then I can do that. (We would also want a similar note on the system-information page)

javabster:
> I slacked you about this too, but I don't trust myself to dash off a reasonable intro before our imminent release and will open up a new issue to do so afterward! But if it's literally just a sentence, "we use backend and system interchangeably", then I can do that. (We would also want a similar note on the system-information page)

ok let's just do it in a followup PR, can you open an issue to track it?

abbycross:
> > ah so Qiskit already accepts either? How smart Qiskit is! 😆
> 
> Its more that the user gets to call it whatever it wants haha
> 
> take this code example, that you're probably familiar with:
> 
> ```
> from qiskit_ibm_runtime import QiskitRuntimeService, Estimator 
> service = QiskitRuntimeService() 
> backend = service.get_backend("ibmq_qasm_simulator")
> ```
> 
> I can then call `backend.properties()` like this doc page says and get the config info I need
> 
> But I could also do this:
> 
> ```
> from qiskit_ibm_runtime import QiskitRuntimeService, Estimator 
> service = QiskitRuntimeService() 
> penguin = service.get_backend("ibmq_qasm_simulator")
> ```
> 
> and then call `penguin.properties()` and get the same info, I chose to assign the backend object to the word `penguin`. I could also choose to assign it to the word `backend` or `system` or whatever I want

whoa. Of course! we're defining those terms in each program. 😮

abbycross:
@javabster I moved the View backend info section under Properties -- does it look right? 

I also added a blurb at the top of both `system-information` and `get-backend-information` about systems/backends being used interchangeably.... let me know if that will suffice for tomorrow :)

mtreinish:
Tbh, this guide is a bit dated and is relying on legacy access patterns from `BackendV1`, and that interface will be phased out in the core data representations provided in qiskit before too long. I realize fundamentally this is just updating some of the language for consistency and not specifically about the underlying information. But I thought it was important to point out that the information here is specific to the IBM provider (specifically it's custom fields on top of the`BackendV2` interface it inherets from). The more modern access patterns for this information are a lot easier to deal with because it's a flatter structure, without duplicated or contradictory fields. There's an existing migration guide outlining some of the differences here: https://qiskit.org/documentation/apidoc/providers.html#migrating-between-backend-api-versions

javabster:
@mtreinish thanks for the update, is the content here just completely wrong now or do some users still need this?

What would you recommend would be most helpful to users:
- update this page to include both backend v1 and v2 info
- completely replace the backend v1 info with v2 info?

javabster:
Ok as we're on a tight deadline today lets go ahead an merge this PR as is so we can get the language consistent, I've opened a followup issue here to update this page to use backendV2 info: https://github.com/Qiskit/documentation/issues/195

javabster (approved):
LGTM 🚀

Issue #63, opened 2023-09-28
Ideas to simplify API pipeline


Eric-Arellano:
Follow up to https://github.com/Qiskit/documentation/issues/9. The initial focus was simply copying it over as is.

Some possible ideas:

* Get rid of the GitHub API client code and have another way of determining the versioning, even the user specifying it.
* Don't pull HTML docs from published websites. Instead do something like getting it from CI at specified commits. That will make the script deterministic, remove a synchronization problem of requiring the docs to already be published, and facilitate us having multiple versions of the API docs.
* Have the user specify which of the 3 packages to regenerate.

frankharkins:
> Don't pull HTML docs from published websites. Instead do something like getting it from CI at specified commits. That will make the script deterministic, remove a synchronization problem of requiring the docs to already be published, and facilitate us having multiple versions of the API docs.

I imagine you've already considered this, but maybe we can get Sphinx to output something nicer than HTML? A quick search brings up [`sphinx-markdown-builder`](https://github.com/liran-funaro/sphinx-markdown-builder)), and Sphinx supports a few different [builders](https://www.sphinx-doc.org/en/master/usage/builders/index.html) that may (or may not) be easier to work with.

Eric-Arellano:
> A quick search brings up [sphinx-markdown-builder](https://github.com/liran-funaro/sphinx-markdown-builder)), and Sphinx supports a few different [builders](https://www.sphinx-doc.org/en/master/usage/builders/index.html) that may (or may not) be easier to work with.

Yeah, unfortunately they all aren't very robust :( the markdown builder fails to parse Qiskit, and the builtin Sphinx builders have issues too. 

The medium term plan is that we will probably want to migrate to https://vemel.github.io/handsdown/ rather than this custom Sphinx -> MDX parser.

Eric-Arellano:
Closing in favor of more granular tracking:

* you can already choose which project to run
* we'll no longer get the version from GitHub as part of https://github.com/Qiskit/documentation/issues/77
* https://github.com/Qiskit/documentation/issues/311 tracks building from CI rather than the live site.

PR #196, opened 2023-10-18
Allow generating API docs for specific projects

Eric-Arellano:
One of the ideas from https://github.com/Qiskit/documentation/issues/63. This is useful to me right now because I want to only redo qiskit-ibm-provider.

`yargs` is a popular command line library in JavaScript. I've used it in a personal project and had a great experience with it.

There will be future config we need to add, like probably having users tell us what CI URL to look at when generating the docs, whereas right now we pull from the live websites (see #63). I think it's likely that config will be best set via a config file, rather than command line arguments. But, even if we have that config file in the future, I still think we'd want `--packages` to be a CLI arg since that is a quick thing that you want to be flexible with setting, vs updating a file. If we switch 100% to a file, that's fine; we can always remove `yargs`.


scripts/commands/updateApiDocs.ts
```diff
@@ -154,18 +180,18 @@ zxMain(async () => {
     await downloadHtml({ baseUrl: pkg.baseUrl, initialUrls: pkg.initialUrls, destination });
   }
 
-  // Sphinx html to markdown
-  console.log('Deleting existing markdowns');
-  await $`rm -rf ${getRoot()}/docs/api/`;
```
Eric-Arellano, line 159: This was deleting our new section `/api/migration-guides`, which is not auto-generated. We need to be more granular.

frankharkins (approved):
- Logic is easy to follow
- Tested locally and it works nicely

Nice work; LGTM! 🚀

Issue #33, opened 2023-09-07
[ NEW PAGE ] Build/ Control flow with Qiskit


javabster:
## What the new page should cover:

```[tasklist]
### Tasks
- [ ] if statement
- [ ] for loop
- [ ] switch statement
- [ ] a warning box: dynamic=true,check out run section
```

## SME collaborators (involve for planning and review):
@jakelishman

kevinsung:
I'm working to enable the for loop, switch statement, and while loop; see https://github.ibm.com/IBM-Q-Software/vt-qcf/issues/1419. As described in that issue, part of the task is to update the dynamic circuits documentation to demonstrate these new capabilities. Since this issue involves the same documentation, I think it would make sense to track that task in this issue to avoid duplicating effort. What do you all think?

Issue #148, opened 2023-10-11
[ NEW PAGE ] Running Dynamic Circuits on Hardware


javabster:
Note: this page should be written with the intention of it being removed as soon as possible once it is no longer needed. By including this information as a separate page (rather than integrating the info amongst existing run pages) we can more easily remove it later.

## What the new page should cover:

- [ ] how to run with dynamic circuits (backend.run, dynamic=true etc etc.)
- [ ] understanding the limitations / considerations for running dynamic circuits on different hardware types

## SME collaborators (involve for planning and review):
@taalexander @jakelishman

Issue #100, opened 2023-10-08
Create new sidebar structure for Build


javabster:
The build section should have the following sidebar structure:

```
├── Introduction
├── Build Circuits with Qiskit
│   ├── Circuit Library
│   └── Circuit Construction
│   ├── Circuit Visualisation
│   ├── Classical feedforward and control flow
│   └── Order of Qubits in Qiskit
│
├── Build Operators with Qiskit
│   ├── Specifying Observables in the Pauli basis (to be written)
│   ├── Operators module Overview
│   └── Synthesizing circuits from operators (to be written)
│
├── Other Circuit Building tools
    ├── Tune Pulse Waveforms
    └── OpenQASM
        ├── What is OpenQASM
        └──  OpenQASM2 and Qiskit
        ├── OpenQASM3 and Qiskit
        ├── OpenQASM3 Feature table
        └── OpenQASM spec
```

Before completing this work, the following PRs should be merged (otherwise theres not enough content to make the folder structure not look weird):
- https://github.com/Qiskit/documentation/pull/156
- https://github.com/Qiskit/documentation/pull/217

also t[his page](https://docs.quantum-computing.ibm.com/build/operators_overview) should be re-titled "Operators module Overview"

abbycross:
Question @javabster -- for [this page](https://docs.quantum-computing.ibm.com/build), when we change the title in the toc to Introduction, we should also change the title on the page itself so that they match. Do you prefer just "Introduction" as the title, or "Introduction: Build circuits with Qiskit" or something like that? I'm fine either way.

PR #198, opened 2023-10-18
add how-to-guide on classical feedforward and control flow

kevinsung:
Fixes https://github.com/Qiskit/documentation/issues/33.

That issue mentioned that we would add a warning box on limitations and link to a page on limitations in the run section, which will be created as part of https://github.com/Qiskit/documentation/issues/148. In the end I decided to keep this guide self-contained and omit the warning. If we still want to add the warning box, we can do it as part of the latter issue.


kevinsung:
The CI complains
```
      Invalid Jupyter notebook metadata. Every .ipynb file needs to 
      set 'title' and 'description' in the file metadata.
```
Can we make this more lenient? I created the notebook with VS Code and it opens in Jupyter Lab. Otherwise, what's an easy way to ensure notebooks pass the linter?

Eric-Arellano:
> Can we make this more lenient?

No, we must have the metadata set for the docs infra to work. I need to improve the instructions though, https://github.com/Qiskit/documentation/issues/200.

To fix, open the notebook in text mode in VSCode and scroll to the very bottom of the file to its metadata section - make sure it's metadata for the notebook and not an individual cell. Then add `title` and `description`. The `title` is the HTML page title, and description is a ~1 sentence summary.

javabster (changes requested):
Overall looking great thanks @kevinsung!

I left some comments below, also could you please update the toc so this file will appear in the sidebar, @abbycross or @beckykd or @Eric-Arellano can help you if you don't know how to do this

abbycross:
> Overall looking great thanks @kevinsung!
> 
> I left some comments below, also could you please update the toc so this file will appear in the sidebar, @abbycross or @beckykd or @Eric-Arellano can help you if you don't know how to do this

I will add this to the toc according to our [Build nav bar plan](#100), and just FYI it is going to be first in the Build section since the pages above it haven't been published yet.

docs/build/dynamic-circuits.ipynb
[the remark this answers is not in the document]
kevinsung: - I've dropped the aka
- I put the switch statement after the if statement because it is similar to it. If and switch are branching constructs; for and while are looping constructs.
- "Context manager" is a technical term that refers to a specific Python concept, see https://docs.python.org/3/reference/datamodel.html#with-statement-context-managers. I added a link to this documentation page.
- I added explanations to all the examples.
- I added explanations for the black and white dots in the drawing.
- I added explanation in the prose for the switch case that uses two conditions.

kevinsung:
@javabster I've addressed your comments and updated the toc.

docs/build/_toc.json
```diff
@@ -8,6 +8,10 @@
     {
       "title": "Build circuits with Qiskit",
       "children": [
+        {
+          "title": "Classical feedforward and control flow",
+          "url": "/build/dynamic-circuits"
```
Eric-Arellano, line 13: Can this URL be made more specific? Right now it's highly generic. It's better to use a future-proof URL like `classical-feedforward-and-control-flow` or `dynamic-circuits-feedforward`.

You'll update here and the file name.

kevinsung: done

javabster (commented):
> The return value of the context manager can be stored and used subsequently as a context manager to create an `else` block, which is executed whenever the contents of the `if` block are *not* executed.

Is there a way to rewrite this line without saying context manager twice? maybe something like "the measurement outcome of the if statement can be stored and used subsequently as a context manager to create an `else` block..."

javabster (commented):
>  Note that for clarify, we chose

do you mean "for clarity"?

javabster:
other than the small language things above this LGTM 🚀

javabster (approved):

PR #207, opened 2023-10-19
copy-edit Order of qubits in Qiskit

abbycross:
Add Order of qubits in Qiskit to Build section

Related to #100


docs/build/qubit-order.mdx
```diff
@@ -5,35 +5,38 @@ description: How endianness works in Qiskit
...
+While most physics textbooks represent an $n$-qubit system as the tensor product $Q_0\otimes Q_1 \otimes ... \otimes Q_{n-1}$, where $Q_j$ is the $j^{\mathrm{th}}$ qubit, Qiskit uses the inverse order, that is, $Q_{n-1}\otimes ... \otimes Q_1 \otimes Q_{0}$. 
 
-This means that if we have, for example, a 3-qubit system with qubit 0 in state $|1\rangle$ and qubits 1 and 2 in state $|0\rangle$, Qiskit would represent this state as $|001\rangle$ while most physics textbooks would represent this state as $|100\rangle$.
+As explained in [this video](https://www.youtube.com/watch?v=EiqHj3_Avps) from [Qiskit's YouTube channel](https://www.youtube.com/@qiskit), this is done to follow the convention in classical computing, in which the $n^{\mathrm{th}}$ bit or most significant bit (MSB) is placed on the left (with index 0), while the least significant bit (LSB) is placed on the right (index $n-1$). This ordering convention is called *little-endian*, while the one from the physics textbooks is called *big-endian*.
```
frankharkins, line 10: > As explained in \[this video\](https://www.youtube.com/watch?v=EiqHj3_Avps) 

Sadly that video is actually not quite right. I'll read the rest of the page tomorrow but I know for sure the video is wrong for reasons [mentioned in a previous PR on the topic](https://github.com/Qiskit/qiskit/pull/10205#discussion_r1221646250).

frankharkins (changes requested):

docs/build/qubit-order.mdx
```diff
@@ -5,35 +5,38 @@ description: How endianness works in Qiskit
...
 
 The matrix representation of any multi-qubit gate is also affected by this different qubit ordering. For example, if we consider the single-qubit gate
 
 $$
-U = \begin{pmatrix} u_{00} & u_{01} \\ u_{10} & u_{11} \end{pmatrix}
+U = \begin{pmatrix} u_{00} & u_{01} \\ u_{10} & u_{11} \end{pmatrix},
 $$
 
-And we want a controlled version $C_U$ whose control qubit is qubit 0 and whose target is qubit 1, following Qiskit's ordering its matrix representation would be
+and we want a controlled version $C_U$ whose control qubit is qubit 0 and whose target is qubit 1, following Qiskit's ordering its matrix representation would be
```
beckykd, line 20:
```suggestion
and we want a controlled version $C_U$ whose control qubit is qubit 0 and whose target is qubit 1, following Qiskit's ordering, its matrix representation would be
```

beckykd (approved):

frankharkins (commented):
Ok I think there's some different things going on which is where the confusion comes from. Here's my understanding.

***

If a circuits has $n$ qubits, we usually label the qubits $0 \rightarrow n$ (so far, no confusion; this is the same everywhere).

The things at play are:

  - **What the index of the qubit _means_.**

    If your qubits are not representing numbers (for example, if each qubit represents the state of a node on a graph), then this is determined by the user.

    If your qubits do represent a number, then Qiskit's convention is that qubit 0 is the _least_-significant (qu)bit. This is "little-endian" as the least-significant bit is at the smallest address (see [Wikipedia](https://en.wikipedia.org/wiki/Endianness)).

  - **How we display the qubits (and their measurement results).**

    Convention is to display a number with the _most_-significant bit on the right. So Qiskit displays qubits in this order:

    $|q_n \cdots q_1 q_0 \rangle$

    That is, measurement results are displayed in a big-endian format. This is where Qiskit differs from "physics textbooks", as they often aren't considering their bits as numbers, so write their statevectors and measurement results in the other order.

    This also applies to the data structures. For example, when accessing individual bits from the result in Python, you'll find the indexes of the bitstrings are reversed from the indexes of the qubits in the circuit.

    ```python
    bitstring[0]  # measurement result of qubit n
    ```

    To keep consistency with this convention, Qiskit always expects qubit 0 on the _right_ in unlabelled strings of bits/qubits. For example:

    ```python
    Statevector.from_label('01')  # creates Statevector with q0 = |1> and q1 = |0>
    ```

To summarize: Qiskit always interprets qubit 0 as the _as the last element of a string_. This can occasionally be awkward as qubit indices in circuits are not the same as their indices in measurement results, or when displayed on screen.

docs/build/qubit-order.mdx
frankharkins: Sorry – I had to sit down properly and get my head round it again. The video isn't as bad as I remember; I don't think I say anything technically incorrect. I'm ok with leaving it in; it doesn't really explain _why,_ but it does give useful examples of how the circuit corresponds to the measurement result.

jakelishman: My main comment here is that I'd personally remove the discussion about the word "endian" and the terms "most significant" and "least significant" (and MSB and LSB).  The term "significance" only really makes sense when you're drawing the analogy to classical computing that Qiskit does with its ordering conventions, because _everybody_ writes integers with the "most significant" bit on the left in English.  Us using "endianness" is needlessly confusing, because the typical use of the word in classical computing actually refers to something completely unrelated.

jakelishman (commented):
Frank mentioned above about the "why" of all this stuff.  Let me try and explain, but it's abstract minutiae of a cross-section of quantum physics, numerical computing and classical hardware architecture, so it's a bit gross.  This is long, but I'm putting it here more for a reference.

---

There's unfortunately several different mathematical operations at play here, and Qiskit's documentation (and tbh most physicists) don't distinguish them well at all.  Along with that, we match up those mathematical operations with classical computer-science terminology by making some analogies, so there's a lot of confusion kicking around.


### Endianness

First up, "endianness".  In Qiskit docs, this PR, Frank's comment, and tbh just about every interaction I've had on the topic, we're all a bit lax about using the word.  There's lots of related concepts, and for clarity right now, I only refer to quantum parts explicitly - everything else is classical computing:

- **byte order in memory**: this is _actually_ what "endianness" means usually (classically), and note that it's about bytes not bits[^1].  This has no relation to Qiskit or IBM Quantum's view of quantum computing hardware / software.
- **bit ordering**: if bits can be addressed individually, such as if they're being sent bit-by-bit over a network, then you have to define an order.  Some people confusingly use the term "bit endianness" to refer to this, where "big endian" means that the most-significant bit comes first.  Qubits _can_ be addressed individually, which is why people use this "endianness" terminology, but beware: we haven't yet defined what "most significant" means, so this doesn't mean anything yet.
- **signficance**: something is "more significant" if it's somehow "worth more".  When we write numbers, like 3276, the "3" is the most significant because it's worth three thousands, while the "6" is only worth six ones so it's in the least significant position.  "Endianness" is about whether the most significant or least significant subcomponent is stored or received first.  When representing numbers in English, we _always_ write the most significant digit on the left.  That applies to classical bitstrings of integers as well, because the bitstring is just a binary representation of a number.  Unlike digits in the number 3276, there's no fixed concept of "place value" for qubits put together, so we still have to define "significance" for qubits - we haven't even defined how you "put together" several qubits yet.
- **bit/digit labelling**: let's say we've been given a number "3276".  *Labelling* is the concept of giving numbers to the digits.  You might say we'll label them like this:
  ```
  Number:      3          2          7          6
               ^          ^          ^          ^
               |          |          |          |
  Label:    digit 1    digit 2    digit 3    digit 4
  ```
  which feels natural for English speakers.  You can think of the "physics textbooks" we refer to as doing this sort of thing with qubits.

  However, that's not actually very natural mathematically or computationally.  Instead, we can label the digits like this:
  ```
  Number:      3          2          7          6
               ^          ^          ^          ^
               |          |          |          |
  Label:    digit 3    digit 2    digit 1    digit 0
  ```
  This has the mathematical advantage that the "place value" of "digit $n$" is $10^n$.  So our number is made up of $(3\times10^3) + (2\times10^2) + (7\times10^1) + (6\times10^0) = 3276$, where you can see that "digit 1" matches with $10^1$ and so on.

  Classical computing always uses the second form for labelling bits.  Qiskit uses the second form for labelling classical bits, so it seems logical to want to arrange things such that the natural ways of using Qiskit cause "qubit 0" (whatever that means) to have some relation to "classical bit 0".

For _classical_ bits, when people say "Qiskit's bit endianness", what is meant is actually along the lines of "Qiskit writes binary numbers in the same way as everyone else, and labels the right-most bit as '0'".

Ok, that's all the classical-computing terminology.  In all this, I was using the example of regular numbers for familiarity, but classical bitstrings work the same way (they're just numbers in base-2, not base-10).  There's no concept of a "byte" in regular numbers, just like there isn't in our quantum-computing memory models, so we're ok ignoring that.

The glaring omission is that we understand what it means if I write "3267" - the individual digits "3", "2", "6" and "7" work together to make a number - but there's no inalienable meaning to trying to just stuff individual qubits together like we did with digits.


### Multiple-qubit systems

Imo, a lot of the confusion here comes from physicists often treating two distinct things as if they're completely interchangeable (we're not renowned for being rigorous).  These are "abstract linear algebra" (the _actual_ mathematics of quantum physics) and "one particular set of conventions for representing quantum objects as matrices" (how we do most of the abstract maths on classical computers).  The mathematical language of quantum physics _does not_ require matrices, and much of it actually doesn't actually work with matrices[^2], but the quantum-computing bits do.

When we use the "quantum"-looking symbols like $\lvert0\rangle$ or $X$, we're using the *abstract* form of things.  These *do not* have a unique matrix representation - we have to choose one.  Whenever matrices are involved, we're using one particular *concrete* matrix form, that we chose.  The matrix form can even change within the code, while still representing the same quantum things (but don't worry about this).

#### Quantum states

For the quantum state $\lvert0\rangle$, it's completely valid to choose _any_ 2-row column vector to describe it.  It's very strong convention for quantum computing software to do this, though:
```math
\text{The matrix representation of $\lvert0\rangle$ is chosen to be }\begin{pmatrix}1\\0\end{pmatrix},\text{ which is the same as $[1,\ 0]^T$.}
```
Qiskit principally uses this matrix representation, as do textbooks[^3].

When we want to consider two or more qubits together, though, there's no universal convention for which order you write them in either the abstract notation or as concrete matrices.  The abstract notation matters less in a software package like Qiskit because it's not something a user can interact with, unlike a matrix.  The only time it comes up is when we try to use it as a short-hand to explain the *matrix* representation we chose.

Qiskit chose the matrix representation $\lvert0\rangle := [1,\ 0]^T$ because it has the convenient property that if `m` is a programming variable containing $\text{\texttt{matrix}}(\lvert \psi\rangle)$ (where $\text{\texttt{matrix}}(x)$ gets the matrix representation of $x$ in our conventions): `m[0] == 1` if $\lvert\psi\rangle = \lvert0\rangle$, whereas `m[1] == 1` if $\lvert\psi\rangle = \lvert1\rangle$.  In other words the `1` in the matrix is in the index that corresponds to the "numeric value" of $\psi$.

This is where classical bit labelling comes in, and where Qiskit diverges from physics textbooks.  _Let's pretend_ that an individual qubit can only be in state 0 or state 1 (this is not at all true, but what we did with $\psi$ above).  If so, we pretend that we can write several qubits together as if they're binary digits, and treat the result as a number.  By analogy to writing "regular" numbers, and using the "maths-friendly" bit-labelling conventions (zero on the right) for classical bits we described above, Qiskit chooses to write several qubits $q_0$, $q_1$, $q_2$ together in one abstract sense as $\lvert q_2 q_1 q_0\rangle$ so that if we replaced the $q_0$ labels, etc with digits 0 or 1, we'd be able to read the whole thing as a binary number.  This is _heavily_ by analogy to classical computing, and this is Qiskit's convention.  Because of this analogy, Qiskit calls "qubit 0" the "least significant bit", because it's treating mashing all the qubits together as if you can read it like a number.

Textbooks just say "I've got $q_a$ and $q_b$, so I'll write them together as $\lvert q_a q_b\rangle$", which is the "English-reading-order-friendly" way of labelling things.  There is no concept of significance here, because we're not reading the qubits like digits in a number.  Textbooks sometimes switch to the Qiskit conventions when they _do_ want to read things like numbers.

When we write $\lvert q_a q_b\rangle$ for two qubits, this is a shorthand for the (still abstract maths) form $\lvert q_a\rangle \otimes \lvert q_b\rangle$.  In this, $\otimes$ is called the "tensor product" operator, and to stress again: it alone does not say anything about matrices.  To do that, you have to find a way of representing the abstract tensor product with matrices.  This is almost universally done using a _matrix_ operation called the Kronecker product.  In Numpy, you do that using `np.kron`, so I'll call that operation $\text{\texttt{kron}}(a,\ b)$.  You're technically still free to choose the left-to-right order of arguments for the representations, but both textbooks and Qiskit define<span id="kron-order"></span>
```math
\text{\texttt{matrix}}(\lvert a\rangle \otimes \lvert b\rangle) := \text{\texttt{kron}}\bigl(\text{\texttt{matrix}}(\lvert a\rangle),\ \text{\texttt{matrix}}(\lvert b\rangle)\bigr),
```
it's just that we assign integer labels to $a$ and $b$ backwards to each other.  Because of this, people (unfortunately) often use the symbol $\otimes$ as if it's the exact same thing as $\text{\texttt{kron}}$, but it's not, and when we're trying to compare bit-labelling / ordering conventions in terms of it, it's very confusing.

With Qiskit's qubit-labelling/ordering convention, and with the above definition for translating the tensor product into matrices, we end up in the state that if we "read" a multi-qubit label like an integer, the "1" in $\text{\texttt{m}} = \text{\texttt{matrix}}\bigl(\lvert q_3 q_2 q_1 q_0\rangle\bigr)$ is in the array index that corresponds to the numeric value of $q_3 q_2 q_1 q_0$ if we replaced all the $q$ with digits, like we did with the single-digit case.

#### Controlled gates

As an example, $CX$ is a 2-qubit mathematical operation that applies $X$ to a "target" qubit if the "control" qubit is in the $\lvert0\rangle$ state.  The order you juxtapose the "target" and "control" qubits is _another_ convention, and one that's not necessarily related to the convention for the order of writing multiple qubits in a state.  Qiskit's documentation completely confuses these two things, and worse, it describes things in terms of "most" and "least" significant qubits, which we said doesn't make universal sense for textbooks.

Qiskit's two conventions are:
- The CX function and gate take their arguments like `QuantumCircuit.cx(control, target)`.
- In all Qiskit functions and gates, we take the arguments to a function as `f(qubit_0, qubit_1, qubit_2, ...)`.

Note that the "function labels" look different to the way we choose to write the states in abstract notation, which is $\lvert q_2\rangle\otimes\lvert q_1\rangle \otimes \lvert q_0\rangle$, and our matrix representation of that, using [this equation](#kron-order).  This is because _lists_ appear in programming with "item 0" on the left, whereas _integers_ appear in programming with "bit 0" on the right.  We're being consistent with programmatic access, to the objects, but people (including our documentation) frequently mistake our functions by thinking that they should _look_ like the abstract mathematical notation.

Let's look at [the Qiskit documentation for `CXGate`](https://qiskit.org/documentation/stubs/qiskit.circuit.library.CXGate.html):

Screenshot 2023-10-24 at 16 32 15

What our documentation (in the first equation) is _trying_ to say here is along the lines of:

> When adding a $CX$ gate to a `QuantumCircuit`, put the "control" qubit on the left of the call and the "target" on the right of the function call or qubit list.  Because, in function calls and lists, Qiskit calls the left-hand argument "0" and the right-hand argument "1", we identify the "control" with qubit 0 and the "target" with qubit 1.  In abstract mathematical notation using Qiskit's ordering conventions for qubit labelling we write the joint state of qubit 0 and qubit 1 as $\lvert q_1\rangle\otimes \lvert q_0\rangle$, and therefore the $CX$ operation is written as $I\otimes\lvert0\rangle\langle0\rvert + X \otimes \lvert 1\rangle\langle1\rvert$.  Qiskit then defines the matrix form using [this convention for resolving the tensor product into matrices](#kron-order), so the resulting matrix looks like `<...>`.

Now the note is trying to say something along the lines of:

> Textbooks typically use a convention where the control is their "qubit 1" and the target is their "qubit 2". They use a qubit-labelling convention that the joint state of qubits 1 and 2 is written as $\lvert q_1\rangle \otimes \lvert q_2\rangle$, and [use the same left-to-right convention for resolving the tensor product into matrices](#kron-order), so their resulting matrix looks like `<...>`.  The exact same matrix in Qiskit represents a $CX$ operation with the control and target qubits switched.

The parts about "more significant" and "less significant", I would argue, are at best not useful, and at worst seriously confusing.  The concept of "significance" of qubits _only_ makes sense in the analogy that Qiskit draws by putting "qubit 0" on the right.  It does not make sense in the textbook convention; they're presented with the assumption of "the control is on the left as written", and there's no such thing as "qubit significance".


#### Strings of Paulis

When we want to write the operator like "$X$ on qubit 0, $Y$ on qubit 1, $Z$ on qubit 2", Qiskit writes the _string_ as "ZYX", by analogy to a _bitstring_ and to how Qiskit writes its qubit states in abstract notation.

---

That was a complete essay, sorry.

That level of detail clearly isn't appropriate for the documentation, but if there's anything useful in it, let me know.


[^1]: More specifically, it's about the ordering of the smallest individually addressable units of memory.  On modern classical hardware, the smallest unit of addressable memory is a byte (not a bit), but we want to store numbers that are (say) 4 bytes long.  When we write it to memory, we can either write the most significant byte at the smallest memory address (big endian) or the least significant byte at the smallest memory address (little endian).  It doesn't make _complete_ sense to talk about "which bit comes first in a byte?" because they're not individually addressable, but in practice, the image most people would have in their head is that the "most significant bit" comes first/left-most, regardless of whether we're doing big- or little-endian byte ordering.  This makes some people feel like big endian is "more natural" and that little endian is "inside out".
[^2]: For physicists: I'm meaning infinite-dimensioned Hilbert spaces, for example.
[^3]: Until we start considering changes of basis, but that's not important for this discussion.

jakelishman:
In _very_ brief, Qiskit's conventions are chosen such that:

- `Statevector.from_int(n, (2,)*num_qubits).data[n] == 1` for all integers `n`.
- In `QuantumCircuit.append(op, qubits)`, `qubits[0]` is referred to as $\lvert q_0\rangle$, and all specific-gate methods are called like `QuantumCircuit.cx(*qubits)` $\leftrightarrow$ `QuantumCircuit.append(CXGate(), qubits)`.
- In controlled gates, the target qubit is the rightmost one in the function call.

Give or take, that leads to everything else.

docs/build/qubit-order.mdx
abbycross, line 10:
```suggestion
```

abbycross: I've removed the discussion about endian and MSB/LSB -- lmk if more is needed! Thanks!

abbycross:
BTW Jay just opened [this PR](https://github.com/Qiskit/documentation/pull/248/files) on a related note -- does this affect this file @frankharkins ?

jakelishman:
That note is duplicating some of what's in this file, and it's a better use of terminology than the existing documentation.  I don't have an issue with any of the text in that note, though it alone is just about labelling conventions, and doesn't say anything about matrix forms or controlled-gate conventions, which this file should continue to address.

beckykd (approved):

abbycross:
@frankharkins @jakelishman are we still waiting for more edits to this file, or can we merge?

jakelishman:
Abby: I'm not sure what Frank's plan is between this PR and #273, so I'll leave it to him.  I think the file this PR is touching still contains information on controlled gates that #273 doesn't (currently) address, and it'd be good to have both of those bits, but the parts that #273 does have are technically more sound.

Issue #135, opened 2023-10-09
Figure out why there are API files with `.rst.md` ending


Eric-Arellano:
```
docs/api/qiskit/qiskit.pulse.library.Constant_class.rst.md
docs/api/qiskit/qiskit.pulse.library.Cos_class.rst.md
docs/api/qiskit/qiskit.pulse.library.Drag_class.rst.md
docs/api/qiskit/qiskit.pulse.library.Gaussian_class.rst.md
docs/api/qiskit/qiskit.pulse.library.Sawtooth_class.rst.md
docs/api/qiskit/qiskit.pulse.library.Sech_fun.rst.md
docs/api/qiskit/qiskit.pulse.library.Sin_class.rst.md
docs/api/qiskit/qiskit.pulse.library.Square_fun.rst.md
docs/api/qiskit/qiskit.pulse.library.Triangle_class.rst.md
```

PR #210, opened 2023-10-19
Regenerate Qiskit documentation

Eric-Arellano:
Even though the API version is the same, this brings in a bunch of spelling improvements from @abbycross.

It also seems to fix https://github.com/Qiskit/documentation/issues/135!


javabster (approved):
LGTM 🚀

Issue #213, opened 2023-10-19
Migrate Qiskit migration guides to MDX


Eric-Arellano:
We now have the files copied over from open source, but we need to finish converting to MDX. The tool we used to convert from RST to MDX could only handle certain things.

## Step 1: finish fixing cross-references

Sphinx uses cross-references like this:

```
:meth:`qiskit.my_module`
:meth:`~qiskit.my_module`
{meth}`qiskit.my_module`
{meth}`~qiskit.my_module`
```

We need to convert those to Markdown links. We have this script to automate the rewriting, but you need to manually set up the `Conversion` list because there have been enough edge cases of problems like the replacement not existing, or there being typos, or the replacement living in another repository. To run this script, you'll:

1. Copy the file `migrate-references.py` to the root of your repository
2. Add to the list `CONVERSIONS`. Tip that running `black` on the file makes it more readable
3. `python3 migrate-references.py`

```python
from __future__ import annotations

from dataclasses import dataclass
from pathlib import Path


GUIDES = [
    Path(f"docs/api/migration-guides/__migration-guide-{guide}.md.txt")
    for guide in ("algorithms", "opflow", "qi")
]


@dataclass(frozen=True)
class Conversion:
    qualified_name: str
    new_reference: str | None
    fixed_qualified_name: str | None = None

    def sphinx_values(self) -> set[str]:
        prefixes = []
        for role in ["class", "meth", "mod", "func"]:
            for tilde in ["", "~"]:
                for punctuation in [(":", ":"), ("{", "}")]:
                    initial, end = punctuation
                    # E.g.:
                    #   * :meth:`
                    #   * :meth:`~
                    #   * {meth}`
                    #   * {meth}`~
                    prefixes.append(f"{initial}{role}{end}`{tilde}")
        return {f"{prefix}{self.qualified_name}`" for prefix in prefixes}

    def new_value(self) -> str:
        name = self.fixed_qualified_name or self.qualified_name
        if self.new_reference is None:
            return f"`{name}`"
        return f"[`{name}`]({self.new_reference})"

    def validate_new_reference(self) -> None:
        """Check that the reference exists."""
        if self.new_reference is None or not self.new_reference.startswith("../"):
            return
        file_name = self.new_reference.split("#")[0]
        normalized_path = Path("docs/api/migration-guides").joinpath(f"{file_name}.md")
        if not normalized_path.exists():
            raise ValueError(
[...]

Issue #77, opened 2023-10-05
Generate historical API versions for Qiskit


Eric-Arellano:
Blocked by https://github.com/Qiskit/documentation/issues/8.

How far back do we go? Update: as far as we can. But at least 2022 and 2023 releases if possible.

PR #218, opened 2023-10-20
Add qiskit.opflow and qiskit.algorithms API docs

Eric-Arellano:
@javabster and I decided that it's important we have these API docs in OneDocs, even though the modules are deprecated. We want them for historical versioning https://github.com/Qiskit/documentation/issues/77.

We also need them for the migration guides: https://github.com/Qiskit/documentation/issues/213.


scripts/commands/updateApiDocs.ts
```diff
@@ -122,36 +122,12 @@ const PACKAGES: Pkg[] = [
     githubSlug: "qiskit/qiskit",
     baseUrl: `https://qiskit.org/documentation`,
     initialUrls: [`https://qiskit.org/documentation/apidoc/index.html`],
-    ignore(id: string): boolean {
```
Eric-Arellano, line 125: Manual changes in this file.

javabster (approved):
LGTM 🚀

PR #225, opened 2023-10-23
Fix link checker for notebooks

frankharkins:
Turns out the script was not correctly extracting markdown from notebooks. We didn't notice because `markdown-link-extractor` seems to be able to find the links from the JSON source 99% of the time.

Despite this, `markdown-link-extractor` doesn't offically support notebooks, so I've fixed the code. This found one more broken link, which I've ignored:

```
❌ docs/build/pulse.ipynb: Could not find link 'pulse_modulation.png'
```


Eric-Arellano (approved):
Thank you!

PR #229, opened 2023-10-23
Optimization levels 2 & 3 behave like 1

### LINKED ISSUE #215: Update optimization to clip at level 1 at most

https://docs.quantum-computing.ibm.com/run/advanced-runtime-options

### Select all that apply

- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [x] other

### Describe the fix.

Server-side (primitives) clip optimization level to 1 at most. This temporarily re-defines optimization levels 1, 2, and 3 to be identical for the Primitives. This is what we want updated in the documentation to reflect that.  Qiskit transpiler optimization level 1 will still do layout, routing, and basis gate translation so at this step we have decreased the cost of transpilation but we haven't stopped accepting abstract or virtual circuits.

beckykd:
closes #215


beckykd:
I changed the examples to use level 1, but I'm not sure that's the right thing to do.  Are the secondary level optimization settings different for the different levels?  Do the secondary level settings even work now?

docs/run/advanced-runtime-options.mdx
```diff
@@ -105,53 +105,51 @@ suppression mechanisms. Level 0 only involves basic transpilation.
...
-options.optimization_level = 2
+options.optimization_level = 1
 
 estimator = Estimator(session=session, options=options)
 ```
 
-
-You might want to configure your transpilation strategy further, and for this, there are advanced transpilation
-options you can set up. These are "second-level options", and can be set as follows:
+The `skip_transpilation` option is an advanced transpilation option, set as follows:
```
jyu00, line 126: nit - I don't know why `skip_transpilation` is moved up. But this grouping becomes a bit odd:

1. `optimization_level` and `initial_layout`/ `routing_method` are all about tuning the transpiler. Whereas `skip_transpilation` is about skipping the transpiler. So it's odd to have this in the middle.
2. The words "second-level options" now appear later, even though `skip_transpilation` is also a second level option.

javabster (commented):
I agree with Jessie's comment, otherwise I think it's good to go :)

jyu00:
> I changed the examples to use level 1, but I'm not sure that's the right thing to do. 

I think that makes sense. To make it more explicit everything is level 1 now.

> Are the secondary level optimization settings different for the different levels? Do the secondary level settings even work now?

They'll work with all levels (except maybe 0).

docs/run/advanced-runtime-options.mdx
beckykd: I moved skip transpilation up because it's part of that note.  Maybe I can just add a link to that information.

lerongil (commented):
Thanks for the update. This looks good to me!

jyu00 (approved):
LGTM!

PR #177, opened 2023-10-16
Move migration guides to API section

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/90.

Note the new URLs:

* /api/migration-guides
* /api/migration-guides/backend-run
* /api/migration-guides/backend-run-comparison
* /api/migration-guides/backend-run-examples

A follow up PR will add the Qiskit migration guides once they are ready to go.

PR #230, opened 2023-10-23
Fix broken links for migration guides

Eric-Arellano:
I accidentally broke this in https://github.com/Qiskit/documentation/pull/177.

This PR fixes the broken links and also a few others, such that our link checker now checks every non-API file 🎉 To fix some errors, it adds a `SYNTHETIC_FILES` list, which are files that only exist in closed source but not open source; so the links are valid, even though the files aren't present here.


abbycross (approved):

javabster (approved):
LGTM 🚀

PR #245, opened 2023-10-25
Make link checker asynchronous

frankharkins:
Small refactor to make the link checker asynchronous.

***

Co-authored-by: Eric Arellano <14852634+Eric-Arellano@users.noreply.github.com>


scripts/commands/checkLinks.ts
```diff
@@ -11,7 +11,8 @@
 // that they have been altered from the originals.
 
 import { globby } from "globby";
-import { existsSync, readFileSync } from "fs";
+import { existsSync } from "fs";
```
Eric-Arellano, line 14: Thanks. It'd be great to also remove this by instead pre-reading what all the images are in your `main` function so that you already have all valid files known in your file cache. Then you can remove this call to `existsSync` inside your check function.

arnaucasau (approved):
I like the use of Promise.all(), and making `checkLinksInFile` async will be useful when implementing the external links :+1: 

LGTM! Thanks!

scripts/commands/checkLinks.ts
frankharkins: Thanks! Should be fixed in 17e853d

Eric-Arellano (approved):

PR #248, opened 2023-10-26
Update hello-world.ipynb

jaygambetta:
I would rather use something like this “little endian” as strictly endianness is a description of byte-ordering and since qiskit doesn’t have the ability to address bytes in memory, endianness does not apply;


abbycross:
Just FYI -- [related work](https://github.com/Qiskit/documentation/pull/207) going on in a new file to be added.

jaygambetta:
ok but we do need to stop calling it endian as it really is not.

abbycross:
Yep we are removing that terminology. Also, you broke the notebook 😆 so I'm going to revert your changes and reapply them, so don't worry if you see reverted changes in the history

abbycross:
@jaygambetta our preview is temporarily broken, so here's a screenshot of what I did offline -- did I get all the latex math formatting correct?
Screenshot 2023-10-26 at 9 23 40 AM

jaygambetta:
ok do what you need and what your posted looked good.

abbycross (approved):

PR #252, opened 2023-10-26
misc session fixes

### LINKED ISSUE #247: Misc issues with session docs

https://docs.quantum-computing.ibm.com/run/sessions
https://docs.quantum-computing.ibm.com/run/run-jobs-in-session

### Select all that apply

- [ ] typo
- [ ] code bug
- [X] out-of-date content
- [ ] broken link
- [X] other

### Describe the fix.

https://docs.quantum-computing.ibm.com/run/sessions:

1. > Since data from the first session job is cached and used by subsequent jobs, if the first job is canceled, subsequent session jobs will all fail.

  This sentence is only half true. Data from the first session job is no longer cached (but subsequent jobs would still fail if the root job is cancelled).

2. > Jobs that fit within the maximum session time run back-to-back on hardware.

This line suggests that jobs in batch mode run serially. But in reality the classical processing part is actually done in parallel. We should also point that out as an advantage of using batch mode.

3. We should have code showing how to use session in iterative and batch modes. For example:

Batch:

The following example shows how you can divide up a long list of circuits into multiple jobs and run them as a batch to take advantage of the parallel processing. 

```python

backend = service.backend("ibm_sherbrooke")

with Session(backend=backend):
    estimator = Estimator()
    start_idx = 0
    jobs = []
    while start_idx < len(circuits):
        end_idx = start_idx + backend.max_circuits
        jobs.append(estimator.run(circuits[start_idx:end_idx], obs[start_idx:end_idx], params[start_idx:end_idx]))
        start_idx = end_idx
```

4. > The quantum device is locked to the session user unless the TTL is reached.

"Locked" can be misleading. Other jobs (such as calibration or jobs already running when a session started) can still run. It'd be more accurate to say a session is active and its jobs get priority until ITTL or max timeout is reached. 

Under https://docs.quantum-computing.ibm.com/run/run-jobs-in-session:

1. > Data from the first session job is cached and used by subsequent jobs.

Same as 1 above. 

2. > Directly specify a string with the backend name. Example:

The code example should include service, i.e. 

```python
with Session(service=service, backend="ibmq_qasm_simulator"):
    ...
```

beckykd:
closes #247

### DIFF
--- a/docs/run/sessions.mdx
+++ b/docs/run/sessions.mdx
@@ -151,7 +151,7 @@ Any session job submitted within the interactive timeout, also known as interact
     There might be a limit imposed on the ITTL value depending on whether your hub is Premium, Open, and so on. 
 </Admonition>
 
-This is an example of running an iterative workload that uses the classical Scipy optimizer to minimize a cost function. In this model, Scipy uses the output of the cost function to calculate its next input. 
+This is an example of running an iterative workload that uses the classical SciPy optimizer to minimize a cost function. In this model, SciPy uses the output of the cost function to calculate its next input. 
 
 ```python
     def cost_func(params, ansatz, hamiltonian, estimator):
--- a/docs/run/sessions.mdx
+++ b/docs/run/sessions.mdx
@@ -37,7 +37,7 @@ There are several benefits to using sessions:
 
     <Admonition type="note" title="Notes">
        -   The queuing time does not decrease for the first job submitted within a session. Therefore, a session does not provide any benefits if you only need to run a single job.
-       -   Since data from the first session job is cached and used by subsequent jobs, if the first job is canceled, subsequent session jobs will all fail.
+       -   If the first session job is canceled, subsequent session jobs will all fail.
     </Admonition>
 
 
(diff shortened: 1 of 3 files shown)


docs/run/sessions.mdx
```diff
@@ -151,21 +151,55 @@ Any session job submitted within the interactive timeout, also known as interact
...
+
 ### Batch
 
 Ideal for running experiments closely together to avoid device drifts,
 that is, to maintain device characterization.
 
 -   Suitable for batching many jobs together.
--   Jobs that fit within the maximum session time run back-to-back on
-    hardware.
+-   The parts of the jobs that are processed classically run in parallel, and the quantum pieces run sequentially on hardware, which saves you time.
```
jyu00, line 180:
```suggestion
-   The classical computation, such as compilation, of the jobs is run in parallel. This means running multiple jobs in a batch would be significantly faster than running them serially. 
```

jyu00 (approved):
One small change on wording but otherwise LGTM. Thanks @beckykd for the quick fix!

Issue #228, opened 2023-10-23
Link checker won't validate Admonitions from MDX syntax


Eric-Arellano:
From what I can tell, MDX syntax like this is not checked:

https://github.com/Qiskit/documentation/blob/00d462fd689a62e544e3c9f14eafee793c1607e3/docs/start/index.mdx#L45-L56

This would presumably be because the MDX component is not being parsed, since it's non-standard markdown. I'm not sure how we'd work around this.

--

@frankharkins had the suggestion to pre-process the Markdown files to simply remove the `<Admonition>` lines. I think that's a great idea. Make sure you pre-process in memory and don't actually rewrite the file live.

arnaucasau:
The bug should be solved in #261 by adding an apostrophe surrounding the admonition tags. Trying to solve this issue I came across another bug. There are links preceded by lines starting with space that won't be validated either. One example can be found in this file:

https://github.com/Qiskit/documentation/blob/00d462fd689a62e544e3c9f14eafee793c1607e3/docs/start/install.mdx#L249-L268

In line 249, we have a line starting with a tabulation or multiple spaces. With those spaces, the link in line 267 won't be validated.

In this case, the link is inside an admonition tag but if you delete the tag the problem doesn't disappear. One solution that I found to work, is to add a dash (-) followed by a space in front of every line. This fix can substitute the use of the apostrophe used in the last commit of #261.

The change is to replace the lines added in that PR:

```
const source_raw = readFileSync(filePath, { encoding: "utf8" });
const source = source_raw.replace(/<Admonition/g,"'<Admonition")
                         .replace(/Admonition>/g,"Admonition>'");
```

by:
```
const source_raw = readFileSync(filePath, { encoding: "utf8" });
const source = source_raw.replace(/^[a-zA-Z0-9<]/gm,"- ");
```

This seems to work fine on my end solving the two bugs, and it finds multiple links that weren't validated before. I want to test different scenarios that perhaps are not happening in the documentation we have right now, but I wanted to know your opinion @Eric-Arellano and @frankharkins before changing anything in #261 or before creating an independent issue (if you think is worth it having two independent issues).

axelhzf:
This does not look like a good solution for the long term. We will have the same problem with other mdx component. I think we should use a mdx parser instead.

frankharkins:
Ok looks like we use remark internally, I'll work with @arnaucasau to use that and come up with something more robust

axelhzf:
[...]

PR #261, opened 2023-10-27
Link checker - validate admonitions

arnaucasau:
This PR is part of the issue https://github.com/Qiskit/documentation/issues/228.

### Summary

This PR solves the issue with the Admonitions tags changing the way in which we read every link. Instead of using the `markdownLinkExtractor` directly at the files, we can use a unified plugin to find every link by parsing each file as HTML.

### Details

To solve the problem, the first approach was to pre-process the markdown removing the tabs in every line inside the admonitions tag, and then removing the tag itself. Doing that, the `markdownLinkExtractor` function is able to validate the links inside the tags. However, to avoid having problems with other mdx components in the future, this PR changes the way in which we read each link by using a [unified](https://github.com/unifiedjs/unified#description) plugin.

Using [rehype](https://github.com/rehypejs/rehype) as a plugin, we can parse every file as HTML and convert it to markdown. Because we are reading the file as HTML, the syntax tree associated is a [hast](https://github.com/syntax-tree/hast), and we have different nodes that we can take advantage of by using different visitors.

In our files, we have two different cases:

1. The text we are reading is markdown: In this case, because we are assuming that is HTML, the node in the tree that it uses to represent this is raw [text](https://github.com/syntax-tree/hast#text) (things outside the HTML tags are just text).
2. The text is actually HTML embedded: In this case, if the tag is for example a paragraph `<p>` or something like that (`<Admonition>` included), we are back in the first case again (The node used is text). If we have different tags, like `` or `<a>`, then we have to access their respective nodes, in this case, "image", and "link" which have an attribute called url with the link.

We need three different visitors to read every link, the first one for the links in regular text inside or outside the HTML tags, another one for links in the href attribute of an `<a>` tag, and the last one for links in the src attribute of an `` tag.

We have to continue using the `markdownLinkExtractor` in the first visitor because the values that we are reading from the text nodes are written in markdown (or raw test, which in this case is equivalent)

Like before, all the links found in the file are stored in an array to continue with its validation.


frankharkins (approved):
Really great work! This is _much_ more robust than the delete-tags hack, and just as simple.

Eric-Arellano (approved):
Nice!

Eric-Arellano (approved):

PR #264, opened 2023-10-31
Clarify session job execution order

beckykd:
Clarify why jobs don't always run in the order they were submitted.


beckykd:
38 messages later -- never mind :)

Issue #104, opened 2023-10-08
[ NEW PAGE ] Transpiler Stage Details / Preset Passmanagers


javabster:
## What the new page should cover:

- [ ] Which stages do or do not preserve unitarity equivalence
- [ ] what stuff you need to know from the start: e.g. initial layout before the stage
- [ ] pre/post stages

## SME collaborators (involve for planning and review):
@1ucian0 @haimeng-zhang @mtreinish 

## Existing resources to build upon
- https://qiskit.org/documentation/apidoc/transpiler.html#working-with-preset-pass-managers
- https://qiskit.org/documentation/apidoc/transpiler.html
- https://qiskit.org/documentation/apidoc/transpiler_plugins.html

beckykd:
Here's a discussion from the Defaults topic (https://github.com/Qiskit/documentation/pull/267#discussion_r1379270616) that probably fits   better with the Transpiler stage details section:

(Nick): One thing I think is super confusing is the difference between Layout and Routing (together I call these Mapping, but that might be a personal thing). At a functional level, I think of them as

Routing: insert SWAPs due to the limited connectivity of the lattice
Layout: pick the set of qubits in the lattice which to use based on their noise properties
However, in practice these seem to be somewhat conflated, likely do to needing to pick a set of qubits before you minimize the number of SWAPs? And I think of Sabre as being used for Routing and VF2 as used for Layout. Is this the right place to discuss this, or should be be linking to another doc where this is discussed.

Member
@[mtreinish](https://github.com/mtreinish) mtreinish [19 hours ago](https://github.com/Qiskit/documentation/pull/267#discussion_r1379306207)
Tl;dr It might be worth saving this for some of the details on stages section of the docs, because this makes more sense there talking about how all the stages are constructed. For the definition of what's in the different optimization levels by default even the list of trivial -> vf2 -> sabre might be too much detail for this document.

[...]

PR #267, opened 2023-10-31
 Transpilation defaults and configuration options

### LINKED ISSUE #101: [ NEW PAGE ] Transpilation defaults and configuration options

beckykd:
Closes #101 

LINK TO TOPIC IN PREVIEW: https://qiskit-docs-preview-pr-267.1799mxdls7qz.us-south.codeengine.appdomain.cloud/transpile/defaults-and-configuration-options (FYI @mtreinish @nbronn )


beckykd:
@nbronn

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -0,0 +1,150 @@
...
+Different gate optimizations are turned on with different `optimization_level` values. In general, the higher the number, the more optimized it is, at the expense of longer processing time. Choosing the best optimization level might take trial and error, as it depends heavily on the circuit being transpiled and the backend being targeted. By default, `optimization_level=1` is used.
+
+Levels: 
+
+
+The `optimization_level` setting specifies how much optimization to perform on the circuits. Higher levels generate more optimized circuits, at the expense of longer compile times.
+
+- `optimization_level=0`: Trivial optimization, which maps the circuit to the backend with no explicit optimization (except whatever optimizations the mapper does).
+- `optimization_level=1`: Light optimization, which maps the circuit and does simple optimizations by collapsing adjacent gates. This is the default value.
+- `optimization_level=2`: Medium optimization, including a noise-adaptive layout and a gate-cancellation procedure based on gate commutation relationships.
```
mtreinish, line 61: All optimization levels >= 1 are noise aware now, they actual all use the same basic layout and routing algorithms, we just use increasing the heuristic effort spent trying to optimize. There is an a doc PR on qiskit that was looking at adding a how to guide for optimization levels where we had a very similar discussion about these descriptions: https://github.com/Qiskit/qiskit/pull/10303

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -0,0 +1,150 @@
...
+    </Tr>
+    <Tr>
+      <Td>1</Td>
+      <Td>
+        Light optimization (default):
+
+        -   Layout (trivial → vf2 → SabreLayout if routing is required)
+        -   Routing (SabreSWAPs if needed)
+        -   1Q gate optimization
+        -   Error suppression: dynamical decoupling
```
mtreinish, line 89: There is no dynamical decoupling by default in Qiskit's optimization levels. You can refer to all the optimization passes as error suppression.

 Also, `VF2PostLayout` is error suppression too, which runs at levels 1, 2, and 3: https://qiskit.org/documentation/stubs/qiskit.transpiler.passes.VF2PostLayout.html

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -0,0 +1,150 @@
...
+      </Td>
+    </Tr>
+    <Tr>
+      <Td>2</Td>
+      <Td>
+        Medium optimization:
+          - Layout/Routing: Optimization level 1 (without trivial) + heuristic optimized with greater
+        search depth and trials of optimization function
+        -   Commutative cancelation
```
mtreinish, line 98: We also run the 1q gate optimization

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -0,0 +1,150 @@
...
+```
+
+
+### Approximation degree
+
+** ~~~Question about this: "approximation_degree (float) - Heuristic dial used for circuit approximation (1.0=no approximation, 0.0=maximal approximation). The default value is 1.0."
+
+It looks like the value is a float in the range [0 - 1]. Should we clarify that in the description?  What do the "in between" measurements mean?**
+
+### Other importatnt options?
```
mtreinish, line 140: I think `seed_transpiler`,  `initial_layout` and the `*_method` arguments are the next most commonly used arguments.

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -0,0 +1,150 @@
...
+    new_circ_lv0 = transpile(ghz, backend=backend, optimization_level=0)
+    plot_circuit_layout(new_circ_lv0, backend)
+```
+
+
+### Approximation degree
+
+** ~~~Question about this: "approximation_degree (float) - Heuristic dial used for circuit approximation (1.0=no approximation, 0.0=maximal approximation). The default value is 1.0."
+
+It looks like the value is a float in the range [0 - 1]. Should we clarify that in the description?  What do the "in between" measurements mean?**
```
mtreinish, line 138: It's a bit nebulous and it's not really well defined at the high level for the transpiler. It's basically an opaque dial that is in theory trading off accuracy of the output vs ease of execution (i.e. less gates) the exact way that's handled has never been concretely defined because it was intended to be left up to certain passes.

The only thing that really uses it by default is two qubit unitary synthesis and it's used to specify the target fidelity of the output decomposition. In other words how much error we're allowed to introduce when we convert a matrix representation of a circuit to discrete gates. If we have a lower value for "more approximation" the output circuit from synthesis will differ more from the input matrix. but also likely have less gates which will be easier to execute (at the cost of not representing exactly what was requested).

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -0,0 +1,150 @@
...
+
+        - Optimization level 2 + heuristic optimized on layout/routing further with greater effort/trials
+        - Two qubit KAK optimization
+        - Error suppression: dynamical decoupling
+      </Td>
+    </Tr>
+  </tbody>
+</Table>
+
+Example, specifying `optimization_level=0`:
```
mtreinish, line 115: It probably would be good to show >1 optimization level to highlight any differences, in the docs I assume this was taken from it uses level 0 and 3: https://qiskit.org/documentation/apidoc/transpiler.html#layout-stage

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -0,0 +1,150 @@
...
+
+# Transpilation default settings and configuration options
+
+Transpilation is the process of rewriting a given input circuit to match the topology of a specific quantum device, to optimize the circuit for execution on present day noisy quantum systems, or both. You can pass circuits to the `transpile()` method and use all default settings, or you can fine-tune the transpilation by using the available parameters. 
+
+## All available parameters:
+
+Following are all of the available parameters for the `transpile()` method.  All parameters except `circuits` are optional.  For full details, see the (Transpiler API documentation)[https://docs.quantum-computing.ibm.com/api/qiskit/transpiler#transpiler-api].
+
+Do we want this whole list?  If so, I'll add descriptions from (https://docs.quantum-computing.ibm.com/api/qiskit/compiler#compilation-routines)
```
mtreinish, line 15: I think it's worth highlighting there that there are two classes of arguments to `transpile()`, those that describe the target of compilation, these are things like `backend` and  `target` (along with the legacy BackendV1 formats like `basis_gates`, `inst_map`). By default most users just use `backend` and provide a backend object, only if you have a more advanced use case do you specify any of the other arguments to describe a backend. The other options are things that purely influence how the transpiler does it's job, these are things like `optimization_level` and the `*_method` arguments which are more commonly used to try and get better, different, or specific output from the transpiler.

nbronn: Agree with Matt. Also I think it is important to specify that abstract circuits need to be transpiled because the backends have only a limit set of basis gates, and cannot execute arbitrary operations. I generally think of the minimum function of the transpiler is to respect connectivity (and introduce SWAP gates where needed) and translate to basis gates.

nbronn, line 15: Would it be worth grouping the arguments according to which stage of transpilation they correspond to?

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -0,0 +1,150 @@
...
+        - Layout (as specified)
+        - Routing (stochastic swaps)
+      </Td>
+    </Tr>
+    <Tr>
+      <Td>1</Td>
+      <Td>
+        Light optimization (default):
+
+        -   Layout (trivial → vf2 → SabreLayout if routing is required)
```
nbronn, line 86: One thing I think is super confusing is the difference between Layout and Routing (together I call these Mapping, but that might be a personal thing). At a functional level, I think of them as

- Routing: insert SWAPs due to the limited connectivity of the lattice
- Layout: pick the set of qubits in the lattice which to use based on their noise properties

However, in practice these seem to be somewhat conflated, likely do to needing to pick a set of qubits before you minimize the number of SWAPs? And I think of Sabre as being used for Routing and VF2 as used for Layout. Is this the right place to discuss this, or should be be linking to another doc where this is discussed.

docs/transpile/defaults_and_configuration_options.mdx
nbronn: Side note: DD is added in the transpiler only for Runtime (probably not necessary to mention).

docs/transpile/defaults_and_configuration_options.mdx
nbronn: What about `CXCancellation`?

docs/transpile/defaults_and_configuration_options.mdx
beckykd: So should I even mention the different processes that take place, or just describe that for each level (1-3),  the same heuristic layout and swap insertion technique is used, but with an increased number of iterations for each level?  And if that's the case, it seems silly to split them out individually in a list or table, unless it's 0 vs 1-3.

docs/transpile/defaults_and_configuration_options.mdx
mtreinish: Tl;dr It might be worth saving this for some of the details on stages section of the docs, because this makes more sense there talking about how all the stages are constructed. For the definition of what's in the different optimization levels by default even the list of trivial -> vf2 -> sabre might be too much detail for this document.

Yeah, they're tightly coupled. Sabre is actually both a routing algorithm and a layout algorithm (it selects a layout by using a routing algorithm but instead of inserting swap gates it swaps the qubits virtually in the layout). The typical pipeline outlined here for level 1 is we first try the trivial layout (0->0, 1->1, etc) if that matches device connectivity we use it and move on (skipping routing). If it doesn't then we run `VF2Layout` (which is different from `VF2PostLayout`) which uses rustworkx to find an isomorphic subgraph on the device to execute the circuit on with no swaps (and if it finds > 1 select the 1 with best noise properties) and also skip routing, then if nothing can be found we run `SabreLayout` to select a good heuristic layout.

Just to make it extra confusing, `SabreLayout` also runs routing and inserts swaps. This is done as an efficiency because the whole algorithm is written in rust and it's much faster to compute the layout, find the swaps with that layout, and then select the layout which gives you the best output and just return that as a single thing. Then to make it even more confusing `VF2PostLayout` runs as part of the routing stage to pick a better layout because it only can be used after routing. But these are implementation details of the specific algorithms we use in the presets and the stages are logically distinct but related at a high level (i.e. you need a layout to run routing because you can't compute where to insert swaps until you know which physical qubits you're running on).

docs/transpile/defaults_and_configuration_options.mdx
mtreinish: Level 0 is using different techniques, but level 1-3 are doing exactly the same thing (with the exception of only level 1 trying trivial first before running vf2). I guess it depends on how we want to structure the documentation. For me personally the table of what exactly is run in each optimization level probably isn't relevant for this guide (I also generally worry about keeping it up to date every release) just the general trend of increasing optimization level value trades off compilation time with a potentially more optimized output circuit.

docs/transpile/defaults_and_configuration_options.mdx
mtreinish: `CXCancellation` only runs at level 1. The optimization it provides is taken care of as part of commutative cancellation. That being said this a good reminder for me that I need to generalize `CXCancellation` because it's not doing anything for backends that don't have CX.

docs/transpile/defaults_and_configuration_options.mdx
beckykd: I see - I tried a version where they're separated out.  Let me know if you like it that way, or one long set.  

Also, I hadn't added the parameter descriptions yet because I'm not sure if we're keeping them, or if we should just leave them all in the API.   I'll do it however people want it.

beckykd: Oooh - I just put out a version like that :)

docs/transpile/defaults_and_configuration_options.mdx
beckykd: OK, I'm going to merge those descriptions.

docs/transpile/defaults_and_configuration_options.mdx
beckykd: Added this discussion to the issue for details about transpiler stages: https://github.com/Qiskit/documentation/issues/104#issuecomment-1790981912

javabster:
hey I know you're still working on this draft, but just one quick comment: can you add a very short code snippet after the intro paragraph, just showing how to use `transpile()` in the most basic way without any of the optional params? Just to give newer users a basic example of how to call the transpile function in their code, and then move on to adding more params

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -1,5 +1,415 @@
...
+  - instruction_durations (List[Tuple[str, Iterable[int], float, Iterable[float], str]] | List[Tuple[str, Iterable[int], float, Iterable[float]]] | List[Tuple[str, Iterable[int], float, str]] | List[Tuple[str, Iterable[int], float]] | InstructionDurations) - Durations of instructions. Applicable only if `scheduling_method` is specified. The gate lengths defined in `backend.properties` are used by default. They are overwritten if `instruction_durations` is specified. The `instruction_durations` format must be as follows. The instruction_durations must be given as a list of tuples [(instruction_name, qubits, duration, unit), …]. | [(‘cx’, [0, 1], 12.3, ‘ns’), (‘u3’, [0], 4.56, ‘ns’)] | [(‘cx’, [0, 1], 1000), (‘u3’, [0], 300)] If `unit` is omitted, the default is `dt`, which is a sample time depending on backend. If the time unit is ‘dt’, the duration must be an integer.
+  - target (Target) - A backend transpiler target. Normally this is specified as part of the backend argument, but if you manually constructed a Target object, you can specify it here. This overrides the target from `backend`.
```
javabster, line 81: I think it would be worth moving the traget param higher up in the list, directly after the backend param.

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -1,5 +1,415 @@
...
+Following are all of the available parameters for the `transpile()` method.  There are two classes of arguments: those that describe the target of compilation, and those that influence how the transpiler works.
+
+
+All parameters except `circuits` are optional.  For full details, see the (Transpiler API documentation)[https://docs.quantum-computing.ibm.com/api/qiskit/transpiler#transpiler-api].
+
+`circuits` (`_CircuitT`) - One or more circuits to transpile. This is the only required parameter.
+
+**Do we want this whole list?  If so, I'll add descriptions from (https://docs.quantum-computing.ibm.com/api/qiskit/compiler#compilation-routines)**
+
+### Parameters used to describe the compilation target:
```
javabster, line 69: For sections like this which include long lists of params I think we should use those markdown expandable arrow section things, so we can hide some of the details unless a user explicitly expands the section.

Example of where we've used these expandable sections: https://docs.quantum-computing.ibm.com/support#frequently-asked-questions

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -1,5 +1,415 @@
...
+
+
+**Scheduling stage**
+- scheduling_method (str) - Name of the scheduling pass. This can also be the external plugin name to use for the scheduling stage. You can see a list of installed plugins by running `list_stage_plugins()` with `scheduling` for the `stage_name` argument.
+  * 'as_soon_as_possible': Schedule instructions greedily, as early as possible on a qubit resource. (alias: 'asap') 
+  * 'as_late_as_possible': Schedule instructions late, that is, keeping qubits in the ground state when possible. (alias: 'alap') 
+
+**UNKNOWN**
+
+- callback (Callable[[BasePass, DAGCircuit, float, PropertySet, int], Any]) - A callback function that is called after each pass execution. The function is called with these keyword arguments:
```
javabster, line 132: I think this callback argument could use a little more explanation, as a new user i don't really understand what this is for

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -1,5 +1,415 @@
...
+
+- output_name (str | List[str]) -  A list with strings to identify the output circuits. The length of the list should be the same length as the circuits parameter.
+
+
+## Commonly used parameters 
+ 
+<span id="optimization-level"></span>
+### Optimization level
+
+Decomposing quantum circuits into the basis gate set of the target device and the addition of SWAP gates needed to match hardware topology causes an increase the depth and gate count of quantum circuits. To mitigate this increased complexity, you can set the `optimization_level`.  Setting this value calls an optimization routine that optimizes the transpilation process by combining or eliminating gates and by optionally using algorithms to find an optimal layout (depending on the level chosen).
```
javabster, line 163:
```suggestion
Decomposing quantum circuits into the basis gate set of the target device and the addition of SWAP gates needed to match hardware topology causes an increase in the depth and gate count of quantum circuits. To mitigate this increased complexity, you can set the `optimization_level`.  Setting this value calls an optimization routine that optimizes the transpilation process by combining or eliminating gates and by optionally using algorithms to find an optimal layout (depending on the level chosen). 
```

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -1,5 +1,415 @@
...
+### Optimization level
+
+Decomposing quantum circuits into the basis gate set of the target device and the addition of SWAP gates needed to match hardware topology causes an increase the depth and gate count of quantum circuits. To mitigate this increased complexity, you can set the `optimization_level`.  Setting this value calls an optimization routine that optimizes the transpilation process by combining or eliminating gates and by optionally using algorithms to find an optimal layout (depending on the level chosen). 
+
+In some cases these methods are so effective the output circuits have lower depth than the inputs. In other cases, not much can be done, and the computation may be difficult to perform on noisy devices. Choosing the best optimization level might take trial and error, as it depends heavily on the circuit being transpiled and the backend being targeted. By default, `optimization_level=1` is used.
+
+The `optimization_level` setting specifies how much optimization to perform on the circuits. Higher levels generate more optimized circuits by increasing the heuristic effort spent trying to optimize, at the expense of longer compile times.  The default level is 1.
```
javabster, line 167: should this sentence be removed? it looks like it just duplicates what is said in the intro

javabster (commented):
Doing a first pass review of this and it's looking great! I feel like I have a better understanding of `transpile()` just from reading this!

The main thing I'm noticing is the page feels very long and quite overwhelming with lots of text, which makes me think maybe it would be worth splitting this into multiple pages, what do you think? Off the top of my head a sidebar structure like this might work:

- Use the transpile() method (not a page, just a dropdown heading)
   - transpile default options and configurations (including all the info you've got about what each of the params mean
   - set optimization_level (just the optimization level section)
   - other commonly used params (approx degree, initial layout and seed transpiler)

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -1,5 +1,411 @@
...
+qc.draw(output='mpl')
+```
+
+![Original circuit](/images/transpile/defaults-and-configuration-options/original-circuit.png)
+
+Transpile the circuit and view the result:
+
+```python
+# Specify the backend to target
+backend = FakeVigo()
```
nbronn, line 44: I think it's probably better not to specify a backend with the old topology. Yet for some examples I would prefer not using a 127 because those have `ecr` as the native 2Q gate.

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -1,5 +1,411 @@
...
+
+# Transpile the circuit
+transpiled_circ = transpile(qc, backend)
+
+# View the transpiled circuit
+transpiled_circ.draw(output='mpl')
```
nbronn, line 50: Also, there are better examples of layout/routing on a 127 for in #308

docs/transpile/defaults_and_configuration_options.mdx
beckykd: Oh yeah; can you suggest some better backends to use?

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -1,5 +1,411 @@
...
+  - inst_map (List[InstructionScheduleMap] | None) - Mapping of circuit operations to pulse schedules. If `None`, the backend’s `instruction_schedule_map` is used.
+
+### Parameters used to influence how the transpiler works:
+
+These parameters impact specific transpilation stages. Some of them might impact multiple stages, but have only been listed under one stage for simplicity. If you specify an argument, such as `initial_layout`  for the qubits you want to use, that value overrides all the passes that could change it. In other words, the transpiler won't change anything that you manually specify. 
+
+The following default values are used if you don't specify any of these parameters: 
+
+```python
+qiskit.compiler.transpile(unitary_synthesis_method='default', translation_method='basis_translator', layout_method='sabre', routing_method='sabre', approximation_degree=1.0, optimization_level=1.0, scheduling_method='as_late_as_possible')
```
nbronn, line 93: We wanted to have the default settings somewhere so they don't get lost in all the possible configurations for each stage

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -1,5 +1,411 @@
...
+```
+
+**Initialization stage**
+
+- hls_config (HLSConfig) - An optional configuration class `HLSConfig` that is passed directly to the `HighLevelSynthesis` transformation pass. This configuration class lets you specify the lists of synthesis algorithms and their parameters for various high-level objects.
```
nbronn, line 98: Probably good to rectify with #283 since there could be duplicate content / stylistic difference.

javabster:
This is generally looking really good, my feedback is largely the same as what I mentioned in my last review: https://github.com/Qiskit/documentation/pull/267#pullrequestreview-1738000805

docs/transpile/defaults_and_configuration_options.mdx
```diff
@@ -1,5 +1,411 @@
...
+
+All parameters except `circuits` are optional.  For full details, see the [Transpiler API documentation](https://docs.quantum-computing.ibm.com/api/qiskit/transpiler#transpiler-api).
+
+`circuits` (`_CircuitT`) - One or more circuits to transpile. This is the only required parameter.
+
+### Parameters used to describe the compilation target:
+
+**Backend parameters** - If you specify `backend`, you don't need to specify `target` or any other backend options. Likewise, if you specify `target`, you don't need to specify `backend` or any other backend options.
+  - backend (Backend) - If this is set, the transpiler compiles the input circuit to this device. If any other option is set that impacts these settings, such as `coupling_map`, it overrides the settings from `backend`.
+  - backend_properties (BackendProperties) - Properties returned by a backend, including information on gate errors, readout errors, qubit coherence times, and so on. Find a backend that provides this information by running `backend.properties()`.
```
javabster, line 68:
```suggestion
  - `backend_properties` (BackendProperties) - Properties returned by a backend, including information on gate errors, readout errors, qubit coherence times, and so on. Find a backend that provides this information by running `backend.properties()`.
```
All of these parameters should be in code blocks like this

docs/transpile/defaults_and_configuration_options.mdx
beckykd: I'll use Sherbrooke

docs/transpile/defaults_and_configuration_options.mdx
beckykd: Added a link and removed these.

docs/transpile/defaults_and_configuration_options.mdx
beckykd: FYI @abbycross

abbycross: We have merged #283 so any needed changes can happen on this PR.

docs/transpile/set-optimization.mdx
```diff
@@ -0,0 +1,154 @@
...
+        - Resynthesis of two-qubit blocks using [Cartan's KAK Decomposition](https://arxiv.org/abs/quant-ph/0507171). 
+        - Unitarity-breaking passes: 
+          * `OptimizeSwapBeforeMeasure`: Moves the measurements around to avoid SWAPs
+          * `RemoveDiagonalGatesBeforeMeasure`: Removes gates before measurements that would not effect the measurements
+      </Td>
+    </Tr>
+  </tbody>
+</Table>
+
+Since CX is the noisiest gate, we can quantify the transpilation's "hardware efficiency" by counting the CX gates in the resulting circuit. We will compare the default transpilation levels given the same circuit.
```
javabster, line 68: Could you add a subheading straight after the table? It felt a bit jarring to go from reading the table and jumping straight into the next piece of info. Maybe add a subheading like:
```suggestion
## Optimization level in action

Since CX is the noisiest gate, we can quantify the transpilation's "hardware efficiency" by counting the CX gates in the resulting circuit. We will compare the default transpilation levels given the same circuit. 
```

or maybe "effect of optimization level on CX count" or whatever you think is best

docs/transpile/set-optimization.mdx
```diff
@@ -0,0 +1,154 @@
...
+```
+gives
+
+```text
+Synthesis fidelity
+Exact: 1.000
+Approximate: 0.992
+```
+
+For examples of the changes to the layout, see Representing quantum computers.
```
javabster, line 147:
```suggestion
Adjusting the optimization level can change other aspects of the circuit too, not just the number of CX gates. For examples of how setting optimization level changes the layout, see [Representing quantum computers](LINK)

Comment (javabster) [reply docs/transpile/set-optimization.mdx:147]: @nbronn I think my suggestion above is factually correct but lmk if not 😅

**Review by javabster — CHANGES_REQUESTED** [review]
This is very nearly there! 

Left a tonne of comments but mostly just typos and rephrases that you can commit directly. Just a few things that need a bit of a closer look

File: docs/transpile/defaults-and-configuration-options.mdx [hunk 21]
```diff
@@ -0,0 +1,209 @@
+---
+title: Transpilation defaults and configuration options
+description: Default settings and configuration options 
+
+---
+# Transpilation default settings and configuration options
+
+Abstract circuits need to be transpiled because backends have a limited set of basis gates and cannot execute arbitrary operations. The transpiler's function is to change arbitrary circuits so that they can run on specified backend.  This is done by translating the circuits to the supported basis gates, and by introducing SWAP gates as needed, so that the circuit's connectivity matches that of the backend. 
+
+You can pass circuits to the `transpile()` method and use all default settings, or you can use parameters to fine tune the transpilation.
```
javabster, line 10:
```suggestion
You can pass circuits to the `transpile()` function and use all default settings, or you can use parameters to fine tune the transpilation.  
```

docs/transpile/defaults-and-configuration-options.mdx
```diff
@@ -0,0 +1,207 @@
+---
+title: Transpilation defaults and configuration options
+description: Default settings and configuration options 
+
+---
+# Transpilation default settings and configuration options
+
+Abstract circuits need to be transpiled because backends have a limited set of basis gates and cannot execute arbitrary operations. The transpiler's function is to change arbitrary circuits so that they can run on specified backend.  This is done by translating the circuits to the supported basis gates, and by introducing SWAP gates as needed, so that the circuit's connectivity matches that of the backend.
```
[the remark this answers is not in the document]
javabster: @beckykd I think this typo fix got missed?

javabster (commented):
I think there are still a bunch of commit suggestions I added earlier on that didn't get committed?

javabster (approved):
LGTM 🚀

PR #269, opened 2023-11-01
update primitives pages to use circuit library and 127q hardware

### LINKED ISSUE #255: [ REFACTOR ] /run/primitives-get-started
To improve:

- [ ] replace usage of random circuits with EfficientSU2 or other examples from the circuit library
- [ ] update examples to use 100Q
- [ ] update examples to use real hardware

### LINKED ISSUE #256: [ REFACTOR ] /run/primitives-examples
To improve:

- [ ]  replace usage of random circuits with EfficientSU2 or other examples from the circuit library
- [ ]  update examples to use 100Q
- [ ]  update examples to use real hardware

kevinsung:
Fixes #255.
Fixes #256.

- Instead of `random_circuit`, use `IQP` from the circuit library
- Instead of `ibmq_qasm_simulator`, use the 127-qubit device `ibm_brisbane`
- Increase circuit sizes to 127 qubits. As part of this, I deleted any code that draws the circuits, as these circuits are too large to display in a reasonable way.


javabster (commented):
This generally looks great!

Before merging I have 2 questions:
- how confident are you that this code is runnable on a real device? Hypothetically could a user copypaste these code snippets and run it?
- we got feedback about not using random circuits but instead using the circuit library, I see now you are mostly using the circuit library but with random hermitians. Are random hermitians much better than random circuits? At least for the get started code snippets could we use something a bit more real?

kevinsung:
>how confident are you that this code is runnable on a real device? Hypothetically could a user copypaste these code snippets and run it?

For all the snippets, I tested that the jobs submitted successfully. Due to queue times I couldn't find out if they would run successfully, but as far as I can tell there's no reason they should fail.

>we got feedback about not using random circuits but instead using the circuit library, I see now you are mostly using the circuit library but with random hermitians. Are random hermitians much better than random circuits? At least for the get started code snippets could we use something a bit more real?

IQP circuits were invented to demonstrate the hardness of simulating quantum circuits, not for any inherently useful purpose, so I think it makes to generate them randomly. For the examples where I used IQP, I needed a non-parameterized circuit, so variational ansatzes like RealAmplitudes or EfficientSU2 were out of the question. The "useful" choices like QFT or PhaseEstimation aren't actually practical to run on noisy devices. This is why I went with IQP.

javabster (approved):
LGTM 🚀

PR #237, opened 2023-10-24
Soft launch Qiskit migration guides

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/213. 

This allows accessing the migration files at:

* http://localhost:3000/api/migration-guides/qiskit-quantum-instance
* http://localhost:3000/api/migration-guides/qiskit-algorithms-module
* http://localhost:3000/api/migration-guides/qiskit-opflow-module

But we don't add the files to `_toc.json` or to the index, so the user needs to know to look for those URLs.

Soft launching will allow us to better make sure we're doing the migration correctly.

PR #270, opened 2023-11-01
fix references in migration guides

kevinsung:
Step 2 of #213 

Some references don't work even on the original page, so I just deleted them.

With this, https://github.com/Qiskit/documentation/pull/237 can be merged.


javabster (approved):
LGTM 🚀

PR #273, opened 2023-11-02
Rewrite bit-ordering page

frankharkins:
Following the discussion on #207, I've attempted to rewrite the bit-ordering page.

I've avoided mentioning how other resources do things, instead just describing how Qiskit handles different decisions when ordering (qu)bits. I also tried to organise these decisions on the page such that one naturally leads to the other.

If accepted, this will supersede #207. All feedback welcome.


jakelishman (commented):
The new text in here looks good to me, and generally easier to understand for beginners (imo, though I'm not always the best judge of that).  It's very good that it's dropped the use of the term "endianness".

I think it's important to talk about ordering conventions for the control and target in controlled gates, because that's a place where people are often confused as well.  We need to overhaul the Qiskit-side API documentation on those too, because our side of the documentation is very confusing as well (mostly because it confuses bit labelling with control-vs-target ordering).

frankharkins:
Thanks Jake! Just to clarify:

> I think it's important to talk about ordering conventions for the control and target in controlled gates, because that's a place where people are often confused as well. 

Is the current discussion enough, or would you like more details? The current page has this:

````markdown
- Each gate in Qiskit can interpret a list of qubits in its own way, but
  controlled gates usually follow the convention `(control, target)`.

  For example, the following cell adds a controlled-X gate where qubit $0$ is
  the control and qubit $1$ is the target.

  ```python
  from qiskit import QuantumCircuit
  qc = QuantumCircuit(2)
  qc.cx(0, 1)
  qc.draw()
  ```

  ```
  q_0: ──■──
       ┌─┴─┐
  q_1: ┤ X ├
       └───┘
  ```
````

**Comment by jakelishman** [comment]
Oh goodness, I totally glazed over that paragraph, sorry.  I think that's a solid start, and then it'd be good to talk about how the consequences of `CX` taking its control as its "zeroth" argument means that the computational-basis matrix form looks like "<explicit matrix>" (given the matrix ordering discussion you already had about statevectors).  It's just a small addition that hits the last remaining confusion, I think.

File: docs/build/qubit-order.mdx [hunk 1]
```diff
@@ -0,0 +1,153 @@
+---
+title: Bit-ordering in Qiskit
```
abbycross, line 2: We should use the same title in all three places (page title, metadata, and toc file). At the moment we have to take into consideration the unfortunate truncation that happens in the left nav (there is an issue out to fix that), so shorter might be better, at least temporarily.

docs/build/qubit-order.mdx
```diff
@@ -0,0 +1,153 @@
...
+description: Learn about Qiskit's ordering conventions and why we chose them
+---
+
+# Order of bits and qubits in Qiskit
+
+If you have a set of $n$ bits (or qubits), you'll usually label each bit $0
+\rightarrow n-1$. Different softwares and resources must choose how they order
+these bits both in computer memory and when displayed on screen.
```
abbycross, line 10:
```suggestion
these bits both in computer memory and when displayed on-screen.
```

docs/build/qubit-order.mdx
```diff
@@ -0,0 +1,153 @@
...
+
+  ```
+  {'00': 0.4999999999999999, '01': 0.4999999999999999}
+  ```
+
+  This occasionally causes confusion when interpreting a string of bits, as you
+  might expect the leftmost bit to be bit $0$, whereas it usually represents bit
+  $n$.
+
+- When representing a state vector as a list of complex numbers (amplitudes),
```
abbycross, line 84:
```suggestion
- When representing a statevector as a list of complex numbers (amplitudes),
```

docs/build/qubit-order.mdx
```diff
@@ -0,0 +1,153 @@
...
+
+  ```
+  q_0: ──■──
+       ┌─┴─┐
+  q_1: ┤ X ├
+       └───┘
+  ```
+
+
+### Changing ordering in Qiskit
```
abbycross, line 119:
```suggestion
### Change ordering in Qiskit
```

docs/build/qubit-order.mdx
```diff
@@ -0,0 +1,153 @@
...
+  ```
+
+  ```
+  Qubit(QuantumRegister(2, 'q'), 0)
+  ```
+
+- On a circuit diagram, qubit $0$ is the topmost qubit, and qubit $n$ the
+  bottommost qubit (you can change this default with the `reverse_bits=True`
+  argument to `QuantumCircuit.draw`; see [Changing
+  ordering](#changing-ordering)).
```
abbycross, line 31:
```suggestion
  argument to `QuantumCircuit.draw`; see [Change
  ordering in Qiskit](#change-ordering-in-qiskit)).
```

docs/build/qubit-order.mdx
```diff
@@ -0,0 +1,153 @@
...
+  ```
+  q_0: ─────
+       ┌───┐
+  q_1: ┤ X ├
+       └───┘
+  ```
+
+- When interpreting bits as a number, bit $0$ is the least significant bit, and
+  bit $n$ the most significant. This is helpful when coding because each bit
+  has the value $2^\text{label}$. For example, the following circuit execution
```
abbycross, line 47: As a total noob I don't understand the 2^label part... does "label" represent the bit's position?

abbycross: Ah, on my third reading I actually digested the part where you define "label" 😆 so I think nothing needs to change!

frankharkins: Even if it's technically correct, do you think it could benefit from rephrasing?

abbycross: If you wanted to help people like me who don't retain that definition half-a-page later, you could add a parenthetical:
 

> 2^label (`label` being the qubit's index in the list).

frankharkins: Will do, thanks! (And thanks for the review)

abbycross (approved):
looks good to me!

jakelishman (approved):
Thanks for this Frank - this reads far more clearly to me than the previous version, and it avoids the dangerous terminology.

PR #288, opened 2023-11-06
Batch mode

### LINKED ISSUE #272: [ NEW PAGE ] Add documentation on Batch mode
### URL, if applicable

_No response_

Qiskit Runtime is introducing a new `Batch` class that works similarly as `Session`, but the former is for submitting multiple jobs at once while the latter is for iterative workloads. 

cc @kt474 

beckykd:
closes #272


abbycross (approved):

beckykd:
@javabster - @jyu00 suggested changing the "Sessions" section to "Scheduling mode", so we can put sessions, batch, and Fair share scheduling in there.

docs/run/sessions.mdx
```diff
@@ -6,7 +6,7 @@ description: Describes what sessions are.
 
 # About sessions
 
-A session allows a collection of jobs to be grouped and jointly scheduled by the Qiskit Runtime service, facilitating iterative use of quantum computers without incurring queuing delays on each iteration. This eliminates artificial delays caused by other users’ jobs running on the same quantum device during the session.
+A session allows a collection of jobs to be grouped and jointly scheduled by the Qiskit Runtime service, facilitating iterative use of quantum computers without incurring queuing delays on each iteration. This eliminates artificial delays caused by other users’ jobs running on the same quantum device during the session. If you want to submit multiple jobs at once, use the [`Batch`](run-primitives-batch) class instead.
```
jyu00, line 9:
```suggestion
A session allows a collection of jobs to be grouped and jointly scheduled by the Qiskit Runtime service, facilitating iterative use of quantum computers without incurring queuing delays on each iteration. This eliminates artificial delays caused by other users’ jobs running on the same quantum device during the session. If you don't need iterative processing and can send all jobs at once, use the [`Batch`](run-primitives-batch) class instead.
```
You can still send multiple jobs under `Session`.

docs/run/run-primitives-batch.mdx
```diff
@@ -0,0 +1,30 @@
+---
+title: Run primitives in a batch
```
jyu00, line 2: Why is this titled Run _primitives_ in a batch, whereas the other is Run _jobs_ in a session ?

jyu00 (commented):
Mostly looks good. We can add more to the batch section when it's more than just a sham class lol

docs/run/run-primitives-batch.mdx
```diff
@@ -0,0 +1,30 @@
...
+description: How to run primitives in batch mode. 
+
+---
+
+# Run primitives in a batch
+
+Batch mode is ideal for running experiments closely together to avoid device drifts, that is, to maintain device characterization.  All jobs must be provided at the outset. If you want to submit iterative job, use [sessions](sessions) instead.
```
javabster, line 9:
```suggestion
Batch mode is ideal for running experiments closely together to avoid device drifts, that is, to maintain device characterization.  All jobs must be provided at the outset. If you want to submit iterative jobs, use [sessions](sessions) instead.
```

docs/run/sessions.mdx
```diff
@@ -134,13 +134,7 @@ If the maximum timeout value has not been reached:
...
     -   The session cannot be reopened.
 
-## Different ways of using sessions
-
-Sessions can be used for iterative or batch execution.
-
-### Iterative
-
-Any session job submitted within the interactive timeout, also known as interactive time to live (ITTL), is processed immediately. This allows some time for variational algorithms, such as VQE, to perform classical post-processing.
+Sessions run _iteratively_.  To run jobs in batch mode, see [batch](run-primitives-batch).   Running iteratively means that any session job submitted within the interactive timeout, also known as interactive time to live (ITTL), is processed immediately. This allows some time for variational algorithms, such as VQE, to perform classical post-processing.
```
javabster, line 137:
```suggestion
Sessions run _iteratively_. Running iteratively means that any session job submitted within the interactive timeout, also known as interactive time to live (ITTL), is processed immediately. This allows some time for variational algorithms, such as VQE, to perform classical post-processing. To run jobs in batch mode, see [batch](run-primitives-batch).
```
It feels a little odd to introduce the idea of iteration then jump to batching then go back to explaining iteration, do i'd recommend rephrase like this

javabster (commented):
Generally looking good to me, I'll let @jyu00  give the final approve before we merge though.

One question I have just for my own curiosity: am I understanding correctly that a user would want to use sessions primarily if they need to do something that requires some classical postprocessing before sending the next job? And a user would want to use batch mode if they don't need to do that kind of post-processing and just want to send the jobs immediately one after the other? so the deciding factor between sessions or batching would be if you need to do classical postprocessing on each job or not?

docs/run/run-primitives-batch.mdx
beckykd: Good question - seems like they should be consistent.  @javabster do you have a preference?

jyu00:
> One question I have just for my own curiosity: am I understanding correctly that a user would want to use sessions primarily if they need to do something that requires some classical postprocessing before sending the next job? And a user would want to use batch mode if they don't need to do that kind of post-processing and just want to send the jobs immediately one after the other? so the deciding factor between sessions or batching would be if you need to do classical postprocessing on each job or not?

Yes, hence Session has a non-zero interactive TTL (i.e. your next job gets priority as long as it's submitted within X mins) and Batch won't.

docs/run/run-primitives-batch.mdx
```diff
@@ -0,0 +1,30 @@
+---
+title: Run primitives in a batch
+description: How to run primitives in batch mode. 
+
+---
+
+# Run primitives in a batch
+
+Batch mode is ideal for running experiments closely together to avoid device drifts, that is, to maintain device characterization.  All jobs must be provided at the outset. If you want to submit iterative jobs, use [sessions](sessions) instead.
```
pandasa123, line 9: Is the note about drift necessarily true? If I have a large experiment that takes several hours, I'll still have drift

beckykd: I'm not sure - it came from the existing Batch content on the Sessions page.  @jyu00 ?

jyu00: @pandasa123 you're right that the statement isn't always true. I think the intent here is to describe when you want to use Batch vs single jobs vs session. Some reasons to use batch would be (1) you want to avoid delays between jobs (e.g. to avoid drift) and (2) to speed things up using classical parallelism. 

tbh I think the opening paragraph can use a bit of rewording. It starts by saying it's ideal for avoiding drift, which can be true but not always. Then it lists 2 bullets points, which I think are meant to demonstrate the benefits, but "Suitable for batching many jobs together" isn't exactly a benefit. To me the biggest selling point of Batch is the speed up you gain by having multiple jobs run in parallel.

docs/run/run-jobs-batch.mdx
```diff
@@ -0,0 +1,30 @@
...
+title: Run jobs in a batch
+description: How to run jobs in batch mode. 
+
+---
+
+# Run jobs in a batch
+
+Batch mode can shorten processing time if all jobs can be provided at the outset. If you want to submit iterative jobs, use [sessions](sessions) instead.  Using batch mode has these benefits:
+
+-  The jobs' classical computation, such as compilation, is run in parallel. Thus, runing multiple jobs in a batch is significantly faster than running them serially.
```
pandasa123, line 11: runing --> running

docs/run/run-jobs-batch.mdx
```diff
@@ -0,0 +1,30 @@
+---
+title: Run jobs in a batch
+description: How to run jobs in batch mode. 
+
+---
+
+# Run jobs in a batch
+
+Batch mode can shorten processing time if all jobs can be provided at the outset. If you want to submit iterative jobs, use [sessions](sessions) instead.  Using batch mode has these benefits:
```
pandasa123, line 9: Is it possible to add links to new tutorials that use batch mode? 

Later today, we'll be updating the [CHSH](https://learning.quantum-computing.ibm.com/tutorial/chsh-inequality) and [Grover](https://learning.quantum-computing.ibm.com/tutorial/grovers-algorithm) one to do exactly that

beckykd: Of course!

beckykd: done

docs/run/_toc.json
```diff
@@ -41,7 +41,7 @@
       ]
     },
     {
-      "title": "Sessions",
+      "title": "Scheduling modes",
```
jyu00, line 44:
```suggestion
      "title": "Execution modes",
```

`Execution modes` is the term we are presenting at summit for batch + session.

docs/run/run-jobs-batch.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+-  The jobs' classical computation, such as compilation, is run in parallel. Thus, running multiple jobs in a batch is significantly faster than running them serially. 
+-  There is no delay between job, which can help avoid drift.
+
+<Admonition type="note">
+    When batching, jobs are not guaranteed to run in the order they are submitted.
+</Admonition>
+
+![Batch execution diagram.](/images/run/batch.png 'Figure 3: Batch execution')
```
jyu00, line 18:
```suggestion
![Batch execution diagram.](/images/run/batch.png 'Figure 1: Batch execution')
```

docs/run/run-jobs-batch.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+The following example shows how you can divide up a long list of circuits into multiple jobs and run them as a batch to take advantage of the parallel processing.
+
+```python
+jobs = []
+with Batch(backend) as batch:
+    estimator = Estimator(batch)
+    # calls within this context are part of the batch.
+    for obs_set in observable_sets:
+        jobs.append(estimator.run(circuits, observables=obs_set))
```
jyu00, line 28: Do we want a full example of using `Batch` (e.g. including import statements)? Since unlike `Session`, there isn't a separate page shows the full code. Or can we just rely on the soon-to-come tutorials?

jyu00 (approved):

docs/run/run-jobs-batch.mdx
beckykd: A full example might be a good idea - @pandasa123 said that the tutorials that I linked to were updated to use Batch today

beckykd: I don't see Batch in the Grover's one though; and the CHSH one is slightly different and doesn't use the with Batch() format that we have in this file

beckykd: They've been updated so I just pointed to them.

PR #307, opened 2023-11-08
Refactor of the link checker

arnaucasau:
### Summary
This PR refactors the link checker and adds unit tests for the Link class.

### Details of the refactor
The class link previously in `scripts/commands/checkLinks.ts` has been moved into `scripts/lib/LinkChecker.ts` together with a new class called File that will be useful to store the information of a file and all its anchors, avoiding having to read the files twice. In the original link checker, we read all the files and stored all the links to check them later. This is a good approach, but the problem comes when we want to extend the functionalities of the script to check anchors as well. In that case, we have the link, but we can't know if an anchor is valid without reading again the destination file looking for the markdown header or the HTML id. With the File class, we can store them at the same moment we extract all the links.

The second change is to read all the links at first, and then check them all together, instead of checking the links of a file on the fly. This change is important to avoid having duplicated links that make us check them multiple times. With this approach, we can reduce the number of links to check, having a big impact when we add the validation of external links (it will reduce the number of fetches). With the changes in this PR, the only links that are validated more than once are internal links. The external links don't depend on their origin to be valid.

In the class Link, we have one change to the attributes. The attribute origin has become an array to store all the origin files where a link appears. This array is necessary because we won't have duplicated links, but we still want to print a custom message for each file in which a link is not valid. The function that validates the links will return an array of error messages instead of only one boolean value. Each error message corresponds with the error found in a link for an origin file.

The main function in `scripts/commands/checkLinks.ts` will now follow the following steps to check all the links:

1. Determine what files we have and separate them into files with links to read and files we don't need to parse.
2. Parse the files with links and get a Map with all the links in all the files without duplications.
3. Create an array with all the valid destinations for a link (This includes files previously parsed, and files without links like images, gifs, etc)
4. Validate all the links
5. Print the results of the validations (custom message for every link and origin in case of error)

### Unit tests
In the file `scripts/lib/LinkChecker.test.ts` we can find the following unit tests:

- Tests for the constructor of the class Link in different scenarios
- Test for the validation of different links (internal with relative paths, internal with absolute paths, external links)
- Test for the generation of the possible paths of a link with a relative path.


Eric-Arellano (commented):
Thanks, Arnau! I'm excited how this will unlock the new features. You did a great job thinking through data structures and writing effective comments and unit tests.

I realized while reading the `Link` class that multiple times, there are essentially two code paths: external vs internal. They don't share much in common, so it's making the code harder to follow. Further, the requirements are quite different between the two types of links.

External links:
* Must de-duplicate across files
* Don't care about anchors
* Don't care about the origin file when checking if it's a valid link

Internal links:
* Doesn't matter if you de-duplicate. It's fast to check
* Does care about anchors
* Does care about the origin file name when checking if it's valid

The code will be a lot simpler if you instead use two classes `ExternalLink` and `InternalLink`. Your main function will treat them differently; for the `ExternalLink`s, it will use something like a Set to deduplicate the links. 

Conventionally, object oriented programming would say you need a common `Link` parent class. I don't think that's necessary and it might result in forcing awkward APIs. It's fine to have two distinct but related classes.

--

That will allow you to make your classes in the lib file much easier to follow by removing side-effects, like mutating `this.origin` and `this.status`. Instead, you would store that in variables inside your main function. 

It's highly desirable to have classes that are simply "data records" that you avoid mutating. All they do is stores some frozen/fixed data, and then have methods that do interesting things with that data. It makes it easier to think about the object because you don't have to do a mental calculus of what the current state of the object is. For example, when reading your code, I was concerned of what happens if `checkLink` is called before `this.origin` is properly initialized.

Data structure wise, having the highly "coupled" `this.origin` and `this.status` is a code smell too. They must be kept in sync, but the code doesn't enforce that; the programmer can mess things up. You can fix that by removing `this.origin` and `this.status` from the class entirely. If you had wanted to keep them there, I would recommend using a data-structure that forces the two to be in sync, such as a `Map`, or an array of tuples.

arnaucasau:
Thank you for your review @Eric-Arellano! After our conversation, I have simplified the Link class by removing `this.status`, and instead returning those boolean values as an array when we are doing the validation. This array is now used in the main function. I have also removed the possibility of modifying `this.origin` and now it's an array defined in the constructor with all the origin files together. To do that, I stored all the necessary information in the `loadFilesAndLinks` function in `scripts/commands/checkLinks.ts` and created each link afterward. Thanks to this change, we can work with an array of Link in the main function instead of a Map, which is more convenient for printing the results!

The last change has been removing the distinction between internal links with relative paths and internal links with absolute paths to work with just internal links. This change simplifies the validation function having now two possible paths (external links and internal links).

scripts/lib/LinkChecker.ts
```diff
@@ -0,0 +1,128 @@
...
+    this.isExternal = linkString.startsWith("http");
+  }
+
+  resolve(origin: string): string[] {
+    /*
+     * Return list of possible paths link could resolve to
+     */
+    if (this.isExternal) {
+      return [this.value];
+    }
```
Eric-Arellano, line 57: This function is never called by the external link checker. So I suggest instead renaming the function to something more clear like `possibleInternalLinks()` or `candidateInternalLinks()`, and then remove this code.

scripts/lib/LinkChecker.ts
```diff
@@ -0,0 +1,128 @@
...
+    this.value = splitLink[0];
+    this.anchor = splitLink.length > 1 ? `#${splitLink[1]}` : "";
+    this.origins = origins;
+    this.isExternal = linkString.startsWith("http");
+  }
+
+  resolve(origin: string): string[] {
+    /*
+     * Return list of possible paths link could resolve to
+     */
```
Eric-Arellano, line 54: Reminder to move this docstring and all others to live above the method, and start with `/**` on the first line. (JSDoc)

scripts/lib/LinkChecker.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  setAnchors(anchors: string[]) {
+    this.anchors = anchors;
+  }
+}
+
+export class Link {
+  readonly value: string;
+  readonly anchor: string;
+  readonly origins: string[];
```
Eric-Arellano, line 35: I realize `origin` is an overloaded name: https://developer.mozilla.org/en-US/docs/Web/API/URL/origin

How about `originFiles` to be unambiguous?

scripts/lib/LinkChecker.ts
```diff
@@ -0,0 +1,128 @@
...
+    /*
+     * True if link is in `existingFiles`, otherwise false
+     */
+    const possiblePaths = this.resolve(origin);
+    for (let filePath of possiblePaths) {
+      if (existingFiles.some((file) => file.path == filePath)) {
+        return true;
+      }
+    }
+    return false;
```
Eric-Arellano, line 103:
```suggestion
    return possiblePaths.some((possibleFile) => (
       existingFiles.some((exitingFile) => existingFile.path == possibleFile)
      )
    )
```

scripts/lib/LinkChecker.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  checkLink(existingFiles: File[]) {
```
Eric-Arellano, line 106:
```suggestion
  checkLink(existingFiles: File[]): boolean[] {
```

scripts/lib/LinkChecker.ts
```diff
@@ -0,0 +1,128 @@
...
+    if (this.isExternal) {
+      // External link
+      const result = this.checkExternalLink();
+      this.origins.forEach(() => results.push(result));
+    } else {
+      // Internal link
+      for (let origin of this.origins) {
+        const result = this.checkInternalLink(existingFiles, origin);
+        results.push(result);
+      }
```
Eric-Arellano, line 123:
```suggestion
      this.origins.forEach((origin) => {
        const result = this.checkInternalLink(existingFiles, origin);
        results.push(result);
        })
```

scripts/lib/LinkChecker.ts
```diff
@@ -0,0 +1,128 @@
...
+     * Returns a boolean for each origin of the link,
+     * true if the link is in `existingFiles`
+     * or is a valid external link, otherwise false
+     */
+    const results = [];
+
+    if (this.isExternal) {
+      // External link
+      const result = this.checkExternalLink();
+      this.origins.forEach(() => results.push(result));
```
Eric-Arellano, line 117: You can early return here.

scripts/lib/LinkChecker.ts
Eric-Arellano: You could also get the set intersection. If there are any elements in both sets, return true. That's probably easier to read and should be more efficient.

scripts/lib/LinkChecker.ts
```diff
@@ -0,0 +1,128 @@
...
+  anchors: string[];
+
+  constructor(path: string) {
+    this.path = path;
+    this.anchors = [];
+  }
+
+  setAnchors(anchors: string[]) {
+    this.anchors = anchors;
+  }
```
Eric-Arellano, line 29: It'd be better for this to be set only in the constructor, and that `this.anchors` is readonly.

scripts/lib/LinkChecker.test.ts
```diff
@@ -0,0 +1,159 @@
...
+      "docs/test/testFile/index.md",
+      "docs/test/testFile/index.mdx",
+      "docs/test/testFile/index.ipynb",
+    ];
+    expect(possiblePaths).toEqual(extectedPaths);
+  });
+  test("Possible links for an internal link with an absolute path", () => {
+    let testLink = new Link("/testFile", ["docs/test/test2/testorigin.mdx"]);
+    let possiblePaths = testLink.resolve("docs/test/test2/testorigin.mdx");
+    let extectedPaths = [
```
Eric-Arellano, line 149:
```suggestion
    let expectedPaths = [
```

Eric-Arellano: Ditto others in the file

scripts/lib/LinkChecker.test.ts
```diff
@@ -0,0 +1,159 @@
...
+      "docs/test/testFile.md",
+      "docs/test/testFile.mdx",
+      "docs/test/testFile.ipynb",
+      "docs/test/testFile/index.md",
+      "docs/test/testFile/index.mdx",
+      "docs/test/testFile/index.ipynb",
+    ];
+    expect(possiblePaths).toEqual(extectedPaths);
+  });
+  test("Possible links for an internal link with an absolute path", () => {
```
Eric-Arellano, line 146: Nit: you want a blank line between each distinct test

scripts/lib/LinkChecker.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    let testLink = new Link("../testpath", [
+      "docs/test/testorigin.mdx",
+      "docs/test/test2/testorigin.mdx",
+      "docs/test/test3/testorigin.mdx",
+      "docs/test/test2/test4/testorigin.mdx",
+    ]);
+    let testFile1 = new File("docs/testpath.mdx");
+    let testFile2 = new File("docs/test/test2/testpath.mdx");
+    let allFiles: File[] = [testFile1, testFile2];
+    const results = testLink.checkLink(allFiles);
```
Eric-Arellano, line 127: Nit: inlining is simpler

```suggestion
    const results = testLink.checkLink([testFile1, testFile2]);
```

Ditto the other tests

scripts/lib/LinkChecker.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+    return false;
+  }
+
+  checkLink(existingFiles: File[]) {
+    /*
+     * Returns a boolean for each origin of the link,
+     * true if the link is in `existingFiles`
+     * or is a valid external link, otherwise false
+     */
```
Eric-Arellano, line 111: This return type is a bit complex for the user because they have to associate the result element with the relevant input. This is similar to my earlier feedback:

> Data structure wise, having the highly "coupled" this.origin and this.status is a code smell too. They must be kept in sync, but the code doesn't enforce that; the programmer can mess things up.

I believe you don't need to do anything with valid links, right? We only care about issues. So, simply don't return the valid ones.

Rather than returning a boolean, could you return structured data? Such as a string or an object that says what the full error is? Then in main, you simply iterate through all of these errors and `console.log()` them.

scripts/lib/LinkChecker.ts
```diff
@@ -0,0 +1,123 @@
...
+    this.isExternal = linkString.startsWith("http");
+  }
+
+  /*
+   * Return list of possible paths link could resolve to
+   */
+  possibleFilePaths(originFile: string): string[] {
+    if (this.value === "") {
+      return [originFile];
+    } // link is just anchor
```
Eric-Arellano, line 56: Nit: a little confusing that this comment is for the above if block. I suggest it being above line 54

scripts/lib/LinkChecker.ts
```diff
@@ -0,0 +1,123 @@
...
+    let baseFilePath;
+    if (this.value.startsWith("/")) {
+      // Path is relative to DOCS_ROOT
+      baseFilePath = path.join(DOCS_ROOT, this.value);
+    } else {
+      // Path is relative to originFile file
+      baseFilePath = path.join(path.dirname(originFile), this.value);
+    }
+    // Remove trailing '/' from path.join
+    baseFilePath = baseFilePath.replace(/\/$/gm, "");
```
Eric-Arellano, line 70: You can use a ternary operator

```suggestion
    const relativeToFolder = this.value.startswith("/") ? DOCS_ROOT : path.dirname(originFile);
    // Also remove trailing '/' from path.join
    const baseFilePath = path.join(relativeToFolder, this.value).replace(/\/$/gm, "");
```

scripts/lib/LinkChecker.ts
```diff
@@ -0,0 +1,123 @@
...
+    let possibleFilePaths = [];
+    for (let index of ["", "/index"]) {
+      for (let extension of CONTENT_FILE_EXTENSIONS) {
+        possibleFilePaths.push(baseFilePath + index + extension);
+      }
+    }
+    return possibleFilePaths;
+  }
+
+  checkExternalLink() {
```
Eric-Arellano, line 84:
```suggestion
  checkExternalLink(): boolean {
```

scripts/lib/LinkChecker.ts
```diff
@@ -0,0 +1,123 @@
...
+    // External link checking not supported yet
+    return true;
+  }
+
+  /**
+   * True if link is in `existingFiles`, otherwise false
+   */
+  checkInternalLink(existingFiles: File[], originFile: string) {
```
Eric-Arellano, line 92:
```suggestion
  checkInternalLink(existingFiles: File[], originFile: string): boolean {
```

scripts/commands/checkLinks.ts
```diff
@@ -122,69 +43,131 @@ function markdownFromNotebook(source: string): string {
...
+
+  // Create an array with all the valid destinations for a link
+  const otherFiles = loadFiles(pathsWithoutLinks);
+  const existingFiles = docsFiles.concat(otherFiles);
+
+  // Validate the links and print the results
+  let allGood = true;
+  linkList.forEach((link) => {
+    const errorMessages = link.checkLink(existingFiles);
+    errorMessages.forEach((errorMessage) => console.log(errorMessage));
```
Eric-Arellano, line 162:
```suggestion
    errorMessages.forEach((errorMessage) => console.error(errorMessage));
```

scripts/commands/checkLinks.ts
```diff
@@ -122,69 +43,131 @@ function markdownFromNotebook(source: string): string {
...
+    allGood &&= errorMessages.length == 0;
+  });
+
+  if (!allGood) {
     console.log("\nSome links appear broken 💔\n");
```
Eric-Arellano, line 167:
```suggestion
    console.error("\nSome links appear broken 💔\n");
```

scripts/commands/checkLinks.ts
```diff
@@ -122,69 +43,131 @@ function markdownFromNotebook(source: string): string {
...
+  filePaths: string[],
+): Promise<[File[], Link[]]> {
+  const fileList: File[] = [];
+  const linkList: Link[] = [];
+
+  // Auxiliary Map to avoid link duplications
+  const linkMap = new Map<string, string[]>();
+
+  for (let filePath of filePaths) {
+    const source = await readFile(filePath, { encoding: "utf8" });
```
Eric-Arellano, line 61: Using `await` in a for loop like this is usually less efficient than you want because it doesn't allow you to leverage your computer's multiprocessing. You can use `Promise.all()` to evaluate a collection/array of `Promise`s so they run in concurrency. That will make this script faster.

I'd recommend creating an `async function` that operates on a single file. It gets all the information you need for this `loadFilesAndLinks`. Then, you can call `await Promise.all(filePaths.map((fp) => loadFile(fp))`

Eric-Arellano (approved):
This is looking great! Good testing too.

Most my comments are nits / small refactors.

PR #266, opened 2023-10-31
Attempted to fix broken API ref links

javabster:
I've done my best to fix up the links that were broken due to https://github.com/Qiskit/documentation/pull/202

There were some links that we just don't have ways to link to in onedocs, because we are missing anchor tags for methods within api ref pages. So for these links I just removed the link entirely

PR #312, opened 2023-11-09
Fix updateApiDocs.ts to handle certain links

### LINKED ISSUE #275: Figure out why links broke for qiskit-ibm-provider
https://github.com/Qiskit/documentation/pull/266 fixed issues manually. Those should instead come from fixes to the generation script.

frankharkins:
PR #266 manually fixed some links in `qiskit-ibm-provider` that were incorrectly transformed by our scripts.

After investigating, it turns out these links were all of the form:

```
https://qiskit.org/documentation/stubs/.*?\.html#
```

None of these links appear in `qiskit-ibm-runtime`.

I fixed this in 43d399e22960bec458f47bcdfe6dce4ea721223d. The problem is the `.html` was not being removed by `removeSuffix` as the link had an anchor. I changed it to split the anchor off, process the link as usual, then add the anchor back on.

I also decided **not** to append the anchor if it's the same as the URL; these anchors don't seem to exist in our docs.

There are two other commits:
  * bbf9275e43f6def07b5ccb50ab46ccc6be197585 in which I re-run the updateApiDocs script with the new fix
  * 6c125faea1d7e9e260079ad346ee1709d11e632b in which I use the same `transformLink` function for both runtime and provider (they were identical beforehand).

***

Resolves #275.


scripts/commands/updateApiDocs.ts
```diff
@@ -50,6 +50,11 @@ interface Arguments {
   packages: string[];
 }
 
+interface Link {
+  url: string;
+  text?: string;
```
Eric-Arellano, line 55: Maybe `description`? `text` is a little vague.

scripts/commands/updateApiDocs.ts
```diff
@@ -61,14 +66,35 @@ type Pkg = {
...
-    text?: string,
-  ) => { url: string; text?: string } | undefined;
+  transformLink?: (url: string, text?: string) => Link | undefined;
 };
 
 type PkgHtml = { pkg: Pkg; version: string; path: string };
 
+function transformLink(
+  url: string,
+  text: string | undefined,
```
Eric-Arellano, line 76: ditto on the name `text`. 

Also FYI in TS, you can use `text?: string` I believe for this same syntax. Either style is fine to me.

Eric-Arellano (approved):
Thank you so much, Frank. Great investigation and clean fix 🙌

scripts/commands/updateApiDocs.ts
frankharkins: I think `text` is the correct term, but I've added a comment in [3d0e8b2](https://github.com/Qiskit/documentation/pull/312/commits/3d0e8b24e5dd65ce0e921f347afeb485f43f2307) to clarify what it means.

scripts/commands/updateApiDocs.ts
frankharkins: that's neater; thanks!

scripts/commands/updateApiDocs.ts
```diff
@@ -78,20 +104,7 @@ const PACKAGES: Pkg[] = [
...
-      let updateText = url === text;
-      const prefix = prefixes.find((prefix) => url.startsWith(prefix));
-      if (prefix) {
-        url = removePrefix(url, prefix);
-        url = removeSuffix(url, ".html");
-        const newText = updateText ? url : undefined;
-        return { url: `/api/qiskit/${url}`, text: newText };
-      }
-    },
+    transformLink: transformLink,
```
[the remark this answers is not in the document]
frankharkins: Nice, thanks!

Eric-Arellano (approved):
Thanks!

PR #313, opened 2023-11-09
Pull in release notes for runtime and provider

frankharkins:
This PR pulls in the release notes for all API docs. The current script seems to handle the most current release notes well.

To do:

- [x] Add metadata for the release note files


Eric-Arellano:
> Render the "Legacy release notes" page for Qiskit API docs (potentially out of scope of this PR)

If it doesn't render, you can add `.txt` to the file ending for now. Otherwise it will cause the docs application to not launch when we sync this PR with the infrastructure.

scripts/lib/sphinx/generateToc.ts
```diff
@@ -125,6 +126,11 @@ export function generateToc(options: {
     tocChildren.push(...orderEntriesByTitle(nestedTocModules));
   }
 
+  tocChildren.push({
+    title: "Release notes",
+    url: pkg.releaseNotesUrl,
+  });
```
Eric-Arellano, line 132: For Qiskit, this needs to be a folder with one file per minor release. I recommend not dealing with Qiskit in this PR yet. Get Runtime and Provider working well since they're much simpler.

For example, Qiskit is complex because we need to split up the legacy HTML page into several pages ourselves, and we need to remove all the non-Terra stuff. That is a one-time migration and probably shouldn't be in this script.

scripts/lib/sphinx/generateToc.ts
```diff
@@ -125,6 +126,11 @@ export function generateToc(options: {
...
+
   tocChildren.push({
     title: "Changelog",
```
Eric-Arellano, line 135: The release notes supersede this.

scripts/lib/sphinx/generateToc.ts
```diff
@@ -126,15 +125,12 @@ export function generateToc(options: {
     tocChildren.push(...orderEntriesByTitle(nestedTocModules));
   }
 
-  tocChildren.push({
-    title: "Release notes",
-    url: pkg.releaseNotesUrl,
-  });
-
-  tocChildren.push({
```
Eric-Arellano: For now, we want Qiskit to still have this Changelog link until the release notes folder is implemented.

Also Runtime and Provider no longer need to specify a changelogUrl. Although that's fine to keep it for now while this code is in transition. We only want to remove changelogUrl once possible.

frankharkins: Can it temporarily be called "release notes"? Just for laziness before I implement that feature

frankharkins:
I've de-scoped this PR so it doesn't cover the Qiskit release notes. The pages aren't loading for me in the preview, but work fine locally 🤔

scripts/lib/sphinx/generateToc.ts
Eric-Arellano: Not a very big deal; but I think you'll simply want the `else` block in the generation script to have the old code?

frankharkins: That's true; done!

scripts/lib/sphinx/addFrontMatter.ts
```diff
@@ -25,6 +29,14 @@ python_api_type: ${result.meta.python_api_type}
 python_api_name: ${result.meta.python_api_name}
 ---
 
+${markdown}
+`;
+    } else if (result.isReleaseNotes) {
+      result.markdown = `---
+title: Release notes
+description: ${pkg.pkg.title} v${pkg.version} release notes
+---
```
frankharkins, line 38: Not sure where the front matter is used and what attributes it needs

scripts/lib/sphinx/sphinxHtmlToMarkdown.test.ts
```diff
@@ -1417,6 +1426,34 @@ compilation flow follows the structure given below:</p>
...
+          </ul>
+          `,
+      url: "http://qiskit.org/docs/release_notes.html",
+      imageDestination: "/images/qiskit",
+    }),
+  ).toMatchInlineSnapshot(`
+        {
+          "images": [],
+          "isReleaseNotes": true,
+          "markdown": "",
```
frankharkins, line 1451: Not sure why this is empty...

scripts/commands/updateApiDocs.ts
```diff
@@ -38,7 +38,7 @@ interface Arguments {
   version: string;
 }
 
-type Pkg = {
+export type Pkg = {
```
Eric-Arellano, line 41: I'm surprised this works. It creates a circular dependency. Does the script run? 

Cleaner would be to move this type definition to `lib/`, e.g. `lib/updateApi.ts`.

scripts/lib/sphinx/SphinxToMdResult.ts
```diff
@@ -16,6 +16,7 @@ export type SphinxToMdResult = {
   markdown: string;
   meta: PythonObjectMeta;
   images: Array<{ src: string; dest: string }>;
+  isReleaseNotes?: boolean;
```
Eric-Arellano, line 19: With the `?`, do you need a tri-state: `false`, `true`, `unknown`? Or is the boolean sufficient?

scripts/lib/sphinx/addFrontMatter.ts
```diff
@@ -25,6 +30,14 @@ python_api_type: ${result.meta.python_api_type}
 python_api_name: ${result.meta.python_api_name}
 ---
 
+${markdown}
+`;
+    } else if (result.isReleaseNotes) {
+      result.markdown = `---
+title: Release notes
+description: ${pkg.title} v${packageVersion} release notes
```
Eric-Arellano, line 38:
```suggestion
title: ${pkg.title} release notes
description: Changes made to ${pkg.title}
```

scripts/lib/sphinx/generateToc.ts
```diff
@@ -32,7 +32,7 @@ export function generateToc(options: {
     title: string;
     name: string;
     version: string;
-    changelogUrl: string;
+    releaseNotesUrl?: string;
```
Eric-Arellano, line 35: Why `?`? You always set it in the script.

scripts/lib/sphinx/generateToc.ts
```diff
@@ -125,10 +125,12 @@ export function generateToc(options: {
     tocChildren.push(...orderEntriesByTitle(nestedTocModules));
   }
 
-  tocChildren.push({
-    title: "Changelog",
-    url: pkg.changelogUrl,
-  });
+  if (pkg.releaseNotesUrl) {
+    tocChildren.push({
+      title: pkg.name === "qiskit" ? "Changelog" : "Release notes",
```
Eric-Arellano, line 130: Nice. Thanks

scripts/lib/sphinx/sphinxHtmlToMarkdown.test.ts
Eric-Arellano: Yeah, I don't think it should...Does debugging when running the test show anything useful? Like console logging it.

Eric-Arellano (commented):
This is looking great.

scripts/commands/updateApiDocs.ts
```diff
@@ -272,7 +278,10 @@ async function convertHtmlToMarkdown(
       title: pkg.title,
       name: pkg.name,
       version,
-      changelogUrl: `https://github.com/${pkg.githubSlug}/releases`,
+      releaseNotesUrl:
+        pkg.name !== "qiskit"
+          ? `/api/${pkg.name}/release_notes`
```
Eric-Arellano, line 283: This is super nitty: can we call it `release-notes`? We've been using `-` for URL and file names in this project.

scripts/lib/sphinx/generateToc.ts
frankharkins: Fixed, thanks!

scripts/commands/updateApiDocs.ts
frankharkins: It does run, but I've moved it to `lib/sharedTypes.ts`

scripts/lib/sphinx/sphinxHtmlToMarkdown.test.ts
frankharkins: It's working now; turns out the html needs to start with `<div role="main">`.

Eric-Arellano (approved):
🎉

PR #315, opened 2023-11-09
Remove manual table of contents from migration guides

Eric-Arellano:
We decided these weren't necessary since the right sidebar (page table of contents) already has this functionality.

Removing it makes the guides a little smaller, so easier to read.


abbycross (approved):
👍

Issue #301, opened 2023-11-08
Change migration guides to use stable API links


Eric-Arellano:
Some of the APIs are being removed, so the "current" version of API docs will break. Switch those to use the stable version.

Issue #281, opened 2023-11-03
Add historical release notes for Qiskit API


Eric-Arellano:
For Qiskit, it historically has a single release note file for everything: https://qiskit.org/documentation/release_notes.html. But we probably want to migrate to one file per version. It's way too big otherwise.

Where should the release note live in the UI? 

Once we have a design, we need to:

1. Port the original release note file from RST to MDX. This will take forever..I'm not sure how we want to pull it off. We may want to instead not include a ton of historical release notes? Point to the RST file?
2. Figure out how Qiskit will do release notes in the future, if they keep using Reno. It's been working well for them and I suspect they'll want to keep using Reno. But how do we integrate that into MDX?

frankharkins:
@Eric-Arellano did we make any decisions on where it'll live in the UI?

Eric-Arellano:
>  did we make any decisions on where it'll live in the UI?

Yes, it will be a simple top-level folder in the left sidebar. You'll replace this here in the `_toc.json` (by changing the generation script):

https://github.com/Qiskit/documentation/blob/a36c9ba94e7b582eb196049165432ce9ba828c00/docs/api/qiskit/_toc.json#L3340-L3344

It should be a folder called "Release notes" and one page per minor release, e.g. "0.44".

Note that for Runtime and Provider, it instead will be a file called Release Notes, since they are so much smaller. https://github.com/Qiskit/documentation/issues/282

PR #329, opened 2023-11-10
Update Qiskit API docs to 0.45

Eric-Arellano:
We'll add back Qiskit 0.44 as part of https://github.com/Qiskit/documentation/issues/281.

This PR switches us to the new version selector mechanism, per https://github.com/Qiskit/documentation/pull/324.

We have to delete from the migration guide the link to 2 APIs. This will be restored when we add the 0.44 historical API docs. Tracked by https://github.com/Qiskit/documentation/issues/301.


javabster (approved):
LGTM 🚀

PR #336, opened 2023-11-12
Link checker: anchor's validation

### LINKED ISSUE #306: Link checker: check internal anchors
See https://github.com/Qiskit/documentation/issues/4 for the design.

arnaucasau:
### Summary

This PR adds support for the validation of anchors using the link checker and fixes some broken anchors. Tests for the validation of the anchors have been included too.

Closes #306


scripts/commands/checkLinks.ts
```diff
@@ -63,8 +63,15 @@ async function loadFilesAndLinks(
       path.extname(filePath) === ".ipynb"
         ? markdownFromNotebook(source)
         : source;
+    const anchors = markdownLinkExtractor(markdown).anchors;
 
-    fileList.push(new File(filePath, []));
+    // Get the anchors from HTML ids.
+    const id_anchors = markdown.match(/(?<=id=")(.*)(?=")/gm);
+    if (id_anchors != null) {
+      id_anchors.map((id) => anchors.push("#" + id));
```
Eric-Arellano, line 71: `map` is intended for returning a new list

```suggestion
      id_anchors.forEach((id) => anchors.push("#" + id));
```

scripts/commands/checkLinks.ts
```diff
@@ -132,7 +139,8 @@ async function loadFilesAndLinks(
 function loadFiles(existingPaths: string[]): File[] {
   const fileList: File[] = [];
   for (let path of existingPaths) {
-    fileList.push(new File(path, []));
+    const synthetic = SYNTHETIC_FILES.includes(path) ? true : false;
+    fileList.push(new File(path, [], synthetic));
```
Eric-Arellano, line 143: Hm, I'm realizing this function doesn't make a ton of sense to me. It creates `File` objects, but they're wrong! They are claiming there are no anchors when some of them do have it.

We're not mutating `File` objects with something like `setAnchors`, which is great. So, these `File` objects aren't very useful.

What would it look like to remove this function?

--

One idea to explore is making `SYNTHETIC_FILES` be a `File[]` rather than `string[]`. You could do this:

```typescript
["my-file1", "file2"].map((fp) => new File(fp, [], true))
```

Eric-Arellano (commented):
Awesome!

scripts/commands/checkLinks.ts
arnaucasau: I like the idea of stopping using a function to create "false" files. I have changed the code to use your suggestion, creating an array of extra files once we have parsed the files we want (line 144).

scripts/commands/checkLinks.ts
```diff
@@ -37,7 +37,7 @@ function markdownFromNotebook(source: string): string {
   let markdown = "";
   for (let cell of JSON.parse(source).cells) {
     if (cell.cell_type === "markdown") {
-      markdown += cell.source;
+      cell.source.map((s: string) => (markdown += s + "\n"));
```
Eric-Arellano, line 40:
```suggestion
      cell.source.forEach((s: string) => (markdown += s + "\n"));
```

Eric-Arellano (approved):
Nice! Please ask Abby C or Becky to verify any link fixes you're not sure about.

arnaucasau:
Hello @abbycross and @beckykd! In this PR, I added support for the link checker to validate the links with anchors, and I found some broken links that I had to change, but I'm not sure I have chosen the more accurate replacements. If you have time, can you give me your opinion, please?

The first one is a non-existing anchor in `docs/run/advanced-runtime-options.mdx` pointing to the same page using the `#skip-transpilation` anchor. Given that `skip_transpilation=True` is an option of the Qiskit transpiler, I changed it for the reference in `docs/api/qiskit-ibm-runtime/qiskit_ibm_runtime.options.TranspilationOptions#skip_transpilation`.

https://github.com/Qiskit/documentation/blob/34ecdbc66fccb25f2c3b30aae1b3703849a759f4/docs/run/advanced-runtime-options.mdx#L108-L110

The second one is a link in `docs/run/run-jobs-in-session.mdx`, redirecting to `docs/run/sessions.mdx` to an anchor called `#active` that doesn't exist either. For the context in the text, I changed it to point to the same file but with the anchor `#the-mechanics-of-sessions-queuing`. On the session page, there are several possible anchors for that one, and I wasn't sure that's the best.

https://github.com/Qiskit/documentation/blob/34ecdbc66fccb25f2c3b30aae1b3703849a759f4/docs/run/run-jobs-in-session.mdx#L112

What do you think? Thanks in advance for looking into this!

abbycross:
Hi @arnaucasau and thanks for your work on this! I'm looking into the history of those links. For the second one, it looks to me like you've chosen the correct place to link to. Looks like a couple sections got combined and the anchor was deleted in the process. @beckykd can probably confirm.

For the first one, I see that the anchor and its related section was deleted [in Blake's PR here](https://github.com/Qiskit/documentation/commit/bad3ad384d1ac9ab72ca7c57bb3feee5f6237e23#diff-90f07cdd079b60b7be1493c4b3e43f9681f0e504d327fe0ea6bfaece790fd975L141), so it looks to me like pointing to the API refs is the right thing to do, though it doesn't give much helpful context. Not sure why the code snippet was removed -- @javabster do you have any insights?

Eric-Arellano (approved):
I agree with Abby C's assessment. These anchors look good. Thanks, Arnau!

javabster:
yep I think linking to the API ref is the right call here 👍 that skip transpilation section was removed in Blake's PR but it is still called out in one of the noteboxes in the runtime compilation page which doesn't have an anchor tag, so i think the api ref is the best alternative.

arnaucasau:
Great! Thanks a lot, @abbycross, @javabster, and @Eric-Arellano for your help!

PR #338, opened 2023-11-12
Link checker: ignore mechanism

### LINKED ISSUE #243: Improve link checker ignore mechanism
Right now, we skip checking the file entirely if it's in the ignore list:

https://github.com/Qiskit/documentation/blob/bc54ab4b0a26f8c247afb0bf5d8fecdb1c3f239c/scripts/commands/checkLinks.ts#L21-L22

https://github.com/Qiskit/documentation/blob/bc54ab4b0a26f8c247afb0bf5d8fecdb1c3f239c/scripts/commands/checkLinks.ts#L136-L138

This is not a good approach because it's too coarse-grained. Even if a file has 1-2 bad links we need to skip, we want to still check the rest of the file.

This should be converted to a map between files to their ignored links, like this:

```typescript
ALLOW_LIST = {
  "my_file.mdx": ["/link1", "/link2"]
}
```

arnaucasau:
### Summary

This PR improves the ignore mechanism of the link checker, allowing us to skip the validation of individual links inside a file. To ignore a link, we can modify the dictionary of line 27. For example:

```
const IGNORED_FILES: { [id: string]: string[] } = {
         'docs/example': ['link1', 'link2'],
};
```

In the previous example, the link checker will check every link in `docs/example` but `link1` and `link2`. Moreover, we can continue ignoring all the links in a file using an asterisk (*). For example:

```
const IGNORED_FILES: { [id: string]: string[] } = {
         'docs/example': ['*'],
};
```

Closes #243


scripts/commands/checkLinks.ts
```diff
@@ -23,8 +23,8 @@ import rehypeRemark from "rehype-remark";
 import rehypeParse from "rehype-parse";
 import remarkGfm from "remark-gfm";
 
-// These files are not searched to see if their own links are valid.
-const IGNORED_FILES: string[] = [];
+// The links in a file are not searched to see if they are valid.
+const IGNORED_FILES: { [id: string]: string[] } = {};
```
Eric-Arellano, line 27: What about calling it `FILES_TO_IGNORES`? It should make it easier to understand call sites of this variable.

Eric-Arellano (commented):
Nice! I like the `"*"` idea. 

You should mention that in the comment on line 27. Those are instructions to developers for how to add an ignore.

scripts/commands/checkLinks.ts
arnaucasau: You are right, thanks!

Eric-Arellano (approved):

PR #340, opened 2023-11-13
Behavior change when root job is canceled

### LINKED ISSUE #190: Behavior when root job is canceled before session activates
### URL, if applicable

These seem applicable (the former links to the latter so maybe only documenting it in one of those places is necessary).

https://qiskit.org/ecosystem/ibm-runtime/sessions.html#what-happens-when-a-session-ends

https://qiskit.org/ecosystem/ibm-runtime/how_to/run_session.html#close-a-session

For the ibm_cloud pay-go channel, the scheduler is going to be making a behavior change [1] such that if the root/first job in a session is canceled before it runs and activates the session, the session will be closed and any remaining queued jobs in that session will be marked as `Failed`. This is similar to what happens today if a user explicitly closes their session before all jobs are run or if the session closes due to the max time expiring and there are still remaining queued jobs in the session.

Similarly, there are plans to make the same behavior change in the ibm_quantum (IQP) channel [2].

We hope to have both changes done in October but let's hold off on publishing docs around this until we have the changes in production for both cloud pay-go and IQP channels.

[1] https://github.ibm.com/IBM-Q-Software/multi-channel-scheduler/pull/1267
[2] https://github.ibm.com/IBM-Q-Software/sw-api-queue/issues/3163

abbycross:
Closes #190.

### DIFF
--- a/docs/run/run-jobs-in-session.mdx
+++ b/docs/run/run-jobs-in-session.mdx
@@ -148,6 +148,10 @@ print(f"Result2: {job2.result()}")
 session.close()
 ```
 
+<Admonition type="warning" title="When the root job is canceled">
+Note that when you cancel the root job in the session (the job which has the same ID as the session), the session closes and fails any remaining queued jobs in the session.
+</Admonition>
+
 <span id="cancel"></span>
 ## Cancel a session 
 


abbycross:
@mriedem I am adding an admonition note to the Close sessions section, and I am not removing any other content since I don't see any current info on canceling root jobs -- please correct me if I overlooked something! ([see preview here](https://qiskit-docs-preview-pr-340.1799mxdls7qz.us-south.codeengine.appdomain.cloud/run/run-jobs-in-session#close-a-session))

PR #348, opened 2023-11-14
Link checker: did you mean feature

arnaucasau:
### Summary

This PR adds a suggestion of a valid link for every broken link that the link checker finds. This feature only works with internal links and anchors.

To calculate the best replacement for a broken link, the function `didYouMean` in `scripts/lib/LinkChecker.ts` uses the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) between the broken link and all the files we have in docs, as it was originally suggested in #4.

The package used to implement the Levenshtein distance calculation is [fast-levenshtein](https://www.npmjs.com/package/fast-levenshtein).


scripts/lib/LinkChecker.ts
```diff
@@ -117,6 +117,52 @@ export class Link {
     );
   }
 
+  /**
+   * Returns a string with a suggested replacement for a broken link
+   */
+  didYouMean(existingFiles: File[], originFile: string): string {
+    const levenshtein = require("fast-levenshtein");
```
Eric-Arellano, line 124: This should be an ES6 import, like the rest of our imports. `require` is the old way of doing imports.

scripts/lib/LinkChecker.ts
```diff
@@ -117,6 +117,52 @@ export class Link {
...
+    let min_score = Number.MAX_SAFE_INTEGER;
+    let suggestion_path: String = "";
+    let suggestion_path_anchors: String[] = [];
+
+    const possiblePaths = this.possibleFilePaths(originFile);
+    const pathNoExtension = possiblePaths[0].replace(/\.[^\/.]+$/, "");
+
+    existingFiles.forEach((file) => {
+      let score = levenshtein.get(pathNoExtension, file.path);
+      if (score < min_score) {
```
Eric-Arellano, line 136: Does this mean we always no matter what give a suggestion, since MAX_SAFE_INTEGER is so high? Generally people recommend only offering a suggestion if you're within a certain distance. Otherwise the suggestion ends up being not at all relevant and is noisy.

scripts/lib/LinkChecker.ts
```diff
@@ -117,6 +117,52 @@ export class Link {
...
+    });
+
+    if (this.anchor == "") {
+      return `❓ Did you mean '${suggestion_path
+        .replace(/\.[^\/.]+$/, "")
+        .replace(/^docs/, "")}'?`;
+    }
+
+    // Find a new valid anchor
+    min_score = Number.MAX_SAFE_INTEGER;
```
Eric-Arellano, line 150: Does it make sense to use a new variable for clarity?

scripts/lib/LinkChecker.ts
```diff
@@ -117,6 +117,52 @@ export class Link {
...
   }
 
+  /**
+   * Returns a string with a suggested replacement for a broken link
+   */
+  didYouMean(existingFiles: File[], originFile: string): string {
+    const levenshtein = require("fast-levenshtein");
+
+    // Find a new valid link
+    let min_score = Number.MAX_SAFE_INTEGER;
```
Eric-Arellano, line 127: In JavaScript, the convention is to use `camelCase`. Ditto the other variables

Eric-Arellano (commented):
Exciting!

scripts/lib/LinkChecker.ts
arnaucasau: Done!

scripts/lib/LinkChecker.ts
arnaucasau: Done!

scripts/lib/LinkChecker.ts
arnaucasau: Yes, maybe it's more readable, done!

scripts/lib/LinkChecker.ts
arnaucasau: Hmm.. you are right! We shouldn't suggest a link that it's not close to the broken one. Instead of modifying MAX_SAFE_INTEGER, I normalized the final score between 0 and 1, to check that we have at least 0.5 of similarity. We can change the 0.5 using the MIN_SIMILARITY constant at the beginning. What do you think?

scripts/lib/LinkChecker.test.ts
```diff
@@ -74,7 +74,7 @@ describe("Validate links", () => {
     let testFile = new File("docs/testpath.mdx", [], false);
     const results = await testLink.checkLink([testFile]);
     expect(results).toEqual([
-      "❌ /testorigin.mdx: Could not find link '/test-alternative-path'",
+      "❌ /testorigin.mdx: Could not find link '/test-alternative-path'.",
```
Eric-Arellano, line 77: Nit: I think it's better to not have the period. It's kinda noisy and makes copying and pasting more annoying

Eric-Arellano (approved):
Nice! While this does add some complexity, you did a good job encapsulating it into a single method.

scripts/lib/LinkChecker.test.ts
arnaucasau: Done!

PR #351, opened 2023-11-14
show how to initialize noise model from hardware backend

kevinsung:
Update "Building noise models" page to include example of initializing noise model from backend.


docs/verify/building_noise_models.ipynb
javabster, line 1: This is a great addition to this page!

could you remove the "Introduction" heading on this page? I know it was there before but I just realised it's not consistent with what we do on other pages. We can launch straight into the paragraph after the title, we don't need the extra "Introduction" heading.

Also for the intro sentence do you think it would be worth adding a sentence about why a user might want to use noise models in the first place?

kevinsung: Done.

javabster (approved):
LGTM 🚀

PR #346, opened 2023-11-14
Handle legacy release notes for qiskit package

frankharkins:
This PR adds release note files for Qiskit v0.44 and lower, and modifies the script to include these files in the TOC.

I used the API generation script to convert [`legacy_release_notes.html`](https://qiskit.org/documentation/legacy_release_notes.html) to markdown, then used the following script to split the markdown file into separate files. This was a one-time process.

<details>
 <summary>Script to convert legacy release notes</summary>

```python
with open("legacy_release_notes.md") as f:
    sections = f.read().split("\n## ")[2:]

sections_dict = {
   # E.g. "0.18.2": markdown_string
   s.split()[1]: f"## {s.split()[1]}\n" + s.split("\n", 1)[1]
   for s in sections
   }

compiled_sections = {}
for patch_version, markdown in sorted(sections_dict.items(), key=lambda x: x[0]):
    print(patch_version)
    minor_version = ".".join(patch_version.split(".")[:2])
    if minor_version not in compiled_sections:
        compiled_sections[minor_version] = f"""---
title: Qiskit {minor_version} release notes
description: New features and bug fixes
---

# Qiskit {minor_version} release notes

"""
    compiled_sections[minor_version] += markdown

for version, markdown in compiled_sections.items():
    with open(f"release-notes/{version}.md", "w") as f:
        f.write(markdown)
```

</details>

This PR also modifies `updateApiDocs.ts` to check for files in `release-notes/` and add them to the TOC.

This PR does not pull in new release notes, that will be covered in #352.

PR #352, opened 2023-11-14
Support new Qiskit release notes

### LINKED ISSUE #281: Add historical release notes for Qiskit API

### LINKED ISSUE #354: Qiskit release notes should have an index page
We realized it's important that `/api/qiskit/release-notes` works and you don't have to go to a dedicated page like `/release-notes/0.45`. That is useful for our redirect logic (https://github.com/Qiskit/documentation/issues/216) and in communication like blogs for us to have a link that always works.

The page should simply be a list of each of the versions we have and link to them. Like duplicating the left sidebar. 

--

Maybe awkwardly, this will be an orphan page, meaning you can only access it from the URL, and not the left side bar. This doesn't work in `_toc.json`:

```json
      "title": "Release Notes",
      "url": "/api/qiskit/release-notes/index",
      "children": [
       ]
```

Abby M and I agreed that's fine. It's still useful to have the stable URL `/release-notes`. We want to be careful to not add meaningful content to the page though, since it's not very discoverable.

frankharkins:
Continuation of #346.

---

Closes #281, closes #354


docs/api/qiskit-ibm-provider/release-notes.md
```diff
@@ -4,7 +4,9 @@ description: Changes made to Qiskit IBM Provider
 in_page_toc_max_heading_level: 2
 ---
 
-# Release Notes
+<span id="release-notes" />
```
frankharkins, line 7: Not sure why this is but it seems harmless

scripts/commands/updateApiDocs.ts
```diff
@@ -323,6 +343,21 @@ async function convertHtmlToMarkdown(
     JSON.stringify(toc, null, 2) + "\n",
   );
 
+  if (pkg.hasSeparateReleaseNotes) {
```
Eric-Arellano, line 346: There's now a lot of release notes logic in this file and the file is already pretty large. Thoughts on creating a new `lib/releaseNotes.ts` file?

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -64,6 +66,10 @@ export async function sphinxHtmlToMarkdown(options: {
     }
   });
 
+  if (isReleaseNotes && releaseNotesTitle) {
+    $page("h1").html(releaseNotesTitle);
+  }
```
Eric-Arellano, line 71: What does this do?

Eric-Arellano (approved):
💜

scripts/commands/updateApiDocs.ts
frankharkins: That's a good idea, I'll merge but an issue for it after summit.

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
frankharkins: This replaces the title "Release Notes" with something like "Qiskit 0.45 release notes", I'll add a comment.

Issue #339, opened 2023-11-13
Fix index pages for Provider and Runtime


Eric-Arellano:
https://docs.quantum-computing.ibm.com/api/qiskit-ibm-provider fails because it expects /api/qiskit-ibm-provider/ibm_provider

Ditto https://docs.quantum-computing.ibm.com/api/qiskit-ibm-runtime because it expects /api/qiskit-ibm-runtime/runtime_service

Qiskit is fine.

--

Note that this is because the URL on qiskit.org has these file names: https://qiskit.org/ecosystem/ibm-provider/apidocs/ibm-provider.html. We need to special case renaming them.

--

I'm not sure if we need to set up redirects for the old pages.

PR #355, opened 2023-11-14
Fix index pages for Runtime and Provider API docs

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/339.

It wasn't intentional that we didn't have `index.md` files and instead used `/runtime_service` and `/ibm_provider`. It was only because that's what the original qiskit.org docs do, and I don't think we realized that was bad and we should fix it.

This PR does mean `/runtime_service` and `/ibm_provider` will now 404. I couldn't find any references to those pages outside of our own docs, so it didn't seem necessary to set up redirects. But I can if we think it's important.


docs/api/qiskit-ibm-provider/_toc.json
```diff
@@ -6,7 +6,7 @@
       "children": [
         {
           "title": "Overview",
-          "url": "/api/qiskit-ibm-provider/ibm_provider"
+          "url": "/api/qiskit-ibm-provider/index"
```
frankharkins, line 9: Should we maybe link to just `/api/qiskit-ibm-provider/` (i.e. without `/index`)?

This is a bikeshed comment though so just go with whatever you prefer.

frankharkins (commented):
Looks good! Are you also updating the links in the nav bar?

Screenshot 2023-11-15 at 15 04 40

docs/api/qiskit-ibm-provider/_toc.json
Eric-Arellano: Eh, couldn't figure out quickly how to do this so I'm going to punt for now: https://github.com/Qiskit/documentation/issues/361. It's behaving this way for Qiskit too

Eric-Arellano (commented):
Yes to updating the top nav bar and home page when syncing these docs to go live. Good catch!

frankharkins (approved):
Nice, thanks!

PR #360, opened 2023-11-15
Rename 'qubit-order' page

frankharkins:
Just a suggestion, but I think we should rename this so the URL matches the title.


Eric-Arellano (commented):
I'm personally a +1. Either way, we need to do this _before_ redirects take place Nov 29. Thoughts @javabster?

javabster (approved):
LGTM 🚀

PR #362, opened 2023-11-15
Fix URL for operators overview page

Eric-Arellano:
URLs should use `-`, not `_`. We need to fix this before setting up URL redirects.


abbycross (approved):
lgtm!

Issue #332, opened 2023-11-10
Allow converting current release to historical


Eric-Arellano:
See https://github.com/Qiskit/documentation/issues/8 for the overall scheme.

Before we switch to a new version of API docs, we need to convert the previous one into a historical release with the subfolder like `0.45`. This should be automated. We need to make sure all links are updated properly, including images.

This issue is a must-have for Runtime and Provider, since they don't publish more than one documentation version to qiskit.org. So, we have no way to get the historical values of those APIs. Once we have https://github.com/Qiskit/documentation/issues/311, we could technically regenerate the API docs, though, by pointing to an old version of CI. But I suspect this script will still be useful, e.g. if the CI artifact no longer exists.

--

We should test that this works by adding back Runtime 0.14 docs from before https://github.com/Qiskit/documentation/issues/332 was merged.

Issue #330, opened 2023-11-10
Historical API versions should automatically be updated with release notes


Eric-Arellano:
For example, when we update `qiskit/release-notes`, we need to also update `qiskit/0.44/release-notes`. That is, all versions should be using the most recent release notes.

See https://github.com/Qiskit/documentation/issues/8 for how the versions are laid out.

Issue #8, opened 2023-09-06
Figure out URL and folder structure for API versioning


Eric-Arellano:
## Background

Part of https://github.com/Qiskit/documentation/issues/77. We need historical versions for the API docs, like Qiskit 0.24 vs 0.25. Distinct versions will be for minor releases, but not patch releases like `0.24.1`.

The non-API docs do not have versioning per-API. Instead, https://github.com/Qiskit/documentation/issues/133 tracks how we will indicate in non-API docs their version compatibility. This ticket is focused on API docs.

### Relationship to translations

API docs have not ever been historically translated and we don't plan to in the future, per @SooluThomas. 

So, while the folder structure we use for translations (https://github.com/Qiskit/docs/issues/7) is relevant, it's somewhat orthogonal.

## Plan

`/api/qiskit/<version>/`, e.g. `/api/qiskit/0.25/circuit`

This URL scheme corresponds to the folder structure, that we'll have a folder `api/qiskit/0.25` etc. That folder will be self-contained, e.g. with its own `_toc.json`. When in the frontend the user changes the currently selected version, we'll switch to rendering that folder.

### Cross-references

We need to update cross-references for stable version folders so that they all link to files in the same version folder, i.e. `0.25` files all cross-link to `0.25` files. 

This will be automated by our API generation script.

### Keep un-versioned API

We'll also continue having an un-versioned "latest" URL, like `/api/qiskit/circuit`, so that people can link to latest.

When generating the API docs for the "latest", it will update both the stable folder and the unstable folders so that they're in sync.

None of our API pages start with a number—which would be a violation of Python naming rules—so we shouldn't have a clash between the version folder and the unversioned docs.

Non-API docs will link to the un-versioned docs in most cases, other than specific edge cases like how `qiskit.algorithms` is being deleted, so the Algorithms migration guide must link to the historical file.

Eric-Arellano:
@axelhzf thought through how best to integrate this with the web app. Each folder, including the top-level one for current, will have a `_package.json` file, similar to `_toc.json`. it will simply be this:

```json
{
  "name": "qiskit",
  "version": "0.25.0"
}
```

Folder structure:

```
qiskit
├── 0.24.0
│   ├── _package.json
│   ├── _toc.json
│   └── ...  // content files for this version
├── _package.json
├── _toc.json
└── ... // content files for the latest version
```

[...]

Issue #370, opened 2023-11-17
Split release notes logic out into separate file


frankharkins:
We should spit the code that handles release notes generation into its own file.

---

> There's now a lot of release notes logic in this file and the file is already pretty large. Thoughts on creating a new `lib/releaseNotes.ts` file?

_Originally posted by @Eric-Arellano in https://github.com/Qiskit/documentation/pull/352#discussion_r1397592377_

PR #377, opened 2023-11-20
Allow generating historical API versions

Eric-Arellano:
First part of https://github.com/Qiskit/documentation/issues/332. A follow up will generate historical versions prior to Qiskit 0.44; this PR is focused on the general infrastructure.

See https://github.com/Qiskit/documentation/issues/8 for the overall of the folder structure.

Users can now run `npm run gen-api -- -p qiskit -v 0.44.0 --historical`.

Runtime and Provider aren't supported since they need https://github.com/Qiskit/documentation/issues/332.

## How links work

We still use relative links for API cross-references. That works because the links will be relative to the current directory, whether that is `0.44/` or the unversioned top-level folder.

However, images use absolute links. We decided to have dedicated folders for images for each API, rather than sharing a common folder. That is because images may change throughout API versions and we don't want to risk merge conflicts.

## Release notes

Per https://github.com/Qiskit/documentation/issues/330, every API version should have identical release notes. We decided that's desirable UI. 

So, this PR copies the top-level release notes when adding a new historical version. That implements the first part of #330; now we only need a script that validates the release notes are all in sync, and also a way to have updates to the top-level release notes automatically update the subfolders.


scripts/commands/updateApiDocs.ts
```diff
@@ -278,9 +316,18 @@ async function convertHtmlToMarkdown(
...
       baseSourceUrl,
-      imageDestination: `/images/api/${pkg.name}`,
+      imageDestination: historical
+        ? `/images/api/${pkg.name}/${versionWithoutPatch}`
+        : `/images/api/${pkg.name}`,
       releaseNotesTitle: `${pkg.title} ${versionWithoutPatch} release notes`,
     });
+
+    result.markdown = transformLinks(result.markdown, (link, _) =>
+      /^\/images.*/.test(link)
```
arnaucasau, line 326: This line should take into account external links and whether we are in the historical version's case or not.

scripts/commands/updateApiDocs.ts
```diff
@@ -207,9 +226,27 @@ zxMain(async () => {
...
+ */
+async function rmFilesInFolder(
+  dir: string,
+  description: string,
+  historical: boolean,
+): Promise<void> {
+  console.log(`Deleting existing markdown for ${description}`);
+  historical
+    ? await $`find ${dir}/* -not -path "*release-notes*" | xargs rm -rf {}`
+    : await $`find ${dir}/* -not -path "*release-notes*" -not -path '*/[0-9]*' | xargs rm -rf {}`;
```
Eric-Arellano, line 247: This is safe to run on folders without historical versions, right? We're going to be adding historical versions soon to runtime and provider - I think it's fine to always use this code. Simpler and more future proof.

scripts/commands/updateApiDocs.ts
```diff
@@ -343,7 +390,7 @@ async function convertHtmlToMarkdown(
     JSON.stringify(toc, null, 2) + "\n",
   );
 
-  if (pkg.hasSeparateReleaseNotes) {
+  if (pkg.hasSeparateReleaseNotes && !historical) {
```
Eric-Arellano, line 393: Instead, we want to delete the generated release note file and replace with whatever the prior notes were: https://github.com/Qiskit/documentation/issues/330

Might want to refactor the code first: https://github.com/Qiskit/documentation/issues/370

You can punt on this in this PR by removing release notes entirely for now. But that may be more complex because you have to make sure `_toc.json` is correct.

scripts/commands/updateApiDocs.ts
```diff
@@ -278,9 +316,18 @@ async function convertHtmlToMarkdown(
...
       baseSourceUrl,
-      imageDestination: `/images/api/${pkg.name}`,
+      imageDestination: historical
+        ? `/images/api/${pkg.name}/${versionWithoutPatch}`
+        : `/images/api/${pkg.name}`,
       releaseNotesTitle: `${pkg.title} ${versionWithoutPatch} release notes`,
     });
+
+    result.markdown = transformLinks(result.markdown, (link, _) =>
+      !historical || /^\/images.*/.test(link) || !link.startsWith("http")
```
Eric-Arellano, line 326: What links are you trying to redirect and what are you not? For example, I think you might want `&& ~link.startswith("http")`, since those will never change; you're using `||`.

I recommend extracting this out to a `lib/` file and writing tests. It's important we get this right and don't break it.

scripts/commands/updateApiDocs.ts
```diff
@@ -369,6 +420,7 @@ async function convertHtmlToMarkdown(
   await downloadImages(
     allImages.map((img) => ({
       src: img.src,
+      // TODO: probably figure out, including the links used
```
Eric-Arellano, line 423: Leftover

scripts/commands/updateApiDocs.ts
arnaucasau: You are right, my intention was using the `||` but without the negation in front `link.startsWith("http")`.

scripts/commands/updateApiDocs.ts
arnaucasau: I changed the code to continue creating the `index.md`, delete the generated release note files and replace them with a copy of the non-historical version release notes. Also, the `_toc.json` contains the new path to each release note in its subfolder.

scripts/commands/updateApiDocs.ts
arnaucasau: After our conversation, I removed this code because we don't have any absolute paths in the API docs, and therefore, this is not necessary anymore.

scripts/commands/updateApiDocs.ts
```diff
@@ -358,6 +410,11 @@ async function convertHtmlToMarkdown(
     await writeFile(`${markdownPath}/release-notes/index.md`, markdown);
   }
 
+  if (historical) {
+    await $`find ${markdownPath}/release-notes/* -not -path "*index.md" | xargs rm -rf {}`;
+    await $`find docs/api/${pkg.name}/release-notes/* -not -path "*index.md" | xargs -I {} cp -a {} docs/api/${pkg.name}/${versionWithoutPatch}/release-notes/`;
+  }
```
Eric-Arellano, line 416: Nit: probably move this to a dedicated helper function like `copyReleaseNotes`

Eric-Arellano (commented):
Awesome! The only issue I noticed is this changing `release-notes/0.44.md`. We don't want that file to change at all.

Also it would be good to run this script on Qiskit 0.45 (not historical) and ensure it doesn't change anything. Like no regressions.

scripts/commands/updateApiDocs.ts
arnaucasau: Oh, I see. We cannot run the same command in both versions because:

- Line 246 run in a non-historical version, will remove all the historical folders.
- Line 247 run in a historical version won't remove anything because "-not -path '*/[0-9]*'" will be true always.

I can investigate if there's another method to use only one command.

Eric-Arellano: Ah. My original thinking was only to remove the _files_ in the folder and don't remove any subfolders. That should handle all the edge cases, including not deleting release notes.

arnaucasau: That's a good idea! I replaced the two lines for `find ${dir}/* -maxdepth 0 -type f | xargs rm -f {}` where we only remove files in the same depth (not files in any subfolder).

scripts/commands/updateApiDocs.ts
```diff
@@ -189,12 +207,20 @@ zxMain(async () => {
...
 
-  const legacyReleaseNoteEntries = await processLegacyReleaseNotes(pkg);
-
-  console.log(`Deleting existing markdown for ${pkg.name}`);
-  await $`find ${outputDir}/* -not -path "*release-notes*" | xargs rm -rf {}`;
+  const legacyReleaseNoteEntries = await processLegacyReleaseNotes(
+    pkg,
+    versionWithoutPatch,
+    args.historical,
+  );
```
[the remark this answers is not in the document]
arnaucasau: Done!

scripts/commands/updateApiDocs.ts
arnaucasau: Done!

arnaucasau:
It's true we were modifying the release note file of the historical version in the non-historical folder. That's the reason for the changes in `release-notes/0.44.md`. It should be fixed now, thanks for your comments @Eric-Arellano  :+1:

arnaucasau:
It's true we were modifying the release note file of the historical version in the non-historical folder. That's the reason for the changes in `release-notes/0.44.md`. It should be fixed now, thanks for your comments @Eric-Arellano  :+1:

scripts/commands/updateApiDocs.ts
```diff
@@ -207,9 +234,25 @@ zxMain(async () => {
...
+/**
+ * Deletes all the files in the folder, but preserves subfolders. That's important
+ * to avoid accidentally deleting other historical versions of the API.
+ *
+ * We delete files when regenerating API docs to capture when APIs have been deleted.
+ */
+async function rmFilesInFolder(
+  dir: string,
+  description: string,
+  historical: boolean,
```
Eric-Arellano, line 250: You can remove this argument.

Eric-Arellano (commented):
Great work, Arnau! I can't approve this because it's my own PR haha. So you'll need to approve it yourself.

Please wait to merge until we have confirmed with the docs infrastructure team it's okay to add this many files.

I added some thoughts to the PR description. Please feel free to make any chanes for anything I left off!

arnaucasau (approved):

PR #381, opened 2023-11-20
Audited qiskit.org links

### LINKED ISSUE #297: Remove links to qiskit.org
We have a couple of these still, such as in the migration guides to tutorials on qiskit.org. Search for `qiskit.org`.

Some of these require fixing the original API docs.

Also deal with `qisk.it` links.

--

Update Nov 21, post https://github.com/Qiskit/documentation/pull/381, we only need to update API docs. Tracked by https://github.com/Qiskit/qiskit/issues/11289 and https://github.com/Qiskit/qiskit-ibm-runtime/issues/1235

javabster:
This PR replaces any qiskit.org links with 1XP or github equivalents. Need to wait to merge until I can update the deprecation policy and maintainer guide links after https://github.com/Qiskit/qiskit/pull/11218/ merges.

This should partially resolve #297 , however for the api links we'll need to make updates for those directly in the repos where the API refs live (maybe this is beyond the scope of the original issue?)

### DIFF
--- a/docs/api/migration-guides/qiskit-runtime.mdx
+++ b/docs/api/migration-guides/qiskit-runtime.mdx
@@ -470,7 +470,6 @@ an algorithm, you might want to tune certain primitive options. For details, see
     - [Get started with Sampler.](../../run/primitives-get-started#start-sampler)
     - Explore [sessions.](../../run/sessions)
     - [Run a primitive in a session.](../../run/run-jobs-in-session)
-    - Experiment with the [migration tutorial.](./qiskit-runtime) 
     - Experiment with the [Run custom transpiled circuits with primitives tutorial.](https://learning.quantum-computing.ibm.com/tutorial/submitting-user-transpiled-circuits-using-primitives)
 
 </Admonition>
--- a/docs/api/migration-guides/qiskit-runtime.mdx
+++ b/docs/api/migration-guides/qiskit-runtime.mdx
@@ -30,7 +30,7 @@ level.
 
 The Qiskit Runtime primitives implement the reference `Sampler` and
 `Estimator` interfaces found in
-[qiskit.primitives](https://qiskit.org/documentation/apidoc/primitives.html).
+[qiskit.primitives](../qiskit/primitives).
 These interfaces let you switch between primitive implementations with
 minimal code changes. Different primitive implementations can be found
 in the `qiskit`, `qiskit_aer`, and `qiskit_ibm_runtime` libraries. Each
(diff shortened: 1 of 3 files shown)


Eric-Arellano:
> however for the api links we'll need to make updates for those directly in the repos where the API refs live (maybe this is beyond the scope of the original issue?)

I intended it to be in scope. But we could create a follow up issue if we want?

javabster:
> I intended it to be in scope. But we could create a follow up issue if we want?

Yeah, like with other API updates I think first and foremost this should be the responsibility of the devs that create the API reference. At least for now as our workload is so high as is

Eric-Arellano:
> like with other API updates I think first and foremost this should be the responsibility of the devs that create the API reference. 

I'm fine with that in general in the future for long-term maintenance. But as the team responsible for the qiskit.org migration, I do think it's appropriate for our team to do. 

It should be fine if this slips from https://github.com/Qiskit/documentation/milestone/3 to https://github.com/Qiskit/documentation/milestone/5 thanks to redirects.

docs/api/migration-guides/qiskit-runtime.mdx
```diff
@@ -470,7 +470,7 @@ an algorithm, you might want to tune certain primitive options. For details, see
     - [Get started with Sampler.](../../run/primitives-get-started#start-sampler)
     - Explore [sessions.](../../run/sessions)
     - [Run a primitive in a session.](../../run/run-jobs-in-session)
-    - Experiment with the [migration tutorial.](https://qiskit.org/ecosystem/ibm-provider/tutorials/Migration_Guide_from_qiskit-ibmq-provider.html) 
+    - Experiment with the [migration tutorial.](./qiskit-runtime)
```
Eric-Arellano, line 473: This is a reference to itself

Eric-Arellano (commented):
thank you!

docs/support.mdx
```diff
@@ -23,10 +23,10 @@ For help with Qiskit, access our Slack community: [Qiskit Slack](https://qisk.it
...
 The following pages are resources for anyone interested in contributing code to Qiskit.
 
-- [Code of conduct](https://github.com/Qiskit/qiskit-metapackage/blob/master/CODE_OF_CONDUCT.md)
-- [Contributing guide](https://qiskit.org/documentation/contributing_to_qiskit.html#contributing-links)
-- [Deprecation policy](https://qiskit.org/documentation/deprecation_policy.html)
-- [Maintainers guide](https://qiskit.org/documentation/maintainers_guide.html)
+- [Code of conduct](https://github.com/Qiskit/qiskit/blob/main/CODE_OF_CONDUCT.md)
+- [Contributing guide](https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md)
+- [Deprecation policy](https://github.com/Qiskit/qiskit/blob/main/DEPRECATION.md)
+- [Maintainers guide](hhttps://github.com/Qiskit/qiskit/blob/main/MAINTAINING.md)
```
Eric-Arellano, line 29:
```suggestion
- [Maintainers guide](https://github.com/Qiskit/qiskit/blob/main/MAINTAINING.md)
```

Eric-Arellano (approved):

PR #394, opened 2023-11-22
Use the minor version in _package.json

arnaucasau:
The PR changes the` _package.json` in every API to use only the minor version instead of the full version.

This is a temporary change to unblock the deployment of Qiskit 0.44, which, in the navigation menu, will only show the minor version without each patch following the directory structure created in https://github.com/Qiskit/documentation/pull/377.


Eric-Arellano (approved):
Thanks! Can you please explain in the PR description why we need to do this?

PR #396, opened 2023-11-22
Update Qiskit platform support for Qiskit 1.0

mtreinish:
As part of Qiskit 1.0 we've increased the minimum requirements and platform support tiers (see #10902). This commit makes these updates to the hosted documentation so that the documented support tiers for each platform reflects the current state for Qiskit 1.0.0.


Eric-Arellano (approved):
Thank you for updating here 💜

PR #400, opened 2023-11-23
Fix Qiskit release notes relative links 

### LINKED ISSUE #386: Qiskit release notes have broken relative links
The issue is that we're using relative links, like `qiskit.circuit.library.CRXGate`:

https://github.com/Qiskit/documentation/blob/f87d9763f8b3bdbe05505918c920f4a355aa328f/docs/api/qiskit/release-notes/0.45.md?plain=1#L147

This normally works well because all API docs are in the same folder. But it breaks with release notes because the Qiskit release notes are in a subfolder. It's not an issue with qiskit-ibm-provider and qiskit-ibm-runtime since `release-notes.md` is in the same folder.

We should probably fix this by using absolute links starting with `/api/qiskit/`. That will ensure that the release notes still work properly when copying the release notes into historical API version folders, like `qiskit/0.44/release-notes`: https://github.com/Qiskit/documentation/issues/330

--

Why did our link checker not find this issue?

arnaucasau:
This PR fixes the broken relative links of the Qiskit release notes. The links are originally valid, but because we have the release notes in a subfolder, we need to update the relative paths.

To update every link, I used the `updateApiDocs.ts` script, but added another function to read every link and add `/api/qiskit` in front of each relative path. This function was executed in line 220, followed by process.exit() to avoid doing anything else.

<details>
 <summary>Function to convert relative links</summary>

```typescript
async function updateLinksReleaseNotes(
  projectName: string,
  legacyReleaseNoteEntries: { title: string; url: string }[],
) {
  for (let entry of legacyReleaseNoteEntries) {
    let source = await readFile(
      `${getRoot()}/docs/api/${projectName}/release-notes/${entry.title}.md`,
      { encoding: "utf8" },
    );

    source = transformLinks(source, (link, _) =>
    link.startsWith("http") || link.startsWith("#") || link.startsWith("/")
      ? link
      : `/api/${projectName}/${link}`,
  );

    await writeFile(
      `${getRoot()}/docs/api/${projectName}/release-notes/${entry.title}.md`,
      source,
    );
  }
}
```
</details>

The PR still needs to modify the `updateApiDocs.ts` to update the relative links of the newest release notes we download when we run the script.

Closes #386


docs/api/qiskit/0.44/release-notes/0.40.md
```diff
@@ -21,39 +21,39 @@ This release officially deprecates the Qiskit IBMQ provider project as part of t
 
 Qiskit Terra 0.23.0 is a major feature release that includes a multitude of new features and bugfixes. The highlights for this release are:
 
-> *   Support for importing OpenQASM 3 programs and creating [`QuantumCircuit`](qiskit.circuit.QuantumCircuit "qiskit.circuit.QuantumCircuit") objects from the input program via two new functions [`qiskit.qasm3.load()`](qasm3#qiskit.qasm3.load "qiskit.qasm3.load") and [`qiskit.qasm3.loads()`](qasm3#qiskit.qasm3.loads "qiskit.qasm3.loads").
+> *   Support for importing OpenQASM 3 programs and creating [`QuantumCircuit`](/api/qiskit/qiskit.circuit.QuantumCircuit "qiskit.circuit.QuantumCircuit") objects from the input program via two new functions [`qiskit.qasm3.load()`](/api/qiskit/qasm3#qiskit.qasm3.load "qiskit.qasm3.load") and [`qiskit.qasm3.loads()`](/api/qiskit/qasm3#qiskit.qasm3.loads "qiskit.qasm3.loads").
```
frankharkins, line 24: The link
```markdown
[`QuantumCircuit`](/api/qiskit/qiskit.circuit.QuantumCircuit "qiskit.circuit.QuantumCircuit")
```

Takes me to

```
http://localhost:3000/api/qiskit/qiskit.circuit.QuantumCircuit
```

rather than

```
http://localhost:3000/api/qiskit/0.44/qiskit.circuit.QuantumCircuit
```

Is that the intended behaviour?

frankharkins (commented):
Nice work! I tried this out locally and every link I clicked worked. I just have one question.

docs/api/qiskit/0.44/release-notes/0.40.md
arnaucasau: Thanks for the review! Yes, we want all the release notes to be exactly the same, and with all the links pointing to the current API, regardless of what version of the API docs we are at that moment :+1:

frankharkins (approved):

Eric-Arellano (commented):
Great work! Thanks.

PR #402, opened 2023-11-23
Move local variables to object

frankharkins:
We're starting to have quite a few local variables that need passing between functions. This PR collects them into the `pkg` object.

This is in anticipation of some upcoming PRs.


scripts/lib/sphinx/generateToc.test.ts
frankharkins, line 1: This file has a large diff but it's mostly because of indentation. The `results` data should be the same. The basic pattern is:

```diff
-   const toc = generateToc({ pkg, results: [...] })
+   const toc = generateToc(pkg, [...])
```

scripts/commands/updateApiDocs.ts
```diff
@@ -178,23 +176,30 @@ zxMain(async () => {
...
+    pkg.baseUrl = `https://qiskit.org/documentation/stable/${pkg.versionWithoutPatch}`;
     pkg.initialUrls = [`${pkg.baseUrl}/apidoc/index.html`];
   }
 
   const destination = `${getRoot()}/.out/python/sources/${pkg.name}/${
-    args.version
+    pkg.version
   }`;
   if (await pathExists(destination)) {
     console.log(`Skip downloading sources for ${pkg.name}:${args.version}`);
```
arnaucasau, line 205: Nit: This line continues using `args` instead of `pkg`.
 
```suggestion
    console.log(`Skip downloading sources for ${pkg.name}:${pkg.version}`);
```

scripts/commands/updateApiDocs.ts
```diff
@@ -206,32 +211,19 @@ zxMain(async () => {
     });
   }
 
-  const baseSourceUrl = `https://github.com/${pkg.githubSlug}/tree/${versionWithoutPatch}/`;
+  const baseSourceUrl = `https://github.com/${pkg.githubSlug}/tree/${pkg.versionWithoutPatch}/`;
   const outputDir = args.historical
```
arnaucasau, line 215: Ditto here
```suggestion
  const outputDir = pkg.historical
```

arnaucasau (approved):
Awesome! It's a good idea to collect the local variables into the `pkg` object! Thank you for the refactor, LGTM! I left two little suggestions for two lines that are still using `args`, but the rest is great.

scripts/commands/updateApiDocs.ts
frankharkins: Good spots, thanks!

Eric-Arellano (approved):
Nice improvement!

PR #405, opened 2023-11-24
[WIP] Converting current API to historical

### LINKED ISSUE #332: Allow converting current release to historical

arnaucasau:
We need a method to convert the current API into a historical version, and for this purpose, we can use the new `convertApiDocsToHistorical.ts` script using the `make-historical` new command. This command expects an argument specifying the name of the API we want to switch over. For example:

> npm run make-historical -- -p qiskit

This feature is useful for `qiskit-ibm-runtime` and `qiskit-ibm-povider` since they don't publish more than one documentation version. Closes #332

### DIFF
--- a/scripts/commands/convertApiDocsToHistorical.ts
+++ b/scripts/commands/convertApiDocsToHistorical.ts
@@ -39,7 +39,7 @@ const readArgs = (): Arguments => {
       type: "string",
       choices: PACKAGES,
       demandOption: true,
-      description: "Which package to update",
+      description: "Which package to convert",
     })
     .parseSync();
 };
@@ -63,7 +63,7 @@ zxMain(async () => {
   const projectNewHistoricalFolder = `${getRoot()}/docs/api/${pkgName}/${versionWithoutPatch}`;
   if (await pathExists(projectNewHistoricalFolder)) {
     console.error(
-      `The package ${pkgName} has already an historical ${versionWithoutPatch} version`,
+      `${pkgName} has already a historical version ${versionWithoutPatch}`,
     );
     process.exit(1);
   }
--- a/scripts/commands/convertApiDocsToHistorical.ts
+++ b/scripts/commands/convertApiDocsToHistorical.ts
@@ -47,29 +47,23 @@ const readArgs = (): Arguments => {
 zxMain(async () => {
   const args = readArgs();
 
-  const pkg = PACKAGES.find((pkg) => pkg === args.package);
-  if (pkg === undefined) {
+  const pkgName = PACKAGES.find((pkgName) => pkgName === args.package);
+  if (pkgName === undefined) {
     throw new Error(`Unrecognized package: ${args.package}`);
   }
 
   const packageFile = await readFile(
-    `${getRoot()}/docs/api/${pkg}/_package.json`,
+    `${getRoot()}/docs/api/${pkgName}/_package.json`,
     { encoding: "utf8" },
   );
   const packageInfo = JSON.parse(packageFile);
-
   const versionMatch = packageInfo.version.match(/^(\d+\.\d+)/);
-  if (versionMatch === null) {
-    throw new Error(
-      `Invalid --version. Expected the format 0.44.0, but received ${args.version}`,
-    );
-  }
   const versionWithoutPatch = versionMatch[0];
 
-  const projectNewHistoricalFolder = `${getRoot()}/docs/api/${pkg}/${versionWithoutPatch}`;
+  const projectNewHistoricalFolder = `${getRoot()}/docs/api/${pkgName}/${versionWithoutPatch}`;
   if (await pathExists(projectNewHistoricalFolder)) {
     console.error(
-      `The package ${pkg} has already an historical ${versionWithoutPatch} version`,
+      `The package ${pkgName} has already an historical ${versionWithoutPatch} version`,
     );
     process.exit(1);
   }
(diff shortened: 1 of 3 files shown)


docs/api/qiskit-ibm-runtime/0.14/release-notes.md
```diff
@@ -0,0 +1,490 @@
+---
+title: Qiskit Runtime IBM Client release notes
+description: Changes made to Qiskit Runtime IBM Client
+in_page_toc_max_heading_level: 2
+---
+
+<span id="release-notes" />
+
+# Qiskit Runtime IBM Client 0.14 release notes
```
Eric-Arellano, line 9: These release notes should ideally be an exact copy of the current release notes, so include 0.15.

scripts/commands/convertApiDocsToHistorical.ts
```diff
@@ -0,0 +1,165 @@
...
+  const packageInfo = JSON.parse(packageFile);
+  const versionMatch = packageInfo.version.match(/^(\d+\.\d+)/);
+  const versionWithoutPatch = versionMatch[0];
+
+  const projectNewHistoricalFolder = `${getRoot()}/docs/api/${pkgName}/${versionWithoutPatch}`;
+  if (await pathExists(projectNewHistoricalFolder)) {
+    console.error(
+      `${pkgName} has already a historical version ${versionWithoutPatch}`,
+    );
+    process.exit(1);
```
Eric-Arellano, line 68: This is a good default. You're following the Principle of Least Surprise. It would be surprising if this script overwrote existing content. Instead, you make it clear to the user what's happening and ask them to decide what to do. Great job!

You could improve the error message by saying something like "Manually delete the existing folder if you intend to overwrite it."

scripts/commands/convertApiDocsToHistorical.ts
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  await mkdirp(projectNewHistoricalFolder);
+  await mkdirp(`${projectNewHistoricalFolder}/apidoc`);
+  await mkdirp(`${projectNewHistoricalFolder}/stubs`);
```
Eric-Arellano, line 73: Not nocessary

scripts/commands/convertApiDocsToHistorical.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  copyApiDocs(pkgName, versionWithoutPatch);
```
Eric-Arellano, line 75: I recommend calling this `copyApiDocsAndUpdateLinks`. Or, directly call `updateLinks` in the `main` function. Right now, you're hiding something important to how the script behaves, that we need to manually fix up links. `copyApiDocs` makes it sound like you solely are running `cp`.

scripts/commands/convertApiDocsToHistorical.ts
```diff
@@ -0,0 +1,165 @@
...
+    pkgName,
+    packageInfo.version,
+    versionWithoutPatch,
+    projectNewHistoricalFolder,
+  );
+  copyImages(pkgName, versionWithoutPatch);
+
+  if (pkgName == "qiskit") {
+    await mkdirp(`${projectNewHistoricalFolder}/release-notes`);
+    copyReleaseNotes(pkgName, versionWithoutPatch);
```
Eric-Arellano, line 86: If it's not qiskit, it should still copy the release notes. Only a file rather than folder. You could probably keep one helper function `copyReleaseNotes` and move this conditional logic into there.

scripts/commands/convertApiDocsToHistorical.ts
```diff
@@ -0,0 +1,165 @@
...
+  await mkdirp(imageDirDest);
+  await $`find ${imageDirSource}/* -maxdepth 0 -type f | xargs -I {} cp -a {} ${imageDirDest}`;
+}
+
+async function updateLinksFile(
+  pkgName: string,
+  versionWithoutPatch: string,
+  source: string,
+  dest: string,
+) {
```
Eric-Arellano, line 154: Thoughts on writing a unit test for this? Particularly the regex and `transformLinks` call.

Eric-Arellano (commented):
Great job! I like how the `main` function reads. Excellent use of abstraction.

docs/api/qiskit-ibm-runtime/0.14/release-notes.md
arnaucasau: Done!

scripts/commands/convertApiDocsToHistorical.ts
arnaucasau: You are right! Thanks!

scripts/commands/convertApiDocsToHistorical.ts
arnaucasau: Currently, the script is also copying the release notes for other projects, but given that we don't have any subfolder, those release notes are treated as every other doc and copied in the `copyApiDocsAndUpdateLinks` function.

It's true that the name `copyReleaseNotes` can be confusing, so I changed it to `copyReleaseNotesFolder` to make explicit that we only copy the folder if it exists.

Eric-Arellano (approved):
This looks great!

PR #417, opened 2023-11-28
Fix broken links qiskit.pulse

arnaucasau:
This PR adds some missing docs for qiskit.pulse. We weren't downloading these files due to an error of redirections in qiskit.org, which is now solved. See https://github.com/Qiskit/qiskit/issues/11239 for more information.


Eric-Arellano (approved):
Huzzah! Thanks, Arnau.

PR #424, opened 2023-11-28
Fix permalinks for old Qiskit API docs

Eric-Arellano:
In https://github.com/Qiskit/documentation/pull/418, we have some lines like this:

```
Qiskit Aer API Reference[¶](#qiskit-aer-api-reference "Permalink to this headline")
```

That's happening because Sphinx changed the word "headline" to "header" in later versions, and we weren't accounting for "headline".


javabster (approved):
LGTM 🚀

PR #431, opened 2023-11-29
Rename terra.md -> index.md

### LINKED ISSUE #430: Missing index page for Qiskit 0.19 to 0.43
This is because the index page was called `terra.html` rather than `index.html` until Qiskit 0.44. I didn't realize we needed to have included the rename here before running the scripts:

https://github.com/Qiskit/documentation/blob/e61f7355a10803240df34d4795010e860001f290/scripts/lib/sphinx/renameUrls.ts#L15-L25

We should fix the script. But it takes too long to regenerate the script. So to fix the problem, instead manually rename the files to `index.html` and update the `_toc.json` files appropriately.

frankharkins:
Resolves #430 as a one-off (script still needs updating)


1ucian0 (approved):
LGTM

@Eric-Arellano , feel free to revert if wrongly merged.

PR #438, opened 2023-11-29
Fix filename issue with Git

frankharkins:
The problem is Git distinguishes between upper and lower case filenames, whereas OSX does not. This PR renames the files to their capitalized equivalents (PR #437 was incorrect as it renamed to the lower case equivalents). This should fix the repo for OSX users. Here are the commands I used.

```zsh
for FILE in $(git diff --name-only); do git mv ${FILE:l} ${FILE}.bak; done
git add .
git commit -m "Temp commit"
for FILE in $(git diff HEAD~ --name-only); do git mv $FILE $FILE:r; done
git commit -m "Finish renaming files"

# EDIT: To fix accidentally reverted changes
for FILE in $(git diff HEAD~ --name-only); do git restore $FILE --source origin/main; done
git add .
git commit -m "Re-add headlines back in"
```

After applying these commits, you may need to run `git restore docs`, but then everything should be sorted.


Eric-Arellano (approved):
Thanks goodness 😅

PR #441, opened 2023-11-29
Fix typo

abbycross:
Fix a typo in a migration guide. I would also like to update the cSpell file with other words that the linter picked up, but they are already in this repo's cSpell. @Eric-Arellano should the cSpell file on the inner repo side get updated by the one in this repo when we sync?


Eric-Arellano (approved):
Yes, ideally we would be keeping the two files in sync.

Issue #442, opened 2023-11-29
Spell check migration guides


Eric-Arellano:
Make this line more precise:

https://github.com/Qiskit/documentation/blob/ebfbdcd33b75c9db1fef1134c35bce7b035598c3/cSpell.json#L276

PR #445, opened 2023-11-30
Spellcheck migration guides

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/442.

### DIFF
--- a/cSpell.json
+++ b/cSpell.json
@@ -270,10 +270,13 @@
     // words joined by underscores. e.g. hello_world, NEW_KEY_VALUE
     "\\S+_\\S+(_\\S+)*",
     // separate line <span id="" /> tags
-    "^<span id=\\S+ />$"
+    "^<span id=\\S+ />$",
+    "^<span id=\\S+></span>$"
   ],
   "ignorePaths": [
     // We cannot easily control these API docs since they live in other repositories.
-    "docs/api"
+    "docs/api/qiskit/**/*.md*",
+    "docs/api/qiskit-ibm-runtime/**/*.md*",
+    "docs/api/qiskit-ibm-provider/**/*.md*"
   ]
 }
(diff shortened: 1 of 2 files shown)


cSpell.json
```diff
@@ -270,10 +270,13 @@
     // words joined by underscores. e.g. hello_world, NEW_KEY_VALUE
     "\\S+_\\S+(_\\S+)*",
     // separate line <span id="" /> tags
-    "^<span id=\\S+ />$"
+    "^<span id=\\S+ />$",
+    "^<span id=\\S+></span>$"
```
Eric-Arellano, line 274: This matches `span id="foo"></span>` for example.

package.json
```diff
@@ -9,7 +9,7 @@
     "check:metadata": "node -r esbuild-register scripts/commands/checkMetadata.ts",
     "check:links": "node -r esbuild-register scripts/commands/checkLinks.ts",
     "check:release-notes": "node -r esbuild-register scripts/commands/validateReleaseNotesSync.ts",
-    "check:spelling": "cspell --relative --no-progress docs/**/*.md*",
+    "check:spelling": "cspell --relative --no-progress docs/**/*.md* docs/api/**/*.md*",
```
Eric-Arellano, line 12: Idk why this was necessary...but it was. Without it, the api folder is skipped.

Eric-Arellano:
fyi @abbycross

Eric-Arellano:
> Does this automatically sync the cSpell.json files in inner/open repos or do we need to update them both manually?

No, they need to be synced manually. This fixes it so that we check the migration guides though, which we weren't doing before in open source so it didn't catch the typo you fixed yesterday.

abbycross:
Actually, the spell-check in the inner repo is what caught it! Good to know that we have to update the cSpell in two places when things are changed... can make that part of the process then. FYI @beckykd

frankharkins (approved):

PR #446, opened 2023-11-30
Reuse release notes from current version

Eric-Arellano:
This PR removes the duplicated release notes on the historical API versions. We switch to reuse the same files as the current API, changing the entries on the `_toc.json` file to point to the right place. With this change, we can remove roughly 1800 duplicated files.

The release notes synchronization test has been removed due to it's no longer necessary. Also, the generation script has been changed to update the `_toc.json` files of all historical versions every time we update the current API. Moreover, when we convert the current API into a historical one, we need to remove the release notes files.


docs/api/qiskit/0.19/_toc.json
```diff
@@ -3789,171 +3789,171 @@
       "children": [
         {
           "title": "0.19",
-          "url": "/api/qiskit/0.19/release-notes/0.19"
+          "url": "/api/qiskit/release-notes/0.19"
         },
```
Eric-Arellano, line 3793: Oohh I didn't catch this until now. Note that we repeat the 0.19 entry twice in this list. This top-level one isn't necessary.

This PR is already so big that maybe it's best to fix in a follow up PR?

scripts/commands/updateApiDocs.ts
```diff
@@ -184,7 +182,7 @@ zxMain(async () => {
     : `${getRoot()}/docs/api/${pkg.name}`;
 
   if (pkg.historical && !(await pathExists(outputDir))) {
-    await createHistoricalFolder(pkg.name, outputDir);
+    mkdirp(outputDir);
```
Eric-Arellano, line 185: Nice job inlining this. Good call.

scripts/commands/updateApiDocs.ts
```diff
@@ -370,15 +369,58 @@ function urlToPath(url: string) {
   return `${getRoot()}/docs${url}.md`;
 }
 
-async function createHistoricalFolder(pkgName: string, outputDir: string) {
-  mkdirp(outputDir);
+/**
+ * Adds a new entry for the release notes of the current API version to the _toc.json
+ * of all historical API versions.
+ */
+async function updateHistoricalTocFiles(pkg: Pkg) {
```
Eric-Arellano, line 376: Nit: this should probably live in the release notes lib file.

Eric-Arellano: Also it's a good idea to put the return type as `Promise<void>`. That allows readers to quickly know this function is intended to do side effects.

scripts/commands/updateApiDocs.ts
```diff
@@ -370,15 +369,58 @@ function urlToPath(url: string) {
...
+    );
+  }
+}
 
-  // All projects have a single release notes file except Qiskit, which has a
-  // subfolder to store the release notes for each historical version.
+/**
+ * Adds a new entry for the current API version to the JSON list of release notes
+ */
+function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
```
Eric-Arellano, line 410: Ditto on moving to release notes file

scripts/commands/updateApiDocs.ts
```diff
@@ -370,15 +369,58 @@ function urlToPath(url: string) {
...
+  // We need to duplicate the first entry when we converted the current version
+  // to historical to ensure having that version at the top.
```
Eric-Arellano, line 412: Maybe add an assertion that the new version is > than the old version? Because otherwise the ordering will be wrong without people realizing.

Eric-Arellano (commented):
Thanks, Arnau!! This is looking great.

Eric-Arellano (commented):
Yay! I can't approve since it's my original PR, but I approve 🙇

scripts/commands/updateApiDocs.ts
arnaucasau: Good idea! However, I deleted that piece of code because the conditional is no longer necessary. That is because we don't have the duplicated entries on each historical `_toc.json` file anymore :+1:

arnaucasau (approved):
Thank you so much for your review @Eric-Arellano!

PR #462, opened 2023-12-01
Use stable links for migration guides

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/301.

I didn't change the links for APIs that will still exist in Qiskit 1.0, like primitives. It seems useful to point to the newest documentation for those replacements.


abbycross (approved):
👍

PR #463, opened 2023-12-01
Add session import statement

### LINKED ISSUE #449: Session import info should be in the top sections

In this page https://docs.quantum.ibm.com/run/run-jobs-in-session#set-up-to-use-sessions we see the instructions to start using the sessions but the `Session` class import is not defined in the first sections so this may lead the user to get an error that the variable is not defined.

image

I see this info is included in other section but I think it would be more useful at the beginning

image

### Select all that apply

- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [X] other

### Describe the fix.

Include the `Session` import at the beginning of the page, for example in _Open a session_ section

beckykd:
closes #449


abbycross (approved):
👍

Issue #166, opened 2023-10-13
Test Jupyter notebooks in CI


Eric-Arellano:
We want to make sure the code still works.

Testing in CI also makes it convenient that we can upload the executed tutorials as an artifact to CI with the latest output of the notebook, from a standardized environment. That allows us to download the result and periodically update the notebooks. Maybe we want to even automate those updates.

I reckon this should be a dedicated CI job, rather than being in the main `lint` job. That will give us better concurrency and maybe make the output more clear, along with the CI config.

We need to decide what version of Qiskit and Qiskit Runtime to test with. This is essentially https://github.com/Qiskit/documentation/issues/133. For now, I think it's sufficient to only test the latest stable version. I strongly encourage us to pin those versions so that our CI is deterministic. That's extremely useful so that we can jump back at any point in time and get things working, and see how the repo was configured at certain snapshots. When you generate API docs for a new version, we'll update the pinned versions.

PR #471, opened 2023-12-04
Add code to test notebook code examples

frankharkins:
This PR adds a script to automatically execute notebooks in CI.

To keep PRs small and focussed, this only adds the scripts to test notebooks, it doesn't actually fix those notebooks or add anything to CI.

<details>
  <summary>Example: Success</summary>

Screenshot 2023-12-04 at 14 22 15

</details>

<details>
  <summary>Example: Failure</summary>

Screenshot 2023-12-04 at 14 23 17

</details>

<details>
  <summary>Why <code>tox</code>?</summary>

An important feature is to be able to run a single notebook in the virtual environment. With `tox`, we can use commands like this:

```sh
tox -- path/to/notebook
```

Whereas with `nox`, this is the simplest I could get it:

```sh
nox --session 'test(path="path/to/notebook")'
```

Note this also doesn't autocomplete the path.

The main reason for choosing `nox` seems to be more flexible configuration, but we only need simple configuration in this repo so this isn't a problem.

</details>

***

Part of #166

### DIFF
--- a/scripts/nb-tester/test-notebook.py
+++ b/scripts/nb-tester/test-notebook.py
@@ -33,20 +33,22 @@ NOTEBOOKS_EXCLUDE = [
 ]
 
 
-def run_notebook(*args, **kwargs):
+def execute_notebook(path: Path, write=False) -> bool:
     """
-    Wrapper function for `_run_notebook` to print status
+    Wrapper function for `_execute_notebook` to print status
     """
     print(f"▶️  {path}", end="", flush=True)
     try:
-        _run_notebook(*args, **kwargs)
-    except Exception as err:
+        _execute_notebook(path, write)
+    except nbconvert.preprocessors.CellExecutionError as err:
         print("\r❌\n")
-        raise err
+        print(err)
+        return False
     print("\r✅")
+    return True
 
 
-def _run_notebook(filepath: Path, write=False) -> None:
+def _execute_notebook(filepath: Path, write=False) -> None:
     """
     Use nbconvert to execute notebook
     """
@@ -88,5 +90,7 @@ if __name__ == "__main__":
 
     notebook_paths = args or find_notebooks()
     print("Executing notebooks:")
-    for path in notebook_paths:
-        run_notebook(path, write)
+    results = [execute_notebook(path, write) for path in notebook_paths]
+    if not all(results):
+        sys.exit(1)
+    sys.exit(0)


scripts/nb-tester/requirements.txt
```diff
@@ -0,0 +1,6 @@
+nbconvert==7.11.0
+nbformat==5.9.2
+ipykernel==6.27.1
+qiskit[all]==0.45.1
+qiskit-aer==0.13.1
+qiskit-ibm-runtime==0.16.1
```
Eric-Arellano, line 6: Thanks for pinning everything! I think this is fine to allow patch updates, since these projects follow semantic versioning.

```suggestion
# We pin to exact versions for a more reproducible and
# stable build.
nbconvert~=7.11.0
nbformat~=5.9.2
ipykernel~=6.27.1
qiskit[all]~=0.45.1
qiskit-aer~=0.13.1
qiskit-ibm-runtime~=0.16.1
```

scripts/nb-tester/test-notebook.py
```diff
@@ -0,0 +1,93 @@
...
+    "docs/api/**",
+    "**/.ipynb_checkpoints/**",
+    # Following notebooks have code errors
+    "docs/build/circuit-construction.ipynb",
+    "docs/build/pulse.ipynb",
+    "docs/transpile/custom-transpiler-pass.ipynb",
+    "docs/transpile/transpiler-stages.ipynb",
+    # Following notebook has some non-python dependency I can't figure out
+    "docs/build/circuit-visualization.ipynb",
+    # Following notebooks make requests so can't be tested yet
```
Eric-Arellano, line 30: Idk how we're going to test this. But I really like that you're not letting it get in the way. Testing 80% of notebooks is better than 0%.

scripts/nb-tester/test-notebook.py
```diff
@@ -0,0 +1,93 @@
...
+    print(f"▶️  {path}", end="", flush=True)
+    try:
+        _run_notebook(*args, **kwargs)
+    except Exception as err:
+        print("\r❌\n")
+        raise err
+    print("\r✅")
+
+
+def _run_notebook(filepath: Path, write=False):
```
Eric-Arellano, line 49:
```suggestion
def _run_notebook(filepath: Path, write=False) -> None:
```

scripts/nb-tester/test-notebook.py
```diff
@@ -0,0 +1,93 @@
...
+    Get paths to all notebooks in NOTEBOOKS_GLOB that are not excluded by
+    NOTEBOOKS_EXCLUDE
+    """
+    all_notebooks = Path(".").rglob(NOTEBOOKS_GLOB)
+    filtered_notebooks = [
+        path
+        for path in all_notebooks
+        if not any(path.match(glob) for glob in NOTEBOOKS_EXCLUDE)
+    ]
+    return filtered_notebooks
```
Eric-Arellano, line 81: Nit

```suggestion
    return [
        path
        for path in all_notebooks
        if not any(path.match(glob) for glob in NOTEBOOKS_EXCLUDE)
    ]
```

scripts/nb-tester/test-notebook.py
Eric-Arellano: Also maybe call it `_execute_notebook` to use the terminology from the library?

scripts/nb-tester/test-notebook.py
```diff
@@ -0,0 +1,93 @@
...
+    "docs/transpile/transpiler-stages.ipynb",
+    # Following notebook has some non-python dependency I can't figure out
+    "docs/build/circuit-visualization.ipynb",
+    # Following notebooks make requests so can't be tested yet
+    "docs/run/get-backend-information.ipynb",
+    "docs/start/hello-world.ipynb",
+]
+
+
+def run_notebook(*args, **kwargs):
```
Eric-Arellano, line 36: Given that there are only two args, probably better to be explicit?

scripts/nb-tester/test-notebook.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    Wrapper function for `_run_notebook` to print status
+    """
+    print(f"▶️  {path}", end="", flush=True)
+    try:
+        _run_notebook(*args, **kwargs)
+    except Exception as err:
+        print("\r❌\n")
+        raise err
```
Eric-Arellano, line 45: This will cause us to early return, right? Thoughts on instead printing all errors? That is generally useful, especially from a CI perspective so you can see all the problems without having to restart.

If you think it's useful to sometimes early exit, we could use `argparse` to add a bool flag for people to decide. Sphinx has `--keepgoing` for this behavior toggle.

Eric-Arellano (commented):
This looks great! Thank you.

scripts/nb-tester/test-notebook.py
frankharkins: Good idea, fixed in cbcc714c395ac7a18675e65160d8aa73ceef321e

Eric-Arellano (approved):
💜

PR #474, opened 2023-12-04
update straggler quantum-computing domain urls

abbycross:
Related to #453 

Find all remaining urls using the quantum-computing.ibm.com domain and change to quantum.ibm.com.

This does not remove from the api docs; those will need to be fixed in their home repos.


Eric-Arellano (approved):
Thanks!

PR #483, opened 2023-12-06
Test notebooks on pull request

frankharkins:
Adds a GitHub action to test notebook code examples on pull request.

### Handling notebooks that submit jobs

We only want to submit jobs to IBM Quantum _very_ rarely (such as updating notebook outputs once per qiskit release), but runtime doesn't have a way of restricting the types of job you can send.

In this PR, I've limited cell execution to 100s. This should be short enough to do most local tasks and to get backend information, but not nearly long enough for a job to run on the open provider (usually takes hours).

If someone accidentally makes a PR with a notebook that submits jobs, the script will timeout and raise an error. On exit, the script cancels any jobs created since it started running. This means they won't take up any resources, and the author will be alerted when CI fails.


Eric-Arellano:
> qiskit_ibm_runtime.accounts.exceptions.InvalidAccountError: "Invalid `token` value. Expected a non-empty string, got ''."

Should we skip these notebooks? Or somehow the specific code blocks? We want to be careful that CI doesn't burn through a bunch of compute resources.

.github/workflows/notebook-test.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Install LaTeX dependencies
+        run: |
+          # Add your notebook to this list if it needs latex to run
+          NEEDS_LATEX="docs/build/circuit-visualization.ipynb"
+          for FILE in $NEEDS_LATEX; do
+            if [[ "${{ steps.changed-files.outputs.all_changed_files }}" = *"$FILE"* ]]; then
+              sudo apt-get install texlive-pictures texlive-latex-extra poppler-utils
+              break
+            fi
+          done
```
frankharkins, line 44: The latex packages are pretty large and only used by one cell. Check out these commits' runs to see this step working:
- Skip installation https://github.com/Qiskit/documentation/pull/483/commits/1d6872fc77eb27a36c0c0e320fc5066c9356989d
- Install texlive https://github.com/Qiskit/documentation/pull/483/commits/0f386392e62cd21c7a42519efb8e690f9cbed867

scripts/nb-tester/test-notebook.py
```diff
@@ -14,78 +14,149 @@
...
-WRITE_FLAG = "--write"
 NOTEBOOKS_GLOB = "docs/**/*.ipynb"
 NOTEBOOKS_EXCLUDE = [
     "docs/api/**",
     "**/.ipynb_checkpoints/**",
-    # Following notebooks have code errors
+    # Following notebooks are broken
     "docs/transpile/transpiler-stages.ipynb",
-    # Following notebooks make requests so can't be tested yet
     "docs/run/get-backend-information.ipynb",
```
frankharkins, line 28: This notebook can now be tested, but is broken as `ibm_lagos` doesn't exist any more. I'll fix in a follow-up.

.github/workflows/notebook-test.yml
```diff
@@ -0,0 +1,67 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+          cache: "pip"
+
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@af2816c65436325c50621100d67f6e853cd1b0f1
```
Eric-Arellano, line 33: Nice. And thanks for pinning this. Good idea to use an action rather than trying to implement it ourselves.

.github/workflows/notebook-test.yml
```diff
@@ -0,0 +1,67 @@
...
+
+      - name: Install LaTeX dependencies
+        run: |
+          # Add your notebook to this list if it needs latex to run
+          NEEDS_LATEX="docs/build/circuit-visualization.ipynb"
```
Eric-Arellano, line 38: I hate Bash. For this to work with future notebooks, I think it will be better to use an array of strings. From https://www.freecodecamp.org/news/bash-array-how-to-declare-an-array-of-strings-in-a-bash-script/, this would be

```bash
NEEDS_LATEX=(
    "docs/build/circuit-visualization.ipynb"
)

for FILE in ${NEEDS_LATEX[@]}; do
```

> The @ symbol in the square brackets indicates that you are looping through all of the elements in the array. If you were to leave that out and just write for str in ${myArray}, only the first string in the array would be printed.

.github/workflows/notebook-test.yml
```diff
@@ -0,0 +1,67 @@
...
+          for FILE in $NEEDS_LATEX; do
+            if [[ "${{ steps.changed-files.outputs.all_changed_files }}" = *"$FILE"* ]]; then
+              sudo apt-get install texlive-pictures texlive-latex-extra poppler-utils
+              break
+            fi
+          done
+
+      - name: Install Qiskit IBM Runtime to save account
+        # To save account; this is re-installed during the "Run tox" step
```
Eric-Arellano, line 47:
```suggestion
        # This is to save our account in the next step. Note that the
        # package will be re-installed during the "Run tox" step.
```

.github/workflows/notebook-test.yml
```diff
@@ -0,0 +1,67 @@
...
+          from qiskit_ibm_runtime import QiskitRuntimeService
+          QiskitRuntimeService.save_account(
+              channel="ibm_quantum",
+              instance="ibm-q/open/main",
+              token="${{ secrets.IBM_QUANTUM_TEST_TOKEN }}",
+              set_as_default=True
+          )
+
+      - name: Run tox
+        uses: lsst-sqre/run-tox@97818256d9fa3c72d0c12f31660718adb495a1cb
```
Eric-Arellano, line 63: This is not a very popular action. While it looks legit, there is some risk using it, including supply chain security. Probably use Tox explicitly like we normally do? While more verbose, it's not very complex code, unlike changed-files.

scripts/nb-tester/test-notebook.py
```diff
@@ -14,78 +14,149 @@
 from pathlib import Path
 import nbconvert
 import nbformat
+import nbclient
```
Eric-Arellano, line 17: Total nit: Python convention is to split first-party and third-party library imports. Start with first-party, then blank line, then third-party.

(In a Python repo, I would have set up isort to automate this.)

scripts/nb-tester/test-notebook.py
```diff
@@ -14,78 +14,149 @@
...
-    # Following notebooks make requests so can't be tested yet
     "docs/run/get-backend-information.ipynb",
+]
+NOTEBOOKS_THAT_SUBMIT_JOBS = [
     "docs/start/hello-world.ipynb",
 ]
 
 
-def execute_notebook(path: Path, write=False) -> bool:
+class ExecuteOptions:
```
Eric-Arellano, line 35:
```suggestion
@dataclass(frozen=True)
class ExecuteOptions:
```

scripts/nb-tester/test-notebook.py
```diff
@@ -14,78 +14,149 @@
...
+        job.cancel()
+    return False
 
-    notebook_paths = args or find_notebooks()
+
+parser = argparse.ArgumentParser(
+    prog="Notebook executor",
+    description="For testing notebooks and updating their outputs",
+    epilog="For help, contact Frank Harkins",
+)
```
Eric-Arellano, line 127: I like to have a helper function called `create_parser() -> ArgumentParser`. Then in your main, you'll use `create_parser().parse_args()`.

scripts/nb-tester/test-notebook.py
```diff
@@ -14,78 +14,149 @@
...
+    submit_jobs: bool
+
+
+def execute_notebook(path: str, options: ExecuteOptions) -> bool:
     """
     Wrapper function for `_execute_notebook` to print status
     """
+    path = Path(path)
+    if path.suffix != ".ipynb":
+        print(f"⏭️ {path} is not a notebook; skipping")
```
Eric-Arellano, line 46: Does this result in a lot of noise? Also I think this makes more sense to filter before you pass in the argument. I would expect `execute_notebook` to be passed in what we already know to be a valid notebook. This looks like filtering in your list comprehension in your `main()`. Likewise, you should have this function take in a `Path` directly.

scripts/nb-tester/test-notebook.py
```diff
@@ -14,78 +14,149 @@
...
+        nbclient.exceptions.CellTimeoutError,
+    )
     try:
-        _execute_notebook(path, write)
-    except nbconvert.preprocessors.CellExecutionError as err:
+        _execute_notebook(path, options)
+    except possible_exceptions as err:
         print("\r❌\n")
         print(err)
+        with open("latex_error.log") as f:
```
Eric-Arellano, line 58: Will this file always exist?

Eric-Arellano (commented):
🙌

scripts/nb-tester/test-notebook.py
frankharkins: My bad, this was left in from debugging 🤦

.github/workflows/notebook-test.yml
```diff
@@ -0,0 +1,73 @@
...
+        with:
+          # Add your notebook to this list if it needs latex to run
+          files: |
+            docs/build/circuit-visualization.ipynb
+
+      - name: Install LaTeX dependencies
+        if: steps.latex-changed-files.outputs.any_changed == 'true'
+        run: |
+          sudo apt-get update
+          sudo apt-get install texlive-pictures texlive-latex-extra poppler-utils
```
frankharkins, line 47: I've replaced the bash logic with this, should be easier to understand, and the step will now appear as "skipped" in the action log if it doesn't run.

frankharkins:
See https://github.com/Qiskit/documentation/pull/483/commits/358eb2fd4624c765cb1ad7a72e47c3f316881014 and https://github.com/Qiskit/documentation/pull/483/commits/1ff5fed7ff212d2cd23b50b6741c666b9b765e50 for examples of this working.

scripts/nb-tester/test-notebook.py
```diff
@@ -10,82 +10,166 @@
...
 
 
-def execute_notebook(path: Path, write=False) -> bool:
+@dataclass(frozen=True)
+class ExecuteOptions:
+    write: bool
+    submit_jobs: bool
+
+
+def execute_notebook(path: str, options: ExecuteOptions) -> bool:
```
Eric-Arellano, line 43:
```suggestion
def execute_notebook(path: Path, options: ExecuteOptions) -> bool:
```

scripts/nb-tester/test-notebook.py
```diff
@@ -10,82 +10,166 @@
...
+        "--write",
+        action="store_true",
+        help="overwrite notebooks with execution outputs",
+    )
+    parser.add_argument(
+        "--submit-jobs",
+        action="store_true",
+        help=(
+            "run notebooks that submit jobs to IBM Quantum and wait indefinitely "
+            "for jobs to complete"
```
Eric-Arellano, line 154:
```suggestion
            "Run notebooks that submit jobs to IBM Quantum and wait indefinitely "
            "for jobs to complete. Warning: this has a real cost because it uses "
            "quantum resources! Only use this argument occasionally and intentionally." 
```

scripts/nb-tester/test-notebook.py
```diff
@@ -10,82 +10,166 @@
...
+            "notebooks in `docs/`. To exclude a notebook from this process, add it "
+            "to `NOTEBOOKS_EXCLUDE` in the script."
+        ),
+        nargs="*",
+    )
+    parser.add_argument(
+        "-w",
+        "--write",
+        action="store_true",
+        help="overwrite notebooks with execution outputs",
```
Eric-Arellano, line 147: Nit: I was a little confused at first if this meant the `output` portions of cells in notebooks. I think technically it does? But the general idea still applies, to overwrite the entire notebook with the new result.

```suggestion
        help="Overwrite notebooks with the results of this script's execution.",
```

scripts/nb-tester/test-notebook.py
```diff
@@ -10,82 +10,166 @@
...
+    for job in jobs:
+        job.cancel()
+    return False
+
+
+def create_argument_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        prog="Notebook executor",
+        description="For testing notebooks and updating their outputs",
+        epilog="For help, contact Frank Harkins",
```
Eric-Arellano, line 132: I don't think that's necessary nor healthy in an open source project. Although I do appreciate you thinking of the detail :) Any of us should be able to help, and this note might not apply in some future. 

Instead, that's why there's so much focus in open source on high code quality and documentation, along with training other teammates on how the code works.

scripts/nb-tester/test-notebook.py
```diff
@@ -10,82 +10,166 @@
...
         return
 
     for cell in nb.cells:
+        # Remove execution metadata to avoid noisy diffs.
         cell.metadata.pop("execution", None)
     nbformat.write(nb, filepath)
 
 
-def find_notebooks() -> list[Path]:
+def find_notebooks(submit_jobs=False) -> list[Path]:
```
Eric-Arellano, line 85: Nit. I like to use `*` when you have boolean args like this. It forces the caller to use a keyword arg (kwarg) rather than positional arg. It's much more readable to see `find_notebooks(submit_jobs=True)` than `find_notebooks(True)`.

```suggestion
def find_notebooks(*, submit_jobs: bool = False) -> list[Path]:
```

Eric-Arellano (commented):
This is looking excellent 🙌

scripts/nb-tester/test-notebook.py
frankharkins:
```suggestion
```

Eric-Arellano (approved):
Looks great! But we probably want Arnau to review it first

scripts/nb-tester/test-notebook.py
```diff
@@ -10,82 +10,166 @@
...
+
 if __name__ == "__main__":
-    args = sys.argv[1:]
-    write = WRITE_FLAG in args
-    if write:
-        args.remove(WRITE_FLAG)
+    args = create_argument_parser().parse_args()
 
-    notebook_paths = args or find_notebooks()
+    paths = map(Path, args.filenames or find_notebooks(args.submit_jobs))
```
arnaucasau, line 163: After adding the `*` in the function definition, we need to specify the argument here. Otherwise, it will fail when running the script without arguments.
```suggestion
    paths = map(Path, args.filenames or find_notebooks(submit_jobs=args.submit_jobs))
```

arnaucasau (commented):
Great job Frank! I've been testing the script with different arguments and files, and besides a small thing (see below), all worked really well! I like a lot the way you wrote the test using actions to get the changed files, and good thinking adding the conditional to install the latex packages only when necessary!

scripts/nb-tester/test-notebook.py
frankharkins: Good spot, thanks!

PR #484, opened 2023-12-06
Remove undefined spans from historical API docs

Eric-Arellano:
I'm not sure why the script set these and still does: https://github.com/Qiskit/documentation/issues/485. While they aren't rendered to the user, we shouldn't have multiple HTML elements with the same ID.

Updated with this:

```bash
rg '<span id="undefined" />' -l | xargs sd -s '<span id="undefined" />' '' 
```

We should find a proper fix in https://github.com/Qiskit/documentation/issues/485, but this is a good workaround for now. We're unlikely to regenerate historical API docs, so this will workaround those ones. For current docs, we can regenerate once we have a fix.


arnaucasau (approved):
LGTM!

PR #486, opened 2023-12-06
Fix `IterativeAmplitudeEstimation` pages crashing due to bad Latex

Eric-Arellano:
The pages were crashing because we had a bad `|` character inside a `$latex$` expression in a table. Normally the `|` would be okay to represent the pipe character, but Markdown thought that it was saying the table column was done.

Result:

Screenshot 2023-12-06 at 3 06 56 PM

Note that I had to use `\vert` for the pipe character. Using `\|` resulted in the Latex interpreting as `||`, which results in the `parallel` symbol being used:

Screenshot 2023-12-06 at 3 08 45 PM

--

While this PR hotfixes the bad files, we should fix the underlying generation script to handle this edge case. See https://github.com/Qiskit/documentation/issues/488

--

This change was generated with:

```bash
rg -F '\$\mathcal' -l | xargs sd -s '\$\mathcal{Q}^k \mathcal{A} | 0\rangle\$' '$\mathcal{Q}^k \mathcal{A} \vert0\rangle$'

rg -F '\$\mathcal' -l | xargs sd -s '\$\mathcal{Q}^k \mathcal{A}                                                                                 | 0\rangle\$' '$\mathcal{Q}^k \mathcal{A} \vert0\rangle$'
```


arnaucasau (approved):
Looks good to me! I tested it locally, and all works well!

PR #492, opened 2023-12-07
API generation from CI artifacts

### LINKED ISSUE #311: API generation can build from CI, not a published site
This will be necessary when qiskit.org goes away! 

This is more complex than it sounds (but still doable) because we rely on a web server. We will need to download the built files from CI, unzip them, then start up a local web server.

arnaucasau:
Closes #311.

It adds a new method to generate the API docs using a CI artifact. The script has a new mandatory argument `--artifact` or `-a` to specify a link to a specific artifact we want to download. This change is necessary because we can't use qiskit.org anymore.

To be able to download the artifacts, we need to have installed and configured the [GitHub command line](https://docs.github.com/en/github-cli/github-cli/quickstart).

Example of a generation for the Qiskit 0.45 API docs using an artifact:

`npm run gen-api -- -p qiskit -v 0.45.0 -a https://github.com/Qiskit/qiskit/suites/17881600359/artifacts/1026798160`


arnaucasau:
Thank you for the review, @Eric-Arellano! I incorporated all the changes you suggested and I tried the script with `qiskit`, `qiskit-ibm-runtime`, and `qiskit-ibm-provider`. It works with all three projects using the current API and also the historical argument. 

Looking at the script, we could instead of downloading the images, copy them from the unzipped folder directly, but because we don't want every image and it would add too many extra files that we don't want, I decided to open the web server always and continue downloading them.

I also changed the `initialUrls` to be a string, and I changed its name to `initialUrl`.

The commands I used to test the generation are:

```
npm run gen-api -- -p qiskit -v 0.45.0 -a https://github.com/Qiskit/qiskit/suites/17881600359/artifacts/1026798160
npm run gen-api -- -p qiskit -v 0.45.0 -a https://github.com/Qiskit/qiskit/suites/17881600359/artifacts/1026798160 --historical

npm run gen-api -- -p qiskit-ibm-provider -v 0.7.2 -a https://github.com/Qiskit/qiskit-ibm-provider/suites/17328261790/artifacts/990337776
npm run gen-api -- -p qiskit-ibm-provider -v 0.7.2 -a https://github.com/Qiskit/qiskit-ibm-provider/suites/17328261790/artifacts/990337776 --historical

npm run gen-api -- -p qiskit-ibm-runtime -v 0.15.0 -a https://github.com/Qiskit/qiskit-ibm-runtime/suites/18288926026/artifacts/1057652367
npm run gen-api -- -p qiskit-ibm-runtime -v 0.15.0 -a https://github.com/Qiskit/qiskit-ibm-runtime/suites/18288926026/artifacts/1057652367 --historical
```

scripts/lib/downloadArtifacts.ts
```diff
@@ -0,0 +1,38 @@
...
+  artifactUrl: string,
+  destination: string,
+) {
+  const artifactId = getArtifactID(artifactUrl);
+
+  if (!(await pathExists(destination))) {
+    await mkdirp(destination);
+  }
+
+  await $`gh api \
```
Eric-Arellano, line 32: I'm not sure this is actually worth adding the code for it. We already say in the prereqs it's required. But FYI in some situations you'd want to proactively check that `gh` is installed and give a good error message to the user if it's not which instructs them to install it.

Not necessary for other binaries like `unzip` since unix users should have it already.

Eric-Arellano (approved):
Nice!

In a follow up PR, can you please add the new docs for Runtime 0.16? You'll want to convert 0.15 to become historical.

Issue #461, opened 2023-12-01
Profile link checker


Eric-Arellano:
`npm run check:links` is taking about 42 seconds on my fast laptop. Profile where the slowdown is coming from and see if there are any ways to make this faster.

We don't care about `--external` to check external docs. Only internal links.

I imagine this is due to historical API versions. If so, we should add a `--historical` argument and only check those when specified, like in CI.

Eric-Arellano:
Looks like all the time is spent in `loadFilesAndLinks`:

```
read doc paths: 49.597ms
load links: 34.776s
other files: 42.891ms
```

That is using this benchmark:

```
  // Determine what files with links we want to parse
  console.time('read doc paths');
  const pathsWithLinks = await globby("docs/**/*.{ipynb,md,mdx}");
  console.timeEnd('read doc paths');

  // Parse the files with links and get a list with all the links
  // in all the files without duplications.
  console.time('load links');
  const [docsFiles, internalLinkList, externalLinkList] =
    await loadFilesAndLinks(pathsWithLinks);
  console.timeEnd('load links');

  // Define extra files that we don't want to parse
  console.time('other files');
  const otherFiles = [
    ...(await globby("{public,docs}/**/*.{png,jpg,gif,svg}")).map(
      (fp) => new File(fp, [], false),
    ),
    ...SYNTHETIC_FILES.map((fp) => new File(fp, [], true)),
  ];
  console.timeEnd('other files');
```

I first noticed this slowdown when we added all historical API versions, which makes sense. They make up ~90% of the docs. Right now, we're reading every single one of those files and getting their anchors:

https://github.com/Qiskit/documentation/blob/09372889c7f9aec81eddfa88389b650843301fad/scripts/commands/checkLinks.ts#L155-L161

We need to know what anchors a file has so that we can validate links from other files are using the anchor correctly:

https://github.com/Qiskit/documentation/blob/09372889c7f9aec81eddfa88389b650843301fad/scripts/lib/LinkChecker.ts#L108-L119

--

But the thing is, we almost never link to the old API docs except for within the old API docs themselves. So, unless we want to check the historical API docs themselves, it is a waste of time to be building this mapping. 

_Except_ that there are a few edge cases where we do actually link to historical API versions, specifically in our migration guides. That's necessary since the APIs are deleted in modern versions: https://github.com/Qiskit/documentation/pull/462.

We need to decide how to handle those specific links to historical API versions. Some ideas:

[...]

Issue #495, opened 2023-12-07
Check links for API docs (Qiskit 0.46+)


Eric-Arellano:
At least the internal links to start. @axelhzf has a good point that it would be ideal to avoid 404s in our site, even if they are for outdated docs not many are looking at.

## [Outdated] adding the mechanism to check

We should add an argument like `--historical-apis` to the link checker and only use it in CI, probably the cron job.

We want to be careful to not slow down performance too badly when implementing this. Per https://github.com/Qiskit/documentation/pull/496, the key is to avoid loading more files into memory than necessary. 

We should process each historical API version atomically. For example, `0.44` docs should not be linking out to anything else but other `0.44` files and external links. So, we should entirely process `0.44` docs at once, then drop it from memory. Then go on to `0.43`, etc. 

## How to fix broken links

Some of the broken links have been due to bugs in our script, like https://github.com/Qiskit/documentation/issues/542 and https://github.com/Qiskit/documentation/issues/732. Legit issues will need to be fixed via a mechanism like https://github.com/Qiskit/documentation/issues/719.

## Update Aug 30 2024

This is high priority. We should have zero broken links for new versions of docs, meaning Qiskit 0.46+. For Runtime, aim to do all releases in 2024, but triage how bad that will be and revisit if it's going to take too much time.

Infrastructure-wise, tweak our commands to only check the targeted dev versions.

PR #496, opened 2023-12-07
Speed up link checker by not loading historical API docs

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/461.

Before, the script took me about 42 seconds. Now, about 5.7:

```
❯ hyperfine 'npm run check:links'
Benchmark 1: npm run check:links
  Time (mean ± σ):      5.680 s ±  0.143 s    [User: 7.285 s, System: 0.471 s]
  Range (min … max):    5.548 s …  5.964 s    10 runs
 
```

This PR differentiates between files we want to check vs. files that we only need to know they exist because other files might link to them. For the latter, we do still need to read in those files to determine their anchors. 

The key insight of this PR is that 99% of the historical API docs can be ignored because the stable docs don't link to them. We only need to load a couple of historical API docs because our migration guides link to them. 

--

I think we probably will want to add an argument to allow checking links for historical API docs: https://github.com/Qiskit/documentation/issues/495. This should work well with that change; we'd update the list based off of the `Arguments` passed in.


scripts/commands/checkLinks.ts
```diff
@@ -116,7 +142,6 @@ const readArgs = (): Arguments => {
     .version(false)
     .option("external", {
       type: "boolean",
-      demandOption: false,
```
Eric-Arellano, line 119: This is the default. No need to set it.

scripts/commands/checkLinks.ts
```diff
@@ -134,57 +159,60 @@ function markdownFromNotebook(source: string): string {
   return markdown;
 }
 
+async function getMarkdownAndAnchors(
```
Eric-Arellano, line 162: Same logic as before.

scripts/commands/checkLinks.ts
```diff
@@ -25,9 +25,39 @@ import remarkGfm from "remark-gfm";
...
+  "docs/start/*",
+  "docs/transpile/*",
+  "docs/verify/*",
+  "docs/support.mdx",
+  "docs/api/migration-guides/*",
+  "docs/api/qiskit/*",
+  "docs/api/qiskit-ibm-runtime/*",
+  "docs/api/qiskit-ibm-provider/*",
+  "!docs/**/*.json",
+];
```
arnaucasau, line 40:
```suggestion
const DOCS_GLOBS_TO_CHECK = [
  "docs/**/*.{ipynb,md,mdx}",
  "!docs/api/*/[0-9]*/*", // Historical versions
  "!docs/api/qiskit/release-notes/*", // Qiskit release notes
];
```

I like the way you defined `DOCS_GLOBS_TO_CHECK` because we can remove all the conditionals we had inside the `processDocsFiles` function. Just one question: do you think it might be a problem to define the files like this if someday we add another folder and forget to update the script? Forgetting to ignore a file is detectable through CI, but forgetting to add one can be easily overlooked. 

Maybe we can use a more general pattern, similar to what we had before, and explicitly write what files we are ignoring.

In any case, I still like your definition, and I'm good with it if you think it's best.

arnaucasau (approved):
Thank you Eric! On my end, the execution time goes from 1 minute 40 seconds to roughly 15 seconds! Awesome work!!

Indeed, the problem with the script was the amount of historical files we were parsing its anchors. Having an extra glob to decide which ones we want to load is a really good idea, and now we can avoid having extra logic to ignore files in addition to the speed up.

PR #500, opened 2023-12-08
Fix span tag id=undefined and fix /api/qiskit/index entry

### LINKED ISSUE #227: Don't refer to `/index` in API table of contents for Qiskit

https://docs.quantum-computing.ibm.com/api/qiskit/index

### Select all that apply

- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [X] broken link
- [ ] other

### Describe the fix.

When on the page https://docs.quantum-computing.ibm.com/api/qiskit and I try to click "qiskit" section under the API References (see screenshot), I am taken to the link https://docs.quantum-computing.ibm.com/api/qiskit/index and getting a 404. 

image

### LINKED ISSUE #485: Triage what is causing `undefined` spans in API docs
https://github.com/Qiskit/documentation/pull/484 works around the problem, which is fine for historical API versions since we won't regenerate those docs. But it's not a good solution for long term. We're still getting this issue even with modern docs versions:

https://github.com/Qiskit/documentation/blob/58f83f190d0fc6ef9ed56350761d319b242b3b9a/docs/api/qiskit/qiskit.providers.BackendV1.md?plain=1#L59

https://github.com/Qiskit/documentation/blob/58f83f190d0fc6ef9ed56350761d319b242b3b9a/docs/api/qiskit/qiskit.transpiler.passes.NoiseAdaptiveLayout.md?plain=1#L23-L27

https://github.com/Qiskit/documentation/blob/58f83f190d0fc6ef9ed56350761d319b242b3b9a/docs/api/qiskit/qiskit.transpiler.passes.NoiseAdaptiveLayout.md?plain=1#L35-L71

https://github.com/Qiskit/documentation/blob/58f83f190d0fc6ef9ed56350761d319b242b3b9a/docs/api/qiskit/qiskit.providers.basicaer.StatevectorSimulatorPy.md?plain=1#L31-L35

https://github.com/Qiskit/documentation/blob/58f83f190d0fc6ef9ed56350761d319b242b3b9a/docs/api/qiskit/qiskit.providers.basicaer.QasmSimulatorPy.md?plain=1#L31-L35

https://github.com/Qiskit/documentation/blob/58f83f190d0fc6ef9ed56350761d319b242b3b9a/docs/api/qiskit/qiskit.algorithms.eigensolvers.VQD.md?plain=1#L69-L73

https://github.com/Qiskit/documentation/blob/58f83f190d0fc6ef9ed56350761d319b242b3b9a/docs/api/qiskit/qiskit.algorithms.eigensolvers.VQD.md?plain=1#L95-L99

https://github.com/Qiskit/documentation/blob/58f83f190d0fc6ef9ed56350761d319b242b3b9a/docs/api/qiskit/qiskit.providers.basicaer.UnitarySimulatorPy.md?plain=1#L31-L35

arnaucasau:
This PR fixes the issue with the undefined span IDs when generating the documentation and transforms the `/api/qiskit/index` entry in the toc file to `/api/qiskit`. A new test has been added to verify we don't generate span tags with an undefined id.

The Qiskit docs have been regenerated to verify the correct behavior of the script for the current version. The changes in the docs files are empty lines matching the great work done by @Eric-Arellano in #484 removing the tags manually.

Closes #485
Closes #227


scripts/commands/updateApiDocs.ts
```diff
@@ -336,6 +336,14 @@ async function convertHtmlToMarkdown(
     await writeFile(path, result.markdown);
   }
 
+  if (pkg.name == "qiskit") {
```
Eric-Arellano, line 339: It's an issue for Provider and Runtime too:

https://github.com/Qiskit/documentation/blob/0b60acf1a5da5c0653fb3ed4bc1dc04700e5e1f7/docs/api/qiskit-ibm-provider/_toc.json#L9

I suspect that instead the fix is going to be inside `generateToc`, that it should special-case `/index`. 

I also remember wondering if the fix belongs in https://github.com/Qiskit/documentation/blob/main/scripts/lib/sphinx/renameUrls.ts, but I think I convinced myself a few weeks ago that no. It's important we keep the `index` file name there so that we create a file called `index.md`. We only need to not use `/index` in `generateToc`.

scripts/lib/sphinx/sphinxHtmlToMarkdown.test.ts
```diff
@@ -1470,6 +1470,57 @@ test("identify release notes", async () => {
 `);
 });
 
+test("test HTML tag without id", async () => {
```
Eric-Arellano, line 1473: Maybe add a comment linking to the original issue. It's hard to know what the problem is here since there is so much HTML.

Some test authors will even put the issue # explicitly in the test description.

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -193,7 +193,8 @@ export async function sphinxHtmlToMarkdown(options: {
       .map((child) => {
         const $child = $page(child);
         $child.find(".viewcode-link").closest("a").remove();
-        const id = $dl.find("dt.sig-object").attr("id");
+        const findID = $dl.find("dt.sig-object").attr("id");
+        const id = findID != undefined ? findID : "";
```
Eric-Arellano, line 197: I believe this will work:

```suggestion
        const id = $dl.find("dt.sig-object").attr("id") || "";
```

Eric-Arellano (commented):
Thanks for looking into this all!

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
arnaucasau: That's better!

scripts/commands/updateApiDocs.ts
arnaucasau: I removed the changes in `updateApiDocs.ts` and updated the `generateToc.ts` file to not use /index in any URL.

Eric-Arellano (approved):
🚀

PR #513, opened 2023-12-11
Why doesn't Tobi appear as an assignee in the feedback issue template?

abbycross:
The code in the yml file seems correct; trying to rearrange the names to see if there is a limit to the number of assignees


abbycross:
We figured out the issue - Tobi needed to be added to our repo. Closing!

PR #514, opened 2023-12-12
[Refactor]: Pull update url logic into own function

frankharkins:
Minor refactor in anticipation of upcoming work

### DIFF
--- a/scripts/lib/sphinx/updateLinks.test.ts
+++ b/scripts/lib/sphinx/updateLinks.test.ts
@@ -11,7 +11,7 @@
 // that they have been altered from the originals.
 
 import { describe, expect, test } from "@jest/globals";
-import { updateLinks } from "./updateLinks";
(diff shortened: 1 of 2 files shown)


Eric-Arellano (commented):
Great cleanup! 

Could you please take a look at `scripts/lib/sphinx/updateLinks.test.ts`? Most of them should now use this dedicated unit test. It's important we ensure this new helper function works robustly since it will be used in multiple places.

But it'd be good to have at least one test still for `updateLinks`, specifically checking that the logic remaining in that helper function works properly. Don't focus on edge cases within updateUrl, but rather edge cases within updateLinks.

frankharkins:
Thanks. I've added a test for the `updateUrl` function, but also left the other tests in too as I only expect to use `updateUrl` within `updateLinks`.

Eric-Arellano (approved):

PR #516, opened 2023-12-12
Suggestion: Do not create preview if no content changed

frankharkins:
I believe content changes are the only changes that should affect the preview. If so, it's a bit wasteful to create a preview for non-content changes, plus it clutters up the PR thread.

### DIFF
--- a/.github/workflows/preview-docs.yml
+++ b/.github/workflows/preview-docs.yml
@@ -23,6 +23,7 @@ on:
     types: [opened, reopened, synchronize, closed]
     paths:
       - "docs/**/*"
+      - "public/images/**/*"
 
 jobs:
   setup:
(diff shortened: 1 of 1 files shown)


.github/workflows/preview-docs.yml
```diff
@@ -21,6 +21,8 @@ name: Preview
 on:
   pull_request:
     types: [opened, reopened, synchronize, closed]
+    paths:
+      - "docs/**/*"
```
Eric-Arellano, line 25:
```suggestion
      - "docs/**/*"
      - "public/images/**/*"
```

Eric-Arellano (approved):
This is a great idea. CI has a real environmental and monetary impact!

PR #526, opened 2023-12-15
Fix notebook testing CI

frankharkins:
This PR addresses a couple of edge cases I didn't consider in #483.

- ~~Adds a token for `qiskit_ibm_provider`, and checks for trailing jobs as we already do for runtime.~~
  
  Turns out provider and runtime share saved account info, and `QiskitRuntimeService.jobs` includes provider jobs, which means the script already handles this fine.
- Don't run job-submitting notebooks without the `--submit-jobs` flag, even if passed explicitly.

  Currently, PR tests will fail if we edit these notebooks as the script will try to run them and they'll time out.
  To avoid having two lists of "notebooks that submit jobs" (one in the script and one in the action), I'm requiring the flag.

### DIFF
--- a/.github/workflows/notebook-test.yml
+++ b/.github/workflows/notebook-test.yml
@@ -55,6 +55,7 @@ jobs:
         if: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
         shell: python
         run: |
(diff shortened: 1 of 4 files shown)


Eric-Arellano (approved):
Thanks!

PR #527, opened 2023-12-15
Upload notebooks in CI

### LINKED ISSUE #523: Upload CI artifacts for executed notebooks
Follow up to https://github.com/Qiskit/documentation/issues/166. 

It's useful to upload the result of executing notebooks because it allows authors to get a copy of the notebooks from a standardized environment. For example, if they're having issues on macOS, they will be able to use the Linux environment from CI.

It's also useful from a transparency perspective: we can see how notebooks rendered at any point of time (outside of artifact) expiration), so we can e.g. look for when regressions were introduced into the notebook.

frankharkins:
Closes #523.

See https://github.com/Qiskit/documentation/actions/runs/7225258563 for an example of this working.


.github/workflows/notebook-test.yml
```diff
@@ -70,4 +73,10 @@ jobs:
...
 
       - name: Run tox
-        run: tox -- ${{ steps.all-changed-files.outputs.all_changed_files }}
+        run: tox -- ${{ steps.all-changed-files.outputs.all_changed_files }} --write
+
+      - name: Upload executed notebooks
+        uses: actions/upload-artifact@v4
+        with:
+          name: Executed notebooks
+          path: ${{ steps.all-changed-files.outputs.all_changed_files }}
```
Eric-Arellano, line 82: Instead, what we did in Qiskit CI is write the notebooks to a dedicated folder `executed_tutorials`. Rather than mutating the existing files.

This currently will upload more files than relevant. If you make this change, I think you can revert your changes to defining `all_changed_files`.

If you still want the script by default to modify in-place, then maye add an argument like `--output-dir`.

Eric-Arellano (commented):
Thanks!

.github/workflows/notebook-test.yml
frankharkins: Good call, the action wasn't handling multiple notebooks correctly. It should be fixed now; see the artifact for [run#7298447436](https://github.com/Qiskit/documentation/actions/runs/7298447436).

README.md
```diff
@@ -130,6 +130,12 @@ pipx install tox
 > add it to the "Check for notebooks that require LaTeX" step in
 > `.github/workflows/notebook-test.yml`.
 
+When you make a pull request with a changed notebook, you can get an version of
```
Eric-Arellano, line 133:
```suggestion
When you make a pull request with a changed notebook, you can get a version of
```

.github/workflows/notebook-test.yml
```diff
@@ -70,4 +74,15 @@ jobs:
           key: ${{ hashFiles('scripts/nb-tester/requirements.txt') }}
 
       - name: Run tox
-        run: tox -- ${{ steps.all-changed-files.outputs.all_changed_files }}
+        shell: python
+        run: |
+          import subprocess
+          files = """${{ steps.all-changed-files.outputs.all_changed_files }}"""
+          args = ["tox", "--"] + files.split("\n") + ["--write"]
+          subprocess.run(args)
```
Eric-Arellano, line 82: Otherwise you won't propagate the error code:

```suggestion
          subprocess.run(args, check=True)
```

.github/workflows/notebook-test.yml
```diff
@@ -31,6 +31,10 @@ jobs:
       - name: Get all changed files
         id: all-changed-files
         uses: tj-actions/changed-files@af2816c65436325c50621100d67f6e853cd1b0f1
+        with:
+          files: docs/**/*.ipynb
+          files_ignore: docs/api/**
```
Eric-Arellano, line 36: Not actually necessary. API docs never have notebooks. And if we had them in `api/migration_guides`, those should be checked.

```suggestion
```

Eric-Arellano (approved):
Thanks!

PR #531, opened 2023-12-18
Update qubit index in  qiskit.quantum_info.Pauli.md

Gistbatch:
I assume this is a typo in the docs.


Eric-Arellano:
Thank you @Gistbatch - I appreciate you taking the time to improve docs!

Because this is for API docs, we need to update Qiskit itself. We'd then regenerate the API docs with your fix to pull in the change here.

Can you please open a PR in https://github.com/Qiskit/qiskit to modify these lines and wherever else is relevant:

https://github.com/Qiskit/qiskit/blob/7e73fcf51e1d3557e0f291bb757248ba14d39ebe/qiskit/quantum_info/operators/symplectic/pauli.py#L82-L86

PR #537, opened 2023-12-19
[WIP] Handle Qiskit release notes structure correctly

### LINKED ISSUE #359: Handle Qiskit release notes structure correctly to split out the notes by version
Follow up to https://github.com/Qiskit/documentation/issues/294 and https://github.com/Qiskit/documentation/issues/281. 

@frankharkins found that it works well using the HTML generation script. A key change from Qiskit itself is we have one page per minor release, like `release-notes/0.45`. We need to make two improvements:

1. Restructure Qiskit/qiskit to have one release note file per release for future releases. Whereas right now they have the file `release_notes` for all notes from Qiskit 0.45 forward
2. Update our script `updateApiDocs.ts` to handle new release notes automatically, including the wonkiness of the prior release note files (`legacy_release_notes.html`).

arnaucasau:
**This is a work in progress**

This PR changes how we write the Qiskit release-notes.html to correctly split the notes by version, updating the historical files if we find any change. If we already have a file for the version we updated, the script adds the new entries without overwriting the file and sorts the entries by patch. If the patch we modified was already in the version file, it substitutes its section for the new one.

Qiskit v0.45 has been regenerated using the new script to split correctly versions 0.45 and 0.25, which we can find in the release notes file of that version. Command used:

`npm run gen-api -- -p qiskit -v 0.45.0 -a https://github.com/Qiskit/qiskit/suites/17881600359/artifacts/1026798160`

Closes #359


scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +151,147 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
     });
   }
 }
+
+/**
+ * Sorts the release notes dictionary by the patch version
+ */
+export function sortReleaseNotesVersions(markdownByPatchVersion: {
+  [id: string]: string;
+}) {
```
Eric-Arellano, line 160: Needs a return type annotation

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +151,147 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
     });
   }
 }
+
+/**
+ * Sorts the release notes dictionary by the patch version
```
Eric-Arellano, line 156:
```suggestion
 * Sorts the release notes dictionary by the patch version
 * in descending order, e.g. 0.25.2 then 0.25.1.
```

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +151,147 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
...
+  const markdownByPathEntries = Object.entries(markdownByPatchVersion);
+
+  const markdownByPatchSortedArray = markdownByPathEntries.sort(
+    ([version1], [version2]) => {
+      const versionPatch1 = version1.split("rc").slice(0, 1)[0];
+      const versionPatch2 = version2.split("rc").slice(0, 1)[0];
+      const comparison = version2.localeCompare(version1);
+
+      if (versionPatch1 == versionPatch2) {
+        // The release candidates should appear first
```
Eric-Arellano, line 170: I don't think this is fully considering that it is legal to have `0.45.1rc1`, i.e. release candidates for patch releases. `0.45.1rc1` needs to be seen as a newer release than `0.45.0`, but it is older than `0.45.1`. This would be good to unit test and use test-driven development for.

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +151,147 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
...
+  const sections: string[] = sectionsSplit.slice(1, sectionsSplit.length);
+
+  const minorVersionsFound = new Set<string>();
+  const markdownByPatchVersion: { [id: string]: string } = {};
+
+  for (let section of sections) {
+    const versionPatch = section.split("\n").slice(0, 1)[0];
+    const versionMinor = versionPatch.split(".").slice(0, 2).join(".");
+
+    // Skip legacy release notes
```
Eric-Arellano, line 206:
```suggestion
    // TODO(https://github.com/Qiskit/qiskit/pull/11436). Remove if/when that PR is merged.
    // See that PR for why we have this special logic.
```

This code is a bummer because it breaks the abstraction that releaseNotes.ts had. The code works with any
arbitrary package. But here we're hardcoding in Qiskit-specific logic.

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +151,147 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
...
+    markdownByPatchVersion[versionPatch] = `## ${versionPatch}\n${content.join(
+      "\n",
+    )}`;
+  }
+
+  return [minorVersionsFound, markdownByPatchVersion];
+}
+
+/**
+ * Updates the release notes files by adding the notes to their corresponding version
```
Eric-Arellano, line 224:
```suggestion
 * Updates the release notes folder by adding the notes to their corresponding version
```

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +151,140 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
...
+ * file.
+ */
+export async function writeReleaseNotes(pkg: Pkg, releaseNoteMarkdown: string) {
+  // Dictionary to store the file header in case we have to reconstruct a historical file
+  const FilesHeaders: { [id: string]: string } = {};
```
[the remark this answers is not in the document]
Eric-Arellano: Ah. I see now what this variable is used for. I recommend calling it something like `filePathToFrontMatter`. "Front matter" is the "term of art" for this metadata. You could also call it `filePathToMetadataBlock`. 

"Header" confused me because I was thinking of a markdown header like `#`.

Eric-Arellano (commented):
Thanks for diving into this! I know it's pretty complex logic. You did a great job here abstracting it well and writing effective tests so we can be confident it works.

scripts/lib/releaseNotes.ts
arnaucasau: You are totally right! I think we should keep the conditional until qiskit `0.45` is not a historical version. Given that we use CI artifacts, we will have the `0.25` version even with the PR merged when regenerating that version. I changed the comment, let me know what you think :+1:

scripts/lib/releaseNotes.ts
arnaucasau: Tested in the new test, and the case is handled. We only enter in the if condition if the two versions have the same patch, like `0.45.1` and `0.45.1rc1`. If the patch is different, the comparison we use is the one with localCompare.

scripts/lib/releaseNotes.test.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    expect(
+      Object.entries(sortReleaseNotesVersions(markdownByVersionPatch)),
+    ).toEqual([
+      ["0.46.1", "test"],
+      ["0.45.1", "test"],
+    ]);
+  });
+
+  test("Sort bigger version and smaller version", () => {
```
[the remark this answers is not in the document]
arnaucasau: Oh, I don't know why I didn't think about it. I removed all the tests and combined them into one with different scenarios. Thanks for the comment!

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +151,140 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
...
+    extractMarkdownReleaseNotesPatches(releaseNoteMarkdown);
+
+  // Read the current release notes for each version found
+  for (let version of minorVersionsFound) {
+    const versionPath = `${basePath}${version}.md`;
+
+    if (!(await pathExists(versionPath))) {
+      // We don't have any release note file for that version
+      continue;
+    }
```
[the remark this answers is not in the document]
arnaucasau: This is an edge case possible if, for some reason, we generate two versions at the same time. For example, if we add a version using the historical argument, the release notes will not be created, and when we add the current version, the file for the former won't exist.

This shouldn't happen often, but if someday we have to regenerate several versions, for some reason, the script will handle the case correctly.

scripts/lib/releaseNotes.ts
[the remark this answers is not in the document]
arnaucasau: Oh, I didn't know the name of the metadata block! FilesHeaders also includes the markdown we add manually before the notes section, like the table in the `0.44.md file`. Maybe we will not modify the files manually ever again, but if we do, the markdown will be included. Do you think filesFrontMatter is still a good name in that case?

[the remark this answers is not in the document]
Eric-Arellano: Ah okay. `filesToInitialHeaders`? It's long, but unambiguous.

scripts/commands/updateApiDocs.ts
```diff
@@ -318,22 +318,25 @@ async function convertHtmlToMarkdown(
...
-      path = currentReleaseNotesPath(pkg);
+    if (!pkg.hasSeparateReleaseNotes || !path.endsWith("release-notes.md")) {
+      await writeFile(path, result.markdown);
+      continue;
+    }
+
+    // Historical versions use the same release notes files as the current API
+    if (pkg.historical) {
+      continue;
     }
```
Eric-Arellano: Oh hm, I'm no longer convinced this is true. Imagine that current is 1.0.0, which we released last week. We just released 0.46.2 today, and now we're generating 0.46.2 with `--historical`. The release notes from last week for 1.0.0 will not yet include 0.46.2 because it didn't exist at the time.

I think that we do need to consider historical release notes. We only need to be careful to not delete the notes for future versions; when generating `-v 0.46.2 --historical`, we shouldn't touch `release-notes/1.0.md`.

--

I'm fine if this is fixed in a followup. This PR already has some complexity and it might be easier to address this edge case in a dedicated change.

scripts/lib/releaseNotes.test.ts
```diff
@@ -0,0 +1,94 @@
...
+// that they have been altered from the originals.
+
+import { describe, expect, test } from "@jest/globals";
+import {
+  sortReleaseNotesVersions,
+  extractMarkdownReleaseNotesPatches,
+} from "./releaseNotes";
+
+describe("sortReleaseNotesVersions", () => {
+  test("Test versions have the correct order", () => {
```
Eric-Arellano, line 20: No need to use `describe` since it's a single test. You can use `test` at the top-level.

scripts/lib/releaseNotes.test.ts
```diff
@@ -0,0 +1,94 @@
...
+      ["0.46.1rc2", ""],
+      ["0.46.1rc1", ""],
+      ["0.45.1", ""],
+      ["0.45.1rc1", ""],
+      ["0.45.0", ""],
+    ]);
+  });
+});
+
+describe("extractMarkdownReleaseNotesPatches", () => {
```
Eric-Arellano, line 47: Ditto on removing `describe`.

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +151,161 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
...
+/**
+ * Sorts the release notes dictionary by the patch version
+ * in descending order, e.g. 0.25.2 then 0.25.1.
+ *
+ * Returns a dictionary with the corresponding markdown for
+ * each patch with the entries sorted by release.
+ */
+export function sortReleaseNotesVersions(markdownByPatchVersion: {
+  [id: string]: string;
+}) {
```
Eric-Arellano, line 164: The comment is helpful - but this still needs a type annotation for the return type.

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +151,161 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
...
+  const markdownByPathEntries = Object.entries(markdownByPatchVersion);
```
Eric-Arellano, line 165: Nit: it's probably clearer to inline this, since you're only using the variable once. I personally find `Object.entries(markdownByPatchVersion).sort()` clearer because I'm thinking in terms of the `Object.entries()` API, rather than having to remember what the variable `markdownByPathEntries` looks like.

scripts/commands/updateApiDocs.ts
```diff
@@ -331,9 +331,10 @@ async function convertHtmlToMarkdown(
           : `/api/${pkg.name}/${link}`,
       );
 
-      path = currentReleaseNotesPath(pkg);
+      await writeReleaseNotes(pkg, result.markdown);
+    } else {
+      await writeFile(path, result.markdown);
     }
```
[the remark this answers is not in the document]
arnaucasau: Looking again at that `if` block, I think there's a bug. It wasn't doing what we wanted because, for runtime and provider, we are generating a release notes file even with the historical argument. I divided it into two early returns, one for the historical versions, and another for the separate release notes. The general case is writing the files as usual (including all API docs).

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +151,161 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
...
+
+  const markdownByPatchSortedArray = markdownByPathEntries.sort(
+    ([version1], [version2]) => {
+      const versionPatch1 = version1.split("rc").slice(0, 1)[0];
```
Eric-Arellano, line 169: A comment would probably help to explain how this code works, including maybe an example string.

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +151,161 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
...
+      const versionPatch2 = version2.split("rc").slice(0, 1)[0];
+      const comparison = version2.localeCompare(version1);
+
+      if (versionPatch1 == versionPatch2) {
+        // The release candidates should appear first
```
Eric-Arellano, line 174:
```suggestion
        // The release candidates within the same patch should appear first.
```

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +151,161 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
...
+export function sortReleaseNotesVersions(markdownByPatchVersion: {
+  [id: string]: string;
+}) {
+  const markdownByPathEntries = Object.entries(markdownByPatchVersion);
+
+  const markdownByPatchSortedArray = markdownByPathEntries.sort(
+    ([version1], [version2]) => {
+      const versionPatch1 = version1.split("rc").slice(0, 1)[0];
+      const versionPatch2 = version2.split("rc").slice(0, 1)[0];
+      const comparison = version2.localeCompare(version1);
```
Eric-Arellano, line 171: I recommend inlining this and only defining it where actually used, on line 181. Right now, the reader has to wonder what this will be used for.

scripts/lib/releaseNotes.ts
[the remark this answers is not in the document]
Eric-Arellano: Oh - I think this probably speaks to my other comment about realizing I think we have a bug how `--historical` stops us from generating release notes?

Eric-Arellano (commented):
Looking good!

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +149,167 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
...
+
+  for (let section of sections) {
+    const versionPatch = section.split("\n").slice(0, 1)[0];
+    const versionMinor = versionPatch.split(".").slice(0, 2).join(".");
+
+    // TODO(https://github.com/Qiskit/qiskit/pull/11436). Remove if/when qiskit 0.45
+    // has been moved into a historical folder.
+    if (+versionMinor < 0.45) {
+      continue;
+    }
```
Eric-Arellano, line 218: This PR has been merged and cherry-picked to the stable/0.45 branch.

arnaucasau: Removed! Thanks Eric!

scripts/commands/updateApiDocs.ts
```diff
@@ -318,21 +318,29 @@ async function convertHtmlToMarkdown(
...
   for (const result of results) {
     let path = urlToPath(result.url);
-    if (pkg.hasSeparateReleaseNotes && path.endsWith("release-notes.md")) {
-      // Historical versions use the same release notes files as the current API
-      if (pkg.historical) {
-        continue;
-      }
 
+    // Historical versions with a single release notes file should not
+    // modify the current API's file.
```
Eric-Arellano, line 323: Great comment and code change. This is now much more clear to me! It's "less DRY", but substantially clearer.

scripts/lib/releaseNotes.ts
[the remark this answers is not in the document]
Eric-Arellano: I continue to be confused about this point. 

> we generate two versions at the same time.

That's not technically possible. The script only can generate one release at a time. 

> For example, if we add a version using the historical argument, the release notes will not be created, and when we add the current version, the file for the former won't exist.

I don't think I'm understanding. If the historical arg has 0.44 and 0.45, then we should expect both those release note files to exist. Now we generate current, which has 0.44, 0.45, and 0.46: it should generate all three of those files.

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +149,161 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
...
+      .slice(0, 1)[0];
+
+    const [_, markdownByPatchOldVersion] =
+      extractMarkdownReleaseNotesPatches(currentMarkdown);
+
+    for (let [versionPatch, markdownPatch] of Object.entries(
+      markdownByPatchOldVersion,
+    )) {
+      // We keep the release notes for a patch if it hasn't been modified for the current release notes.
+      // Otherwise, we keep the modified version
```
Eric-Arellano, line 269: Nit: `keep` suggests to me that we are not modifying anything

```suggestion
      // Otherwise, we use the modified version.
```

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -84,6 +84,12 @@ export async function sphinxHtmlToMarkdown(options: {
       const dest = `${imageDestination}/${filename}`;
 
       $img.attr("src", dest);
+
+      if (isReleaseNotes && releaseNotesTitle) {
+        // Release notes links should point to the current version
+        $img.attr("src", dest.replace(/[0-9].*\//, ""));
```
Eric-Arellano, line 90: Interesting. What was happening before this change?

Eric-Arellano (commented):
This overall looks great! I only have one doubt still. Thanks for working on this!

scripts/lib/releaseNotes.ts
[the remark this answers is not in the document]
arnaucasau: > That's not technically possible.

Yes, sorry, I meant in the same PR. In a situation where you skipped one version, and you want now to add the current and the previous you skipped.

> I don't think I'm understanding.

The problem that I wanted to describe can't happen anymore because now, the script generates release notes even when we use the historical argument. However, the problem is still there.

When you generate a version for the first time, the release notes for that new version are not present in the release notes folder. The script will read the release notes in order to find old versions we modified, and therefore we need to update from the files in the folder. To update the old versions we need to read the current release notes for those versions. The catch is that the script will try to read the current version as well, but that file doesn't exist yet, so we need to prevent it from breaking.

We need to read the current release notes file because we don't want to overwrite the whole file we had. There are some parts like the front matter or text added manually (like the table in qiskit v0.44) that we want to preserve. In addition, if we don't modify a version, we want to have the notes and not lose them (not sure if this can happen because of the way Qiskit generates the release notes now though).

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
arnaucasau: Before this change, if we used the historical argument, the release notes had all the links correct but the images. The images were pointing to the historical folder instead of the current version. For example, `public/images/api/qiskit/0.44/circuit-1.png` instead of `public/images/api/qiskit/circuit-1.png`

scripts/lib/releaseNotes.ts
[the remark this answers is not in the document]
arnaucasau: > That's not technically possible.

Yes, sorry, I meant in the same PR. In a situation where you skipped one version, and you want now to add the current and the previous you skipped as historical without having it generated as current before.

> I don't think I'm understanding.

The problem that I wanted to describe can't happen anymore because now, the script generates release notes even when we use the historical argument. However, the problem is still there.

When you generate a version for the first time, the release notes for that new version are not present in the release notes folder. The script will read the release notes (html converted to md) in order to find old versions we modified, and we need to update from the files in the folder. To update the old versions, we need to read the current release notes for those versions. The catch is that the script will try to read the current version as well, but the file doesn't exist yet, so we need to prevent it from breaking.

We need to read the current release notes files because we don't want to overwrite everything we had. There are some parts like the front matter or text added manually (like the table in qiskit v0.44 or a custom description) that we want to preserve. Your comment made me realize that we don't need to go through each section because the release notes should contain all the versions by default, so we can just overwrite that part with the new content. In the last commit, I removed some unnecessary code. Thank you Eric.

scripts/lib/releaseNotes.ts
```diff
@@ -150,3 +149,148 @@ function addNewReleaseNoteToc(releaseNotesNode: any, newVersion: string) {
...
+    extractMarkdownReleaseNotesPatches(releaseNoteMarkdown);
+
+  // Read the current release notes for each version found
+  for (let version of minorVersionsFound) {
+    const versionPath = `${basePath}/${version}.md`;
+
+    if (!(await pathExists(versionPath))) {
+      // We don't have any release note file for that version
+      continue;
+    }
```
Eric-Arellano, line 255: Okay, I think I finally get it 😅

```suggestion
    // When we're adding a version, its release note file will not yet exist, so we
    // cannot read it here to determine the prior contents. Instead, skip the relese
    // note since we are going to add it later in updateApiDocs.ts.
    if (!(await pathExists(versionPath))) {
      continue;
    }
```

Eric-Arellano (approved):

Issue #476, opened 2023-12-05
[Epic] Add translations to non-API docs.


Eric-Arellano:
```[tasklist]
### Tasks
- [x] https://github.com/Qiskit/documentation/issues/7
- [x] https://github.com/Qiskit/documentation/pull/541
- [ ] Coordinate with designers to decide if the language switcher will be only for docs, or unified with Learning app
- [ ] Coordinate with ibm.com team to add language switcher & implement URL changing
- [ ] https://github.com/Qiskit/documentation/issues/185 and also ban using header anchors (except API docs)
```

Issue #7, opened 2023-09-06
Figure out folder structure for translations


Eric-Arellano:
We need to figure out URL scheme and folder organization for docs, which are tied together.

## URL/file names will stay the same

That is, we don't try translating `/start/my_file` to `/empieza/mi_archivo` for example. It's a common convention in localization to use consistent URLs and it makes everything much easier to implement.

## Proposed URL scheme

English URLs will still the same w/o language in the URL/file, e.g. `/start`. But English _does_ still need to show up in the language toggle.

Translations will start with `/<lang>` in their URL, e.g. `/es/start` and `/fr/run`. Everything after in the URL will be the same. Each language will have a full copy of the normal documentation but translated, and excluding API docs.

To communicate the language information to the docs app, we could reuse `_package.json` from https://github.com/Qiskit/docs/issues/8? Or add `_language.json`? Whatever makes the most sense for the application.

```json
{"code": "es", "name": "Spanish"}
```

```json
{"code": "zh-TW", "name": "Chinese (traditional)"}
```

Similar to how version toggling works, changing the language via the language toggle would change which top-level language folder is loaded (or English). It changes the language setting for the entire Docs app.

## Locales vs languages

In qiskit.org, we had a concept of locales: the combination of language and region, e.g. `es_UN` for Spanish universal. The region part is useful to do things like show currencies and times/dates differently. 

But in practice, we only had one locale per language, _except for_ Chinese having both simplified and traditional.

Given the nature of the Qiskit documentation, I propose that we simplify and only use langauge codes like `/es` and `/fr`, without locales for every language but Chinese. It seems unlikely that we'll have the bandwidth to differentiate by locale. We didn't have the bandwidth before. It also doesn't seem super critical.

Chinese would be `zh-CN` (simplified) and `zh-TW` (traditional), though.

We could always add locales in the future to the other languages. We could set up redirects from `/es/` to `/es-UN/`, for example.

#### FYI: how to integrate with Crowdin

Crowdin exports the folders with the locale as the prefix, e.g. `es-es/`. I couldn't change this behavior via their `language mapping` mechanism, even though [the docs](https://support.crowdin.com/file-management/#file-export) suggest it should have worked. For all but Chinese, we'll post-process the folder after downloading translations to remove the locale portion from the folder.

### `_toc.json`

This file controls the names of things in the left sidebar, and what URLs they take you to.

We should not change the URLs. But it makes sense to translate the page titles.

```json
  "children": [
    {
[...]

PR #541, opened 2023-12-20
Add Crowdin to set up translations

Eric-Arellano:
Part of  https://github.com/Qiskit/documentation/issues/476. 

This PR sets up two GitHub actions for interfacing with Crowdin.com, the website the translation team uses for translations.

1. Upload English sources. We do this on any change to `docs/` (except API docs), which the translation team so that they can start making updates sooner. For example:

Screenshot 2023-12-20 at 5 45 05 PM

2. Download the translations. This happens once a week, which the translations team requested. It will open up a pull request like https://github.com/Eric-Arellano/documentation/pull/10. 

## How auth works

We only need to set GitHub Secrets for CROWDIN_PROJECT_ID and CROWDIN_PERSONAL_TOKEN. GitHub will automatically set GITHUB_TOKEN using [this mechanism](https://docs.github.com/en/actions/security-guides/automatic-token-authentication).

## Translations saved at `translations/` folder

This adds a top-level `translations/` folder, rather than nesting the translations inside the `docs/` folder. For example:

Screenshot 2023-12-20 at 5 48 09 PM

This top-level folder is useful because it reduces noise in the `docs/` folder. It makes it more obvious to the team what is the original source docs vs. what is translations and should not be modified directly in this repository.

### How this works with the app

To get docs previews working in this repo, we teach the Docker image to mount `translations/` as if its contents were in the folder `docs/`, since the application expects all content to live there.

When we sync the translations from open-source to closed-source for production, we will copy the contents of `translations/` so that it actually lives underneath `docs/`.  For example, `translations/es/support.mdx` becomes `docs/es/support.mdx`.

## Locales

As explained in https://github.com/Qiskit/documentation/issues/7, we generally want to use the two-letter code for languages, like `es` and `fr`. The only exception is Chinese, where we need to distinguish between Traditional and Simplified.

This config works with that, so long as in Crowdin.com, the project sets up this Language Mapping under the Languages settings:

Screenshot 2023-12-20 at 5 59 19 PM

## Doesn't yet add `_languages.json`

As explained in https://github.com/Qiskit/documentation/issues/7, we will probably want a file like `_languages.json` that teaches the docs app what each language is, such as its name to show in the language toggle. That schema still needs to be finalized and it can be added in a follow up.


start
```diff
@@ -13,10 +13,22 @@
...
 
+def translation_volume_mounts() -> Iterator[str]:
+    """Return the CLI args to mount each language in the translations/ folder.
+
+    Unlike the root `Dockerfile`, we cannot use `-v $PWD/translations:/home/node/app/docs` because
+    Docker complains that the volume is already mounted when we mount the `/docs` folder. So, instead
+    we need to be more specific to mount each language's folder.
+    """
+    for folder in PWD.glob("translations/*"):
+        yield from ["-v", f"{folder}:/home/node/app/docs/{folder.name}"]
```
Eric-Arellano, line 29: This `yield from` is a nifty Python feature. This function is a _generator_, meaning it only returns the next element once we iterate to it. That is rather than storing everything all at once in memory. This is also how `PWD.glob()` works.

Here, I mostly use the iterator for concise code. Rather than having a `result: list[str]` variable etc.

Eric-Arellano (commented):
I have a working example over at https://github.com/Eric-Arellano/documentation :) 

See https://github.com/Eric-Arellano/documentation/actions/runs/7281421497 for an English upload, and https://github.com/Eric-Arellano/documentation/pull/11 for a PR to download the translations.

@SooluThomas has helped to set up the Crowdin Token, so this PR should be safe to merge. Once merged, I'll run the GitHub Action to upload English sources to Crowdin.

crowdin.yml
```diff
@@ -0,0 +1,22 @@
...
+files:
+  [
+    {
+      "source": "docs/**/*",
+      "translation": "translations/%two_letters_code%/%original_path%/%original_file_name%",
+      "ignore":
+        [
+          "docs/api/qiskit/**/*",
+          "docs/api/qiskit-ibm-runtime/**/*",
+          "docs/api/qiskit-ibm-provider/**/*",
```
SooluThomas, line 14: wouldn't it simply be easy to have

```
"/docs/api",
```

instead?

docs/placeholder.txt
SooluThomas, line 1: What is this file for?

translations/placeholder.txt
SooluThomas, line 1: 👍

crowdin.yml
Eric-Arellano: No because of `docs/api/migration-guides`.

SooluThomas: Do you want the migration guides to be translated?

Eric-Arellano: I'm not sure - what do you think? See https://docs.quantum.ibm.com/api/migration-guides.

SooluThomas: I'll ask the team 👍

SooluThomas: Ok, let's keep it then :D

SooluThomas (approved):

docs/placeholder.txt
Eric-Arellano: Oh, I didn't realize I added it here too. Oops. Good catch!

arnaucasau (approved):
Looks good to me!

Eric-Arellano:
Going to merge so that we can get Crowdin set up before the new year. I'm happy to make any adjustments in follow up PRs. cc @frankharkins

Issue #542, opened 2023-12-21
Missing method pages for Qiskit API docs before Qiskit 0.44


Eric-Arellano:
Until Qiskit 0.44 via https://github.com/Qiskit/qiskit/pull/10455, in Qiskit docs, methods lived on dedicated HTML pages. 

~For some reason, when generating the historical API docs for Qiskit, the script failed to generate those method pages and it only generated class pages, module pages, and function pages. So, our redirects from qiskit.org to ibm.com are working properly but resulting in 404s because the method pages are not there.~

This would have been caught by https://github.com/Qiskit/documentation/issues/495.

As explained in https://github.com/Qiskit/documentation/pull/543, we intentionally inlined methods/attributes into their class page via the file `mergeClassMembers.ts`. That in general worked great, but there are a couple of instances where it fails, like https://docs.quantum.ibm.com/api/qiskit/0.33/qiskit.result.Result mentioned in that PR description. We need to figure out why those failed and get them to work consistently.

The other issue was that I did not realize this inlining was happening when setting up redirects from qiskit.org to ibm.com, so we were trying to redirect method pages to non-existent method pages. I have a PR up for review to fix our redirect logic to instead go to the parent class's page.

To close this issue, we need to:

1. Properly inline all methods. That can be verified by running our link checker over these API docs and making sure that all internal links work.
2. Land the redirects change.

Eric-Arellano:
Every version of the API docs has been regenerated. This should be fixed. Thanks @arnaucasau 🚀

Issue #479, opened 2023-12-06
[ EPIC ] Improve API Ref UX/UI


javabster:
We've been receiving a lot of feedback about the UX/UI of our API reference pages so I'm tracking them all together as one epic here. Some of these may be due to bugs in the API generation script, others might just be features that we forgot to include that were present in qiskit.org and now users are noticing they've disappeared. Either way they all come under the general umbrella of "things we should address to make the API references better". There may be some issues we could close as duplicates

```[tasklist]
### Information hierarchy / headers
- [ ] #231 
- [ ] #274 
- [ ] #167 
- [ ] #712
- [ ] #714
- [ ] https://github.com/Qiskit/documentation/issues/942
- [ ] https://github.com/Qiskit/documentation/issues/1276
- [ ] https://github.com/Qiskit/documentation/issues/941
- [ ] https://github.com/Qiskit/documentation/issues/932
- [ ] https://github.com/Qiskit/documentation/issues/1098
- [ ] https://github.com/Qiskit/documentation/issues/1215
- [ ] https://github.com/Qiskit/documentation/issues/1272
- [ ] https://github.com/Qiskit/documentation/issues/1196
- [ ] https://github.com/Qiskit/documentation/issues/1352
- [ ] https://github.com/Qiskit/documentation/issues/1388
- [ ] https://github.com/Qiskit/documentation/issues/1389
- [ ] https://github.com/Qiskit/documentation/issues/1395
```

```[tasklist]
### Visual design
- [ ] #521
- [ ] https://github.com/Qiskit/documentation/issues/948
- [ ] https://github.com/Qiskit/documentation/issues/947
- [ ] #735 
- [ ] #736
- [ ] https://github.com/Qiskit/documentation/issues/856
- [ ] https://github.com/Qiskit/documentation/issues/1216
- [ ]  #468 
- [ ] https://github.com/Qiskit/documentation/issues/1131
- [ ] https://github.com/Qiskit/documentation/issues/1287
- [ ] https://github.com/Qiskit/documentation/issues/1361
```

```[tasklist]
### Rendering bugs
- [ ] #59
- [ ] #732
- [ ] #488
- [ ] #485
- [ ] https://github.com/Qiskit/documentation/issues/814
- [ ] https://github.com/Qiskit/documentation/issues/769
- [ ] https://github.com/Qiskit/documentation/issues/755
- [ ] https://github.com/Qiskit/documentation/issues/1200
[...]

PR #543, opened 2023-12-21
Allow dedicated page for each method

arnaucasau:
Part of fixing https://github.com/Qiskit/documentation/issues/542.

## Methods living on dedicated pages

Before Qiskit 0.44, methods lived on dedicated HTML pages. We changed that in Qiskit 0.44 because it resulted in horrible build-time performance when migrating `qiskit.org/documentation` from the old Pytorch theme to the new Furo theme, so that instead methods lived directly on their class pages.

It looks like the script `updateApiDocs.ts` was trying to emulate methods living on class pages before Qiskit 0.44 through the file `mergeClassMembers.ts`. This code attempted to remove the method pages and instead have them live directly on the class pages. 

I believe the original script author had `mergeClassMembers.ts` to try to avoid performance issues. Until a fix a few weeks ago to the docs application, having a huge amount of pages would cause the site to load extremely slowly due to the left side bar being too large. That is now fixed because the left side bar is loaded dynamically rather than statically.

## `mergeClassMembers.ts` was causing problems

1. It looks like it wasn't behaving properly, at least some of the time. We did not exhaustively audit every page, but for example, https://docs.quantum.ibm.com/api/qiskit/0.33/qiskit.result.Result only had properties at the top, then a table summarizing all the methods below, but not the methods actually inlined:

Screenshot 2023-12-21 at 11 43 47 AM

2. We didn't know about this inlining when setting up qiskit.org redirects, which do a 1:1 mapping so that dedicated method pages take you to the dedicated method page on IBM, which was 404ing.

## Solution: allow dedicated method pages

By deleting `mergeClassMembers.ts`, the script will properly generate dedicated HTML pages for each method. 

We took this approach—rather than instead fixing `mergeClassMembers.ts`—for a few reasons:

1. It allows us to avoid messing with our redirect rules, which are the 1:1 mapping. 
2. We've gotten user feedback that the API docs are not as usable as they could be. We are planning to fix that in https://github.com/Qiskit/documentation/issues/479, but it won't be fixed in time for putting out the fire of https://github.com/Qiskit/documentation/issues/542. In the meantime, it is more user-friendly to have dedicated method pages for those historical API docs.

The biggest reason we would not want to take this approach of dedicated method pages is the possible impact on performance mentioned above. But that has been fixed already in the IBM application.

## Impact of this PR

No impact when generating docs where methods already live on class pages, like Qiskit 0.44+.

For earlier Qiskit versions, it results in the method pages being generated. Those will be added in follow up PRs.


Eric-Arellano (approved):
Thanks!

PR #545, opened 2023-12-21
Rename terra.md to index.md

Eric-Arellano:
When generating docs for Terra before 0.44, the index page was called `terra.html`. We need it to be called `index`.


arnaucasau (approved):
LGTM! Thank you!

PR #546, opened 2023-12-21
Add missing method pages for Qiskit 0.19 to Qiskit 0.33

arnaucasau:
Part of fixing https://github.com/Qiskit/documentation/issues/542. We will regenerate Qiskit 0.32 and 0.34-0.43 in a follow up PR because we are still getting the original source docs for those.

Every version was generated like this: `npm run gen-api -- -p qiskit -v 0.31.0 --historical -a ""`.

Note that this also brings in a recent change that results in removing some blank lines we had before.


Eric-Arellano:
This PR made me realize that the approach we took in https://github.com/Qiskit/documentation/pull/543 is not the best one. This PR regresses the UX quite a bit:

1. The method pages don't ever mention the class, so they're very confusing what the page refers to. We'd need to fix this, which takes time
2. It results in the user having to hop around the page a bunch
3. It bloats the left-side bar and makes the specific API version slower to load since there are so many pages.

Instead, the new plan is to revert https://github.com/Qiskit/documentation/pull/543 and figure out why it wasn't consistently working. Fix those edge cases. Concurrently, fix our redirect logic so that methods always go to their parent class.

PR #549, opened 2023-12-22
Fix some async errors in updateApiDocs.ts

Eric-Arellano:
We weren't calling `await` on `startWebServer` and `mkdirp`. This could result in order-of-operation issues.

While fixing that, I realized that we can simplify the code. We only need to start the web server if we're downloading the artifact from CI. So, only have that code in the `else` block. We can reduce the amount of code by moving the server's start to inside `downloadApiSources`.

### DIFF
--- a/scripts/commands/updateApiDocs.ts
+++ b/scripts/commands/updateApiDocs.ts
@@ -120,13 +120,13 @@ const readArgs = (): Arguments => {
     .option("historical", {
       type: "boolean",
       default: false,
-      description: "Is this a prior release? Only works with `-p qiskit`.",
+      description: "Is this a prior release?",
     })
     .option("artifact", {
       alias: "a",
       type: "string",
       demandOption: true,
-      description: "Which artifact from CI to download",
+      description: "The URL for the CI artifict to download. Must be from GitHub Actions.",
     })
     .parseSync();
 };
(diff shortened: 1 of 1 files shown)


scripts/commands/updateApiDocs.ts
```diff
@@ -120,13 +120,14 @@ const readArgs = (): Arguments => {
...
-      description: "Is this a prior release? Only works with `-p qiskit`.",
+      description: "Is this a prior release?",
     })
     .option("artifact", {
       alias: "a",
       type: "string",
       demandOption: true,
-      description: "Which artifact from CI to download",
+      description:
+        "The URL for the CI artifict to download. Must be from GitHub Actions.",
```
Eric-Arellano, line 130:
```suggestion
        "The URL for the CI artifact to download. Must be from GitHub Actions.",
```

scripts/commands/updateApiDocs.ts
```diff
@@ -383,15 +380,15 @@ async function downloadApiSources(
...
     await downloadCIArtifact(pkg.name, artifactUrl, destination);
     await saveHtml({
       baseUrl: pkg.baseUrl,
       initialUrl: pkg.initialUrl,
       destination,
     });
-  } catch (e) {
+  } finally {
     await closeWebServer(listenPort);
-    throw e;
```
frankharkins, line 395: Nice! I was going to make an issue for this; I had the server linger a few times when I encountered an error between this function and the end of `zxMain`.

frankharkins (approved):
Looks good, thanks!

PR #550, opened 2023-12-22
Extract link checker ignores to a dedicated file

Eric-Arellano:
The config added too much noise to the file `checkLinks.ts`. That file should only have high-level logic. That allows us to more cleanly split out the list between `SHOULD_BE_FIXED` and `EXPECTED` so that we know what is a legit problem and we're only punting on it.

No changes were made to the actual list.


frankharkins (approved):

frankharkins:
Thanks!

PR #552, opened 2023-12-22
Inline all methods for Qiskit API

arnaucasau:
The generation script was not inlining all the methods in the class pages because we weren't assigning the correct value to the `python_api_name` metadata. With this PR, we change the way we search for the id we set as the name in the `sphinxHtmlToMarkdown.ts`

The problem comes from we are missing the class `sig-object` in the `dt` tags we used to extract the `python_api_name` metadata in the HTML used to generate the markdown API files. We actually don't need to specify the class because the `dt` tag we are looking for is unique, and this PR fixes that.

Moreover, the changes introduced in #543 are reverted to recover the mergeClassMembers.ts script that we need to inline all the methods in historical versions.

Tested the generation of Qiskit 0.33, where we will inline the methods in a follow-up, and also the generation of Qiskit 0.45 to validate we are not modifying any version that was working previously.

Commands used (Notice we need to have the API docs downloaded in .out):

> npm run gen-api -- -p qiskit -v 0.33.0 -a "" --historical
> npm run gen-api -- -p qiskit -v 0.45.0 -a ""

Part of #542


Eric-Arellano (approved):
🙌

PR #553, opened 2023-12-22
Deduplicate error messages in link checker for same link in same file

Eric-Arellano:
Otherwise, we get a couple of duplicate error messages. This happens when the same file has the same bad link >1 times.


arnaucasau (approved):
Looks good to me. Thanks for the change Eric! Tested locally, and all works well! :rocket:

PR #554, opened 2023-12-27
Fix API generation script

arnaucasau:
The API generation script is not downloading the images because we stopped the server after downloading the HTML sources and never started it again. This PR starts the server before downloading the images and adds validation for the server to be up. Without this validation, the script tries to download the images immediately and fails because the server doesn't have time to start. This validation wasn't necessary until now because we opened the server before downloading the CI artifact, giving enough time to avoid all problems.

### DIFF
--- a/scripts/lib/downloadImages.ts
+++ b/scripts/lib/downloadImages.ts
@@ -17,23 +17,37 @@ import { createWriteStream } from "node:fs";
 import { finished } from "stream/promises";
 import { Readable } from "stream";
 import pMap from "p-map";
+import { startWebServer, closeWebServer } from "../lib/webServer";
 
 export async function downloadImages(
   images: Array<{ src: string; dest: string }>,
+  imagesPath: string,
 ) {
-  await pMap(
-    images,
-    async (img) => {
-      if (await pathExists(img.dest)) return;
-      const response = await fetch(img.src);
-      if (response.ok) {
-        await mkdirp(dirname(img.dest));
-        const stream = createWriteStream(img.dest);
-        await finished(Readable.fromWeb(response.body as any).pipe(stream));
-      } else {
-        console.log(`Error downloading ${img.src} to ${img.dest}`);
-      }
-    },
-    { concurrency: 4 },
+  const missingImages = images.filter(
+    async (img) => !(await pathExists(img.dest)),
   );
+
+  if (missingImages.length == 0) {
+    return;
+  }
+
+  await startWebServer(`${imagesPath}/artifact`, 8000);
+  try {
+    await pMap(
+      missingImages,
+      async (img) => {
+        const response = await fetch(img.src);
+        if (response.ok) {
+          await mkdirp(dirname(img.dest));
+          const stream = createWriteStream(img.dest);
+          await finished(Readable.fromWeb(response.body as any).pipe(stream));
+        } else {
+          console.log(`Error downloading ${img.src} to ${img.dest}`);
+        }
+      },
+      { concurrency: 4 },
+    );
+  } finally {
+    await closeWebServer(8000);
+  }
 }
--- a/scripts/lib/webServer.ts
+++ b/scripts/lib/webServer.ts
@@ -14,6 +14,17 @@ import { $ } from "zx";
 
 export async function startWebServer(directory: string, listenPort: number) {
   $`python3 -m http.server ${listenPort} -d ${directory} -b ::1 &`;
+
+  // Wait until the server is up and able to listen to the requests
+  while (true) {
+    try {
+      const response = await fetch(`http://localhost:${listenPort}`);
+      return;
+    } catch {
+      // Wait 5 ms for the next fetch
+      await new Promise((res) => setTimeout(res, 5));
+    }
+  }
 }
 
 export async function closeWebServer(listenPort: number) {
--- a/scripts/commands/updateApiDocs.ts
+++ b/scripts/commands/updateApiDocs.ts
@@ -364,6 +364,7 @@ async function convertHtmlToMarkdown(
       src: img.src,
       dest: `${getRoot()}/public${img.dest}`,
     })),
+    htmlPath,
   );
 }
 


scripts/lib/downloadImages.ts
```diff
@@ -17,23 +17,37 @@ import { createWriteStream } from "node:fs";
...
-        const stream = createWriteStream(img.dest);
-        await finished(Readable.fromWeb(response.body as any).pipe(stream));
-      } else {
-        console.log(`Error downloading ${img.src} to ${img.dest}`);
-      }
-    },
-    { concurrency: 4 },
+  const missingImages = images.filter(
+    async (img) => !(await pathExists(img.dest)),
   );
```
Eric-Arellano, line 28: Unlike the snippet I sent you on Slack, this will not run things in parallel. But it's probably fast enough to not matter.

--

Either way, this works properly?

scripts/lib/downloadImages.ts
```diff
@@ -17,23 +17,37 @@ import { createWriteStream } from "node:fs";
...
+
+  if (missingImages.length == 0) {
+    return;
+  }
+
+  await startWebServer(`${imagesPath}/artifact`, 8000);
```
Eric-Arellano, line 34: The 8000 is a magic number. How about we remove from `startWebServer` and `closeWebServer` this argument and hardcode it? In the web server file, you can set a constant `PORT = 8000`.

scripts/lib/webServer.ts
```diff
@@ -14,6 +14,17 @@ import { $ } from "zx";
 
 export async function startWebServer(directory: string, listenPort: number) {
   $`python3 -m http.server ${listenPort} -d ${directory} -b ::1 &`;
+
+  // Wait until the server is up and able to listen to the requests
+  while (true) {
```
Eric-Arellano, line 19: Instead, a better approach is to use exponential backoff. It avoids sending too many requests all at once. Whereas right now, once a request fails, you immediately are sending another. This library looks good: https://www.npmjs.com/package/exponential-backoff

But you can also avoid using that library and keep this simpler. Do something like this strategy:

1. No matter what, wait one second
2. Check if valid
3. Wait 3 more seconds
4. Check if valid. Else exit

The setup is simple enough that you can make it explicit and Keep It Simple. I'd recommend this approach.

Eric-Arellano (commented):
Thanks for looking into this!

scripts/lib/downloadImages.ts
arnaucasau: Oops, it's not working because the promise is always true and we always download the images. Good catch! With the snippet you sent me on Slack we have the problem that the array becomes of the type `img | null`, and I tried to do it differently. However, it seems it's not possible to use `filter` directly, thus I used your code but type casting the argument to the type we want.

scripts/lib/webServer.ts
arnaucasau: Thanks for the comment! I also thought it was better to keep it simple, and given that the time needed is close to 500 ms, I used the while loop. It's a good idea to wait before the first validation and then check if the server is listening, so I followed the strategy you said in the comment :+1: The times I used are 500 ms in the first wait and 1 second in the second wait (We should perform 1 or 2 fetches at most).

scripts/lib/downloadImages.ts
```diff
@@ -17,23 +17,44 @@ import { createWriteStream } from "node:fs";
 import { finished } from "stream/promises";
 import { Readable } from "stream";
 import pMap from "p-map";
+import { startWebServer, closeWebServer } from "../lib/webServer";
 
 export async function downloadImages(
   images: Array<{ src: string; dest: string }>,
+  imagesPath: string,
```
Eric-Arellano, line 24: This name is a little ambiguous: is it the images we will be writing, or the original images? Maybe use a name like `originalImagesFolderPath`? And probably have the caller put the `/artifacts` in the argument to begin with.

scripts/lib/downloadImages.ts
```diff
@@ -17,23 +17,44 @@ import { createWriteStream } from "node:fs";
...
+  const missingImagesResults = await Promise.all(
+    images.map(async (img) => {
+      const exists = await pathExists(img.dest);
+      return exists ? null : img;
+    }),
   );
+
+  const missingImages = missingImagesResults.filter(
+    (img): img is { src: string; dest: string } => img !== null,
+  );
```
Eric-Arellano, line 35: Does it work to instead type-cast the overall `missingImages` variable? Because the current type hint is a lie. The `img` argument can be `null`! It's only after the `filter` that we're confident there are no more `null`s.

scripts/lib/webServer.ts
Eric-Arellano: I'm not following how this while loop is safe. If the first time we do the `try` block fails, then we enter the `catch` block and wait 1 second. Then we try again. That can keep happening in an infinite loop, right? There is no code to say at a max, try 1-2 times.

I think you can remove the `while` loop and make your 2 attempts more explicit. It's fine in this case to repeat a little bit of code.

scripts/lib/downloadImages.ts
arnaucasau: I tried it but I got this error in vs code trying with the types `{ src: string; dest: string; }[] ` and `Array<{ src: string; dest: string; }>` : Type 'null' is not assignable to type '{ src: string; dest: string; }

arnaucasau: Using flatMap we have a similar issue to when I used just filter, here there's an example: https://stackoverflow.com/a/74959389. As suggested in that answer, I used map and flat separately

scripts/lib/downloadImages.ts
```diff
@@ -17,23 +17,42 @@ import { createWriteStream } from "node:fs";
...
-        console.log(`Error downloading ${img.src} to ${img.dest}`);
-      }
-    },
-    { concurrency: 4 },
-  );
+  const missingImages = (
+    await Promise.all(
+      images.map(async (img) => {
+        const exists = await pathExists(img.dest);
+        return exists ? [] : img;
```
Eric-Arellano, line 30:
```suggestion
        return exists ? [] : [img];
```

Eric-Arellano (approved):
Looks great! Please double check this executes properly - there's a bug in the current implementation.

PR #555, opened 2023-12-27
Rewrite `./start` script to Python

Eric-Arellano:
In https://github.com/Qiskit/documentation/pull/541, we're going to need to add semi-complex logic to our `./start` script to dynamically discover every subfolder we have in `translations/`, then tell Docker to mount the folder as a volume. I'd much rather write that logic in Python than Bash.

This PR is prework so that https://github.com/Qiskit/documentation/pull/541 is a smaller PR. It makes no logic changes and should be a 1:1 mapping.

### DIFF
--- a/start
+++ b/start
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env python3
 # This code is a Qiskit project.
 #
 # (C) Copyright IBM 2023.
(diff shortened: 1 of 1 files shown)


start
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env python3
```
Eric-Arellano, line 1: This is a really neat trick. Note that the file still is called simply `./start`. Users don't need to do anything differently, as long as they have `python3` installed on their machine. 

This is part of why I called the script `./start` and not `./start.sh`. The other reason is less typing hah. Although if we did call it `./start.sh`, we could do this trick! File extensions can be a lie - which is important to know for security!

start
```diff
@@ -11,10 +11,34 @@
...
+        ],
+        check=True,
+    )
+
+
+if __name__ == "__main__":
+    try:
+        main()
+    except KeyboardInterrupt:
+        pass
```
Eric-Arellano, line 44: Otherwise you get an annoying error message when using ctrl-c to close the server.

arnaucasau (approved):
Looks good to me!

Issue #532, opened 2023-12-18
Ban absolute internal links, other than for API docs


Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/476. As explained in https://github.com/Qiskit/documentation/issues/7, there are two styles of internal links:

```
/start/my_file
../start/my_file
```

The first style starting with `/` would always refer to the English version of the file; you'd need to use `/es/start/my_file` to get the Spanish version. Or we'd need to have the docs app interpret the URL differently, which I think is complex and not very transparent.

Instead, I propose we ban the first style and always use relative links like `./my_file` and `../start/my_file`. That will correctly behave for translations because the link `./my_file` from `/es/start/foo` will take you to `/es/start/my_file`.

We should make an exception for `/api` and `/public/images` though, since those won't be translated.

Eric-Arellano:
@axelhzf proposed instead doing this in the docs application itself. It will correctly relativize links for us. That is a much better approach because 

a) implementing this issue is tricky, and
b) it would be annoying for writers to have to remember to not use absolute links; it adds friction to their workflow

PR #556, opened 2023-12-27
Refactor and test link checker's markdown handling

Eric-Arellano:
The link checker is going to keep getting more complex in https://github.com/Qiskit/documentation/issues/495 and https://github.com/Qiskit/documentation/issues/532. So, this is some prework to make the script more maintainable.

It moves all the `lib/` code for the link checker to a new `links/` subfolder. The markdown logic is moved to `markdown.ts`, refactored and now has tests.


scripts/lib/links/markdown.test.ts
```diff
@@ -0,0 +1,104 @@
...
+                "metadata": {},
+                "source": [
+                    "Line 3."
+                ]
+            }
+        ],
+        "metadata": {}
+    }
+  `);
+  expect(result).toBe("Line 1.\nLine 2.\nLine 3.");
```
Eric-Arellano, line 48: This PR has a change in behavior. Before, we would add a `\n` to every single line, including lines in the same markdown block. But those lines already have `\n` separators, so really they were ending up with `\n\n`. Now, we only add our own `\n` in between distinct cells.

scripts/commands/checkLinks.ts
```diff
@@ -119,61 +82,22 @@ async function processDocsFiles(
...
-            } else {
-              linkMap.set(TreeNode.url, [filePath]);
-            }
-          });
-        };
-      })
-      .use(remarkStringify)
-      .process(markdown);
+    files.push(new File(filePath, anchors, false));
+    await addLinksToMap(filePath, markdown, linksToOriginFiles);
```
Eric-Arellano, line 95: Note that we should have been calling `await` before. I don't think it actually broke things though, fortunately.

scripts/lib/links/markdown.ts
```diff
@@ -0,0 +1,85 @@
...
+  cell_type: string;
+  source: string[];
+}
+
+export function markdownFromNotebook(rawContent: string): string {
+  const cells = JSON.parse(rawContent).cells as JupyterCell[];
+  return cells
+    .filter((cell) => cell.cell_type === "markdown")
+    .map((cell) => cell.source.join(""))
+    .join("\n");
```
Eric-Arellano, line 35: We used to have a for loop that would repeatedly concatenate to the string `markdown`. This uses a more functional style.

scripts/lib/links/markdown.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export function parseAnchors(markdown: string): string[] {
+  const result = markdownLinkExtractor(markdown).anchors;
+  const idAnchors = markdown.match(/(?<=id=")(.*)(?=")/gm);
+  if (idAnchors !== null) {
+    result.push(...idAnchors.map((id) => `#${id}`));
```
Eric-Arellano, line 42: This technique for adding the IDs to `result` is different. Rather than adding one element at a time, we add the whole batch all at once.

scripts/lib/links/markdown.ts
```diff
@@ -0,0 +1,85 @@
...
+    path.extname(filePath) === ".ipynb" ? markdownFromNotebook(source) : source;
+  return [markdown, parseAnchors(markdown)];
+}
+
+export async function addLinksToMap(
+  filePath: string,
+  markdown: string,
+  linksToOriginFiles: Map<string, string[]>,
+): Promise<void> {
+  const addLink = (link: string): void => {
```
Eric-Arellano, line 61: This helper function is new and helps to DRY things.

scripts/lib/links/markdown.ts
```diff
@@ -0,0 +1,85 @@
...
+    } else {
+      entry.push(filePath);
+    }
+  };
+
+  unified()
+    .use(rehypeParse)
+    .use(remarkGfm)
+    .use(rehypeRemark)
+    .use(() => (tree: Root) => {
```
Eric-Arellano, line 74: This used to be a named `function` and now it's an anonymous function, which is more readable imo.

scripts/lib/links/markdown.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export function parseAnchors(markdown: string): string[] {
+  const result = markdownLinkExtractor(markdown).anchors;
+  const idAnchors = markdown.match(/(?<=id=")(.*)(?=")/gm);
+  if (idAnchors !== null) {
+    result.push(...idAnchors.map((id) => `#${id}`));
+  }
+  return result;
+}
```
arnaucasau, line 45: I like this new function, it makes the code simpler and more readable. Good idea with the spread operator! Moreover, we could also get rid of the if condition and the push by using the spread operator in the return. I still like the way you defined the function though, just FYI.

```suggestion
export function parseAnchors(markdown: string): string[] {
  const mdAnchors = markdownLinkExtractor(markdown).anchors;
  const idAnchors = markdown.match(/(?<=id=")(.*)(?=")/gm) || [];
  return [...mdAnchors, ...idAnchors.map((id) => `#${id}`)];
}
```

scripts/lib/links/markdown.ts
arnaucasau: And helps as well with the readability of the unified plugin! :rocket:

arnaucasau (approved):
LGTM! Thank you Eric for the refactor and for adding the tests as well!

scripts/lib/links/markdown.ts
Eric-Arellano: I love it 😍 Thanks!

PR #557, opened 2023-12-28
Refactor link checker to use FileBatch class

Eric-Arellano:
In https://github.com/Qiskit/documentation/issues/495, we will start checking historical API docs. For performance, it's important that we don't load every single file in the project at once in memory. Instead, we want to operate in batches: e.g. check all of 0.44, then drop from memory and move on to 0.43.

This PR is a pre-factor to add a new `FileBatch` class. For now, we only have a single `FileBatch`, the same as before. This class still differentiates between files "to load" vs "to check", per https://github.com/Qiskit/documentation/pull/496.

### DIFF
--- a/scripts/commands/checkLinks.ts
+++ b/scripts/commands/checkLinks.ts
@@ -14,9 +14,8 @@ import { globby } from "globby";
 import yargs from "yargs/yargs";
 import { hideBin } from "yargs/helpers";
 
-import { Link, File } from "../lib/links/LinkChecker";
-import FILES_TO_IGNORES from "../lib/links/ignores";
-import { getMarkdownAndAnchors, addLinksToMap } from "../lib/links/markdown";
+import { File } from "../lib/links/LinkChecker";
+import { FileBatch } from "../lib/links/FileBatch";
 
 const DOCS_GLOBS_TO_CHECK = [
   "docs/**/*.{ipynb,md,mdx}",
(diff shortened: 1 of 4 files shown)


Eric-Arellano:
I didn't add tests for `FileBatch` since it's too complex to test with needing actual files set up etc.

frankharkins (approved):
Thanks!

scripts/commands/checkLinks.ts
```diff
@@ -59,104 +58,21 @@ const readArgs = (): Arguments => {
...
+    DOCS_GLOBS_TO_CHECK,
+    DOCS_GLOBS_TO_LOAD,
+  );
   const otherFiles = [
     ...(await globby("{public,docs}/**/*.{png,jpg,gif,svg}")).map(
-      (fp) => new File(fp, [], false),
+      (fp) => new File(fp, []),
     ),
     ...SYNTHETIC_FILES.map((fp) => new File(fp, [], true)),
   ];
```
arnaucasau, line 73: Just as an idea. What do you think about having the `otherFiles` in the constructor of the `fileBatch` class and using the `fromGlobs` for them? By doing that, we can avoid having an extra argument in the check call. I think that the argument can be confusing because we already provided files to the class.

scripts/lib/links/LinkChecker.ts
```diff
@@ -25,7 +25,7 @@ export class File {
    *    path: Path to the file
    * anchors: Anchors available in the file
    */
-  constructor(path: string, anchors: string[], synthetic: boolean) {
+  constructor(path: string, anchors: string[], synthetic: boolean = false) {
```
arnaucasau, line 28: Good idea with the default value!

arnaucasau (commented):
The refactor looks really good! I tested it locally and all works well. Thank you Eric!

scripts/commands/checkLinks.ts
Eric-Arellano: Yeah, I think that could work too. The reason I didn't do it is we'd end up recomputing `otherFiles` many times in the follow up PR to add historical APIs, since each historical version is its own `FileBatch`.

arnaucasau: Oh, I see, that's a good reason. However, this list loads all the images for all the versions every time, when perhaps it's not necessary to load any image besides the ones belonging to that version, and maybe the current one. For example, Qiskit v0.33 doesn't need to have the images in `public/images/api/qiskit/0.19/` load. Regardless of that, I like a lot you did the refactor, so if you think it's okay because the amount of images is not that large, I'm good with it too :smiley:

PR #560, opened 2023-12-29
Allow checking historical APIs with link checker

Eric-Arellano:
This unblocks https://github.com/Qiskit/documentation/issues/495. We still have to actually fix the broken links, but we can at least now detect them.

To make the script more interactive, we now output a description of each file batch, like this:

```
Checking links for non-API docs


Checking links for qiskit-ibm-runtime v0.14
❌ docs/api/qiskit-ibm-runtime/0.14/ibm-runtime.md: Could not find link 'runtime_service' ❓ Did you mean '/api/qiskit-ibm-runtime/0.14/ibm-runtime'?
❌ docs/api/qiskit-ibm-runtime/0.14/qiskit_ibm_runtime.QiskitRuntimeService.md: Could not find link '/api/qiskit/providers_models'


Checking links for qiskit-ibm-runtime v0.15
❌ docs/api/qiskit-ibm-runtime/0.15/ibm-runtime.md: Could not find link 'runtime_service' ❓ Did you mean '/api/qiskit-ibm-runtime/0.15/ibm-runtime'?
❌ docs/api/qiskit-ibm-runtime/0.15/qiskit_ibm_runtime.QiskitRuntimeService.md: Could not find link '/api/qiskit/providers_models'
```

Note that to fix the broken links, we're going to need to set `toLoad` for some of the projects. For example, you can see in the above `qiskit-ibm-runtime` failures that they need to read in the file `/api/qiskit/providers_models`. I plan to address that in a follow up because we want to load the minimum amount of files possible to reduce memory usage, so I still want to see what we really need to load.

### DIFF
--- a/scripts/commands/checkLinks.ts
+++ b/scripts/commands/checkLinks.ts
@@ -10,6 +10,8 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
+import { readdir } from "fs/promises";
+
 import { globby } from "globby";
 import yargs from "yargs/yargs";
 import { hideBin } from "yargs/helpers";
--- a/scripts/commands/checkLinks.ts
+++ b/scripts/commands/checkLinks.ts
@@ -28,6 +28,7 @@ interface Arguments {
   [x: string]: unknown;
   external: boolean;
   currentApis: boolean;
+  historicalApis: boolean;
   qiskitReleaseNotes: boolean;
 }
 
@@ -45,6 +46,13 @@ const readArgs = (): Arguments => {
       default: false,
       description: "Check the links in the current API docs.",
     })
+    .option("historical-apis", {
+      type: "boolean",
+      default: false,
+      description:
+        "Check the links in the historical API docs, e.g. `api/qiskit/0.44`. " +
+        "Warning: this is slow.",
+    })
     .option("qiskit-release-notes", {
       type: "boolean",
       default: false,
@@ -56,7 +64,7 @@ const readArgs = (): Arguments => {
 async function main() {
   const args = readArgs();
 
-  const fileBatch = await determineFileBatch(args);
+  const fileBatches = await determineFileBatches(args);
   const otherFiles = [
     ...(await globby("{public,docs}/**/*.{png,jpg,gif,svg}")).map(
       (fp) => new File(fp, []),
@@ -64,7 +72,14 @@ async function main() {
     ...SYNTHETIC_FILES.map((fp) => new File(fp, [], true)),
   ];
 
-  const allGood = await fileBatch.check(args.external, otherFiles);
+  let allGood = true;
+  for (const fileBatch of fileBatches) {
+    const allValidLinks = await fileBatch.check(args.external, otherFiles);
+    if (!allValidLinks) {
+      allGood = false;
+    }
+  }
+
   if (!allGood) {
     console.error("\nSome links appear broken 💔\n");
     process.exit(1);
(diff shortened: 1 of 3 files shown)


scripts/commands/checkLinks.ts
```diff
@@ -94,7 +123,43 @@ async function determineFileBatch(args: Arguments): Promise<FileBatch> {
...
+  let description: string;
+  if (args.currentApis && args.qiskitReleaseNotes) {
+    description = "non-API docs, current API docs, and Qiskit release notes";
+  } else if (args.currentApis) {
+    description = "non-API docs and current API docs";
+  } else if (args.qiskitReleaseNotes) {
+    description = "non-API docs and Qiskit release notes";
+  } else {
+    description = "non-API docs";
+  }
```
arnaucasau, line 135: We can generate the description without the conditionals using [Intl.ListFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat) and we can add the historical API docs comment without using an exponential number of conditionals.

```suggestion
  const options: [string, boolean][] = [
    ["non-API docs", true],
    ["current API docs", args.currentApis],
    ["historical API docs", args.historicalApis] , 
    ["Qiskit release notes", args.qiskitReleaseNotes]
  ];

  const optionsEnabled = options.filter(([_, enabled]) => enabled);
  const optionsText = optionsEnabled.map(([text, _]) => text);

  const formatter = new Intl.ListFormat('en', {
    style: 'long',
    type: 'conjunction',
  });
  const description = formatter.format(optionsText);
```

scripts/commands/checkLinks.ts
```diff
@@ -94,7 +123,43 @@ async function determineFileBatch(args: Arguments): Promise<FileBatch> {
...
+
+    const fileBatch = await FileBatch.fromGlobs(
+      [indexPath.replace("index.md", "*")],
+      [],
+      `${projectName} v${versionMatch[0]}`,
+    );
+    result.push(fileBatch);
+  }
+  return result;
 }
```
arnaucasau, line 163: FYI, alternative method: To create all the paths for the historical versions, maybe we can read directly the names of the folders we have. The advantage is that they are already sorted by name, and we don't need to use the replace for the glob expression. Notice that we need to import `readdir` as well: `import { readdir } from "fs/promises";
`
```suggestion
async function determineHistoricalFileBatches(
  projectName: string,
): Promise<FileBatch[]> {
  const historicalFolders = (
    await readdir(`docs/api/${projectName}`, { withFileTypes: true })
  ).filter((file) => file.isDirectory() && file.name.match(/[0-9].*/));

  const result = [];
  for (const folder of historicalFolders) {
    const fileBatch = await FileBatch.fromGlobs(
      [`docs/api/${projectName}/${folder.name}/*`],
      [],
      `${projectName} v${folder.name}`,
    );
    result.push(fileBatch);
  }
  return result;
}
```

arnaucasau (approved):
Awesome! Thank you so much for the PR Eric! I like a lot the validation in batches, it helps a lot to understand in what version we have the errors. The code looks good too, and it works very well (tested locally)! I left you two comments, but I'm happy with the way you did it!

scripts/commands/checkLinks.ts
Eric-Arellano: Much better! Thanks!

scripts/commands/checkLinks.ts
Eric-Arellano: Neat! I think this is too cryptic though, so I'm gonna pass on it. But, this does remind me of pattern matching, which I love. If you haven't yet read about pattern matching, check it out - it was added to Python recently: https://peps.python.org/pep-0636/

Issue #558, opened 2023-12-28
Use correct index page for Runtime and Provider


Eric-Arellano:
We should be using https://github.com/Qiskit/documentation/blob/main/docs/api/qiskit-ibm-runtime/ibm-runtime.md and https://github.com/Qiskit/documentation/blob/main/docs/api/qiskit-ibm-provider/ibm-provider.md as the `/index` files for both projects. They're more useful and they align with the index file for Qiskit.

Instead, we've been renaming the files for the top-level module:

https://github.com/Qiskit/documentation/blob/dabb28cc63fe0228a55c733f00f881e24ca63364/scripts/lib/sphinx/renameUrls.ts#L19-L23

I think fixing this issue means:

1. Updating that script and regenerating the docs
2. Ensuring there aren't broken links by removing these ignores:

https://github.com/Qiskit/documentation/blob/dabb28cc63fe0228a55c733f00f881e24ca63364/scripts/lib/links/ignores.ts#L23-L24

3. Figuring out why those two pages don't have metadata. Don't ignore them here:

https://github.com/Qiskit/documentation/blob/dabb28cc63fe0228a55c733f00f881e24ca63364/scripts/commands/checkMetadata.ts#L17-L25

Issue #66, opened 2023-10-02
API pipeline needs to add "front matter" (title) to several pages


Eric-Arellano:
These never had front matter. Axel said it's fine for now, but we should improve it soon.

> Please fix these files: docs/api/qiskit-ibm-provider/ibm-provider.md,docs/api/qiskit/transpiler_builtin_plugins.md,docs/api/qiskit-ibm-runtime/ibm-runtime.md,docs/api/qiskit-ibm-runtime/options.md,docs/api/qiskit-ibm-runtime/runtime_service.md

We need to an `else` clause here:

https://github.com/Qiskit/documentation/blob/e8724413bfa5d8885f4efa308ce44efe140440bf/scripts/lib/sphinx/addFrontMatter.ts#L19-L30

TBD how we determine the title and description!

axelhzf:
Perhaps we can use the filename as a fallback?

Eric-Arellano:
https://github.com/Qiskit/documentation/pull/821 fixes a legit bug in historical API docs. The other files so far seem to be because they are not using the `automodule` directive. Instead, they're pages like 

https://github.com/Qiskit/qiskit/blob/1e0e4fccd469617489a28e4cfa38271a9b0b617b/docs/apidoc/transpiler_builtin_plugins.rst?plain=1#L1-L10

I'm thinking we could extract the title from the `<h1`>, like `Built-in Transpiler Synthesis Plugins`. Then the description could be `API docs for Built-in Transpiler Synthesis Plugins`. 

We should go through this file list to confirm that this is the correct diagnosis and fix:

<details><summary>Failing files after #821</summary>
<p>
docs/api/qiskit/transpiler_builtin_plugins.md
docs/api/qiskit/0.19/aer.md
docs/api/qiskit/0.19/ibmq_jupyter.md
docs/api/qiskit/0.19/ibmq_visualization.md
docs/api/qiskit/0.19/qiskit.aqua.aqua_globals.md
docs/api/qiskit/0.19/qiskit.optimization.INFINITY.md
docs/api/qiskit/0.19/qiskit.quantum_info.two_qubit_cnot_decompose.md
docs/api/qiskit/0.24/aer.md
docs/api/qiskit/0.24/aqua.md
docs/api/qiskit/0.24/ibmq-provider.md
docs/api/qiskit/0.24/ibmq_jupyter.md
docs/api/qiskit/0.24/ibmq_visualization.md
[...]

PR #561, opened 2023-12-29
Provider and Runtime show modules as top-level in sidebar

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/558.

Before:

Screenshot 2023-12-29 at 1 38 50 PM

After:

Screenshot 2023-12-29 at 1 39 08 PM

This new left sidebar layout aligns with Qiskit. The consistency is useful, but also I think this new layout is more functional anyways. It reduces the amount of nesting. It also allows us to collapse the left sidebar by default, which we do in the rest of the docs.

## Changes index page

Before, the index page was the description for the code components belonging to the top-level modules `qiskit_ibm_provider` and `qiskit_ibm_runtime`, rather than enumerating what all the modules are:

Before:
Screenshot 2023-12-29 at 1 40 19 PM

After:
Screenshot 2023-12-29 at 1 41 30 PM

Note that this index page is not accessible from the left sidebar; it's not clear what we would call it. But you can access it from the top nav-bar:

Screenshot 2023-12-29 at 1 42 13 PM

## Metadata

The new index pages have never had metadata. That's because we don't have any technique to determine from Sphinx programatically its metadata because the Sphinx generated HTML page is missing it:

https://github.com/Qiskit/documentation/blob/92b3fa45c10aa917c3fbc7ada861b2db55203902/scripts/lib/sphinx/PythonObjectMeta.ts#L13-L24

This is an instance of https://github.com/Qiskit/documentation/issues/66. I think we need a generic solution to that problem, such as Axel's suggestion to use filenames as a fallback. But, it's important that the metadata is high quality for these two index pages since they are so high-profile. 

So, I propose hardcoding the metadata for them. Because there are only two files, I think this approach is acceptable; obviously it wouldn't scale if we were hardcoding dozens of files.


docs/api/qiskit-ibm-provider/ibm_provider.md
```diff
@@ -0,0 +1,66 @@
+---
```
Eric-Arellano, line 1: This used to be `/index`. It is the API page for the top-level module.

docs/api/qiskit-ibm-provider/index.md
```diff
@@ -1,66 +1,16 @@
 ---
```
Eric-Arellano, line 1: This used to be `ibm-provider.md` (with `-`, not `_`). The only change we've made is the front matter / metadata.

scripts/commands/checkMetadata.ts
```diff
@@ -15,13 +15,7 @@ import fs from "fs/promises";
 import { globby } from "globby";
 
 const IGNORED_FILES = new Set([
-  "docs/api/qiskit-ibm-provider/ibm-provider.md",
   "docs/api/qiskit/transpiler_builtin_plugins.md",
-  "docs/api/qiskit-ibm-runtime/ibm-runtime.md",
-  "docs/api/qiskit-ibm-runtime/0.14/ibm-runtime.md",
-  "docs/start/__migration-guide-algorithms.md",
-  "docs/start/__migration-guide-opflow.md",
-  "docs/start/__migration-guide-qi.md",
```
Eric-Arellano, line 24: These files don't exist anymore.

scripts/commands/updateApiDocs.ts
```diff
@@ -93,12 +93,6 @@ const PACKAGES: PkgInfo[] = [
     githubSlug: "qiskit/qiskit",
     initialUrl: `/apidoc/index.html`,
     hasSeparateReleaseNotes: true,
-    tocOptions: {
-      collapsed: true,
-      nestModule(id: string) {
-        return id.split(".").length > 2;
-      },
-    },
```
Eric-Arellano, line 101: This is now the universal behavior.

scripts/lib/sphinx/addFrontMatter.ts
```diff
@@ -20,7 +20,14 @@ export function addFrontMatter<T extends SphinxToMdResult>(
 ): void {
   for (let result of results) {
     let markdown = result.markdown;
-    if (result.meta.python_api_name) {
+    if (result.meta.hardcoded_frontmatter) {
```
Eric-Arellano, line 23: The order of this `if` matters. If you hardcode the frontmatter, we use that no matter what. That will be important when we fix https://github.com/Qiskit/documentation/issues/66 to add a fallback like using the filename. We want to make sure we still use the hardcoded metadata for these two index files.

scripts/lib/sphinx/generateToc.test.ts
```diff
@@ -13,293 +13,86 @@
 import { describe, expect, test } from "@jest/globals";
 import { generateToc } from "./generateToc";
 
-describe("generateTocFromPythonApiFiles", () => {
+describe("generateToc", () => {
   test("generate a toc", () => {
     const toc = generateToc(pkg, [
-      { meta: {}, url: "/docs/runtime" },
```
Eric-Arellano, line 19: I changed this test to better reflect reality.

scripts/lib/sphinx/generateToc.test.ts
```diff
@@ -13,293 +13,86 @@
...
-            "title": "Release notes",
-            "url": "/api/qiskit_ibm_runtime/release-notes",
-          },
-        ],
-        "title": "Qiskit Runtime IBM Client",
-      }
-    `);
-  });
-
-  test("nest modules", () => {
```
Eric-Arellano, line 100: This test isn't necessary anymore since the nesting behavior is now consistent. The above test should cover this.

arnaucasau (approved):
Looks good to me! Thank you Eric!

frankharkins (approved):
Thanks!

Will the navbar need updating? Clicking `API reference > Qiskit IBM Runtime Client` still takes me to "Overview"

Eric-Arellano:
> Will the navbar need updating? Clicking API reference > Qiskit IBM Runtime Client still takes me to "Overview"

Is this the staging preview? That image is super out of date and we've had issues updating it. In production, the URL from the top nav bar is https://docs.quantum.ibm.com/api/qiskit-ibm-runtime and it takes you to the correct `/index` file.

PR #569, opened 2024-01-02
Check Runtime historical API docs for broken links

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/495. We now will enforce in CI that historical API docs are valid for Runtime and Provider. To do this, this PR adds the `--skip-broken-historical` argument.

In a follow up, I want to try checking Qiskit docs, at least certain versions. But I'm waiting until we regenerate the docs with https://github.com/Qiskit/documentation/pull/552 applied because I suspect it will fix some issues.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,10 +33,10 @@ jobs:
       - name: Link checker
         run: >
           npm run check:links --
-            --current-apis
-            --qiskit-release-notes
-            --historical-apis
-            --skip-broken-historical
+          --current-apis
+          --qiskit-release-notes
+          --historical-apis
+          --skip-broken-historical
       - name: Formatting
         run: npm run check:fmt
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -31,7 +31,12 @@ jobs:
       - name: Spellcheck
         run: npm run check:spelling
       - name: Link checker
-        run: npm run check:links -- --current-apis --qiskit-release-notes
+        run: |
+          npm run check:links -- \
+            --current-apis \
+            --qiskit-release-notes \
+            --historical-apis \
+            --skip-broken-historical
```
Eric-Arellano, line 39: This will make CI for every PR a little slower. But given that we're making changes to the API generation script the next two months, it's probably worth it. We could always remove it if CI is getting too slow.

Issue #511, opened 2023-12-11
Check in CI that every page renders


Eric-Arellano:
This would have caught the issue in https://github.com/Qiskit/documentation/pull/510.

The only approach I can think of is to use the `./start` Docker image in CI to start up the server, then visit every doc page that we have in the repo. Make sure they each give back a 200.

This will slow down CI, but that's worth it.

--

Note that the closed source repo already have this check because when it builds the app, it checks every page beforehand. But it would be better to catch this error early.

javabster:
I think adding something like selenium tests to our testing suites shouldn't add too much overhead if we keep them minimal: https://www.selenium.dev/documentation/webdriver/getting_started/first_script/

Eric-Arellano:
I don't think we even need something that complex. It should be sufficient to do `python3 -m http.server`, statically find all our URLs by inspecting the `docs/` folder, and then making a GET request to each of those.

Eric-Arellano:
I got started on this in https://github.com/Qiskit/documentation/pull/572. Some questions for @arnaucasau and @frankharkins (and others):

1. Do you think this makes sense for when we run the script?
    - The script is quite slow. On my M1, it takes ~2 minutes to check non-API docs. Granted, that's because I have Apple Silicon so Docker is really slow because it's built for Intel. 
    - So, in CI, only check non-API docs for every PR
    - Have our external link checker workflow check every night API docs & Qiskit release notes, including historical API docs
2. Do we want the script to auto-start the Docker image for you (`./docker`), or is it fine to have users need to start the server and we only eagerly validate at the beginning of the script that it's set up?
    - Normally, I'd say that the script should auto-set it up so it's easier to run the check
    - But, `./start` will fail if you don't have Docker running, and it also can be very slow to start the server if you have a "cold" (un-cached) build. That would complicate our code to ping if the server is up every N seconds. We'd want to use a library for exponential backoff:

https://github.com/Qiskit/documentation/blob/494e6b632e1de74e8d863a321d706494f45556a2/scripts/lib/webServer.ts#L20-L31

    - So, I personally prefer to keep it simple. I doubt people will even run this script much _locally_ and they'll rely on CI.

PR #572, opened 2024-01-02
Add check that all pages render

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/511. Example:

```
$ npm run check-pages-render

Checked 10 / 71 pages
Checked 20 / 71 pages
Checked 30 / 71 pages
Checked 40 / 71 pages
Checked 50 / 71 pages
Checked 60 / 71 pages
Checked 70 / 71 pages
✅ All pages render without crashing
```

## Only checks non-API docs by default

This script is quite slow, at least on my M1 because the Docker images is built with x86. 

So, to avoid CI slowing down too much, we only check non-API pages in PR builds. Our nightly cron job checks everything else.

## Does not auto-start Docker

We no longer have the file `webServer.ts` thanks to https://github.com/Qiskit/documentation/pull/578, which was a great improvement. 

Rather than adding back somewhat complex code for us to auto-start the server—and then to periodically ping if it's ready or time out—we expect the user to start up the server. That's acceptable since usually people will rely on CI to run this check. It's too slow for people to be frequently running locally.


.github/workflows/nightly-extended-checks.yml
```diff
@@ -33,3 +33,15 @@ jobs:
...
+        run: |
+          ./start &
+          sleep 20
+      - name: Check API pages render
+        run: |
+          npm run check-pages-render -- \
+            --qiskit-release-noets \
+            --current-apis \
+            --historical-apis \
+            --translations
```
frankharkins, line 47: Aesthetic nit but I recently learnt `>` removes newlines

```suggestion
        run: >
          npm run check-pages-render --
            --qiskit-release-noets
            --current-apis
            --historical-apis
            --translations
```

scripts/commands/checkPagesRender.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+  if (!args.historicalApis) {
+    globs.push(
+      "!docs/api/{qiskit,qiskit-ibm-provider,qiskit-ibm-runtime}/[0-9]*/*",
+    );
+  }
+  if (!args.qiskitReleaseNotes) {
+    globs.push("!docs/api/qiskit/release-notes/*");
+  }
+  if (args.translations) {
```
frankharkins, line 145:
```suggestion
  if (!args.translations) {
```

.github/workflows/nightly-extended-checks.yml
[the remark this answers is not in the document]
Eric-Arellano: We don't ever have releases in this repo. One idea I was considering though is `branch` builds whenever API docs or translations change. Wdyt?

scripts/commands/checkPagesRender.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+  if (!args.historicalApis) {
+    globs.push(
+      "!docs/api/{qiskit,qiskit-ibm-provider,qiskit-ibm-runtime}/[0-9]*/*",
+    );
+  }
+  if (!args.qiskitReleaseNotes) {
+    globs.push("!docs/api/qiskit/release-notes/*");
+  }
+  if (!args.translations) {
```
Eric-Arellano, line 145: Wait no, this was correct.

```suggestion
  if (!args.translations) {
```

.github/workflows/nightly-extended-checks.yml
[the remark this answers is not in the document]
frankharkins: That works! The case I'm imagining is where we merge a few changes rapidly then push them to production (which a nightly check might not catch in time)

scripts/commands/checkPagesRender.ts
frankharkins: 🤦 sorry

arnaucasau (approved):
Looks good to me! Good idea with the `console.log` every 10 files!

PR #573, opened 2024-01-02
Fix external links in release notes

Eric-Arellano:
We recently started checking release notes for broken links, which found a few 🚀


abbycross (approved):
👍

Issue #223, opened 2023-10-20
Make API generation more maintainable and readable


Eric-Arellano:
Right now, it can be somewhat brittle when e.g. Sphinx changes how HTML is generated, which caused the issues in https://github.com/Qiskit/documentation/issues/67. (Although, Axel did a great job writing tests!)

I think probably this looks like adding more testing where possible.

Originally, I thought we should consider switching from Sphinx to https://vemel.github.io/handsdown/. But we decided this is not a good option, including because cross-references are extremely important for the API docs.

--

Update Jan 3: @frankharkins, @arnaucasau, and I agreed this issue is high priority before we add more complexity via https://github.com/Qiskit/documentation/pull/522, https://github.com/Qiskit/documentation/issues/454, and https://github.com/Qiskit/documentation/issues/479. 

Rough plan:

1. Improve test coverage:
     * https://github.com/Qiskit/documentation/pull/581 adds some
     * Should determine if `mergeClassMembers.ts`, `updateLinks.ts`, and `generateToc.ts` still need better testing. For example, I don't think `generateToc.ts` is teesting all the differente edge cases adequately
     * We have a lot of non-trivial logic in `updateApiDocs.ts` - what can we move to `lib` so that we can test it?
2. Refactor `sphinxHtmlToMarkdown.ts`, our most complex file. Split out the huge function into smaller helper functions & write more focused unit tests for those. Right now our unit tests are a bit hard to read and update since they have to mock so much.
3. Refactor `convertHtmlToMarkdown`, which right now does a lot. Is there a better level of abstraction?

Eric-Arellano:
We've made a lot of progress on this. The remaining work can be tracked by https://github.com/Qiskit/documentation/issues/845 and https://github.com/Qiskit/documentation/issues/715.

PR #581, opened 2024-01-03
Add more tests for API generation script

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/223. 

One way to improve the maintainability of our API generation is to test more. It gives us confidence we aren't breaking things and it makes it more obvious how the function behaves.

This PR also makes some light refactoring like renaming things to be more clear.


scripts/lib/sphinx/addFrontMatter.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      meta: {},
+      images: [],
+      isReleaseNotes: true,
+    },
+  ];
+  const pkg = {
+    versionWithoutPatch: "0.0",
+    hasSeparateReleaseNotes: true,
+    title: "My Quantum Software Project",
+  } as Pkg;
```
Eric-Arellano, line 49: This is an important trick to make tests more readable, rather than us having to give dummy values for attributes that aren't relevant. We should audit existing tests to use this when possible.

scripts/lib/sphinx/addFrontMatter.ts
```diff
@@ -39,11 +39,15 @@ python_api_name: ${result.meta.python_api_name}
 ${markdown}
 `;
     } else if (result.isReleaseNotes) {
+      const versionStr = pkg.hasSeparateReleaseNotes
+        ? ` ${pkg.versionWithoutPatch}`
+        : "";
+      const descriptionSuffix = pkg.hasSeparateReleaseNotes
+        ? `in ${pkg.title}${versionStr}`
+        : `to ${pkg.title}`;
```
Eric-Arellano, line 47: This is a behavior change for Qiskit, but same for Runtime and Provider.

scripts/lib/sphinx/dedupeHtmlIds.test.ts
```diff
@@ -10,23 +10,22 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
-import { describe, expect, test } from "@jest/globals";
-import { dedupeIds } from "./dedupeIds";
+import { expect, test } from "@jest/globals";
 
-describe("dedupeIds", () => {
```
Eric-Arellano, line 16: Same test, I only removed the unnecessary `describe()`.

scripts/lib/sphinx/addFrontMatter.ts
arnaucasau: Could we use `to` for both cases?

scripts/lib/sphinx/flattenFolders.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    {
+      url: "/api/my-pkg/apidocs/my_module2",
+    } as SphinxToMdResultWithUrl,
+    {
+      url: "/api/my-pkg/stubs/my_module.foo.Bar",
+    } as SphinxToMdResultWithUrl,
+    {
+      url: "/api/my-pkg/release_notes",
+    } as SphinxToMdResultWithUrl,
+  ];
```
arnaucasau, line 32: We can use the `as SphinxToMdResultWithUrl[]` at the end to avoid repeating it on each entry :)
```suggestion
  const results = [
    { url: "/api/my-pkg/apidoc/my_module" },
    { url: "/api/my-pkg/apidocs/my_module2" },
    { url: "/api/my-pkg/stubs/my_module.foo.Bar" },
    { url: "/api/my-pkg/release_notes" },
  ] as SphinxToMdResultWithUrl[];
```

scripts/lib/sphinx/specialCaseResults.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      meta: {},
+    } as SphinxToMdResultWithUrl,
+    {
+      url: "/api/my-pkg/ibm-provider",
+      meta: {},
+    } as SphinxToMdResultWithUrl,
+    {
+      url: "/api/my-pkg/other-file",
+    } as SphinxToMdResultWithUrl,
+  ];
```
arnaucasau, line 41: Same here:
```suggestion
  const results = [
    { url: "/api/my-pkg/release_notes" },
    { url: "/api/my-pkg/terra" },
    {
      url: "/api/my-pkg/ibm-runtime",
      meta: {},
    },
    {
      url: "/api/my-pkg/ibm-provider",
      meta: {},
    },
    { url: "/api/my-pkg/other-file" },
  ] as SphinxToMdResultWithUrl[];
```

arnaucasau (approved):
Nice work! Thank you for adding all these tests, it will help a lot ensuring that everything works as intended and understanding some parts of the code better. I added a new test in addFrontMatter.test.ts to verify the order of the conditionals if we have python_api_name defined and also hardcoded_frontmatter. I think this will be useful for #66.

I left a couple of comments, but I approve the PR for when you think it's ready to merge, all worked well on my end!

scripts/lib/sphinx/addFrontMatter.ts
Eric-Arellano: The English reads poorly to say "Changes made to Qiskit 0.45" because it sounds like you're only changing Qiskit 0.45, when in reality you are changing Qiskit 0.45+. It's not wrong, but makes less sense.

(Prepositions are the worst part of languages for me lol)

Issue #519, opened 2023-12-12
Implement GitHub source code links in docs app & generation script


Eric-Arellano:
This is the final part of https://github.com/Qiskit/documentation/issues/454. It's blocked by https://github.com/Qiskit/documentation/issues/518 and https://github.com/Qiskit/documentation/issues/517.

Once we have a design, we need to implement it. We'll need to decide on what CSS class names to use etc and make sure that they're preserved by our API generation script. This will relate to the overall improvement of API docs, where we also are going to need to start preserving class names so that we can have specific styling rules: https://github.com/Qiskit/documentation/issues/479.

Eric-Arellano:
Closed by https://github.com/Qiskit/documentation/pull/620. We'll need to add an additional implementation for the `linkcode` extension as part of https://github.com/Qiskit/documentation/issues/517, but that is tracked by #517.

PR #584, opened 2024-01-04
First refactor of sphinxHtmlToMarkdown.ts

arnaucasau:
Part of https://github.com/Qiskit/documentation/issues/223.

First small refactor of the `sphinxHtmlToMarkdown.ts` script. This PR divides the `sphinxHtmlToMarkdown` function into different helper functions, reducing its length and allowing the possibility of adding tests for different parts of the code.

Changes in this PR:

- Renamed the `$page` variable to `$` to reduce the noise introduced by the number of calls we had and to follow the [Cheerio](https://cheerio.js.org/docs/basics/loading#load) convention.
- Moved the code to find all the images into a helper function.
- Grouped similar transformations to the HTML (done previous to its conversion to markdown) into a helper function called `preprocess`.
- Moved the conversion from HTML to markdown process to a different function to simplify the `sphinxHtmlToMarkdown` function (divide more in the future).
- Refactored the recursive search of methods, adding early returns and merging some parts.
- Moved the unified plugin to the `unifiedParser.ts` file in order to abstract the transformation, which can now be called from other files. That script can group together all the unified plugins, such as the link checker, the mergeClassMembers.ts, and this one.

In this PR, the code has been copied almost identically to how it was before, except for some little adjustments, like in the unified plugin where we now use an anonymous function and have the two visitors merged.

More changes will be introduced in a follow-up.


scripts/lib/sphinx/unifiedParser.ts
```diff
@@ -10,7 +10,82 @@
...
+import { removePrefix, removeSuffix } from "../stringUtils";
+
 export const remarkStringifyOptions = {
   fences: true,
   resourceLink: true,
 };
+
+export async function sphinxHtmlToMarkdownUnifiedPlugin(
+  htmlToProcess: string,
+  handlers?: Record<string, Handle>,
```
arnaucasau, line 33: Let me know your opinion about moving the unified plugin to this file. I'm not sure it's the best option because to end up with the same result, we need all the transformations done in the `sphinxHtmlToMarkdown.ts`. The only advantage I see is to reduce the amount of code we have in that script, but all our unified plugins are very different from each other to make something general.

Even though I'm more on board with not moving it, I wanted to show you the change in this PR to know your opinion.

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -351,8 +219,145 @@ export async function sphinxHtmlToMarkdown(options: {
...
+      $link.attr("href", href.replaceAll(".html", ""));
+    }
+  });
+
+  // Custom heading for release notes
+  if (isReleaseNotes && releaseNotesTitle) {
+    $("h1").html(releaseNotesTitle);
+  }
+
+  // remove permalink links
```
arnaucasau, line 281: Same code as before. I just created the helper function here.

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -53,130 +39,12 @@ export async function sphinxHtmlToMarkdown(options: {
...
-    $quote.replaceWith(`<h3>${$quote.html()}</h3>`);
-  });
-
-  $main.find(".sd-card-body blockquote").each((_, quote) => {
-    const $quote = $page(quote);
-    $quote.replaceWith($quote.children());
-  });
+  const $ = load(html);
+  const main = $(`[role='main']`);
+  const $main = $(main);
```
Eric-Arellano, line 44: Do you happen to know what the difference is? Either way, it's good to keep the code the same as before like you are doing. Makes this PR easier to review and less risky.

scripts/lib/sphinx/unifiedParser.ts
Eric-Arellano: Yeah I wouldn't move it until we in the same PR switch other code to use the same config.

We might still want to split out sphinxHtmlToMarkdown.ts into several smaller follows. But that would be splitting out more than only this code.

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -581,3 +507,41 @@ function buildSpanId(id: string): MdxJsxFlowElement {
     children: [],
   };
 }
+
+// translate type headings to titles
```
Eric-Arellano, line 511: This comment is confusing because the function is more generic than this. I recommend using proper docstring and something like this:

> Find the element that both matches the `selector` and whose content is the same as `text`.

No need to mention how this function gets used. It could be used for many purposes.

--


Also the return type should be `Cheerio`. Ditto that all the functions should have a return type.

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -581,3 +507,41 @@ function buildSpanId(id: string): MdxJsxFlowElement {
...
+    return "function";
+  } else if ($dl.hasClass("exception")) {
+    return "exception";
+  } else if ($dl.hasClass("property")) {
+    return "property";
+  } else if ($dl.hasClass("method")) {
+    return "method";
+  } else if ($dl.hasClass("attribute")) {
+    return "attribute";
+  }
```
Eric-Arellano, line 544: TS might be mad, but you could deduplicate this with something like this:

```typescript
for (const className of ["function", ...]) {
  if ($d1.hasClass(className)) {
      return className
  }
}
```

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -351,8 +203,145 @@ export async function sphinxHtmlToMarkdown(options: {
...
+  return images;
+}
+
+function preprocessHtml(
+  $: CheerioAPI,
+  $main: Cheerio<any>,
+  baseSourceUrl: string | undefined,
+  isReleaseNotes: boolean,
+  releaseNotesTitle: string | undefined,
+) {
```
Eric-Arellano, line 250:
```suggestion
): void {
```

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -351,8 +203,145 @@ export async function sphinxHtmlToMarkdown(options: {
...
+    const href = $a.attr("href");
+    if (href?.startsWith("http:")) return;
+    if (href?.includes(`/_modules/`)) {
+      //_modules/qiskit_ibm_runtime/ibm_backend
+      const match = href?.match(/_modules\/(.*?)(#|$)/);
+      if (match) {
+        const newHref = `${baseSourceUrl ?? ""}${match[1]}.py`;
+        $a.attr("href", newHref);
+      }
+    }
```
Eric-Arellano, line 306: This technically should be preserving source links to GitHub. Interesting..To investigate more in https://github.com/Qiskit/documentation/issues/519.

Eric-Arellano (commented):
Nice work!

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
arnaucasau: I wasn't sure, but I think it's the same thing, at least with all the tests I did. I removed the second one and renamed the first to `$main`.

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
arnaucasau: I agree. I kept all the original comments like that one, and it's confusing. I was planning to add the docstring to all functions and improve all the comments in a follow-up. This one, in particular, was inside the "main" function before, and it's only copied outside. Nonetheless, I added the comment you wrote, thanks :+1:

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
arnaucasau: Much better. To avoid nesting a lot of long lists of types, I added a type definition in PythonObjectMeta.ts that perhaps we can use in more places. This is an example of what I mean by a long list of types (without the `as` in the return, the compiler complains about the return type being a string):
```ts
function getPythonApiType(
  $dl: Cheerio<any>,
):
  | "function"
  | "class"
  | "exception"
  | "method"
  | "property"
  | "attribute"
  | "module"
  | undefined {
  for (const className of [
    "function",
    "class",
    "exception",
    "method",
    "property",
    "attribute",
    "module",
  ]) {
    if ($dl.hasClass(className)) {
      return className as
        | "function"
        | "class"
        | "exception"
        | "method"
        | "property"
        | "attribute"
        | "module";
    }
  }
```

scripts/lib/sphinx/unifiedParser.ts
arnaucasau: I moved it back to the `sphinxHtmlToMarkdown.ts` file!

scripts/lib/sphinx/PythonObjectMeta.ts
```diff
@@ -10,15 +10,17 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
+export type pythonApiType =
```
Eric-Arellano, line 13: Types are capitalized. This requires changing more lines though

```suggestion
export type PythonApiType =
```

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -192,24 +72,22 @@ export async function sphinxHtmlToMarkdown(options: {
...
     const replacement = $dl
       .children()
       .toArray()
       .map((child) => {
-        const $child = $page(child);
+        const $child = $(child);
         $child.find(".viewcode-link").closest("a").remove();
         const id = $dl.find("dt").attr("id") || "";
 
-        if (child.name === "dt" && $dl.hasClass("class")) {
```
Eric-Arellano, line 204: This is the only part of the refactor that is tricky to review because it mixes a few refactors:

* changing variable names like `$page` to `$`
* using `getPythonApiType` helper (great!)
* using early returns and changing the control flow

That last one is making it hard to review. What do you think about reverting that change for now and only making the first two changes in this PR? If that's going to be hard to pull off, lmk and I can more closely review this to convince myself this is safe.

arnaucasau: Yes, no problem. I brought back the code we had before just adding the minimum changes possible to make it work (the new variable names and two arguments for one helper function). 

The way I test these changes is by regenerating the three APIs and seeing if we introduce any differences. The refactor changed the flow of that part because there are several if conditions with the same code, so I moved the default return to be an early return at the beginning and the duplicate code to be the default at the end. Once you do that, there are redundant ifs that can be removed as well.

I think it's an interesting refactor to do, but I agree we can add it in a follow-up.

Eric-Arellano (approved):
Beautiful!

PR #585, opened 2024-01-04
Minor typo cleanup in transpile section

abbycross:
Came across some little fixes.


docs/transpile/qiskit-transpiler-service.mdx
```diff
@@ -69,7 +69,8 @@ The following examples demonstrate how to transpile circuits using the Qiskit tr
     )
     transpiled_circuit = cloud_transpiler_service.run(random_circ)
     ```
-​## Next steps
+
+## Next steps
```
abbycross, line 73: This was a really weird one. Not sure why the text as originally written did not properly render as a heading, but only by completely deleting the line and starting it over did it render properly. 🤷

beckykd (approved):

docs/transpile/qiskit-transpiler-service.mdx
Eric-Arellano: Because there wasn't a blank line between the code block and the header, so markdown was failing to recognize it correctly. Technically headers should always have one blank line above and below them.

abbycross: Believe it or not but I tried that... it was as if there was an invisible character somewhere in there. Which might have been the case, actually.

PR #587, opened 2024-01-05
Fix release notes order

### LINKED ISSUE #538: Releases notes should have their order inverted
I didn't realize when we added release notes that we have release notes in ascending order, like 0.25.0 then 0.25.1. They should be in descending order so that it's chronological, like 0.25.1 then 0.25.0.

arnaucasau:
This PR changes the release notes to have a chronological descending order.

#### How to regenerate the release notes

To regenerate the release notes and fix the order of each version, I used the API generation script with some changes. We need to modify the writeSeparateReleaseNotes function from `releaseNotes.ts` and the `updateApiDocs.ts` script. 

The first step is to comment the `extractMarkdownReleaseNotesPatches` call in line 245, given that we don't want to write the latest release notes file. Also, we need to substitute the two arrays returned by that call for the release notes we want to regenerate and an empty dictionary. I did the regeneration in batches of 6 files:

```ts
//const [minorVersionsFound, markdownByPatchVersion] =
//  extractMarkdownReleaseNotesPatches(releaseNoteMarkdown);

const markdownByPatchVersion: {
  [id: string]: string;
} = {};

const minorVersionsFound = ['0.10', '0.15', '0.14', '0.13', '0.12', '0.11'];
```

The following step is to add some logic to read all the files and store each patch version we find in each file. With that information, we will re-write each release notes file at the end of the function with the correct order. That piece of code was extracted from the first version of the script which can be found in this commit https://github.com/Qiskit/documentation/pull/537/commits/7aa586f0540c74d5f092580313137ea91ddc4fe3. It was copied in the same place it appear in the commit.

```ts
const [_, markdownByPatchOldVersion] =
extractMarkdownReleaseNotesPatches(currentMarkdown);

for (let [versionPatch, markdownPatch] of Object.entries(
  markdownByPatchOldVersion,
)) {
  // We keep the release notes for a patch if it hasn't been modified for the current release notes.
  // Otherwise, we use the modified version.
  if (!markdownByPatchVersion.hasOwnProperty(versionPatch)) {
    markdownByPatchVersion[versionPatch] = markdownPatch;
  }
}
```

Finally, to speed up the process, I commented the files we regenerate in the `updateApiDocs.ts` script. This step is optional, and if used we have to make sure we restore all the docs before creating the PR.

```ts
async function convertHtmlToMarkdown(
  htmlPath: string,
  markdownPath: string,
  baseSourceUrl: string,
  pkg: Pkg,
) {
  const files = await globby(
    [
      //"apidocs/**.html",
      //"apidoc/**.html",
      //"stubs/**.html",
      "release_notes.html",
    ],
    {
      cwd: htmlPath,
    },
  );
```
Closes #538

### DIFF
--- a/docs/api/qiskit/release-notes/0.18.md
+++ b/docs/api/qiskit/release-notes/0.18.md
@@ -5,6 +5,71 @@ description: New features and bug fixes
 
 # Qiskit 0.18 release notes
 
+## 0.18.3
+
+<span id="terra-0-13-0" />
+
+### Terra 0.13.0
+
+No Change
+
+<span id="id488" />
+
+### Aer 0.5.1
+
+<span id="release-notes-0-5-1-upgrade-notes" />
+
+<span id="id489" />
+
+#### Upgrade Notes
+
+*   Changes how transpilation passes are handled in the C++ Controller classes so that each pass must be explicitly called. This allows for greater customization on when each pass should be called, and with what parameters. In particular this enables setting different parameters for the gate fusion optimization pass depending on the QasmController simulation method.
+*   Add `gate_length_units` kwarg to `qiskit.providers.aer.noise.NoiseModel.from_device()` for specifying custom `gate_lengths` in the device noise model function to handle unit conversions for internal code.
+*   Add Controlled-Y (“cy”) gate to the Stabilizer simulator methods supported gateset.
+*   For Aer’s backend the jsonschema validation of input qobj objects from terra is now opt-in instead of being enabled by default. If you want to enable jsonschema validation of qobj set the `validate` kwarg on the `qiskit.providers.aer.QasmSimualtor.run()` method for the backend object to `True`.
+
+<span id="release-notes-0-5-1-bug-fixes" />
+
+<span id="id490" />
+
+#### Bug Fixes
+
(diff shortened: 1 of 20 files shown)


Eric-Arellano:
Thanks! Can you please put in the PR description an explanation of how you did this? For example, if automated, then a copy of the source code.

arnaucasau:
Of course! done!

docs/api/qiskit/release-notes/0.18.md
```diff
@@ -5,6 +5,71 @@ description: New features and bug fixes
 
 # Qiskit 0.18 release notes
 
+## 0.18.3
```
Eric-Arellano, line 8: Weird that there's no 0.18.1 and 0.18.2 in this file. But that's how it already was. This looks right.

Eric-Arellano (approved):
This looks great! I manually inspected ~70% of the releases. 

Excellent job describing how you generated the change :) For future reference, it can also be useful to run `git diff` or `git show` to get the code diff. Then you can use a codeblock with the language set as `diff`. This is really useful because GitHub will color additions as green and removals as red. Other developers can also save the diff and run `git apply` on it to have Git automatically make the code change.

docs/api/qiskit/release-notes/0.18.md
arnaucasau: Yeah, I noticed too. I checked in the legacy_release_notes.md and there are no 0.18.1 and 0.18.2 either

PR #592, opened 2024-01-05
Run extended checks on API updates

Eric-Arellano:
As of https://github.com/Qiskit/documentation/pull/569 and https://github.com/Qiskit/documentation/pull/572, our extended check PR now checks more than simply external links. It also checks:

* that all pages render for translations and API docs
* the internal links for API docs

So, we should run these extended checks when we make changes to the API and translations folders.


frankharkins (approved):
Nice, thanks!

PR #596, opened 2024-01-08
Split sphinxHtmlToMarkdown into helper functions

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/223. There are still improvements we can make like better testing. 

This PR tries to avoid making code changes and only moves the code to helper functions.

### DIFF
--- a/scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
+++ b/scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
@@ -36,28 +36,208 @@ import { Root } from "mdast";
 export async function sphinxHtmlToMarkdown(options: {
   html: string;
   url: string;
-  imageDestination?: string;
+  imageDestination: string;
   // url links to a fixed version and ending in /
   // https://github.com/Qiskit/qiskit-ibm-runtime/tree/0.9.2/
-  baseSourceUrl?: string;
-  releaseNotesTitle?: string;
+  baseSourceUrl: string;
+  releaseNotesTitle: string;
 }): Promise<SphinxToMdResult> {
-  const {
-    html,
-    url,
-    imageDestination = "/images/api/",
-    baseSourceUrl,
-    releaseNotesTitle,
-  } = options;
-  const meta: PythonObjectMeta = {};
-  const isReleaseNotes = url.endsWith("release_notes.html") ? true : false;
-
+  const { html, url, imageDestination, baseSourceUrl, releaseNotesTitle } =
+    options;
   const $ = load(html);
   const $main = $(`[role='main']`);
+
+  const isReleaseNotes = url.endsWith("release_notes.html");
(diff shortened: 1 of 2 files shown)


scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -36,28 +36,208 @@ import { Root } from "mdast";
...
   html: string;
   url: string;
-  imageDestination?: string;
+  imageDestination: string;
   // url links to a fixed version and ending in /
   // https://github.com/Qiskit/qiskit-ibm-runtime/tree/0.9.2/
-  baseSourceUrl?: string;
-  releaseNotesTitle?: string;
+  baseSourceUrl: string;
+  releaseNotesTitle: string;
```
Eric-Arellano, line 43: These were only optional for the sake of testing. But that's a dangerous strategy because it's really important that production never forget to set the arguments.

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -36,28 +36,208 @@ import { Root } from "mdast";
...
+  }
+
+  removeHtmlExtensionsInRelativeLinks($, $main);
+  removePermalinks($main);
+  removeDownloadSourceCode($main);
+  handleTabs($, $main);
+  addLanguageClassToCodeBlocks($, $main);
+  replaceSourceLinksWithGitHub($, $main, baseSourceUrl);
+  convertRubricsToHeaders($, $main);
+  processSimpleFieldLists($, $main);
```
Eric-Arellano, line 63: I don't like this name very much, but I still don't understand well all the edge cases of what this function is doing. We should rename it once we have a better understanding of the semantic significance.

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -36,28 +36,208 @@ import { Root } from "mdast";
...
+  removeColons($main);
+  preserveMathBlockWhitespace($, $main);
```
Eric-Arellano, line 65: This used to happen later in the sequence. I moved it here because it "fits" better here with the other processing.

scripts/lib/sphinx/sphinxHtmlToMarkdown.ts
```diff
@@ -36,28 +36,208 @@ import { Root } from "mdast";
...
+  $main.find('a[title="Link to this heading"]').remove();
+  $main.find('a[title="Link to this definition"]').remove();
+}
+
+function removeDownloadSourceCode($main: Cheerio<any>): void {
+  $main.find("p > a.reference.download.internal").closest("p").remove();
+}
+
+/**
+ * Convert sphinx-design tabs.
```
Eric-Arellano, line 144: This comment is new.

frankharkins (approved):
I find this much easier to read, thank you!

PR #598, opened 2024-01-08
Sessions don't work on sims

### LINKED ISSUE #583: Update session doc to indicate it doesn't work with simulator
**What is the expected feature or enhancement?**

Session documentation today uses simulator as an example, but session doesn't actually work with simulator backends, since there is no queue. It's usually not a problem since simulator jobs usually run immediately, as they would in a session. But occasionally they do back up, which lead to users questioning why their active session jobs do not run right away.

**Acceptance criteria**

Clarify how simulator jobs behave in a session.

beckykd:
Closes #583

### DIFF
--- a/docs/run/sessions.mdx
+++ b/docs/run/sessions.mdx
@@ -9,7 +9,7 @@ description: An overview of sessions and when to use them.
 A session is a Qiskit Runtime feature that lets you efficiently run multi-job iterative workloads on quantum computers. Using sessions helps avoid delays caused by queuing each job separately, which can be particularly useful for iterative tasks that require frequent communication between classical and quantum resources.
 
 <Admonition type="note">
-The queuing time does not decrease for the first job submitted within a session. Therefore, a session does not provide any benefits when running a single job. Additionally, sessions do not work on simulators because they do not have a queue.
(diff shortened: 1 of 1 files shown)


Eric-Arellano (approved):

docs/run/sessions.mdx
```diff
@@ -9,7 +9,7 @@ description: An overview of sessions and when to use them.
 A session is a Qiskit Runtime feature that lets you efficiently run multi-job iterative workloads on quantum computers. Using sessions helps avoid delays caused by queuing each job separately, which can be particularly useful for iterative tasks that require frequent communication between classical and quantum resources.
 
 <Admonition type="note">
-The queuing time does not decrease for the first job submitted within a session. Therefore, a session does not provide any benefits when running a single job.
+The queuing time does not decrease for the first job submitted within a session. Therefore, a session does not provide any benefits when running a single job. Additionally, sessions do not work on simulators because they do not have a queue.
```
abbycross, line 12:
```suggestion
The queuing time does not decrease for the first job submitted within a session. Therefore, a session does not provide any benefits when running a single job. Additionally, sessions do not work on simulators because simulators do not have a queue.
```

abbycross: small clarification

abbycross (approved):

PR #601, opened 2024-01-09
Automate notebook execution

frankharkins:
<sub>(Sorry for all the noise, this is difficult to test locally)</sub>

This PR adds two features to the notebook testing action.

### 1. Update PRs with executed notebooks

Adding the "Auto execute" label to a PR will trigger the workflow to run any changed notebooks, then push these changes to the PR branch. This makes it very easy to get a copy of the notebook executed in CI.

### 2. Execute all notebooks and make a PR

You can also now [dispatch this workflow](https://github.com/Qiskit/documentation/actions/workflows/notebook-test.yml), which will execute all notebooks and make a PR with the updates. This will be useful when we update Python packages. See #600 for an example.

### DIFF
--- a/.github/workflows/notebook-test.yml
+++ b/.github/workflows/notebook-test.yml
@@ -31,7 +31,6 @@ on:
       - "docs/**/*.ipynb"
       - "!docs/api/**/*"
     types: [opened, synchronize, labeled]
-  issue_comment:
   workflow_dispatch:
 jobs:
   execute:
--- a/.github/workflows/notebook-test.yml
+++ b/.github/workflows/notebook-test.yml
@@ -12,15 +12,32 @@
 
 name: Test notebooks
 on:
+  # Three event types can trigger this workflow:
+  #  1. Pull request (opened, synchronize)
+  #     On creating or pushing to a PR branch, this action runs any changed
+  #     notebooks and reports pass / fail. This is to catch broken code
+  #     examples.
+  #  2. Pull request (labeled)
+  #     On adding the `🤖 Auto execute` label (id: 6409509633) to a PR, this
+  #     workflow executes any changed notebooks, then pushes the updates to the
+  #     PR branch. This makes it easy to quickly get notebooks executed in the
+  #     CI environment.
+  #  3. Workflow dispatch
+  #     When triggered by a workflow_dispatch event, we execute all notebooks,
+  #     and make a PR to the branch with the updated notebooks. This is handy
+  #     when upgrading qiskit package versions.
   pull_request:
     paths:
       - "docs/**/*.ipynb"
       - "!docs/api/**/*"
+    types: [opened, synchronize, labeled]
+  issue_comment:
   workflow_dispatch:
 jobs:
   execute:
     name: Execute notebooks
     runs-on: ubuntu-latest
+    if: github.event.action != 'labeled' || github.event.label.id == 6409509633
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v4
@@ -29,6 +46,7 @@ jobs:
           cache: "pip"
 
       - name: Get all changed files
+        if: github.event_name == 'pull_request'
         id: all-changed-files
         uses: tj-actions/changed-files@af2816c65436325c50621100d67f6e853cd1b0f1
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/notebook-test.yml
```diff
@@ -12,15 +12,32 @@
...
+  #  3. Workflow dispatch
+  #     When triggered by a workflow_dispatch event, we execute all notebooks,
+  #     and make a PR to the branch with the updated notebooks. This is handy
+  #     when upgrading qiskit package versions.
   pull_request:
     paths:
       - "docs/**/*.ipynb"
       - "!docs/api/**/*"
+    types: [opened, synchronize, labeled]
+  issue_comment:
```
frankharkins, line 34:
```suggestion
```

.github/workflows/notebook-test.yml
```diff
@@ -12,15 +12,32 @@
...
   workflow_dispatch:
 jobs:
   execute:
     name: Execute notebooks
     runs-on: ubuntu-latest
+    if: github.event.action != 'labeled' || github.event.label.id == 6409509633
```
frankharkins, line 40: I chose to go by label ID so we can change the name if needed. Note if you delete the label this feature will stop working.

Eric-Arellano:
Thanks, Frank! I know you spent a lot of time figuring this out. It's nifty.

I have some concerns if it's worth the downsides.

1. The auto-created PR will not run CI because of GitHub's restriction that a workflow cannot trigger other workflows, unless it was triggered by workflow_dispatch.
2. It decouples the notebook update from the PR making the relevant changes. For example, if we were to update Qiskit versions, then the notebooks should be updated in the same PR. They're tightly coupled. 
3. This adds substantial complexity to the workflow. Beyond the lines of code added, it makes you reason about multiple different execution paths: is this a workflow_dispatch, or a PR label, or a sync? 

Instead, while a little clunky, I suspect the original approach isn't That Bad. When you want to update notebooks, download the changes from CI and push a commit with the update.

--

> You can also now [dispatch this workflow](https://github.com/Qiskit/documentation/actions/workflows/notebook-test.yml), which will execute all notebooks and make a PR with the updates.

You could do this beforehand. Only you'd need to manually create the PR.

--

What do you think?

frankharkins:
@Eric-Arellano thanks for the review and honest opinion!

> 1. The auto-created PR will not run CI because of GitHub's restriction that a workflow cannot trigger other workflows, unless it was triggered by workflow_dispatch.
> 2. It decouples the notebook update from the PR making the relevant changes. For example, if we were to update Qiskit versions, then the notebooks should be updated in the same PR. They're tightly coupled.

I imagined you'd trigger the workflow from the "feature" branch that updates the package. This would make a PR to the feature branch, you'd review it, merge it, then merge the feature branch to main. A simpler option is to have the action just push to the feature branch (no pr).

> 3. This adds substantial complexity to the workflow. Beyond the lines of code added, it makes you reason about multiple different execution paths: is this a workflow_dispatch, or a PR label, or a sync?

I agree and am OK dropping this if you think the complexity outweighs it's usefulness. This was a fun project for me but the end result is more complex than I hoped. No one asked for this feature and I don't want to impose unnecessary burden on future maintainers.

>> You can also now [dispatch this workflow](https://github.com/Qiskit/documentation/actions/workflows/notebook-test.yml), which will execute all notebooks and make a PR with the updates.
>
> You could do this beforehand. Only you'd need to manually create the PR.

While the action had a `workflow_dispatch` trigger, it didn't actually work as the "changed files" was empty. Another option is to remove the PR-label feature and just keep the dispatch feature.

Eric-Arellano:
> While the action had a workflow_dispatch trigger, it didn't actually work as the "changed files" was empty. 

Ah, that would be great to fix! If a workflow_dispatch, ideally we would regenerate every single notebook. No need to do other stuff though like auto-opening a PR; the change is fixing "changed files" to handle the edge case.

> I imagined you'd trigger the workflow from the "feature" branch that updates the package.

Historically our team hasn't used many feature branches. So this would require changing their workflow.

> I agree and am OK dropping this if you think the complexity outweighs it's usefulness. This was a fun project for me but the end result is more complex than I hoped. No one asked for this feature and I don't want to impose unnecessary burden on future maintainers.

Okay, sounds good! I think that's the right call here. +1 to this being nifty and a cool project nevertheless :) 

It would be very useful still to fix workflow_dispatch to work properly.

PR #602, opened 2024-01-09
Reorganize API docs files

Eric-Arellano:
Before, files were split between `scripts/lib` and `scripts/lib/sphinx` even though they were all specifically relevant to the API generation script. Now they all live under `scripts/lib/api`.

This PR also:

* Renames `SphinxToMdResult` to `HtmlToMdResult`
* Renames `sphinxHtmlToMd.ts` to `htmlToMd.ts`
* Renames `PythonObjectMeta` to `Metadata` and renames its attribute names.
* Separates imports from first-party and third-party

TypeScript was useful at making sure everywhere was correctly updated.

--

This isn't necessarily the final file organization. I think we still might want to split `htmlToMd.ts` into smaller files. But this is some progress.


frankharkins (approved):
Thanks!

PR #612, opened 2024-01-10
Reflect changes in all templates

beckykd:


.github/ISSUE_TEMPLATE/ui.yml
```diff
@@ -11,6 +11,22 @@ body:
   - type: markdown
     attributes:
       value: Thank you for reporting!  Please also use the search to see if there are any other relevant issues or pull requests.
+  - type: markdown
```
Eric-Arellano, line 14: I don't think one makes sense. UI issues should always be in this repo.

.github/ISSUE_TEMPLATE/user-feedback.yml
```diff
@@ -11,6 +11,22 @@ body:
   - type: markdown
     attributes:
       value: Thank you for offering feedback! This form is intended for general user impressions and reactions to the Qiskit and IBM Quantum documentation. If you would like to report a bug or typo, request new content, or tell us about a problem with the UI, please return to the <a href="https://github.com/Qiskit/documentation/issues/new/choose">issues</a> page and select one of the other issue types. If you have questions about this form or more information to share, please reach out to <a href="mailto:quantum.design@us.ibm.com">quantum.design@us.ibm.com</a>.
+  - type: markdown
```
Eric-Arellano, line 14: I'm not certain this one makes sense either. I want to make sure that our team is aware of overall user feedback.

.github/ISSUE_TEMPLATE/ui.yml
beckykd: OK - I was on the fence with this one; in my experience people sometimes don't know what the UI really is, so it might be a problem with any "web page" they find, but we can always transfer issues if necessary.

Eric-Arellano (approved):
Looks good, but please update the PR title before merging.

PR #616, opened 2024-01-10
improve readability

beckykd:
Just playing with some formatting


beckykd:
The table view didn't turn out well, so for now I'm changing the subtitles so they are all the same.  Dev is looking into making different text sizes for h3 and bold, so I'll watch that and see if it gives further improvement.

abbycross (approved):
👍

PR #618, opened 2024-01-11
Add some tests for `processHtml.ts`

Eric-Arellano:
It's useful for us to have unit tests for each helper function in `processHtml.ts`, whereas right now are only tests are more like integration tests in `htmlToMd.ts`.

1. Unit tests make it easier to test all the edge cases.
2. Unit tests make are a form of documentation for what each helper function is doing.
3. Unit tests make it more obvious when something breaks where the break comes from. With an integration test, you have to look at all the code in the pipeline.

I'm envisioning that the tests in `htmlToMd.ts` should be rewritten in a follow up to be more full-on integration tests, like choose 2-3 pages and convert the entire pages.

There are still some missing tests in `processHtml.ts` that we should add in a follow up.


scripts/lib/api/htmlToMd.test.ts
```diff
@@ -829,78 +829,6 @@ By default this is sys.stdout.</p></li>
     `);
   });
 
-  test("extract module metadata", async () => {
```
Eric-Arellano, line 832: This is now tested in processHtml.test.ts

scripts/lib/api/processHtml.test.ts
```diff
@@ -0,0 +1,300 @@
...
+      {
+        src: "http://localhost:3000/_static/images/view-page-source-icon.svg",
+        dest: "/my-images/view-page-source-icon.svg",
+      },
+    ]);
+    doc.expectHtml(`<img src="/my-images/view-page-source-icon.svg">`);
+  });
+});
+
+test("handleSphinxDesignCards()", () => {
```
Eric-Arellano, line 101: This test has a lot of HTML and it's hard to see what's happening. Idk if I should add comments to the test? Or is it reasonable that people look at the source code to see what's happening? For example, the source code already explains what it does in the docstring.

scripts/lib/api/processHtml.test.ts
```diff
@@ -0,0 +1,300 @@
...
+});
+
+test("addLanguageClassToCodeBlocks()", () => {
+  const doc1 = Doc.load(`<p><strong>Circuit symbol:</strong></p>
+    <div class="highlight-default notranslate"><div class="highlight"><pre><span></span>     ┌──────────┐
+    q_0: ┤ U(ϴ,φ,λ) ├
+        └──────────┘
+    </pre></div>
+    </div>`);
+  // TODO(EA): This behavior is bad. We shouldn't be setting the language as Python because it's not always correct, such
```
Eric-Arellano, line 210: I'm not fixing this here because it will result in changes when we regenerate all the API docs.

scripts/lib/api/processHtml.ts
```diff
@@ -39,24 +39,24 @@ export function processHtml(options: {
   const isReleaseNotes = url.endsWith("release_notes.html");
```
Eric-Arellano, line 39: I renamed some functions in this file to make it more clear what's happening.

scripts/lib/api/processHtml.ts
```diff
@@ -70,30 +70,25 @@ export function loadImages(
...
-      const dest = `${imageDestination}/${filename}`;
-
-      $img.attr("src", dest);
-
+      let dest = `${imageDestination}/${filename}`;
       if (isReleaseNotes) {
         // Release notes links should point to the current version
-        $img.attr("src", dest.replace(/[0-9].*\//, ""));
+        dest = dest.replace(/[0-9].*\//, "");
       }
```
Eric-Arellano, line 87: This is a change. We before returned the wrong `dest` in the `Image` type and we were only fixing the `dest` for the HTML itself.

Eric-Arellano (commented):
We still need more tests. But I want to land this since Arnau might pick up refactors again soon, and I want to try getting source code links to work.

scripts/lib/api/processHtml.ts
arnaucasau: Good catch! This change is important to have in mind when we regenerate the API docs in the future because we will override the images of the current version, but I don't see why not with the new release notes mechanism.

arnaucasau (approved):
Awesome work! Thank you Eric!

scripts/lib/api/processHtml.ts
```diff
@@ -39,24 +39,24 @@ export function processHtml(options: {
   const isReleaseNotes = url.endsWith("release_notes.html");
   const images = loadImages($, $main, url, imageDestination, isReleaseNotes);
   if (url.endsWith("release_notes.html")) {
-    setReleaseNotesHeading($, releaseNotesTitle);
+    renameAllH1s($, releaseNotesTitle);
```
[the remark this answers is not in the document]
Eric-Arellano: I think there's still a little bit of use of this being a dedicated function because it keeps the main block extremely _declarative_ rather than _imperative_.

PR #619, opened 2024-01-11
Fix API generation inlining methods mechanism

arnaucasau:
This PR fixes a problem we have when regenerating the API docs with pages that already have the methods inlined. The `mergeClassMembers.ts` script replaces all the content under the methods and attributes sections for the inlined members, but if there are none, the sections end up empty.

For example, in Qiskit v0.26, we can see how files, like `docs/api/qiskit/0.26/qiskit.extensions.HamiltonianGate.md`, have the methods and attributes sections with a table of contents of all its elements, and after the sections, we find all methods/attributes mixed up and ordered alphabetically.

The result of applying the script to files like that one was to have the methods and attributes sections empty, and the actual members listed below without any change.

```md
## Methods
## Attributes

{all methods and attributes alphabetically sorted}
```

This PR adds an if condition to check when we have new methods/attributes to inline and when we need to keep the old content, like the table of contents mentioned before. It also adds a new test to verify that situation.


Eric-Arellano (commented):
Yay! With the test, I would have expected the table to be replaced with the inlined methods/atributes. Is that not happening because you don't have `HtmlToMdResult`s describing those code objects?

arnaucasau:
Yes, because I wanted to test the situation where we don't have any other dedicated method page. In those cases, we keep the table and not overwrite the content under the existing sections, so the result is the same as the input. The actual method would be in the same file somewhere, and the table has the links to them. It's true that I should change the links to make it more clear :+1:

Eric-Arellano (approved):

Eric-Arellano (approved):

Issue #518, opened 2023-12-12
Design GitHub source code links


Eric-Arellano:
This ticket is to track the visual design of https://github.com/Qiskit/documentation/issues/454.

# Current implementation

We simply use an external link on the same line as the code object's description. It always appears at the end of the code and is normal text - we don't do any funky floating or anything.

<img width="586" alt="Screenshot 2024-01-11 at 4 49 15 PM" src="https://github.com/Qiskit/documentation/assets/14852634/7828fec0-2126-45dd-a176-c1e8090f4e1d">

I suspect this design is probably fine other than weirdness with the padding. Note how the underline extends directly to the code. I couldn't fix this by adding extra whitespace in the content - it wouldn't work. I think we need to adjust the CSS. 

<img width="104" alt="Screenshot 2024-01-11 at 4 50 32 PM" src="https://github.com/Qiskit/documentation/assets/14852634/b35c9811-3174-4398-9592-8086b675c24a">

But what is tricky about adjusting the CSS is it will impact all inline code highlighting:

<img width="338" alt="Screenshot 2024-01-11 at 4 51 25 PM" src="https://github.com/Qiskit/documentation/assets/14852634/5b4d2fc7-9bd2-472a-94a3-0ad6f3b655b0">

Note that we're using padding-right for inline code, rather than margin-right, which would result in the background not being the shaded gray.

# Original notes

## What the old qiskit.org design looks like

<img width="714" alt="Screenshot 2023-12-12 at 12 46 18 PM" src="https://github.com/Qiskit/documentation/assets/14852634/bca8048f-3ec8-4978-83c9-43c610a0ebb4">

When you click the `[source]` link, it takes you a new webpage on qiskit.org with a copy of the source code:

<img width="794" alt="Screenshot 2023-12-12 at 12 46 54 PM" src="https://github.com/Qiskit/documentation/assets/14852634/fbdb3c5f-047a-4c06-8988-a52659d27f7b">

Our implementation will instead take you to GitHub.com, so be an external link.

Note how the old design handles the `#` icon to get the permalink for the current code object. This only happens on hover:

<img width="515" alt="Screenshot 2023-12-12 at 12 47 51 PM" src="https://github.com/Qiskit/documentation/assets/14852634/6aac227f-a9c9-4134-8177-1f063fced91d">

## Which code objects have links?

Note that not every single code object has a link, For example, in the first image in this issue, only `exception QiskitNatureError` has it, but not its method `with_traceback()`.

[...]

Issue #517, opened 2023-12-12
Switch Qiskit, Runtime, and Provider to have more precise source code links


Eric-Arellano:
Thanks to https://github.com/Qiskit/documentation/pull/620, we now have source code links for these three projects back to GitHub. But those links only take you to the overall file and it would be better if we took you to the specific lines of code.

See https://github.com/Qiskit/qiskit_sphinx_theme/issues/186#issuecomment-1651974704 for how we'd do this. We'd also need to update this repository to know how to consume `sphinx.ext.linkcode`.

melechlapson:
Hi Eric, Could this not jus tbe resolved by adding #L<line number> to the end of the URLs so that it will ink to the line in the file not just the file itself?

Eric-Arellano:
> Could this not jus tbe resolved by adding #L to the end of the URLs so that it will ink to the line in the file not just the file itself?

That's the end state that we want, indeed. But we need an _automated_ way to get the specific line numbers. It's not realistic for a human to go and add the line numbers for every code object.

So, we need to switch the way that Sphinx generates the GitHub URLs. Right now, we're using `sphinx.ext.viewcode` which only takes us to the overall file. `sphinx.ext.linkcode` should let us instead have the specific line numbers.

melechlapson:
I think I'd like to take on this issue. It would be my first contribution

melechlapson:
@Eric-Arellano can you assign me this issue?

Eric-Arellano:
Hey @melechlapson, great to hear you're interested in this one! So there will be two stages to implementing:

1. Figure out how to get `sphinx.ext.linkcode` working well in the source repositories themselves.
2. Improve our API generation script so that we can consume Sphinx's output and convert it into something understood by the docs application.

Those are sequential, so let's start with step 1. Even if things get busy and you don't have time to do step 2, it would still be a big help to figure out step 1.

To do this, you'll want to choose a Sphinx-based project and experiment with how to get useful URLs generated that link to specific lines of code. Reminder that the goal is specific lines of code; we already know how to link to the overall file. You may find it convenient to test with our example docs in `qiskit-sphinx-theme` since it's a simple set up designed to experiment: https://github.com/Qiskit/qiskit_sphinx_theme. You'd be using the `example_docs` directory.

[...]

Issue #454, opened 2023-11-30
Epic: link to source code / GitHub in API docs


aeddins-ibm:
### Feedback

Before the API docs migrated to ibm.com, IIRC the doc pages often included links to the relevant Github source code. [edit: apparently it may not have been github, but some other hosted version of the source code?] These are useful for verifying details of how the code behaves. It would be nice to have those back if it's not too technically difficult.

An example workflow: 
A user needs to know something about something in the API, say SparsePauliOp. They google "SparsePauliOp" and get the doc page (usually the top result), which might already have the info they need. If it doesn't, then the Github link on that page means with 1 click they get to the relevant source code, where they can find out what they need to know.

Now with the new docs, if the doc page does not have the desired info, then the user will probably do a new Google search "github qiskit", select the second result for qiskit/qiskit, then in that page use the Github search box to search for "SparsePauliOp", pick the most promising search-result snippet, etc. Of course it's doable but it adds time/clicks to something a user/developer might do many times, so direct links from the docs would be nice to have.

javabster:
Yeah I agree I think a GitHub link would be preferred, I'm guessing our link checker would be able to catch if any of the  links get out of sync? I'm imagining a case where a branch name changes (e.g when it goes from main to stable to a historical version number) or something and then the url no longer works

Eric-Arellano:
> I'm guessing our link checker would be able to catch if any of the links get out of sync?

Yes, although only as a nightly cron job because external link checking was really slow and sometimes flaky.

>  I'm imagining a case where a branch name changes (e.g when it goes from main to stable to a historical version number) or something and then the url no longer works

The link would always be a permalink to the exact commit used when the docs were built. That's important because code can change between versions.

Eric-Arellano:
Thanks again for the upvotes everyone! 

FYI I split this into three tasks so that we can more easily track how things are going:

- https://github.com/Qiskit/documentation/issues/518
- https://github.com/Qiskit/documentation/issues/517
- https://github.com/Qiskit/documentation/issues/519

The designers are going to start soon on figuring out a design. Let us know if anyone wants to try out https://github.com/Qiskit/documentation/issues/517! Otherwise, one of us core maintainers will try to get to it near the end of December or earlier January.

javabster:
question - is the plan for this source code link to just be available for the api-ref pages? If possible I think it could be a nice addition to the non-api-ref pages as well, what do you think?

Eric-Arellano:
[...]

PR #620, opened 2024-01-11
Add source links when using `sphinx.ext.viewcode`

Eric-Arellano:
Implements https://github.com/Qiskit/documentation/issues/454 for projects that are using `sphinx.ext.viewcode`. All 3 of our projects were historically using it until I turned it off after removing qiskit.org, so the historical API docs are good to go. I'll restore it for current versions of these projects and regenerate the docs. 

This implementation is not a perfect implementation:

1. The visual design is a little awkward, especially the lack of padding. This can be improved via https://github.com/Qiskit/documentation/issues/518

Screenshot 2024-01-11 at 12 28 07 PM

2. The link only takes you to the overall code page, not the specific lines. This could be improved via https://github.com/Qiskit/documentation/issues/517.

But it's good enough to not block on these improvements.

This PR regenerates all Runtime historical versions, but not any current versions nor Qiskit historical versions.

## How source code URLs are determined

`sphinx.ext.viewcode` embeds a copy of every Python file used in API docs and uses internal relative links like `../modules/qiskit_ibm_runtime/ibm_backend.html`. They correspond to Python files we can be confident exist. We transform those relative links into GitHub links here:

https://github.com/Qiskit/documentation/blob/790e9372f64ab7d5f15eaccc229b4d0765781d44/scripts/lib/api/processHtml.ts#L94-L105

https://github.com/Qiskit/documentation/blob/790e9372f64ab7d5f15eaccc229b4d0765781d44/scripts/lib/api/processHtml.ts#L163-L183

Our links assume that there is a branch called `stable/<versionWithoutPatch` in GitHub, like `stable/0.8`.

## Replacing `[source]` with `GitHub ↗︎`

We remove the original link from Sphinx and replace it with our own. This is important so that the link is not included in the `<code>` HTML element incorrectly. It also allows us to set a custom link label and `title` (the text when highlighting).

### DIFF
--- a/docs/api/qiskit-ibm-runtime/0.14/qiskit_ibm_runtime.Estimator.md
+++ b/docs/api/qiskit-ibm-runtime/0.14/qiskit_ibm_runtime.Estimator.md
@@ -10,7 +10,7 @@ python_api_name: qiskit_ibm_runtime.Estimator
 
 <span id="qiskit_ibm_runtime.Estimator" />
 
-`Estimator(backend=None, session=None, options=None)`
+`Estimator(backend=None, session=None, options=None)`[GitHub](https://github.com/qiskit/qiskit-ibm-runtime/tree/stable/0.14/qiskit_ibm_runtime/estimator.py "view source code")
 
 Class for interacting with Qiskit Runtime Estimator primitive service.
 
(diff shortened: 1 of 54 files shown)


arnaucasau (approved):
Looks good to me! Thank you Eric! Tested locally and all works well!

Eric-Arellano:
"Extended checks" fails but not due to anything from this PR. It was already broken. I'll fix in a follow up.

PR #624, opened 2024-01-11
Add GitHub links for Provider API docs

Eric-Arellano:
I didn't realize that Provider already is using the `viewcode` Sphinx extension for current docs, so we didn't need to make any change to that repository to get GitHub source links working.

Note that this required adding a new mechanism to special-case certain files because they don't map 1:1 between GitHub and sphinx.ext.viewcode.

Other than that change, this PR was generated automatically by running `npm run gen-api -- -p qiskit-ibm-provider -v 0.7.3 -a https://github.com/Qiskit/qiskit-ibm-provider/actions/runs/7330787504/artifacts/1135354934`.

### DIFF
--- a/scripts/lib/api/processHtml.ts
+++ b/scripts/lib/api/processHtml.ts
@@ -187,12 +187,12 @@ export function replaceViewcodeLinksWithGitHub(
     ) {
       return;
     }
-    //_modules/qiskit_ibm_runtime/ibm_backend
-    let match = href.match(/_modules\/(.*?)(#|$)/)![1];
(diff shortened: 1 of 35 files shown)


Eric-Arellano:
Hm, this has a legit broken link! Let's wait to merge this. Good job link checker :)

scripts/lib/api/processHtml.ts
frankharkins, line 1: Neat solution, thanks

frankharkins (approved):
Thanks Eric! I tried a few links and they all pointed to the correct files

Issue #547, opened 2023-12-21
External link checker should create a GitHub issue with errors


Eric-Arellano:
I didn't realize that we've had broken links for a couple of days. That's because right now you have to manually check https://github.com/Qiskit/documentation/actions/workflows/external-link-checker.yml.

<img width="1157" alt="Screenshot 2023-12-21 at 2 43 49 PM" src="https://github.com/Qiskit/documentation/assets/14852634/186f3b5f-f7cf-4173-a9b4-8d43dc4b9868">

It'd be neat if instead the action would create a GitHub issue with any problems. Ideally, it would update the existing issue rather than creating a new issue every day.

Eric-Arellano:
@arnaucasau @frankharkins can I get your thoughts on what you think is desirable? The question is how to handle multiple days in a row of failures. If Monday was broken and we already opened an issue, what do we do on Tuesday if it's not yet fixed?

1. Create a new issue?
2. Update an existing issue?

I believe that updating an existing issue is harder than it sounds - we need to preserve state across runs of what the latest issue was. The only viable/simple alternative for 2 I can think of us is to hardcode a specific issue number. Our action will make sure that issue is open and either add a comment or edit the issue description. Because it reopens the issue if closed, it's fine for people to close the issue whenever it's fixed.

What do you think? If you like my plan for updating an existing issue, do you think it's better to rewrite the PR description or add a new comment? The new comment will trigger a new GitHub notification, whereas editing is silent.

arnaucasau:
I like your plan of updating an existing issue rather than creating a new one every day. I think we can have one fixed issue that through GitHub actions will be re-opened when necessary to edit the description. In my opinion, editing is better than creating a new comment to avoid having multiple notifications for the same links, but we can still add a new comment if we want only when we have a new broken link, and not when the broken links are the same as the day before. Something like:

```bash
Link checker searches for broken links

if found and the issue is closed:
    Re-open the issue 

Update the description with the broken links

if new broken links
    Post a comment to inform about the new broken links
```

The last step can be done by comparing the output of the link checker and the issue description. To avoid having multiple comments informing about the new broken links, maybe we can delete the previous one (if existed) and create a new comment to trigger the notifications.

However, having one issue with the description updated with all the broken links and not having more notifications than the first one (re-open the issue), I think is possibly enough. What do you think?

Some information I found about the GitHub actions:
1. https://github.com/peter-evans/create-or-update-comment
2. https://github.com/peter-evans/find-comment

[...]

PR #625, opened 2024-01-12
Use dedicated workflow for page rendering check for historical API docs

Eric-Arellano:
It takes over 3 hours for us to check that every page renders in the app. So, it was too slow to have that check every time we make a change to API docs.

Still, it's a useful check and we can't expect humans to run that check, realistically. So, we keep the check but only run it once a week and when translations are changed.


frankharkins (commented):
Thanks! For the scheduled job, do you want to create an issue or comment if it fails? Otherwise we might miss it.

Eric-Arellano:
Yes, but it's a little more work than I first expected. See https://github.com/Qiskit/documentation/issues/547. I don't think we should block on it.

frankharkins:
Since this only runs once a week, what you think of just making a new issue each time? Then the command should be a one-liner

Eric-Arellano:
> Since this only runs once a week, what you think of just making a new issue each time? Then the command should be a one-liner.
>
> EDIT: To be clear, I don't have a strong opinion on this

I think it'd be good to add, but I'd rather not block on this and it's blocking https://github.com/Qiskit/documentation/pull/624 because the current CI check is so slow.

frankharkins (approved):

PR #522, opened 2023-12-14
Add objects.inv

frankharkins:
- Added `objectsInv.ts` to read, decompress, and parse `objects.inv` files (and vice versa)
- Updated `updateApiDocs.ts` to copy the file over from the artifact
- Added logic to transform links and delete any references that don't exist in our docs

### Explanation

The `objects.inv` file is an index that [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) uses to create cross-references to our documentation. To allow other projects to cross-reference Qiskit, we need to host this file.

For a description of the file's data structure, see the [sphobjinv documentation](https://sphobjinv.readthedocs.io/en/stable/syntax.html), whose implementation I copied when working out how to transform this file.

PR #628, opened 2024-01-12
Make link extraction from markdown more explicit

Eric-Arellano:
This is prework for https://github.com/Qiskit/documentation/pull/522, which adds `sphinx.inv`. We want to check the links contained in `sphinx.inv`, but we don't expect anyone to have links pointing to `sphinx.inv`. 

Our original abstractions made it hard to model `objects.inv` correctly with the link checker. Now, `markdown.ts` is renamed to `extractLinks.ts` and it solely deals with parsing files. It no longer has a bad coupling to the complex `linksToOriginFiles` variable from `FileBatch.ts`. 

This should make it much more obvious how to handle `objects.inv`.

### DIFF
--- a/scripts/lib/links/extractLinks.ts
+++ b/scripts/lib/links/extractLinks.ts
@@ -22,6 +22,13 @@ import rehypeRemark from "rehype-remark";
 import rehypeParse from "rehype-parse";
 import remarkGfm from "remark-gfm";
 
+export type ParsedFile = {
+  /** Anchors that the file defines. These can be linked to from other files. */
+  anchors: string[];
+  /** Links that this file has to other places. These need to be validated. */
+  links: string[];
+};
+
 interface JupyterCell {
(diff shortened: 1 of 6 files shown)


scripts/lib/links/extractLinks.ts
```diff
@@ -36,47 +43,38 @@ export function markdownFromNotebook(rawContent: string): string {
...
-    const entry = linksToOriginFiles.get(link);
-    if (entry === undefined) {
-      linksToOriginFiles.set(link, [filePath]);
-    } else {
-      entry.push(filePath);
-    }
-  };
-
-  unified()
+export async function parseLinks(markdown: string): Promise<string[]> {
```
frankharkins, line 53: This feels like a more natural separation of logic

frankharkins (approved):
Thanks!

PR #637, opened 2024-01-17
Regeneration Qiskit API historical docs v0.44-v0.42

arnaucasau:
Part of #542.

Regeneration of the Qiskit API docs (historical versions) with the methods inlined and the GitHub source links (thank you @Eric-Arellano). Given the number of files modified, this PR only updates the versions 0.44, 0.43, and 0.42. The rest of the versions will be added in a follow-up.


Eric-Arellano:
The new plan is to have a dedicated PR for 0.44, then subsequent PRs are 2 versions per batch. We want to closely review things because it's important we get the quality right.

PR #638, opened 2024-01-17
Simplify `Pkg` type by removing unused properties

Eric-Arellano:
We're not using `initialUrl`, `baseUrl`, and `ignore` anymore. I need to add some new complexity to `Pkg` soon, so I plan to make it a proper `class`. This is prework.

This also renames `lib/sharedTypes.ts` to `lib/api/Pkg.ts`. The types weren't shared with any other project and the abstractions are specific to API docs generation.

Finally, this simplifies our implementation of handling `__init__.py` with GitHub source code links.


scripts/lib/api/Pkg.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { join } from "path/posix";
+
+import { Link } from "../sharedTypes";
+
+/**
+ * Python package (e.g. qiskit, qiskit-ibm-runtime).
+ * This information does not depend on the generation script
+ */
+export type PkgInfo = {
```
Eric-Arellano, line 21: No longer has `ignore` and `initialUrl`

scripts/lib/api/Pkg.ts
```diff
@@ -0,0 +1,48 @@
...
+  hasSeparateReleaseNotes?: boolean;
+  transformLink?: (link: Link) => Link | undefined;
+};
+
+/**
+ * Information about the specific package version we're dealing with.
+ * This stuff _does_ depend on the parameters passed to the generation script,
+ * and may be updated at the script runs.
+ */
+export type Pkg = PkgInfo & {
```
Eric-Arellano, line 34: No longer has `baseUrl`

scripts/lib/api/generateToc.test.ts
```diff
@@ -148,10 +149,4 @@ const pkg = {
...
   releaseNoteEntries: [],
   historical: false,
-
-  // These properties are not used but included to appease typescript
-  githubSlug: "qiskit/qiskit-ibm-runtime",
-  baseUrl: `https://qiskit.org/ecosystem/ibm-runtime`,
-  initialUrl: `https://qiskit.org/ecosystem/ibm-runtime/apidocs/ibm-runtime.html`,
-  transformLink: undefined,
-};
+} as unknown as Pkg;
```
Eric-Arellano, line 152: Idk why TypeScript made me do this `as unknown` here but not in another test.

scripts/commands/updateApiDocs.ts
```diff
@@ -184,27 +176,13 @@ zxMain(async () => {
...
   await convertHtmlToMarkdown(
-    `${destination}/artifact`,
+    `${artifactFolder}/artifact`,
     outputDir,
     baseGitHubUrl,
     pkg,
   );
 });
 
-/**
```
Eric-Arellano, line 194: Moved to `fs.ts` so that this file is smaller.

scripts/lib/api/generateToc.test.ts
frankharkins: Maybe you need `GithubSlug`?

Eric-Arellano: Yeah, that's what the error message was saying. But to clarify, I don't want to have to specify irrelevant types here which IMO makes the test harder to read. We get around this here:

https://github.com/Qiskit/documentation/blob/1ae2fd56e55573395897326ecb3913839d456c4c/scripts/lib/api/addFrontMatter.test.ts#L56-L58

arnaucasau: Making `GithubSlug` optional allows us to drop the cast to unknown, but maybe it doesn't make sense because all the projects are required to have it, right?
```ts
export type PkgInfo = {
  name: string;
  githubSlug?: string;
```

frankharkins: Hmm... or maybe the type in the function?

```ts
(): Pkg => {
```

Eric-Arellano: I don't want to make it optional because it's important that we set `githubSlug` in production. Instead, I'm planning to make `Pkg` a proper class in a follow up PR and to have a static constructor like `Pkg.test()` that uses default arts for every option so that tests are easier for us to work with and we avoid these "incomplete types".

scripts/commands/updateApiDocs.ts
```diff
@@ -223,14 +200,12 @@ async function convertHtmlToMarkdown(
...
 
-  const ignore = pkg.ignore ?? (() => false);
-
   let results: Array<HtmlToMdResult & { url: string }> = [];
   for (const file of files) {
     const html = await readFile(join(htmlPath, file), "utf-8");
     const result = await sphinxHtmlToMarkdown({
       html,
-      url: `${pkg.baseUrl}/${file}`,
+      url: `http://localhost:8000/${file}`,
```
Eric-Arellano, line 208: Great job @arnaucasau realizing that we still need to set `localhost:8000` for the image code to work. But note that we don't actually ever use that URL! I will simplify this code in a follow up PR -- the simplification impacts too many lines of code.

scripts/lib/api/generateToc.test.ts
frankharkins: The constructor is a good idea

arnaucasau (approved):
Looks good to me, and I like the idea of making `Pkg` a class in a follow-up! Thank you Eric!

frankharkins (approved):
Just tested and it worked for me, thanks!

PR #639, opened 2024-01-17
Simplify image extraction code

Eric-Arellano:
We can simplify our image extraction code now that we copy images from the downloaded CI artifact rather than downloading it from a URL. Before, we stored `src` as a URL; now we just need the `fileName`.

### DIFF
--- a/scripts/lib/api/htmlToMd.test.ts
+++ b/scripts/lib/api/htmlToMd.test.ts
@@ -111,38 +111,6 @@ describe("sphinxHtmlToMarkdown", () => {
     ).toMatchInlineSnapshot(`""`);
   });
 
-  test("extract images", async () => {
-    expect(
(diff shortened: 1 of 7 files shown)


scripts/lib/api/htmlToMd.test.ts
```diff
@@ -111,38 +111,6 @@ describe("sphinxHtmlToMarkdown", () => {
     ).toMatchInlineSnapshot(`""`);
   });
 
-  test("extract images", async () => {
```
Eric-Arellano, line 114: Already tested in `processHtml.test.ts`

arnaucasau (approved):
Awesome, thank you Eric! Tested it locally with Qiskit v0.45 (historical and non-historical ) and all works well. I left one suggestion to avoid using the doesnt-matter link.

PR #641, opened 2024-01-17
Refactor `Pkg` to be a class for API generation

Eric-Arellano:
To correctly generate source code links for Qiskit historical docs, we need complex logic to determine the `baseGithubUrl` because the versions between Qiskit and Qiskit Terra were different. Rather than putting that logic in `upsideApiDocs.ts`, it's cleaner to abstract it behind a method `Pkg.baseGithubUrl()`.

There are a few other methods that fit naturally on `Pkg`, rather than being free functions.

This also allows us to make the `Pkg` interface immutable. Before, we would build up the `Pkg` over time by modifying attributes. Now, every attribute is `readonly` and we set up everything in the constructor. That's useful because you no longer have to worry if you're dealing with a properly initialized `Pkg`; it will now always be properly initialized.

### DIFF
--- a/scripts/commands/updateApiDocs.ts
+++ b/scripts/commands/updateApiDocs.ts
@@ -31,7 +31,7 @@ import { updateLinks } from "../lib/api/updateLinks";
 import { specialCaseResults } from "../lib/api/specialCaseResults";
 import addFrontMatter from "../lib/api/addFrontMatter";
 import { dedupeHtmlIdsFromResults } from "../lib/api/dedupeHtmlIds";
-import { Pkg, VALID_PACKAGE_NAMES } from "../lib/api/Pkg";
+import { Pkg } from "../lib/api/Pkg";
 import { zxMain } from "../lib/zx";
 import { pathExists, getRoot, rmFilesInFolder } from "../lib/fs";
 import { downloadCIArtifact } from "../lib/api/downloadCIArtifacts";
@@ -56,7 +56,7 @@ const readArgs = (): Arguments => {
     .option("package", {
       alias: "p",
       type: "string",
-      choices: VALID_PACKAGE_NAMES,
+      choices: Pkg.VALID_NAMES,
       demandOption: true,
       description: "Which package to update",
     })
--- a/scripts/commands/updateApiDocs.ts
+++ b/scripts/commands/updateApiDocs.ts
@@ -101,19 +101,14 @@ zxMain(async () => {
   );
   pkg.releaseNoteEntries = await findLegacyReleaseNotes(pkg);
 
-  const artifactFolder = `${getRoot()}/.out/python/sources/${pkg.name}/${
-    pkg.version
-  }`;
-
+  const artifactFolder = pkg.ciArtifactFolder();
   if (await pathExists(artifactFolder)) {
     console.log(`Skip downloading sources for ${pkg.name}:${pkg.version}`);
   } else {
     await downloadCIArtifact(pkg.name, args.artifact, artifactFolder);
   }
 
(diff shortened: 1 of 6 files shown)


scripts/commands/updateApiDocs.ts
```diff
@@ -52,54 +50,13 @@ interface Arguments {
   artifact: string;
 }
 
-function transformLink(link: Link): Link | undefined {
```
Eric-Arellano, line 55: Moved without any changes.

scripts/lib/api/Pkg.ts
```diff
@@ -12,6 +12,16 @@
...
 import { join } from "path/posix";
 
+import { findLegacyReleaseNotes } from "./releaseNotes";
+import { removePrefix, removeSuffix } from "../stringUtils";
+import { getRoot } from "../fs";
+
+export const VALID_PACKAGE_NAMES = [
+  "qiskit",
+  "qiskit-ibm-runtime",
+  "qiskit-ibm-provider",
```
Eric-Arellano, line 22: Hardcoding the 3 values is less general than we had before. But the set is small and stable enough that I think it's best to keep it simple.

scripts/lib/api/Pkg.ts
```diff
@@ -20,35 +30,156 @@ export interface Link {
...
+      hasSeparateReleaseNotes: kwargs.hasSeparateReleaseNotes ?? false,
+      transformLink: kwargs.transformLink,
+      version: kwargs.version ?? "0.1.0",
+      versionWithoutPatch: kwargs.versionWithoutPatch ?? "0.1",
+      historical: kwargs.historical ?? false,
+      releaseNoteEntries: kwargs.releaseNoteEntries ?? [],
+    });
+  }
+
+  outputDir(parentDir: string): string {
```
Eric-Arellano, line 150: I got rid of the default arg `parentDir = "docs"`. It wasn't used and imo the default was dangerous because you might fail to realize you need to set it.

scripts/lib/api/releaseNotes.ts
```diff
@@ -10,29 +10,23 @@
...
 import { readFile, writeFile, readdir } from "fs/promises";
 
-interface releaseNoteEntry {
-  title: string;
-  url: string;
-}
+import { $ } from "zx";
+
+import { getRoot, pathExists } from "../fs";
+import type { Pkg, ReleaseNoteEntry } from "./Pkg";
```
Eric-Arellano, line 19: The `type` is necessary to avoid a circular dependency.

Eric-Arellano (commented):
It's annoying how much boilerplate there is with TS classes. But I think this change is worth it.

scripts/commands/updateApiDocs.ts
```diff
@@ -31,13 +31,11 @@ import { updateLinks } from "../lib/api/updateLinks";
 import { specialCaseResults } from "../lib/api/specialCaseResults";
 import addFrontMatter from "../lib/api/addFrontMatter";
 import { dedupeHtmlIdsFromResults } from "../lib/api/dedupeHtmlIds";
-import { removePrefix, removeSuffix } from "../lib/stringUtils";
-import { Link, Pkg, PkgInfo, getPkgRoot } from "../lib/api/Pkg";
+import { Pkg, VALID_PACKAGE_NAMES } from "../lib/api/Pkg";
```
frankharkins, line 34: Nit: Could `VALID_PACKAGE_NAMES` be a static attribute of `Pkg`?

frankharkins (approved):
Thanks, this is much neater 🚀

PR #642, opened 2024-01-18
Report notebook warnings

frankharkins:
The notebook testing script does not report any warnings raised in notebooks. It's important to report these so we can fix deprecations before they're removed from Qiskit, and to spot any unnecessary warnings that might harm the user experience.

I've tried this in the past and the best method is to search the executed notebook outputs for warnings. That's what I do here.

### Screenshots

Screenshot 2024-01-18 at 12 45 19

### DIFF
--- a/scripts/nb-tester/test-notebook.py
+++ b/scripts/nb-tester/test-notebook.py
@@ -12,6 +12,8 @@
 
 import argparse
 import sys
+import warnings
+import textwrap
 from dataclasses import dataclass
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


scripts/nb-tester/test-notebook.py
```diff
@@ -39,6 +41,39 @@ class ExecuteOptions:
...
+    def report(self):
+        message = f"Warning detected in cell {self.cell_index}:\n"
+        for line in self.msg.splitlines():
+            message += (
+                textwrap.fill(
+                    line, width=77, initial_indent=" │ ", subsequent_indent=" │ "
+                )
+                + "\n"
+            )
+        print(f"\033[0;33m{message}\033[0m", flush=True)
```
Eric-Arellano, line 58: It'd be good to de-magic the escape codes by creating class CONSTANTS that say what they correspond to. Another approach is to create a helper function like this:

https://github.com/pantsbuild/pants/blob/e7b4b251b74f5e8332e6d527d0c7a5a3ffc878ab/build-support/migration-support/convert_source_to_sources.py#L95-L96

Eric-Arellano (approved):
Thanks!

scripts/nb-tester/test-notebook.py
```diff
@@ -49,16 +84,23 @@ def execute_notebook(path: Path, options: ExecuteOptions) -> bool:
...
     except possible_exceptions as err:
         print("\r❌\n")
         print(err)
         return False
+
+    notebook_warnings = extract_warnings(nb)
+    if notebook_warnings:
+        print("\r⚠️")
+        [w.report() for w in notebook_warnings]
+        return True
```
Eric-Arellano, line 97: Should this return `False`? We want to error if there were warnings imo.

frankharkins: I set it as `True` because some warnings are out of our control, e.g. if a Qiskit function uses a deprecated numpy function.

It depends if you'd rather have false positives or false negatives.

Eric-Arellano: What if we require you to allow-list deprecations? I'm concerned if we don't error, then no one will ever notice the warnings. People don't check CI logs proactively.

frankharkins: Good point, maybe I'm over-planning; I'll set it to `False` and we can revisit it if we need.

Eric-Arellano: > because some warnings are out of our control, e.g. if a Qiskit function uses a deprecated numpy function.

Ideally that will prompt us to go fix Qiskit itself. In the meantime to them fixing it, we could allow-list it.

Eric-Arellano (approved):
Thanks!

PR #644, opened 2024-01-18
Separate GitHub URL logic from HTML processing

Eric-Arellano:
This is prework for fixing GitHub source code URLs for Qiskit back when it was a metapackage. We need to link out to multiple GitHub repositories for historical versions of Qiskit. So it's no longer viable to have a single `baseGithubUrl` per `Pkg`—the `baseGithubUrl` depends on the specific file in question, like `qiskit.aer.my_file` should go to the Aer repository whereas `qiskit.transpiler` should go to Qiskit Terra.

To prepare for that, this PR moves all the computation of the full GitHub URL to live in `Pkg`. This is a cleaner interface anyways; it separates out the URL computation from the HTML processing done inside `processHtml.ts`.


arnaucasau (approved):
Awesome, good thinking with the `determineGithubUrlFn()` function! Tested it locally with historical and non-historical versions and works well!

PR #645, opened 2024-01-18
Extend make-historical script to use objects.inv

arnaucasau:
The API generation script is now able to work with `objects.inv` files thanks to @frankharkins at https://github.com/Qiskit/documentation/pull/522. This PR allows us to copy over the objects.file when we want to make a version historical running `npm run make-historical -- -p <projectName>`.


Eric-Arellano (approved):
Great attention to detail!

PR #647, opened 2024-01-18
Handle historical Qiskit versions with GitHub source links

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/454. Historical versions of Qiskit docs are tricky because of qiskit-metapackage referring to multiple distinct GitHub repositories.

### DIFF
--- a/scripts/lib/api/Pkg.ts
+++ b/scripts/lib/api/Pkg.ts
@@ -167,18 +167,169 @@ export class Pkg {
     // file `my_module.py`. We need to add back the `/__init__.py` when linking to GitHub.
     const convertToInitPy = new Set([
       "qiskit_ibm_provider",
+      "qiskit/qasm",
       "qiskit/qasm2",
       "qiskit/qasm3",
       "qiskit/transpiler/preset_passmanagers",
     ]);
-    const baseUrl = `https://github.com/${this.githubSlug}/tree/stable/${this.versionWithoutPatch}/`;
-    return (fileName) => {
-      if (convertToInitPy.has(fileName)) {
-        fileName = `${fileName}/__init__`;
-      }
-      return `${baseUrl}${fileName}.py`;
-    };
+    const normalizeFile = (fp: string) =>
+      convertToInitPy.has(fp) ? `${fp}/__init__` : fp;
+
+    // Provider, Runtime, and Qiskit 0.45+ are simple: there is a branch called `stable/<version>`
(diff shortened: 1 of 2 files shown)


scripts/lib/api/Pkg.ts
```diff
@@ -167,18 +167,169 @@ export class Pkg {
...
+      throw new Error(
+        `No compatible version found for the file ${fileName} with qiskit-metapackage ${metapackageVersion}}`,
+      );
+    }
+    return result;
+  };
+
+  let slug: string;
+  let version: string;
+  if (
```
Eric-Arellano, line 312: Matt T reviewed the `if` statements to make sure this covers all the import paths.

scripts/lib/api/Pkg.ts
```diff
@@ -167,18 +167,169 @@ export class Pkg {
...
+    // multiple packages. Refer to the version table in api/qiskit/release-notes/0.44.md.
+    return (fileName) =>
+      determineHistoricalQiskitGithubUrl(
+        this.versionWithoutPatch,
+        normalizeFile(fileName),
+      );
+  }
+}
+
+const QISKIT_METAPACKAGE_TO_TERRA = new Map([
```
Eric-Arellano, line 197: These versions come from https://docs.quantum.ibm.com/api/qiskit/release-notes/0.44. I double checked them and tried to be careful - I don't think you need to spend the time rechecking.

scripts/lib/api/Pkg.ts
```diff
@@ -167,18 +167,169 @@ export class Pkg {
...
+      };
+    }
+
+    // Otherwise, we have to deal with Qiskit <0.45, when we had the qiskit-metapackage comprised of
+    // multiple packages. Refer to the version table in api/qiskit/release-notes/0.44.md.
+    return (fileName) =>
+      determineHistoricalQiskitGithubUrl(
+        this.versionWithoutPatch,
+        normalizeFile(fileName),
+      );
```
Eric-Arellano, line 193: This pattern is really neat. We're using partial application. We want to return a function `(string) => string`, but we need to pass in the extra argument `metapackageVersion` for it to work.

arnaucasau (approved):
Nice use of the partial application! This will unblock the regeneration of the historical versions, and later the ibm.com redirects. I tested it locally and triple-checked the versions in the package map, and all looks good to me.

PR #648, opened 2024-01-19
Fix setting class name for videos

Eric-Arellano:
Even though this looks like an HTML element, it's rendered by React so apparently needs to be called `className`.


arnaucasau (approved):
LGTM! Thanks!

PR #654, opened 2024-01-19
Fix some broken learning links

Eric-Arellano:
learn.qiskit.org recently stopped working so the redirects are failing, breaking CI on main.


arnaucasau (approved):
LGTM! Thank you Eric!

PR #656, opened 2024-01-19
Disable objects.inv for historical API docs

Eric-Arellano:
This is meant to unblock regenerating the Qiskit API docs. But more generally, I think it's not a necessary feature to support for now. 

1. I've never seen a project's conf.py not point to latest
2. I'm not confident it even works with pointing to the correct subfoler. (That's ignorance - it might work!)
3. We can add this mechanism in the future whenever we want


arnaucasau (approved):
Thank you Eric. Tested it with historical and non-historical versions.

PR #664, opened 2024-01-22
Regeneration Qiskit v0.41 API docs

arnaucasau:
Part of #542.

Regeneration of the Qiskit API docs (historical versions) with the methods inlined and the GitHub source links. Given the number of files modified, the regeneration is split into different PRs. This one only updates the version 0.41.


Eric-Arellano (approved):
Thanks! I still can't find any instances of where inlining had failed previously. Maybe that was only broken for certain versions of Sphinx. That makes sense to me - the original script author hadn't run the script on old versions of Qiskit/Sphinx.

This looks good to me. Did you have a chance to use the link checker to see if any of the GitHub links are broken? To speed things up, we could also wait to fix those broken links in one dedicated PR if easier - like regen everything, then have a dedicated PR to clean up after the broken links.

PR #665, opened 2024-01-22
API generation script - Fix loadImages function

arnaucasau:
The `processHtml.ts` script assumes that we always have an `src` attribute in the `img` tags, and even though this is true for the latest API versions, for Qiskit historical versions, we can find cases like the following example:

**Qiskit v0.41 release_notes.html line 23360**
```html
<div class="animation">
  <img id="_anim_imga6548212aea7457ca5d2f940fabeaea3">
  <div class="anim-controls">
    <input id="_anim_slidera6548212aea7457ca5d2f940fabeaea3" type="range" class="anim-slider"
```

This PR adds an early return to the `loadImage` function to skip the images without a `src` attribute given that we can't find the image in the `_images` folder of the GitHub artifact.

### DIFF
--- a/scripts/lib/api/processHtml.ts
+++ b/scripts/lib/api/processHtml.ts
@@ -73,11 +73,16 @@ export function loadImages(
   imageDestination: string,
   isReleaseNotes: boolean,
 ): Image[] {
-  return $main
+  const images: Image[] = [];
+  $main
     .find("img")
     .toArray()
-    .map((img) => {
+    .forEach((img) => {
       const $img = $(img);
+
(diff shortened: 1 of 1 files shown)


frankharkins (commented):
Nice fix! This logic looks good but maybe using `filter` would be neater?

Eric-Arellano (approved):
+1 to `.filter().map()` being more idiomatic. Good find!

arnaucasau:
Good idea @frankharkins! Thank you both for the review

Eric-Arellano (approved):

PR #666, opened 2024-01-22
Minor fixes for markdown code examples

### LINKED ISSUE #609: Audit all non-api ref code blocks in preparation for qiskit 1.0
before the release of qiskit 1.0 in feb we should check that all code cells in docs are runnable using qiskit 1.0 code. This could be done running the notebooks locally with 0.46 or the 1.0 pre-release  installed (once either is released).

The goal is for none of the code blocks to break or throw deprecation warnings with 1.0

frankharkins:
Minor tweaks to make the code examples run. Closes #609.


Eric-Arellano (approved):
Thank you! Do you know why we didn't know about these failures earlier?

Eric-Arellano:
> then never got tested since.

I thought the CI check would have caught the issues? Is it because the CI check only runs on changed notebooks?

frankharkins:
>  Is it because the CI check only runs on changed notebooks?

Yes that's right; markdown needs testing manually

Eric-Arellano:
Ah, duh. Didn't realize this is MDX. Got it. Thanks!

PR #674, opened 2024-01-22
Address image issues and associated text in Transpiler stages page

### LINKED ISSUE #657: Image issues in transpiler stages topic

https://docs.quantum.ibm.com/transpile/transpiler-stages

### Select all that apply

- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [X] other

### Describe the fix.

There are a couple of issues with images on this page.  

- [ ] The first image is in dark mode.
- [ ] transpiler-11 is in there twice.  I'm pretty sure the one after "The transpilation output varies due to the stochastic SWAP mapper. Therefore, the numbers below will likely change each time you run the code." is supposed to be a different image.

abbycross:
Closes #657


abbycross:
@kaelynj do you still have this image? -- and if so, would it be complicated to switch the background to light mode? We have functionality to automatically change an image based on the user's light/dark mode preferences.

image

Eric-Arellano:
> We have functionality to automatically change an image based on the user's light/dark mode preferences.

If and only if we provide the dark mode variant of the image.

abbycross:
> > We have functionality to automatically change an image based on the user's light/dark mode preferences.
> 
> If and only if we provide the dark mode variant of the image.

The image we have now would serve as the dark mode version -- unless it can't for some reason @Eric-Arellano ?

kaelynj:
I only have the dark mode version unfortunately.  I got it from someone else to make slides from and can ask if they have a light mode version of it.

frankharkins:
If we can't get the light-mode version, you can make a rough one by inverting the image, then inverting the hue to preserve the colours.

298724279-fe821ccb-0f53-47da-83d6-478a1974b162

abbycross:
> If we can't get the light-mode version, you can make a rough one by inverting the image, then inverting the hue to preserve the colours.
> 

Thanks for the tip @frankharkins ! I tried out bringing in the image you created and it works to switch from light to dark.

abbycross:
@kaelynj I've added in the new images, as you can see in this preview: https://qiskit-docs-preview-pr-674.1799mxdls7qz.us-south.codeengine.appdomain.cloud/transpile/transpiler-stages

I'm not thrilled with the sizing and the alignment of the hardware map image... wonder if we can instead produce it directly in the notebook as output? I fussed with it but can't seem to reproduce it myself within the notebook. 

Feel free to write directly to this branch to make the necessary text edits. Thanks!

docs/transpile/transpiler-stages.ipynb
```diff
@@ -125,7 +124,11 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This distribution is quite wide, signaling the difficulty the SWAP mapper is having in computing the best mapping.\n",
+    "This distribution is quite wide, signaling the difficulty the SWAP mapper is having in computing the best mapping.  In fact, we can take a closer look at what hardware qubits were chosen to run the circuit:\n",
+    "\n",
+    "<img src=\"/images/transpile/transpiler-stages/hardware-qubit-graph.png\" alt=\"Layout of qubits on hardware\" width=\"350%\" />\n",
+    "\n",
+    "As you can see this selection of qubits isn't even fully connected to each other!  This means unnecessary SWAP gates must be inserted in order to execute all of the two-qubit gates because we naively chose which hardware qubits to use in our circuit.\n",
```
abbycross, line 131:
```suggestion
    "This distribution is quite wide, which demonstrates how difficult it is for the SWAP mapper to compute the best mapping. Look closely at which hardware qubits were chosen to run the circuit, and note that this selection of qubits isn't fully connected to each other.\n",
    "\n",
    "<img src=\"/images/transpile/transpiler-stages/hardware-qubit-graph.png\" alt=\"Layout of qubits on hardware\" width=\"350%\" />\n",
    "\n",
    "This requires inserting unnecessary SWAP gates to execute all the two-qubit gates because we naively chose which hardware qubits to use in our circuit.\n",
```

kaelynj (approved):
These changes look good to me!

kaelynj:
Ah, @Eric-Arellano, looks like some of the checks are failing b/c the tox environment does not contain the `graphviz` library.  Is it straightforward to add in?

Eric-Arellano:
> Is it straightforward to add in/are there any concerns about the prod environment?

It's installed at the OS-level, not Python package manager. So it's more complex than normal but not too bad. You would add `graphviz` to this line:

https://github.com/Qiskit/documentation/blob/75aaa317a204068fb940cc2776b0ce5935471cb2/.github/workflows/notebook-test.yml#L46-L50

Rename the job title from "Install LaTeX dependencies" to "Install Linux dependencies"

PR #677, opened 2024-01-23
Fix broken release note link

Eric-Arellano:
It started 404ing today.


arnaucasau (approved):
LGTM! Thanks! It seems like the same broken link appears twice in that file, in line 484 you can find the other one

PR #678, opened 2024-01-23
Regeneration Qiskit v0.38 and v0.37 API docs

arnaucasau:
Part of #542.

Regeneration of the Qiskit API docs (historical versions) with the methods inlined and the GitHub source links. Given the number of files modified, the regeneration is split into different PRs. This one only updates the versions 0.38 and 0.37.

All the GitHub source links work!


Eric-Arellano (approved):
Nice. Looks like this is almost entirely adding GitHub links, which is verified by how the number of lines added is almost the same as removed.

Issue #681, opened 2024-01-23
Qiskit's historical API versions don't match their docs


arnaucasau:
We have three Qiskit versions (0.36, 0.35, and 0.34) with the documentation shifted. Those three versions have the documentation of the next version instead of its own:

- Docs in qiskit/0.36 correspond to qiskit/0.37
- Docs in qiskit/0.35 correspond to qiskit/0.36
- Docs in qiskit/0.34 correspond to qiskit/0.35
- We don't have docs for qiskit/0.34

We should shift every version except qiskit/0.36 given that we currently have version 0.37.0 there whereas in qiskit/0.37 we have version 0.37.2.

Once the shifting is done we need to remove qiskit/0.34 which will be duplicated in qiskit/0.35.

To shift every version we only need to rename the version folder in .out with the correct one and regenerate the docs. For example, to shift qiskit/0.35 we can run:

```bash
mv .out/python/sources/qiskit/0.35.0 .out/python/sources/qiskit/0.36.0
npm run gen-api -- -p qiskit -v 0.36.0 -a "" --historical
```

**Versions regenerated:**

- [x] qiskit 0.36
- [x] qiskit 0.35
- [x] qiskit 0.34

Eric-Arellano:
Removed and redirects are working. https://docs.quantum.ibm.com/api/qiskit/0.34 goes to https://docs.quantum.ibm.com/api/qiskit/0.35. Good work @arnaucasau!

<img width="210" alt="Screenshot 2024-01-30 at 11 50 32 AM" src="https://github.com/Qiskit/documentation/assets/14852634/49d5e0c6-b509-4e27-889d-9803e8d71d73">

PR #682, opened 2024-01-23
Regeneration Qiskit v0.35 API docs

arnaucasau:
Part of #542.

Regeneration of the Qiskit API docs (historical versions) with the methods inlined and the GitHub source links. Given the number of files modified, the regeneration is split into different PRs. This one only updates the version 0.35.

In addition, this PR corrects one part of the problem we have with versions 0.36, 0.35, and 0.34 described in #681. Those three versions have the documentation of the next version instead of its own. The rest of the versions will be corrected in a follow-up.

All the GitHub source links work!


Eric-Arellano (approved):
Excellent. This is a huge user improvement

PR #701, opened 2024-01-24
GitHub source link by tag: special case

arnaucasau:
We can't access the file `qiskit/optimization/converters/quadratic_program_to_negative_value_oracle` using the `stable/0.7` branch because the file was removed at patch 0.7.4. Instead, we need a special case to access the file using the tag 0.7.3


Eric-Arellano (approved):
Thanks!

PR #703, opened 2024-01-24
Restore Qiskit 0.45.2 docs

Eric-Arellano:
This reverts commit cd859c2ee75097f29038ff1f2cc4af3caebdbaca.

@kevinsung recently fixed the docs app's rendering of some math expressions. I re-deployed the Docker image we use with `npm run check-pages-render`, so this should confirm that it's now safe to merge this change.


arnaucasau (approved):
LGTM, thanks!

PR #683, opened 2024-01-23
Improve function handling

frankharkins:
This PR improves how the API generation script handles functions.

### Demote some `h2` tags to `strong`

We currently convert all `.rubric` elements to `h2`, but this makes some headings higher-level than they should be (specifically "Example" and "References" headings). For an example see [api/qiskit/converters](https://docs.quantum.ibm.com/api/qiskit/converters).

Screenshot 2024-01-23 at 17 37 33

This PR converts rubrics to `strong` (rather than `h2`) if the name is "Example(s)" or "Reference(s)". This is a bit hacky, but it seems to work and I can't think of a better way of doing it. The Sphinx LaTeX writer also takes a [similar approach](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-rubric).

I've regenerated the docs using this new logic and I can see no false positives (that is, cases in which we use `strong` where we should use `h2`).

### Add function names as headings

Functions currently have no headings. This PR adds a `h3` with the function name, which improves readability, gives anchor tag icons, and will add more functions to the right sidebar.

I've also included some logic to avoid adding function names when the page is already dedicated to the function (such as [least_busy](https://docs.quantum.ibm.com/api/qiskit-ibm-provider/qiskit_ibm_provider.least_busy)).

One quirk with this I spotted was some function names not appearing in the sidebar in `/api/qiskit/utils` (see `add_deprecation_to_docstring` in the screenshot). This is possibly because the title hierarchy skips `h2`.

Screenshot 2024-01-23 at 18 00 37

***

Closes #231

PR #709, opened 2024-01-25
Add back _package.json

frankharkins:
- Restore `_package.json` that was accidentally deleted in #683

### DIFF
--- a/scripts/lib/api/releaseNotes.ts
+++ b/scripts/lib/api/releaseNotes.ts
@@ -11,7 +11,7 @@
 // that they have been altered from the originals.
 
 import { parse } from "path";
-import { readFile, writeFile, readdir } from "fs/promises";
+import { readFile, writeFile, readdir, access } from "fs/promises";
 
 import { $ } from "zx";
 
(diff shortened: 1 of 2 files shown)


scripts/lib/api/releaseNotes.ts
```diff
@@ -108,12 +108,17 @@ export async function updateHistoricalTocFiles(pkg: Pkg): Promise<void> {
   ).filter((file) => file.isDirectory() && file.name.match(/[0-9].*/));
 
   for (let folder of historicalFolders) {
-    let tocFile = await readFile(
-      `${getRoot()}/docs/api/${pkg.name}/${folder.name}/_toc.json`,
-      {
-        encoding: "utf8",
-      },
-    );
+    const tocFilePath = `${getRoot()}/docs/api/${pkg.name}/${
```
Eric-Arellano, line 111: Why this code? I would expect `_toc.json` to always be defined.

Eric-Arellano (commented):
Thanks! Do you know if it was deleted manually or via the script?

scripts/lib/api/releaseNotes.ts
Eric-Arellano: > Handle empty historical version folders

Oh, try running `git clean -fd`.

frankharkins: > Do you know if it was deleted manually or via the script?

The script deleted it then errored due to the empty folder so didn't recreate it.

> Oh, try running git clean -fd.

I've deleted the empty folder now I know that's what caused it, this fix is to prevent anyone else tripping over the same problem.

Eric-Arellano: I don't think it's worth the code complexity. Generally I _would_ want us to error if we had empty folders

frankharkins: Possibly because I generated the docs with the historical flag at some point? Honestly not sure; I can remove it if you prefer.

Eric-Arellano: Ah, yeah possibly. I'd prefer to not have it because it's extra code complexity and I _would_ want us to error if we had a malformed historical folder.

Eric-Arellano (approved):

PR #713, opened 2024-01-26
Add headers for exceptions

### LINKED ISSUE #712: Exceptions are missing headers in API docs
See https://docs.quantum.ibm.com/api/qiskit/exceptions. It would be useful for the exceptions to have headers so that:

1. They show up in the right sidebar
2. They are more visually distinct
3. You can get an anchor link for them

This will look very similar to https://github.com/Qiskit/documentation/pull/683. I suspect we also want to use `###` (h3) for now.

Screenshot 2024-01-25 at 5 02 57 PM

Screenshot 2024-01-25 at 5 03 13 PM

frankharkins:
Quick fix to use the same logic for functions and exceptions. Closes #712.

A known problem with this implementation is that some headings should not be `h3`, but it depends on the page. See https://github.com/Qiskit/documentation/issues/714.

### DIFF
--- a/docs/api/qiskit/algorithms.md
+++ b/docs/api/qiskit/algorithms.md
@@ -193,6 +193,8 @@ Algorithms that compute the fidelity of pairs of quantum states.
 
 ### Exceptions
 
+### AlgorithmError
+
 <span id="qiskit.algorithms.AlgorithmError" />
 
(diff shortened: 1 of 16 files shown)


Eric-Arellano (approved):
💜

Issue #673, opened 2024-01-22
500 error in historical API docs file due to Katex


Eric-Arellano:
https://docs.quantum.ibm.com/api/qiskit/0.42/qiskit.algorithms.linear_solvers.AbsoluteAverage has a 500 error. I suspect due to invalid math as we changed recently some of our code to pre-process math expressions.

--

I think this was not caught by our CI job to check that all pages render because we are using an outdated Docker image in `./start` that doesn't have those math expressions. The page renders fine on `localhost:3000` when using `./start`. So, the first step is to re-release the Docker image used by `./start` and then rerun our Pages Render CI job to see what other pages are also broken.

Update: this was indeed true. It is now fixed.

--

Note that we cannot change the source code for historical Qiskit docs easily because we can't regenerate those docs. So we have a few options:

1. Change the copy of the files from qiskit.org that we've been using. Make sure everyone uses the updated values.
2. Add special-casing code to our API generation script in this repo
3. Add special-casing code to the app itself in the internal repo

This is going to be a problem for fixing historical link 404s, too: https://github.com/Qiskit/documentation/issues/495

I strongly suspect that the best fix is changing iqp-channel-docs. Note that the file used to render fine, so I believe this was an issue with the special casing we added to the docs app to handle MathJax -> Katex. We probably are mis-applying a rule.

Eric-Arellano:
This is still broken even with a recent change made to the docs platform.

```
Error compiling file.
ParseError: KaTeX parse error: Undefined control sequence: \abs at position 1: \̲a̲b̲s̲{\frac{1}{N}\su…
    at new ParseError (/home/node/app/packages/preview/.next/server/chunks/108.js:37:524676)
    at Parser.parseGroup (/home/node/app/packages/preview/.next/server/chunks/108.js:37:802815)
    at Parser.parseAtom (/home/node/app/packages/preview/.next/server/chunks/108.js:37:797200)
    at Parser.parseExpression (/home/node/app/packages/preview/.next/server/chunks/108.js:37:796118)
    at Parser.parse (/home/node/app/packages/preview/.next/server/chunks/108.js:37:795592)
    at parseTree (/home/node/app/packages/preview/.next/server/chunks/108.js:37:805474)
    at renderToDomTree (/home/node/app/packages/preview/.next/server/chunks/108.js:37:806448)
    at Object.renderToString (/home/node/app/packages/preview/.next/server/chunks/108.js:37:806588)
[...]

PR #718, opened 2024-01-29
Improve the API script to handle `|` characters

### LINKED ISSUE #488: Improve API script to handle `|` inside table element values
See https://github.com/Qiskit/documentation/pull/486. We need to escape the `|` symbol. 

I'm unclear if this is only a problem within Latex `$ $` expressions. Also, we had to use `\vert` there instead of `\|`. It might be fine to only handle Latex for now.

It's important we unit test this. That will require getting the original Sphinx HTML. We could recreate in RST a test by looking at this offending RST, then generating the HTML by running a Sphinx build:

https://github.com/Qiskit/qiskit/blob/fa05da3fd8bd5ae95fc8dad08139edfa85378f93/qiskit/algorithms/amplitude_estimators/iae.py#L189-L190

That RST needs to be inside a Sphinx table.

arnaucasau:
### Summary

This PR changes how we handle pipes inside math expressions when we convert sphinx HTML into markdown. Some math expressions use pipes to define quantum states using the Dirac notation, and we need to escape those characters to avoid breaking the page when the pipe characters are used inside a markdown table.

### Details

One solution could only handle the `|` characters used inside a table, but given that the math expressions could be used in nested tags (e.g `<td> <p> <span class="math"> SOME_EXPRESSION </span></p></td>`), and would need to make the script more complex without ensuring we fix all the cases where we could make the page to fail to render, I decided to handle that character differently in all math expressions.

The PR replaces the `|` character with `\vert ` which will represent the same character. `\vert` needs extra space at the end to handle cases where the pipe was next to a non-numerical character. In those cases, we should avoid converting `|x` to `\vertx` given that the latter is not a valid command (`\vert x` is the correct conversion).

We also need to take into account that, in some cases, we still need to use the `|` characters because when escaped (`\|`), it represents a double pipe (`||`), which could be used in different mathematical expressions like the length of a vector.

This is the regex used, which only matches pipe characters not preceded by a backslash:
```ts
/(?<!\\)\|/gm
```

A new test was added to verify different cases where we can find a pipe character. The tests checks `|` characters outside math expressions, in math expressions inside a table, and in math expressions outside the table. We can also find a case where we intentionally want to use `\|` to create a double pipe. In the following screenshot, we can see the rendered result of the test.

test-example

Closes #488

### DIFF
--- a/scripts/lib/api/htmlToMd.test.ts
+++ b/scripts/lib/api/htmlToMd.test.ts
@@ -1042,7 +1042,7 @@ bits.</p>
 <div role='main'>
 <div class="math notranslate nohighlight">
 \\[\\begin{split}CCX q_0, q_1, q_2 =
-    I \\otimes I \\otimes |0 \\rangle \\langle 0| + CX \\otimes |1 \\rangle \\langle 1| =
+    I \\otimes I \\otimes \vert 0 \\rangle \\langle 0\vert  + CX \\otimes \vert 1 \\rangle \\langle 1\vert  =
    \\begin{pmatrix}
         1 &amp; 0 &amp; 0 &amp; 0 &amp; 0 &amp; 0 &amp; 0 &amp; 0\\\\
         0 &amp; 1 &amp; 0 &amp; 0 &amp; 0 &amp; 0 &amp; 0 &amp; 0\\\\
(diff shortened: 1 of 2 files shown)


frankharkins (approved):
Nice work! I've re-generated the docs with the new script and it all LGTM 🚀

Eric-Arellano:
Unfortunately, this resulted in some pages crashing, per https://github.com/Qiskit/documentation/issues/673#issuecomment-1928099938. This is because we broke column alignment settings for matrices, like this from `PauliTable.md`:

```
\begin{split}\left(\begin{array}{ccc|ccc}
```

If you use `\vert` for `{ccc|ccc}`, you get:

```
ParseError: KaTeX parse error: Unknown column alignment: \vert at position 37: …egin{array}{ccc\̲v̲e̲r̲t̲ ̲ccc}
    x_{0,0…
```

Maybe we were over-zealous with always replacing `|` with `\vert`. What about we fix https://github.com/Qiskit/documentation/issues/488 solely be fixing the problematic line? It's the only instance of this problem we've found! We actually could now even fix this by changing the HTML directly to use `\vert`. That works because these are historical API docs so we can modify the input HTML thanks to https://github.com/Qiskit/documentation/pull/737. The APIs were removed from modern Qiskit, so no worries about fixing qiskit/qiskit.

PR #723, opened 2024-01-30
Switch notebook outputs to SVG

### LINKED ISSUE #721: Switch notebook outputs to SVGs
We should consider switching all notebook outputs to SVG. This will give our cell outputs basically infinite resolution and reduce the transferred size of the images (equals faster load times for users).

frankharkins:
The documentation website now supports SVGs in notebook outputs.

Advantages are:
  - Very crisp images: SVG diagrams have basically infinite resolution.
  - Faster load times: Looking at `docs/build/circuit-library` (a page with lots of drawings), the transferred size goes from 220kb -> 70kb (70%) after switching to SVGs.

The one caveat is that Matplotlib-generated SVGs use lots of lines, which makes for large diffs. In the past, I've solved this problem by minifying the SVGs to one line, which makes them as easy to work with as PNGs. I'm using a package I created (squeaky) to do this. If bus-factor is a concern, I'm happy to add other developers as maintainers, or we could vendor the relevant code in this repo. I can also split the linter addition out into another PR if needed.

***

Closes #721


Eric-Arellano:
Yay! Probably worth adding `squeaky` in a dedicated PR to help us make sure we're doing it well, such as documenting in the README.

frankharkins:
> Probably worth adding squeaky in a dedicated PR 

Sounds good, I'll do that before adding the new outputs

PR #724, opened 2024-01-30
Add notebook linting

frankharkins:
Uses squeaky to lint notebooks.

While this has some minor benefits (removing empty cells etc.), it will become much more useful when we switch notebook outputs to SVG.


Eric-Arellano:
>  IMO we can ignore it.

I'd much rather that we allow-list it. I've been bitten in the past by merging on red CI. Once you get in the habit, it is hard to break and it's easy to not realize that there were new regressions you weren't aware of.

tox.ini
```diff
@@ -7,3 +7,7 @@ skipsdist = true
 deps = -r scripts/nb-tester/requirements.txt
 setenv = PYDEVD_DISABLE_FILE_VALIDATION=1
 commands = python scripts/nb-tester/test-notebook.py {posargs}
+
+[testenv:lint]
+deps = squeaky==0.5.0
+commands = squeaky --check {posargs}
```
Eric-Arellano, line 13: `--check` won't modify the files, right? That's good. 

If that's true, it'd be good to have a `fmt` environment so that `squeaky` makes changes. Or a more appropriate name could be `fix` since it is making actual changes to source code like removing metadata; it's not only changing how things are formatted.

README.md
```diff
@@ -145,6 +145,19 @@ click "Show all checks" in the info box at the bottom of the pull request page
 on GitHub, then choose "Details" for the "Test notebooks" job. From the job
 page, click "Summary", then download "Executed notebooks".
 
+## Lint notebooks
```
Eric-Arellano, line 148: If you take my advise below, you'll want to also mention how to format/fix notebooks. It'd be helpful to mention how CI will have the formatted/fixed notebook for you, too.

Eric-Arellano (commented):
Nice job with Squeaky! I love your use of emojis. That's a cool tool :)

tox.ini
```diff
@@ -7,3 +7,11 @@ skipsdist = true
 deps = -r scripts/nb-tester/requirements.txt
 setenv = PYDEVD_DISABLE_FILE_VALIDATION=1
 commands = python scripts/nb-tester/test-notebook.py {posargs}
+
+[testenv:lint]
+deps = -r scripts/nb-tester/requirements.txt
```
Eric-Arellano, line 12: I don't think you need to keep specifying this since it's in the default `[testenv]`

Eric-Arellano (approved):
Nice! Please ensure that CI passes for all notebooks so we don't merge a failure into main.

frankharkins:
> Please ensure that CI passes for all notebooks so we don't merge a failure into main.

Yes, I'll wait for #734 first which should fix CI

tox.ini
frankharkins: Fixed, thanks!

frankharkins:
@Eric-Arellano I think the diff is the same as when you last approved but just want to double-check it's ok before merging

Eric-Arellano (commented):
Yes, good to merge. But please make sure you have pulled main first and re-linted things - GitHub doesn't prompt you to do that anymore due to the merge queue.

Issue #719, opened 2024-01-29
API docs generation should download the original Sphinx HTML from Box


Eric-Arellano:
# Original issue, which was trying to use Git. See below comment for new plan with Box

## Motivation

After a few months of API generation, we've found a few issues with the current workflow:

1. We regenerate docs more than expected, and it's annoying when regenerating docs to trace down what the prior CI artifact URL was. If you get it wrong, then you'll have unintentional changes from the prior generation.
2. The historical API docs don't have any CI artifacts, so the only way to get the original Sphinx is to ask @arnaucasau or me. This is bad for bus factor / sustainability.
3. CI artifacts expire after 90 days, so we won't be able to easily get the artifacts in the future.
4. We need to make changes to the historical Qiskit API docs to fix issues like https://github.com/Qiskit/documentation/issues/673 or broken links. 
    - This is because we can't easily regenerate the original Sphinx docs due to "bit rot".
    - For new API docs, we should fix the source repository directly.
    - We could do this either by a) changing the original HTML or b) adding special-case rules in our pipeline to do string replacing. Several of us agreed that a) is more sustainable because the string replacing risks being overly general and making unintentional changes. But, if we change the original HTML, we need to make sure every other developer gets those edits.

Another framing of this: *API docs generation should be deterministic*. If you run the docs generation today vs in 6 months, you should get the same result, regardless of what machine you're running on. Whereas right now people can have different copies of the original HTML docs locally, or not know how to get them in the first place.

### Why upload to Git?

We considered alternatives like uploading to file storage like Box. It's simpler to upload it to Git and it's useful so that we have proper versioning. 

### Store artifacts by minor version, not patch

I recommend storing at paths like `scripts/api-artifacts/qiskit/0.19.zip`. (Feel free to suggest other paths!) 

I think it's more useful to store as `0.19` rather than the patch version `0.19.1` because it reflects reality: we only ever have one patch version per minor version. When upgrading to a new patch version, there's no need to store the original HTML docs. 

### Store compressed artifact

I heavily lean towards us storing the HTML docs as compressed like a zip file or tar file. 

Compression makes the file size smaller and fewer number of files. That's useful so the repo doesn't get even slower to use with Git. We could also look into [Git LFS](https://git-lfs.com) (large file storage) to make it faster.

Compression also results in a less noisy Git diff when adding a new version of the docs. 

[...]

PR #728, opened 2024-01-30
[WIP] Use Git LFS for zip files

arnaucasau:
Part of #719

The CI artifacts used for our API generation script will be stored in a directory called `/scripts/api-artifacts/<pkg-name>/<versions-without-patch>/`. We need to use Git LFS to avoid storing big zip files and slowing down the performance of the repo. Different compressing methods were tested and we decided to keep using zip files given the little improvement in space and to simplify our script due to the fact that the artifacts are downloaded as a zip file from GitHub.

This decision can be revisited! We thought it was the most sensible for now, but it might make sense to change it later.

This PR is a precursor for another PR that will store all API versions' artifacts.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.zip filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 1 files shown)


Eric-Arellano:
Hm, thanks for pointing that out, @mtreinish! I wasn't aware.

PR #730, opened 2024-01-30
Update details on finding the supported Python versions

mtreinish:
This commit updates the isntall guide to point to a more permenant location on where to find the supported python versions for a release. Previously the guide pointed people to setup.py on github. This was problematic for two reasons, first the supported version information isn't present there anymore as we moved to using the pyproject.toml file to contain all the static metadata about the project, second the data on main refers to the next release not the current one. Instead this commit points people to project page on PyPI which is where the metadata from the previous location gets hosted for each release. This is the cannonical source for which python versions are supported of any given Qiskit release (or the corresponding metadata in the package file itself).

### DIFF
--- a/docs/start/install.mdx
+++ b/docs/start/install.mdx
@@ -11,7 +11,7 @@ Whether you will work locally or in a cloud environment, the first step for all 
 <span id="local"></span>
 ## Install and set up Qiskit with the Qiskit Runtime client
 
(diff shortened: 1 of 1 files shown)


abbycross (approved):
👍

abbycross:
@mtreinish side note: I noticed that even though the current readme on qiskit/qiskit has the correct URL, the old qiskit.org link is still listed on the qiskit PyPI page. Assuming that PyPI is pulling from the readme, is there a way to get it to reload so that it shows the up-to-date link?

mtreinish:
The metadata on the pypi page comes directly from the metadata in package published. The only way to update the links is with a new release. It's been fixed in the package metadata in source: https://github.com/Qiskit/qiskit/commit/1d45d472fcdf229160b92af7d72e40c01d0eec9d so once we publish a new release the page will update to reflect the new links

PR #731, opened 2024-01-30
Tweaks to intro to primitives page

### LINKED ISSUE #646: Small tweaks to Introduction to primitives page
https://docs.quantum.ibm.com/run/primitives

On this page, second paragraph, I'd like to remove the bolded words "useful" and "abstraction". If additional formatting is deemed important here, I propose formatting the entire definition of primitive like so:

> In this context, a **primitive** is the smallest processing instruction, _the simplest building block from which one can create something useful for a given abstraction level._

Also, the word "provider" is used [in this section](https://docs.quantum.ibm.com/run/primitives#how-to-use-qiskit-primitives). May need to replace with "system" since "provider" is no longer used in this context, but I'd like to verify this is correct.

abbycross:
Closes #646 

Remove distracting formatting and review use of the word "provider" on this page.

### DIFF
--- a/docs/run/primitives.mdx
+++ b/docs/run/primitives.mdx
@@ -6,7 +6,7 @@ description: Introduction to primitives in Qiskit and Qiskit Runtime, and an exp
 
 # Introduction to primitives
 
-Computing systems are built upon multiple layers of abstraction. Abstractions allow to focus on a
+Computing systems are built upon multiple layers of abstraction. Abstractions allow us to focus on a
 particular level of detail relevant to the task at hand. The closer you get to the hardware,
 the lower the level of abstraction you'll need (for example, you could
 want to manipulate electrical signals), and vice versa, the more complex the task you want to perform,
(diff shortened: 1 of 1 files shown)


docs/run/primitives.mdx
```diff
@@ -6,15 +6,15 @@ description: Introduction to primitives in Qiskit and Qiskit Runtime, and an exp
 
 # Introduction to primitives
 
-Computing systems are built upon multiple layers of **abstraction**. Abstractions allow to focus on a
+Computing systems are built upon multiple layers of abstraction. Abstractions allow to focus on a
```
frankharkins, line 9:
```suggestion
Computing systems are built upon multiple layers of abstraction. Abstractions allow us to focus on a
```

Eric-Arellano (approved):
Other than Frank's suggestion, lgtm

abbycross:
I will merge this now and change the term provider later if it seems needed -- but I've put out a proposal to change the term Provider to Channel on the fair-share page, which should help remove some of the confusion.

PR #733, opened 2024-01-31
Update tox dependencies to include graphivz

kaelynj:
PR #674 had an issue with automated checks and needs the `graphviz` package to run `plot_circuit_layout()` function.


.github/workflows/notebook-test.yml
```diff
@@ -42,12 +42,13 @@ jobs:
           # Add your notebook to this list if it needs latex to run
```
Eric-Arellano, line 42: Oh, can you also please update the values here to say "Linux" rather than LaTeX? In the comment, you could say "If it needs latex or graphviz"

when changing the `id`, be careful to change line 48 too.

kaelynj: Done!

Eric-Arellano (approved):
thanks!

PR #734, opened 2024-01-31
Improve file filtering in CI

frankharkins:
The `transpiler-stages` notebook is broken, but will be tested in CI any time someone edits it (even if just changing copy etc.). This breaks PRs that are perfectly fine.

This PR ignores excluded files completely. I also took the opportunity to:
- improve the code readability somewhat (i.e. less `path for path in paths if path`),
- print a message when notebooks are skipped, otherwise someone running `tox` locally might believe their notebook was fine even though it never ran.

### DIFF
--- a/scripts/nb-tester/test-notebook.py
+++ b/scripts/nb-tester/test-notebook.py
@@ -243,12 +243,12 @@ if __name__ == "__main__":
     args = create_argument_parser().parse_args()
 
     paths = map(Path, args.filenames or find_notebooks(submit_jobs=args.submit_jobs))
-    paths = filter_paths(paths, submit_jobs=args.submit_jobs)
+    filtered_paths = filter_paths(paths, submit_jobs=args.submit_jobs)
 
     # Execute notebooks
     start_time = datetime.now()
     print("Executing notebooks:")
-    results = [execute_notebook(path, args) for path in paths]
+    results = [execute_notebook(path, args) for path in filtered_paths]
     print("Checking for trailing jobs...")
     results.append(cancel_trailing_jobs(start_time))
     if not all(results):
--- a/scripts/nb-tester/test-notebook.py
+++ b/scripts/nb-tester/test-notebook.py
@@ -35,6 +35,36 @@ NOTEBOOKS_THAT_SUBMIT_JOBS = [
 ]
 
 
+def filter_paths(paths: list[Path], submit_jobs: bool) -> list[Path]:
+    """
+    Filter out any paths we don't want to run, printing messages.
+    """
+    good_paths = []
+    for path in paths:
+        if path.suffix != ".ipynb":
+            print(f"ℹ️ Skipping {path}; file is not `.ipynb` format.")
+            continue
+
+        if any(path.match(glob) for glob in NOTEBOOKS_EXCLUDE):
+            this_file = Path(__file__).relative_to(Path(".").resolve())
+            print(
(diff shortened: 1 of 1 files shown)


scripts/nb-tester/test-notebook.py
```diff
@@ -35,6 +35,36 @@
 ]
 
 
+def filter_paths(paths: list[Path], submit_jobs: bool) -> list[Path]:
```
Eric-Arellano, line 38: Nicer is to return `Iterator[Path]`. Every time you want to return a path, you will `yield path`. No need to have `good_paths` as a collection - remove that. No return statement.

https://github.com/Qiskit/documentation/blob/ff3500a3926b4b9663dee35adf007b0e036cd34a/start#L21-L29

frankharkins: Much nicer, thanks!

scripts/nb-tester/test-notebook.py
```diff
@@ -35,6 +36,34 @@
...
+    """
+    Filter out any paths we don't want to run, printing messages.
+    """
+    for path in paths:
+        if path.suffix != ".ipynb":
+            print(f"ℹ️ Skipping {path}; file is not `.ipynb` format.")
+            continue
+
+        if any(path.match(glob) for glob in NOTEBOOKS_EXCLUDE):
+            this_file = Path(__file__).relative_to(Path(".").resolve())
```
Eric-Arellano, line 49:
```suggestion
            this_file = Path(__file__).resolve()
```

scripts/nb-tester/test-notebook.py
```diff
@@ -35,6 +36,34 @@
...
+            print(
+                f"ℹ️ Skipping {path}; to run it, edit `NOTEBOOKS_EXCLUDE` in {this_file}."
+            )
+            continue
+
+        if (
+            any(path.match(glob) for glob in NOTEBOOKS_THAT_SUBMIT_JOBS)
+            and not submit_jobs
```
Eric-Arellano, line 57: Nit: don't waste time doing an O(n) operation if the boolean is false anyways. n is so small this is nitty, though.

```suggestion
            not submit_jobs
            and any(path.match(glob) for glob in NOTEBOOKS_THAT_SUBMIT_JOBS)
```

scripts/nb-tester/test-notebook.py
```diff
@@ -214,15 +243,12 @@ def create_argument_parser() -> argparse.ArgumentParser:
     args = create_argument_parser().parse_args()
 
     paths = map(Path, args.filenames or find_notebooks(submit_jobs=args.submit_jobs))
-    if not args.submit_jobs:
-        paths = [path for path in paths if not any(path.match(glob) for glob in NOTEBOOKS_THAT_SUBMIT_JOBS)]
+    paths = filter_paths(paths, submit_jobs=args.submit_jobs)
```
Eric-Arellano, line 246: Nit: you're changing the variable type of `paths` here, which is subtle and discouraged in Python. (Fine in other languages like Rust.) Consider calling it `specified_paths` and `filtered_paths`.

Eric-Arellano (commented):
Nice!

scripts/nb-tester/test-notebook.py
frankharkins: Sounds good, thanks!

Eric-Arellano (approved):

PR #737, opened 2024-02-01
API generation script uses Box to download CI artifacts

arnaucasau:
Part of #719

This PR modifies the API generation script to download the CI artifacts stored in the Box folder https://ibm.ent.box.com/folder/246867649571 and removes the `-a <artifact-url>` argument.

The PR also renames the old `scripts/lib/api/downloadCIArtifacts.ts` to `scripts/lib/api/apiArtifacts.ts` to make it more generic, and it removes the associated test. The test file only was related to the `-a` argument that was removed.


scripts/lib/api/apiArtifacts.ts
frankharkins, line 1: Nice! Only suggestion is to re-add the ability to skip downloading if the artifact is already in `.out`; I found that quite useful when I was on flaky internet, or when I wanted to modify the artifact to see how the script reacted.

scripts/commands/updateApiDocs.ts
```diff
@@ -100,11 +92,7 @@ zxMain(async () => {
   );
 
   const artifactFolder = pkg.ciArtifactFolder();
-  if (await pathExists(artifactFolder)) {
-    console.log(`Skip downloading sources for ${pkg.name}:${pkg.version}`);
-  } else {
-    await downloadCIArtifact(pkg.name, args.artifact, artifactFolder);
-  }
+  await loadCIArtifact(pkg, artifactFolder);
```
frankharkins, line 95: I think `download` makes more sense as a name here (`load` makes me think you're loading it into memory).

scripts/lib/api/apiArtifacts.ts
arnaucasau: Thanks a lot for the review Frank!

I agree that is a nice feature. With this change, one of the goals is also to make the script more deterministic and ensure you will regenerate the latest versions available in Box. This is useful for the automation of the script too.

However, the idea is to add the feature back in a follow-up and allow you to skip the download when you have the more recent artifact. Box keeps the same shared link when you modify a file, so we were thinking of checking the hash of the zip file (more details in the [issue](https://github.com/Qiskit/documentation/issues/719#issuecomment-1919300586)) to avoid having an outdated artifact in the `.out` folder. Other ideas are welcome :)

When regenerating multiple versions we will be able to automate the script when #739 is merged, so maybe the extra time downloading won't be that important there. 

In the case you described though, in the meantime, we can tweak the `scripts/lib/api/apiArtifacts.ts` script locally, commenting lines 39 and 59 to skip the download and the removal of the artifact from `.out`. Not the ideal case, but I'll start working on that afterward. What do you think?

README.md
```diff
@@ -238,28 +238,39 @@ If you're working on our support code in `scripts/`, run `npm run fmt` to automa
...
+7. Select `People with the link` and go to `Link Settings`.
+8. Under `Link Expiration` select `Disable Shared Link on` and set an expiration date of ~10 years into the future.
+9. Copy the direct link at the end of the `Shared Link Settings` tab.
+10. Modify the `scripts/lib/api/api-html-artifacts.json` file adding the new versions with the direct link from step 9.
+11. Run `npm run gen-api -- -p <pkg-name> -v <version>`,
+    e.g. `npm run gen-api -- -p qiskit -v 0.45.0`
 
 In case you want to save the current version and convert it into a historical one, you can run `npm run make-historical -- -p <pkg-name>` beforehand.
 
+### Regenerate an existing API docs version
```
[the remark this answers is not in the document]
arnaucasau: You are right, thanks! Order reversed!

scripts/commands/updateApiDocs.ts
arnaucasau: I changed it back to `download`. I agree it makes more sense. Given that I was using the same name for a helper function, the result is changing `loadCIArtifact` and `downloadCIArtifact` to `downloadCIArtifact` and `downloadFromBox` respectively.

scripts/lib/api/apiArtifacts.ts
frankharkins: I see, thanks for the explanation :)

Eric-Arellano: We could do a compromise. Add an argument `--skip-download` with a default to `false`. I trust the developer to know when to use that. As long as the default is `false`.

arnaucasau: In the last commit, I added the `--skip-download` argument. We can skip the download using the argument if we have a folder called artifact in `.out` with the docs. Otherwise, the script will download from Box.

README.md
```diff
@@ -238,25 +238,36 @@ If you're working on our support code in `scripts/`, run `npm run fmt` to automa
 
 To check that formatting is valid without actually making changes, run `npm run check:fmt` or `npm run check`.
 
-## Regenerate the API docs
+### Regenerate an existing API docs version
```
Eric-Arellano, line 241: This should still be h2 since it's not a subsection.

```suggestion
## Regenerate existing API docs
```

Eric-Arellano: It'd be helpful to add below this sentence:

> This is useful when we make improvements to the API generation script.

README.md
```diff
@@ -238,25 +238,36 @@ If you're working on our support code in `scripts/`, run `npm run fmt` to automa
...
-3. Determine the current version of the published stable documentation, e.g. at https://github.com/Qiskit/qiskit/releases
-4. Find a link to a CI artifact with the project's documentation, e.g at https://github.com/Qiskit/qiskit/suites/17881600359/artifacts/1026798160. To find this:
+1. Choose which documentation you want to generate (`qiskit`, `qiskit-ibm-provider`, or `qiskit-ibm-runtime`) and its version.
+2. Run `npm run gen-api -- -p <pkg-name> -v <version>`,
+   e.g. `npm run gen-api -- -p qiskit -v 0.45.0`
+
+If the version is not for the latest stable minor release series, then add `--historical` to the arguments. For example, use `--historical` if the latest stable release is 0.45.\* but you're generating docs for the patch release 0.44.3.
+
+## Generate new API docs
+
```
Eric-Arellano, line 250:
```suggestion

This is useful when new docs content is published, usually corresponding to new releases or hotfixes for content issues.

```

README.md
```diff
@@ -238,25 +238,36 @@ If you're working on our support code in `scripts/`, run `npm run fmt` to automa
...
+1. Choose which documentation you want to generate (`qiskit`, `qiskit-ibm-provider`, or `qiskit-ibm-runtime`) and its version.
+2. Determine the current version of the published stable documentation, e.g. at https://github.com/Qiskit/qiskit/releases
```
Eric-Arellano, line 252:
```suggestion
2. Determine the full version, such as by looking at https://github.com/Qiskit/qiskit/releases
```

README.md
```diff
@@ -238,25 +238,36 @@ If you're working on our support code in `scripts/`, run `npm run fmt` to automa
...
-If the version is not for the latest stable minor release series, then add `--historical` to the arguments. For example, use `--historical` if the latest stable release is 0.45.\* but you're generating docs for the patch release 0.44.3.
+   5. Download the artifact by clicking on its name.
+4. Rename the downloaded zip file with its version number, e.g. `0.45.zip` for an artifact from `qiskit v0.45.2`.
+5. Upload the renamed zip file to https://ibm.ent.box.com/folder/246867649571
+6. Share the file by clicking the `Copy shared link` button
+7. Select `People with the link` and go to `Link Settings`.
+8. Under `Link Expiration` select `Disable Shared Link on` and set an expiration date of ~10 years into the future.
+9. Copy the direct link at the end of the `Shared Link Settings` tab.
+10. Modify the `scripts/lib/api/api-html-artifacts.json` file adding the new versions with the direct link from step 9.
+11. Run `npm run gen-api -- -p <pkg-name> -v <version>`,
```
Eric-Arellano, line 269: Needs to also mention `--historical` because of patch releases. But maybe instead have an `h3` like this:

```
### Generate patch releases

For example, if the current docs are for 0.45.2 but you want to generate 0.45.3.

Follow the same process above for new API docs, other than:

* When uploading the artifact, overwrite the existing file with the new one. Be careful to follow the above steps to change "Link Expiration"!
* Add `--historical` to the end of the `npm run gen-api` command.

File: scripts/commands/updateApiDocs.ts [hunk 7]
```diff
@@ -47,7 +47,7 @@ interface Arguments {
   package: string;
   version: string;
   historical: boolean;
-  artifact: string;
+  "skip-download": boolean;
```
Eric-Arellano, line 50: this can be skipDownload, e.g.

https://github.com/Qiskit/documentation/blob/7439931636ba19620200508ea9283a0ec7888d62/scripts/commands/checkLinks.ts#L32

scripts/commands/updateApiDocs.ts
```diff
@@ -71,12 +71,10 @@ const readArgs = (): Arguments => {
...
-    .option("artifact", {
-      alias: "a",
-      type: "string",
-      demandOption: true,
-      description:
-        "The URL for the CI artifact to download. Must be from GitHub Actions.",
+    .option("skip-download", {
+      type: "boolean",
+      default: false,
+      description: "Skip the download of the artifact from Box?",
```
Eric-Arellano, line 77:
```suggestion
      description: "Rather than downloading the artifact from Box, reuse what is already downloaded. This can save time, but it risks using an outdated version of the docs.",
```

scripts/commands/updateApiDocs.ts
```diff
@@ -100,10 +98,13 @@ zxMain(async () => {
   );
 
   const artifactFolder = pkg.ciArtifactFolder();
```
Eric-Arellano, line 100: The artifactFolder's verbosity annoys me - it's such a long path. But probably best for us to change in a follow up, since it's not strictly necessary here.

scripts/lib/api/apiArtifacts.ts
```diff
@@ -0,0 +1,66 @@
...
+    fs.readFileSync(
+      `${getRoot()}/scripts/lib/api/api-html-artifacts.json`,
+      "utf-8",
+    ),
+  );
+
+  if (!(`${pkg.versionWithoutPatch}` in artifactJson[`${pkg.name}`])) {
+    throw new Error(
+      `Package ${pkg.name} version ${pkg.versionWithoutPatch} doesn't have an artifact stored. You can add one to https://ibm.ent.box.com/folder/246867452622
+      following the steps detailed in the \`Generate the API docs\` section on the repo's README`,
```
Eric-Arellano, line 54: Nice message! It'd be good to add:

> If you are not an IBMer with access to the Box folder, you can ask in your pull request for a maintainer to help you. In the meantime, you can use another URL in `api-html-artifacts.json`, such as GitHub or even `localhost` for a server you start up; the URL needs to result in downloading the zip file.

Eric-Arellano (commented):
Awesome! 👏

scripts/lib/api/api-html-artifacts.json
```diff
@@ -0,0 +1,35 @@
+{
```
Eric-Arellano, line 1: I think this file is probably better at `scripts/api-html-artifacts.json`.  It is not normal library code. We expect you to change this file! We want it to be more easily discoverable and make it clear it's not normal helper code.

arnaucasau: Done!

README.md
arnaucasau: I added this new section, but without the last point that talks about the `--historical` argument. Generating patches of the current docs shouldn't be always historical. Instead, I added back the paragraph we had before at the end of the `Generate new API docs` section

README.md
```diff
@@ -273,6 +275,16 @@ This is useful when new docs content is published, usually corresponding to new
 
 In case you want to save the current version and convert it into a historical one, you can run `npm run make-historical -- -p <pkg-name>` beforehand.
 
+If the version is not for the latest stable minor release series, then add `--historical` to the arguments. For example, use `--historical` if the latest stable release is 0.45.\* but you're generating docs for the patch release 0.44.3.
```
Eric-Arellano, line 278: This belongs in the `Generate patch releases` docs because it's only relevant for that.

If you're worried that people won't realize they need to read that section, then I would add to the top of this current section something like this:

> If you're generating a patch release, also see the below subsection for additional steps.

Eric-Arellano (approved):

PR #739, opened 2024-02-01
Automate the regeneration of the API docs

### LINKED ISSUE #720: Automate regenerating all API docs
Thanks to https://github.com/Qiskit/documentation/issues/719, it should be possible to fully automate regenerating every API doc version! This is useful when we make improvements to how the API generation script works so that we can apply improvements to the past.

### Create dedicated commit per version

Sometimes API doc changes can be _huge_. It's often useful to split out the regeneration into several smaller PRs. 

So, treat each package's version as an atomic commit, e.g. the message:

> Regenerate Qiskit 0.45

Then, the developer can `git cherry-pick` whichever commits they want.

### Validate good Git state

They should be a branch other than `main`, so that we don't add commits to their `main` improperly. I recommend we only error with a good message; don't do something magical like checking out a new branch for them. 

Ensure there are no pending changes. Example check from closed source:

```javascript
  const initialStatus = await gitStatus();
  if (initialStatus !== '') {
    console.error(`
Repository must have clean Git state when calling
validate-open-source.ts. 'git status' returned:\n\n${initialStatus}`);
    process.exit(1);
  }
```

### Maybe CLI args? 

It's probably useful to allow skipping historical API docs since they're so slow? But also this could be added later after we try things out.

### Needs to compute the `-v` argument

With https://github.com/Qiskit/documentation/issues/719, the user still needs to specify the full patch version with the `-v` argument. So, the script will need to determine that for you by reading in the `_package.json` of existing API docs.

arnaucasau:
### Summary
Thanks to #737, the API generation script doesn't need an argument to specify the CI artifact URL used in the generation. Because of that, we can automatize the script to regenerate all API docs in the repository.

### New command
The PR adds a new command `regen-apis` where we can regenerate multiple versions of our three APIs with one command:
```bash
npm run regen-apis
```

By default, all minor releases of each API will be regenerated, but we can add the `--current-apis-only` argument to regenerate only the latest stable versions.

```bash
npm run regen-apis -- --current-apis-only
```

Alternatively, you can regenerate only one of the three available APIs. We can use the `-p` argument, which could be combined with the other arguments, similarly to the API generation script:

```bash
npm run regen-apis -- -p <pkg-name>
```

### Git instructions
To run the script, we need to create a dedicated branch and have a clean git status. The script will create a commit per version regenerated and the developers can git cherry-pick them how they want to create different PRs.

If the regeneration fails for a specific version, the script will call `git restore` to avoid mixing changes from two different versions when regenerating the next one.

### Output
At the end of each call to the new command, we will see a summarization of all versions regenerated with three different possible outcomes per version:

<table style="width:100%">
  <tr>
    <th>Possible Output</th>
    <th>Meaning</th>
    <th>Creates a new commit</th>
    <th>Restores the files</th>
  </tr>
  <tr>
    <td>✅ <pkg-name> <version> regenerated correctly</td>
    <td>The regeneration was successful</td>
    <td>Yes</td>
    <td>No</td>
  </tr>
  <tr>
    <td>☑️ <pkg-name> <version> is up-to-date</td>
    <td>The regeneration was successful, but no files were modified</td>
    <td>No</td>
    <td>No</td>
  </tr>
  <tr>
    <td>❌ <pkg-name> <version> failed to regenerate</td>
   <td>The regeneration failed</td>
   <td>No</td>
   <td>Yes</td>
 </tr>
</table>
                 
      

Closes #720


scripts/commands/regenerateApiDocs.ts
```diff
@@ -0,0 +1,158 @@
...
+        \`npm run regenerate-apis\`. 'git status' returned:\n\n${initialStatus}`);
+    process.exit(1);
+  }
+
+  const currentBranch = await gitBranch();
+  if (currentBranch == "main\n") {
+    console.error(`
+        Please create a dedicated branch to regenerate the API docs correctly.
+        Use 'git checkout -b <name-branch>' to continue`);
+    process.exit(1);
```
Eric-Arellano, line 63: Would be a good helper function: `await validateGitState()`

scripts/commands/regenerateApiDocs.ts
```diff
@@ -0,0 +1,158 @@
...
+async function regenerateVersions(
+  pkgName: string,
+  versions: string[],
+): Promise<string[]> {
+  const results: string[] = [];
+
+  for (const [index, version] of versions.entries()) {
+    try {
+      if (index == versions.length - 1) {
+        // The last version is always the latest stable minor release
```
Eric-Arellano, line 94: More clear would be to take 3 args in the function: `currentVersion` and `historicalVersions`. That's a type-driven API - the type system enforces your "invariant" for you.

scripts/commands/regenerateApiDocs.ts
```diff
@@ -0,0 +1,158 @@
...
+        fs.readFileSync(`${pkgDocsPath}/${folder.name}/_package.json`, "utf-8"),
+      );
+      historicalVersions.push(historicalVersion.version);
+    }
+  }
+
+  const currentVersion = JSON.parse(
+    fs.readFileSync(`${pkgDocsPath}/_package.json`, "utf-8"),
+  );
+  return [...historicalVersions, currentVersion.version];
```
Eric-Arellano, line 138: If you take my above suggestion, then you could either return a tuple or an object if you want it more structured.

scripts/commands/regenerateApiDocs.ts
```diff
@@ -0,0 +1,158 @@
...
+    .option("package", {
+      alias: "p",
+      type: "string",
+      choices: Pkg.VALID_NAMES,
+      demandOption: false,
+      description: "Which package to update",
+    })
+    .option("historical", {
+      type: "boolean",
+      default: false,
```
Eric-Arellano, line 41: I think it makes more sense for this to be true by default. It's important that we remember to apply fixes to historical API docs.

scripts/commands/regenerateApiDocs.ts
```diff
@@ -0,0 +1,158 @@
...
+    const result = await regenerateVersions(pkgName, pkgVersions);
+    results.set(pkgName, result);
+  }
+
+  console.log("");
+  results.forEach((result: string[], pkgName: string) => {
+    console.log(`Regeneration of ${pkgName}:`);
+    result.forEach((msg) => console.error(msg));
+    console.log("");
+  });
```
Eric-Arellano, line 82: maybe log at the end some tips:

> Each regenerated version has been saved as a distinct commit. If the changes are too large for one single PR, consider splitting it up into multiple PRs by using `git cherry-pick` or `git rebase -i` so each PR only has the commits it wants to target.

Eric-Arellano (commented):
Exciting 🚀

scripts/commands/regenerateApiDocs.ts
```diff
@@ -0,0 +1,158 @@
...
+  const historicalVersions: string[] = [];
+
+  if (historical) {
+    const historicalFolders = (
+      await readdir(`${pkgDocsPath}`, { withFileTypes: true })
+    ).filter((file) => file.isDirectory() && file.name.match(/[0-9].*/));
+
+    for (const folder of historicalFolders) {
+      const historicalVersion = JSON.parse(
+        fs.readFileSync(`${pkgDocsPath}/${folder.name}/_package.json`, "utf-8"),
```
Eric-Arellano, line 129: This should use `await` to be consistent

scripts/commands/regenerateApiDocs.ts
Eric-Arellano: I think `--historical` still makes more sense! You'd specify it as `--no-historical`.

Generating `--current-apis` doesn't imply you don't also generate historical. That would only work if you call it `--current-apis-only`.

package.json
```diff
@@ -14,6 +14,7 @@
     "fmt": "prettier --write .",
     "test": "jest",
     "typecheck": "tsc",
+    "regenerate-apis": "node -r esbuild-register scripts/commands/regenerateApiDocs.ts",
```
Eric-Arellano, line 17: Regenerate is a hard word to spell, imo. How about `regen-apis`? Matches `gen-api`.

scripts/commands/regenerateApiDocs.ts
Eric-Arellano: I mean `await fs.readFile`, not `readFileSync`

Eric-Arellano (approved):
Great work! It'd be helpful to update the README. I think in most cases, you'll want to regenerate all API docs. So you could start with explaining how to use this script in the first part of the Regen Docs section. Then say something like:

> If you only want to regenerate a specific version, then ...

scripts/commands/regenerateApiDocs.ts
```diff
@@ -0,0 +1,181 @@
...
+  const pkgDocsPath = `docs/api/${pkgName}`;
+  const historicalVersions: string[] = [];
+
+  if (!currentApisOnly) {
+    const historicalFolders = (
+      await readdir(`${pkgDocsPath}`, { withFileTypes: true })
+    ).filter((file) => file.isDirectory() && file.name.match(/[0-9].*/));
+
+    for (const folder of historicalFolders) {
+      const historicalVersion = await JSON.parse(
```
Eric-Arellano, line 132:
```suggestion
      const historicalVersion = JSON.parse(
```

scripts/commands/regenerateApiDocs.ts
```diff
@@ -0,0 +1,181 @@
...
+        await readFile(`${pkgDocsPath}/${folder.name}/_package.json`, "utf-8"),
+      );
+      historicalVersions.push(historicalVersion.version);
+    }
+  }
+
+  const currentVersion = await JSON.parse(
```
Eric-Arellano, line 139:
```suggestion
  const currentVersion = JSON.parse(
```

Eric-Arellano (approved):

PR #752, opened 2024-02-02
API generating script: historical images

arnaucasau:
Some historical versions don't have the `_images` folder in the artifact store in Box (https://ibm.ent.box.com/folder/246867452622), and we don't want the script to fail when trying to save them, especially in the automatic regeneration.

### DIFF
--- a/scripts/commands/updateApiDocs.ts
+++ b/scripts/commands/updateApiDocs.ts
@@ -245,7 +245,7 @@ async function convertHtmlToMarkdown(
     JSON.stringify(pkg_json, null, 2) + "\n",
   );
 
-  if (await pathExists(`${htmlPath}/_images`)){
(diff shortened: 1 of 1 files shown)


scripts/commands/updateApiDocs.ts
```diff
@@ -245,8 +245,11 @@ async function convertHtmlToMarkdown(
     JSON.stringify(pkg_json, null, 2) + "\n",
   );
 
-  console.log("Saving images");
-  await saveImages(allImages, `${htmlPath}/_images`, pkg);
+  if (await pathExists(`${htmlPath}/_images`)){
```
Eric-Arellano, line 248: We should only allow this for historical:

```suggestion
  if (!pkg.historical || await pathExists(`${htmlPath}/_images`)){
```

Eric-Arellano (approved):

PR #759, opened 2024-02-05
Regeneration Qiskit 0.39-0.30 API docs

arnaucasau:
This PR regenerates some Qiskit historical API docs versions. Given the amount of files modified, the regeneration of all versions is done in batches:

Versiones regenerated:

- qiskit v0.39.5
- qiskit v0.38.0
- qiskit v0.37.2
- qiskit v0.36.0
- qiskit v0.35.0
- qiskit v0.33.1
- qiskit v0.32.1
- qiskit v0.31.0
- qiskit v0.30.1


Eric-Arellano (approved):
Thanks! I think there are so few new lines because functions used to almost exclusively live on dedicated pages. We don't inline those into method pages, which we should continue to not do.

I did find one instance of a function now getting a header. This looks good.

Issue #755, opened 2024-02-02
Release notes for historical versions should link to stable API docs


Eric-Arellano:
Qiskit release notes before Qiskit 0.45 were stored in the file `legacy_release_notes.rst`, which was deleted upstream. This means that it's safe for us to make manual changes to release notes for Qiskit before v0.45, which we've been doing without issue.

But starting with Qiskit 0.45, the release notes are included in `release_notes.rst`. This means that generating the release notes for 1.0 will regenerate the release notes for 0.45. So we can't make manual changes anymore! 

Not being able to make manual changes is a problem because we sometimes have to fix things like broken links (see https://github.com/Qiskit/documentation/pull/747/commits/427efccc7dcc7e09d7805f3265330a83cb489d3e and https://github.com/Qiskit/documentation/pull/747/commits/22b05917e2b8a4d6aeaf91b7cdf3299d280a410a). But then those manual changes get reverted next time we regenerate the docs, like https://github.com/Qiskit/documentation/pull/753/commits/ba0fb97bec6cd16be08a0a3db4ea4aa73ce69a64.

I'm pretty sure the correct solution is for us to fix the original source code, like the Reno release notes folder in Qiskit. We need to fix the broken links to stable links.

PR #767, opened 2024-02-06
Revert vertical line Latex conversion

Eric-Arellano:
Reverts https://github.com/Qiskit/documentation/pull/718. Turns out that breaks certain pages, as explained at https://github.com/Qiskit/documentation/pull/718#issuecomment-1928630060.

Instead, we fix the problematic pages by changing the original Sphinx HTML in Box directly in this PR's last commit.

We have to manually revert the 0.45 docs due to https://github.com/Qiskit/documentation/issues/755.


arnaucasau (approved):
LGTM, thanks!

Issue #316, opened 2023-11-09
Add dev version of the API docs


Eric-Arellano:
This has been really useful for qiskit.org/documentation since we added it ~2 months ago. We should include a dev version of the docs. Maybe set up a cron job for that to update the docs every night, 3 days, or week.

Eric-Arellano:
To do this, we need a way to automatically get the artifact URL for the latest commit. That relates to https://github.com/Qiskit/documentation/issues/501.

I'm going to remove this from the Qiskit 1.0 milestone. It would still be great to have, but there are even bigger priorities like https://github.com/Qiskit/documentation/issues/479, https://github.com/Qiskit/documentation/issues/454, and https://github.com/Qiskit/documentation/issues/476.

Eric-Arellano:
We decided to start with adding a `/dev` version of the docs that is manually generated for now. https://github.com/Qiskit/documentation/issues/745 tracks switching it to CD.

We need to make these changes:

1. `updateApiDocs.ts` will add a `/dev` folder and takes the `--dev` arg.
    - check that our support code, like link checker, works on `--dev`
2. teach the docs app how to handle `/dev`
     - don't use it in search index
     - the version dropdown should add the label `(dev)` and add it to the top

Issue #702, opened 2024-01-24
Add warning when not on current version of API docs


Eric-Arellano:
Our SEO was bad for qiskit.org, so old versions of qiskit.org API pages are often the top results. Those then get redirected to ibm.com and users don't realize they're on the old version of the docs unless they open up the version selector

<img width="276" alt="Screenshot 2024-01-24 at 12 18 21 PM" src="https://github.com/Qiskit/documentation/assets/14852634/a7491c9e-bc99-4556-b51c-e13a45f0ffc6">

We've heard a lot of user complaints about the docs being out-of-date, when really they're only looking at the old version. So we should make it more obvious to users by adding a warning. This is only for API pages.

Example:

<img width="953" alt="Screenshot 2024-01-24 at 12 19 30 PM" src="https://github.com/Qiskit/documentation/assets/14852634/64de5482-9850-473b-85a4-791fa13112fb">

beckykd:
Text suggestion:  "This document is for an unsupported version of Qiskit. To view the current documentation, choose the version marked (latest) from the dropdown at the top of the page. "

Maybe @abbycross can think of a way to shorten it, but I do think it's important to tell them HOW to view the current version.

abbycross:
> Text suggestion: "This page refers to an unsupported version of Qiskit. To view the current documentation, choose the version marked (latest) from the version menu. "

Just a couple small changes, and changing "top of the page" in case the dropdown ends up not being at the top of the page for mobile users 🤷

gracelindsell:
Designs are completed for the banner on old API's and can be found [here](https://www.figma.com/file/sNXucK1uMaZ8XXyLTZrdrs/OneDocs?type=design&node-id=4302%3A8562&mode=design&t=gqDwyWc6QiaANYmt-1).

Eric-Arellano:
@gracelindsell @Vanguardbots @javabster, do we want this dev warning to show up for Qiskit 0.46 when 1.0 is released? What about Qiskit 1.0 when Qiskit 2.0 is released? That is, should we have the notion of "stable" docs versions that aren't "latest" but also don't have a dev warning saying there are newer docs available?

My opinion: this nuance is not necessary. Even when Qiskit 0.46 is released, we do still want to encourage people to switch to Qiskit 1.0. We want them to realize they are not on the latest.

gracelindsell:
@Eric-Arellano I agree with your opinion. The goal is to have people realize they are not on the latest version and thus to update that language upon release

[...]

PR #771, opened 2024-02-06
Add `type` attribute to `_package.json`

arnaucasau:
Part of #316

This PR adds the attribute `type` in the `_package.json` files to be able to differentiate between the latest, historical, and the dev version once is implemented.

This new attribute will be used to determine when to add a label in the version selector of the API docs

### DIFF
--- a/scripts/commands/updateApiDocs.ts
+++ b/scripts/commands/updateApiDocs.ts
@@ -240,7 +240,7 @@ async function convertHtmlToMarkdown(
 
   console.log("Generating version file");
   const type = pkg.historical ? "historical" : "latest";
-  const pkg_json = { name: pkg.name, version: pkg.version, type: type };
+  const pkg_json = { name: pkg.name, version: pkg.version, type };
   await writeFile(
     `${markdownPath}/_package.json`,
     JSON.stringify(pkg_json, null, 2) + "\n",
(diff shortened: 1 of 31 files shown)


scripts/commands/updateApiDocs.ts
```diff
@@ -239,7 +239,8 @@ async function convertHtmlToMarkdown(
   }
 
   console.log("Generating version file");
-  const pkg_json = { name: pkg.name, version: pkg.version };
+  const type = pkg.historical ? "historical" : "latest";
+  const pkg_json = { name: pkg.name, version: pkg.version, type: type };
```
Eric-Arellano, line 243: You can use a nifty JS feature:

```suggestion
  const pkg_json = { name: pkg.name, version: pkg.version, type };
```

Eric-Arellano (approved):
Thanks! 

We might want to introduce another category called `stable` for API docs that aren't latest but also shouldn't have the dev warning from https://github.com/Qiskit/documentation/issues/702 when viewing. Unclear - I asked in https://github.com/Qiskit/documentation/issues/702#issuecomment-1930439160. Either way, that doesn't block this PR because I think it would be a new variant of the enum.

arnaucasau:
Closed in favor of using the path to determine the type of each version instead of adding a new attribute to the `_package.json`

PR #772, opened 2024-02-06
Always check if pages render in extended CI check

Eric-Arellano:
Sometimes the link checker fails due to flakes. It's annoying that that means the page render check won't run.

### DIFF
--- a/.github/workflows/extended-checks.yml
+++ b/.github/workflows/extended-checks.yml
@@ -51,8 +51,10 @@ jobs:
         run: |
           ./start &
           sleep 20
+        if: always()
       - name: Check current API pages render
         run: >
           npm run check-pages-render --
           --qiskit-release-notes
           --current-apis
+        if: always()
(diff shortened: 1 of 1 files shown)


.github/workflows/extended-checks.yml
```diff
@@ -51,8 +51,10 @@ jobs:
         run: |
           ./start &
           sleep 20
+        if: always()
       - name: Check current API pages render
         run: >
           npm run check-pages-render --
           --qiskit-release-notes
           --current-apis
+        if: always()
```
frankharkins, line 60:
```suggestion
        if: !cancelled()
      - name: Check current API pages render
        run: >
          npm run check-pages-render --
          --qiskit-release-notes
          --current-apis
        if: !cancelled()
```

Nit as we probably won't encounter this issue, but `always` will run even if the job is cancelled.

frankharkins (approved):

PR #775, opened 2024-02-07
Refactor of `Pkg` to add a package type

arnaucasau:
This PR replaces the `historical` attribute of the class Pkg for an attribute named `type` which will allow us to know if the package version is the latest one or the historical one. This small refactor is convenient to avoid having a bunch of boolean attributes when the version types `dev` and `stable` are introduced.

Different version types will have different behaviors in our API generation script like a custom path (e.g. numerical subfolder or `dev` subfolder).

### DIFF
--- a/scripts/commands/updateApiDocs.ts
+++ b/scripts/commands/updateApiDocs.ts
@@ -91,11 +91,13 @@ zxMain(async () => {
     );
   }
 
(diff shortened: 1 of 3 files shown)


Eric-Arellano (approved):
Nice!

PR #779, opened 2024-02-07
Extend the API generation script to generate dev versions

arnaucasau:
This PR adds support to generate dev versions for the API generation script.

For the moment, the automatic regeneration will not take into account dev versions to keep this PR easy to review. I'll add support for that in a follow-up, together with the documentation of the dev versions.

Part of #316

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ npm run check:links -- --external
 
 # By default, only the non-API docs are checked. You can add any of the
 # below arguments to also check API docs and/or Qiskit release notes.
-npm run check:links -- --current-apis --historical-apis --qiskit-release-notes
+npm run check:links -- --current-apis --dev-apis --historical-apis --qiskit-release-notes
 
 # However, `--historical-apis` currently has failing versions, so you may
 # want to add `--skip-broken-historical`.
@@ -227,7 +227,7 @@ To check that all the non-API docs render:
 1. Start up the local preview with `./start` by following the instructions at [Preview the docs locally](#preview-the-docs-locally)
 2. In a new tab, `npm run check-pages-render`
 
-You can also check that API docs and translations render by using any of these arguments: `npm run check-pages-render -- --qiskit-release-notes --current-apis --historical-apis --translations`. Warning that this is exponentially slower.
+You can also check that API docs and translations render by using any of these arguments: `npm run check-pages-render -- --qiskit-release-notes --current-apis --dev-apis --historical-apis --translations`. Warning that this is exponentially slower.
 
 CI will check on every PR that non-API docs correctly render. We also run a nightly cron job to check the API docs and
 translations.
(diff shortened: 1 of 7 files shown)


scripts/commands/updateApiDocs.ts
```diff
@@ -91,7 +97,20 @@ zxMain(async () => {
...
+  if (args.historical && args.dev) {
+    throw new Error(
+      `${args.package} ${args.version} cannot be historical and dev at the same time. Please remove at least only one of these two arguments: --historical, --dev.`,
+    );
+  }
+
+  const devRegex = /[0-9](-rc|-dev)/;
+  if (args.dev && !args.version.match(devRegex)) {
+    throw new Error(
+      `${args.package} ${args.version} is not a correct dev version. Please make sure the version has one of the following suffixes with a hyphen immediately following the patch version: -rc, -dev. e.g. 1.0.0-rc1 or 1.0.0-dev`,
```
arnaucasau, line 109: Following the semver specification we need to have a hyphen between the patch version and the prerelease identifiers. I tested it and without it, we won't be able to sort the versions without adding extra logic on the version sector code.

For more information: https://semver.org/#spec-item-9

README.md
```diff
@@ -287,6 +287,8 @@ This is useful when new docs content is published, usually corresponding to new
 11. Run `npm run gen-api -- -p <pkg-name> -v <version>`,
     e.g. `npm run gen-api -- -p qiskit -v 0.45.0`
 
+If you are regenerating a dev version, then you can add `--dev` as an argument and the documentation will be built at `/docs/api/<pkg-name>/dev`.
```
Eric-Arellano, line 290: Should add here and above:

> For dev versions, end in the `--version` in `-dev`, e.g. `-v 1.0.0-dev`. If a release candidate has already been released, use `-v 1.0.0rc1`, for example.

scripts/commands/updateApiDocs.ts
Eric-Arellano: For rc1, there is no `-`, e.g. `1.0.0rc1`. See https://pypi.org/project/qiskit/#history. We need to use the version from the actual release because that's what users have to put into their Python install.

So we'll need to adapt the internal code to work around this.

Eric-Arellano (commented):
Looks great! The link checker and page render-er need dev support. I can push a commit.

scripts/commands/updateApiDocs.ts
arnaucasau: Thanks for pointing that out! I changed the script to accept release candidate versions without the dash, but this is something super important to make sure we correct when we read the version from the `_package.json` on iqp-channel-docs because otherwise, it will break components that are not able to recognize it

Eric-Arellano (approved):
Nice! Please double check my commit to change the link checker.

PR #786, opened 2024-02-08
API generation script uses correct artifact names

arnaucasau:
Part of #316 

This PR adds the artifacts for the dev versions of qiskit and qiskit-ibm-runtime and fixes how we read the JSON file to take into account the new names of the artifacts.

### DIFF
--- a/scripts/api-html-artifacts.json
+++ b/scripts/api-html-artifacts.json
@@ -1,6 +1,6 @@
 {
   "qiskit": {
-    "1.0-dev": "https://ibm.box.com/shared/static/eh9tzx3senx5xkq1vvtsr5h697r6j7gk.zip",
+    "dev": "https://ibm.box.com/shared/static/eh9tzx3senx5xkq1vvtsr5h697r6j7gk.zip",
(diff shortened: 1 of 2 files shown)


scripts/api-html-artifacts.json
```diff
@@ -1,5 +1,6 @@
 {
   "qiskit": {
+    "1.0-dev": "https://ibm.box.com/shared/static/eh9tzx3senx5xkq1vvtsr5h697r6j7gk.zip",
```
Eric-Arellano, line 3: What about simplifying this to be `dev`? While in Box I think the file should be called `1.0-dev`, it doesn't need to be so here. That avoids the confusion when we add `1.1-dev` - what do we do with `1.0-dev`? It wouldn't make sense to keep it here. It's only possible to have a single `dev` version in this repository.

The motivation for Box to name `1.0-dev` is so that we can have a historical record of our past dev docs.

Eric-Arellano (commented):
Smart idea to put `-dev` in the JSON file.

scripts/api-html-artifacts.json
arnaucasau: That's much better, thank you!

Eric-Arellano (approved):

PR #788, opened 2024-02-08
Extend the `regen-apis` command to regenerate dev versions

arnaucasau:
Part of #316 

This PR extends the API automatic regeneration script to regenerate dev versions. When using the `npm run regen-apis` command, the script will read the `_package.json` from the `dev` folder if it exists and regenerate that version as the last one of the package.

### DIFF
--- a/scripts/commands/regenerateApiDocs.ts
+++ b/scripts/commands/regenerateApiDocs.ts
@@ -15,10 +15,10 @@ import { readFile, readdir } from "fs/promises";
 import yargs from "yargs/yargs";
 import { hideBin } from "yargs/helpers";
 import { $ } from "zx";
-import fs from "fs";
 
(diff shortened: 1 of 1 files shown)


scripts/commands/regenerateApiDocs.ts
```diff
@@ -54,14 +54,17 @@ zxMain(async () => {
       continue;
     }
 
-    const [historicalVersions, currentVersion] = await getPackageVersions(
+    const [historicalVersions, currentVersion] = await getReleasedVersions(
       pkgName,
       args.currentApisOnly,
     );
+    const devVersion = await getDevVersion(pkgName);
```
Eric-Arellano, line 61: Nit: consider calling it `maybeDevVersion`.

Eric-Arellano (approved):

Issue #773, opened 2024-02-06
Check that pages render for historical API docs changes


Eric-Arellano:
The job now only takes about 25 minutes, whereas it used to take 3 hours. So it's now reasonable to run in PRs that change historical API docs. 

Although, we have to deal with https://github.com/Qiskit/documentation/issues/705 for this to work. Possibly be removing translations.

PR #792, opened 2024-02-09
API changes now check that API pages render

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/773. 

Now that we've fixed all historical pages not rendering, it's important we don't regress. The check is much faster than it used to be, so we decided it's worth the CI slowdown.

More broadly, it's become clear that there are two types of changes we make: a) API docs, and b) non-API docs. They are very different workflows, since API docs are 100% automated. Our CI should better reflect this reality, whereas before for example `main.yml` would check relative links for API docs. 

To speed up the API CI job, we split it up into two parallel jobs.

Once we start working more with translations, we'll want to consider a `translations.yml` workflow. Those are also fairly isolated in comparison to API docs and non-API docs.


.github/workflows/api-checks.yml
```diff
@@ -10,14 +10,11 @@
...
-# This Action runs checks things that we care about but would be
-# too slow to check in every PR.
-
-name: Extended checks
+name: API checks
 on:
   workflow_dispatch:
   schedule:
-    - cron: "0 0 * * *" # Every day at midnight UTC
+    - cron: "0 0 * * 0,2,4" # Every Sunday, Tuesday, and Thursday at midnight UTC
```
Eric-Arellano, line 17: Switched to 3 times a week to waste fewer resources. Most people aren't checking on Friday night and Saturday what happened, for example.

Eric-Arellano (commented):
See https://github.com/Qiskit/documentation/actions/runs/7850053208/job/21424707125 and https://github.com/Qiskit/documentation/actions/runs/7850053208/job/21424707277 for examples.

frankharkins (approved):
Looks good, thanks!

arnaucasau (approved):
LGTM, thank you Eric!

Issue #803, opened 2024-02-14
Switch all historical API doc links from Qiskit 0.44 to 0.46


Eric-Arellano:
For Qiskit release notes and our migration guides, we have to link to old Qiskit API docs because Qiskit 1.0 removed so many APIs. For now, we've been using Qiskit 0.44. But better is to use Qiskit 0.46 since it's the latest release so will include things like deprecation warnings.

We only couldn't do this yet because there is no `/api/qiskit/0.46` URL since 0.46 is the current version of docs.

To do this, try to automatically replace all `/api/qiskit/0.44/` with `/api/qiskit/0.46` and then run `npm run check:links -- --qiskit-release-notes` to see what breaks. Be careful to not change Qiskit release notes 0.45+, per https://github.com/Qiskit/documentation/issues/755.

PR #804, opened 2024-02-14
Future proof release notes and migration guide links before Qiskit 1.0

Eric-Arellano:
When Qiskit 1.0 is released, a bunch of APIs will be removed from our Qiskit API docs. This breaks internal links we have.

This PR switches all links that will break to point to Qiskit 0.44. Once 1.0 is released, we should switch what we can to 0.46, per https://github.com/Qiskit/documentation/issues/803.

It does not change Qiskit release notes for 0.45 and 0.46, since we can't safely make manual changes to those, per https://github.com/Qiskit/documentation/issues/755. Those still need to be dealt with.

## How change was generated

I found the issues by generating 1.0 docs as the current version, then running `npm run check:links -- --qiskit-release-notes`. I started with manual fixes, then realized this is easily automatable.

I accidentally `git clean`ed the script, but it looked like this:

```python
from pathlib import Path
from typing import DefaultDict
from collections import defaultdict
import re


def parse_error(error: str, file_to_bad_links: DefaultDict[str, list[str]]) -> None:
    lines = error.splitlines()
    bad_link_line = lines[0]
    file_list_lines = lines[1:]

    bad_link = ""  # Regex that captured the link
    files = [line.split("❓")[0].strip() for line in file_list_lines]

    for file in files:
        file_to_bad_links[file].append(bad_link)


def read_errors() -> DefaultDict[str, list[str]]:
    raw_text = Path("api_errors.txt").read_text()
    file_to_bad_links = defaultdict(list)
    for error in raw_text.split("❌"):
        parse_error(error, file_to_bad_links)
    return file_to_bad_links


def rewrite_file(fp: str, bad_links: list[str]) -> None:
   # code to use `string.replace()`
   pass 

def main() -> None:
    file_to_bad_links = read_errors()
    for file, bad_links in file_to_bad_links.items():
      rewrite_file(file, bad_links)


main()
```

Then I had to fix a few links manually that were only in 0.45 and not 0.44.


scripts/commands/checkLinks.ts
```diff
@@ -143,12 +143,13 @@ async function determineCurrentDocsFileBatch(
...
-    "docs/api/qiskit/0.44/qiskit.{algorithms,extensions,opflow}.*",
-    "docs/api/qiskit/0.44/qiskit.utils.QuantumInstance.md",
+    // The 0.44 docs are used by release notes for APIs that were removed in 1.0.
+    "docs/api/qiskit/0.44/*.md",
     "docs/api/qiskit/0.45/qiskit.quantum_info.{OneQubitEuler,TwoQubitBasis,XX}Decomposer.md",
     "docs/api/qiskit/0.45/qiskit.transpiler.synthesis.aqc.AQC.md",
     "docs/api/qiskit/0.45/{tools,quantum_info,synthesis_aqc}.md",
+    "docs/api/qiskit/0.45/qiskit.passmanager.FlowController.md",
+    "docs/api/qiskit/0.45/qiskit.circuit.CommutationChecker.md",
   ];
```
Eric-Arellano, line 153: These changes are telling our link checker that it needs to read in these files because other files link to them.

arnaucasau (approved):
I rerun the link checker with 1.0.0rc1, and all looks good to me! Thank you Eric!

Issue #814, opened 2024-02-14
Missing 1.0 API pages


Eric-Arellano:
* qiskit.pulse.library.SymbolicPulse
* qiskit.primitives.BaseEstimator
* qiskit.primitives.BaseSampler

The HTML files do exist in the uploaded artifacts to Box! So this is an issue with our API generation script. It doesn't look like it's macOS specific (filesystem case sensitivity) because the files download on my mac just fine. 

When fixing this, we should try regenerating every API to see if this has been a problem before.

PR #815, opened 2024-02-15
Don't delete API docs for class aliases

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/814. We shouldn't be deleting attribute pages if they don't correspond to an owning class.

This PR also runs `npm run regen-apis`; as shown, this PR's logic change only impacts Qiskit v1 docs.


scripts/lib/api/mergeClassMembers.ts
```diff
@@ -94,12 +89,7 @@ export async function mergeClassMembers<T extends HtmlToMdResultWithUrl>(
     }
   }
 
-  // remove merged results
-  const finalResults = reject(results, (result) =>
-    includes(["method", "attribute", "property"], result.meta.apiType),
-  );
```
frankharkins, line 100: This is where the pages were being deleted because we were assuming certain `apiType`s were merged. Now we just check if they're in `mergedMemberUrls`.

frankharkins (approved):
Generated output looks good!

I have a couple of non-blocking questions.

scripts/lib/api/mergeClassMembers.ts
```diff
@@ -32,21 +32,16 @@ export async function mergeClassMembers<T extends HtmlToMdResultWithUrl>(
...
-        return result.meta.apiName?.startsWith(`${clazz.meta.apiName}.`);
-      }),
-      (result) => result.meta.apiName,
+    const unsortedMembers = resultsWithName.filter(
+      (result) =>
+        includes(["method", "property", "attribute"], result.meta.apiType) &&
+        result.meta.apiName?.startsWith(`${clazz.meta.apiName}.`),
     );
+    unsortedMembers.forEach((result) => mergedMemberUrls.add(result.url));
+    const members = orderBy(unsortedMembers, (result) => result.meta.apiName);
```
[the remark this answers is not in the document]
Eric-Arellano: That's correct. I don't know why we were looking for `functions` before - they never are members of a class. In practice, it made no difference if we included them or not.

scripts/lib/api/mergeClassMembers.ts
Eric-Arellano: Exactly.

Issue #793, opened 2024-02-09
Metadata check should differentiate between API docs vs non-API


Eric-Arellano:
In the spirit of https://github.com/Qiskit/documentation/pull/792, we shouldn't waste our time with `npm run check` in validating API docs. By definition, humans are not editing these files!

Instead, add an argument like `--apis`. Imo, we don't need the full breadth of arguments we use for the link checker and page renderer, like `--dev-apis` vs `--current-apis`, since the script is already fast enough. Simply `--apis` is enough.

PR #817, opened 2024-02-15
Don't check metadata for API docs by default

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/793. 

This unblocks us starting to check API docs in https://github.com/Qiskit/documentation/issues/66. It also speeds up the workflow for non-API docs, which is valuable because non-API and API docs have such different workflows.

I only added an `--apis` argument rather than e.g. `--current-apis` vs `--historical-apis` because this check is fast enough to not need the nuance.

### DIFF
--- a/scripts/commands/checkMetadata.ts
+++ b/scripts/commands/checkMetadata.ts
@@ -20,6 +20,7 @@ import { globby } from "globby";
 interface Arguments {
   [x: string]: unknown;
   apis: boolean;
+  translations: boolean;
 }
 
 const readArgs = (): Arguments => {
--- a/scripts/commands/checkMetadata.ts
+++ b/scripts/commands/checkMetadata.ts
@@ -10,14 +10,34 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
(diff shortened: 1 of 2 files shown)


scripts/commands/checkMetadata.ts
```diff
@@ -10,14 +10,34 @@
...
+import grayMatter from "gray-matter";
 import { globby } from "globby";
 
-const IGNORED_FILES = new Set([
-  "docs/api/qiskit/transpiler_builtin_plugins.md",
-  "docs/api/qiskit/dev/transpiler_builtin_plugins.md",
-]);
+interface Arguments {
+  [x: string]: unknown;
+  apis: boolean;
```
arnaucasau, line 22: Nit: Add the translations argument
```suggestion
  apis: boolean;
  translations: boolean;
```

arnaucasau (commented):
Tested locally and works well. I only have one comment

scripts/commands/checkMetadata.ts
Eric-Arellano: Thanks!

arnaucasau (approved):

PR #819, opened 2024-02-15
Allow skipping download when regenerating all docs

Eric-Arellano:
I regenerated the docs last night and we haven't changed Box artifacts since then. I want to regenerate the docs again today, so it's convenient to let me skip the download to speed things up.


arnaucasau (approved):
Awesome, good idea! This argument will be useful for those kinds of regenerations.

arnaucasau (approved):
LGTM! :rocket:

PR #821, opened 2024-02-15
Fix metadata detection of modules for historical Qiskit docs

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/66. We were failing to detect modules using old versions of Sphinx that set the module metadata in `<div>` rather than `<span>` or `<section>`.

Some of the module pages had no metadata at all, whereas some were misclassified as `functions` corresponding to the first function in the module's page. Fixing this metadata will fix the HTML meta properties, and it also fixes the table of contents to have correct nesting.


Eric-Arellano:
I recommend viewing by commit. The first commit is the only manual one. I will spend the time this afternoon to closely review each commit - so feel free to do more of a spot check.

arnaucasau (approved):
Checked around 40% of the files and looks good to me! Thanks Eric!

PR #830, opened 2024-02-16
Change notebook output to SVG

frankharkins:
The documentation website now supports SVGs in notebook outputs.

Advantages are:
  - Very crisp images: SVG diagrams have basically infinite resolution.
  - Faster load times: Looking at `docs/build/circuit-library` (a page with lots of drawings), the transferred size goes from 220kb -> 70kb (70%) after switching to SVGs.


.github/workflows/notebook-test.yml
```diff
@@ -86,12 +86,10 @@ jobs:
...
           import subprocess
-          subprocess.run([
-              "echo",
-              "c.InlineBackend.figure_format = 'svg'",
-              ">>",
-              "~/.ipython/profile_default/ipython_kernel_config.py"
-          ])
+          subprocess.run(
+            "echo \"c.InlineBackend.figure_format = 'svg'\" >> ~/.ipython/profile_default/ipython_kernel_config.py",
+            shell=False
```
Eric-Arellano: This would need to be `shell=True`. But better is to use `Path(...).write_text()`

frankharkins: Always get it confused 😢 
Not sure it's actually the right file; need to work out which config file it's reading from

frankharkins: I discovered 84b6514176e800675ddee54b2e1ccf14714513be which is much cleaner and means writers can produce them locally without changing their global config settings.

Eric-Arellano (approved):
🥳

docs/verify/local-testing-mode.ipynb
```diff
@@ -76,14 +76,23 @@
...
      "ignore-warnings"
     ]
    },
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/Users/frankharkins/Documents/documentation/.tox/py311/lib/python3.11/site-packages/qiskit_ibm_runtime/session.py:156: UserWarning: Session is not supported in local testing mode or when using a simulator.\n",
```
frankharkins, line 91:
```suggestion
      "python3.11/site-packages/qiskit_ibm_runtime/session.py:156: UserWarning: Session is not supported in local testing mode or when using a simulator.\n",
```

This notebook was not run before so had no outputs. I think it's helpful to show this warning so users understand it's expected, but I've edited out my device-specific information.

Eric-Arellano (approved):
Thanks!

docs/verify/local-testing-mode.ipynb
frankharkins: Sure!

docs/verify/local-testing-mode.ipynb
```diff
@@ -76,14 +76,23 @@
...
      "ignore-warnings"
     ]
    },
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "python3.11/site-packages/qiskit_ibm_runtime/session.py:156: UserWarning: Session is not supported in local testing mode or when using a simulator.\n",
```
frankharkins, line 91:
```suggestion
      "/Users/qiskit/.venv/site-packages/qiskit_ibm_runtime/session.py:156: UserWarning: Session is not supported in local testing mode or when using a simulator.\n",
```

PR #831, opened 2024-02-16
Better organize link checker ignores

Eric-Arellano:
This adds the idea of "always ignored URLs", rather than making us repeat every file in which the URL appears. That has become necessary now that we have dev and historical API docs.

This PR also removes some stale ignores.

### DIFF
--- a/.github/workflows/api-checks.yml
+++ b/.github/workflows/api-checks.yml
@@ -21,6 +21,8 @@ on:
       - "docs/api/qiskit/**/*"
       - "docs/api/qiskit-ibm-provider/**/*"
       - "docs/api/qiskit-ibm-runtime/**/*"
+      - "scripts/checkLinks.ts"
+      - "scripts/lib/links/ignores.ts"
 
 jobs:
   link-checker:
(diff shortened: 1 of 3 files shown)


arnaucasau (approved):
Good idea with this change!! One question, isn't it dangerous to ignore relative links and anchors (e.g., #id1)? We could have a file with the same name in different APIs that will be ignored unintentionally. I don't think this will happen in the API docs, but maybe with some anchors and section titles. What do you think?

Eric-Arellano:
> isn't it dangerous to ignore relative links and anchors (e.g., #id1)?

I think it depends on the anchor. Certain highly generic ones like `#id1` are a bad idea. The majority of the anchor ignores here are in dedicated file lists. I only globally ignored a few like `#qiskit.utils.optionals.HAS_TESTTOOLS`, which seem well scoped.

arnaucasau (approved):
Agreed. Maybe we could add a little comment below the `// Always ignored URLs` section to have it in mind, just in case. Apart from that, all looks good to me!! Thank you for the change, tested locally and works well too :+1:

PR #844, opened 2024-02-19
Update platform support description

wshanks:
* Mention a few non-supported modes (FreeBSD, WebAssembly, PyPy) by name so they come up more easily for someone searching for them.
* Fix typo (missing "be").
* Downgrade confidence that Qiskit will work on other platforms
* Expand reference to CPython

### DIFF
--- a/docs/start/install.mdx
+++ b/docs/start/install.mdx
@@ -153,9 +153,9 @@ If you are planning to work locally and use simulators built into Qiskit, then y
 <span id="os-support"></span>
 ## Operating system support
 
-Qiskit strives to support as many operating systems as possible, but due to limitations in available testing resources and operating system availability, not all operating systems can be supported. Operating system support for Qiskit is broken into three tiers with different levels of support for each tier. For operating systems outside these, Qiskit is probably still installable, but it is not tested and you will have to build Qiskit (and likely Qiskit’s dependencies) from source.
+Qiskit strives to support as many operating systems as possible, but due to limitations in available testing resources and operating system availability, not all operating systems can be supported. Operating system support for Qiskit is broken into three tiers with different levels of support for each tier. For platforms outside these such as FreeBSD or WebAssembly (WASI), Qiskit may still be installable, but it is not tested and you will have to build Qiskit (and likely Qiskit’s dependencies) from source.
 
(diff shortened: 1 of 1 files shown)


docs/start/install.mdx
```diff
@@ -153,9 +153,9 @@ If you are planning to work locally and use simulators built into Qiskit, then y
 <span id="os-support"></span>
 ## Operating system support
 
-Qiskit strives to support as many operating systems as possible, but due to limitations in available testing resources and operating system availability, not all operating systems can be supported. Operating system support for Qiskit is broken into three tiers with different levels of support for each tier. For operating systems outside these, Qiskit is probably still installable, but it is not tested and you will have to build Qiskit (and likely Qiskit’s dependencies) from source.
+Qiskit strives to support as many operating systems as possible, but due to limitations in available testing resources and operating system availability, not all operating systems can be supported. Operating system support for Qiskit is broken into three tiers with different levels of support for each tier. For platforms outside these such as FreeBSD or WebAssembly (WASI), Qiskit may still be installable, but it is not tested and you will have to build Qiskit (and likely Qiskit’s dependencies) from source.
```
Eric-Arellano, line 156:
```suggestion
Qiskit strives to support as many operating systems as possible, but due to limitations in available testing resources and operating system availability, not all operating systems can be supported. Operating system support for Qiskit is broken into three tiers with different levels of support for each tier. For platforms outside these, such as FreeBSD or WebAssembly (WASI), Qiskit may still be installable, but it is not tested and you will have to build Qiskit (and likely Qiskit’s dependencies) from source.
```

Eric-Arellano (approved):
Thanks!

wshanks:
My main motivation here was to get "pypy" into the documentation so it would show up in a search. In https://github.com/Qiskit/qiskit/issues/11822, I even linked to this page about operating system support to say that I only saw operating system support discussed when really it did mention only supporting CPython. I added FreeBSD and WebAssembly as I what I thought would be the other "odd" cases that someone might actually wonder about.

I wondered if the section should be using "platform" instead of "operating system". It discusses both CPU architecture and operating system for the most part but also includes the mention of Python interpreter (and whatever WebAssembly is...I guess a "runtime"). That would be a larger change, so I didn't do that here.

abbycross (approved):
👍

Issue #732, opened 2024-01-30
API generation should remove Matplotlib figure captions


Eric-Arellano:
[This RST](https://github.com/Qiskit/qiskit-ibm-runtime/blob/99976520e904cf99487a78b78e170b0a23a04be0/qiskit_ibm_runtime/fake_provider/__init__.py#L34-L60) results in this Sphinx-generated HTML:

<img width="359" alt="Screenshot 2024-01-30 at 4 36 11 PM" src="https://github.com/Qiskit/documentation/assets/14852634/9d400920-6982-4441-9f1f-f7fee96baca8">

Note the `Fig 1` caption and how it has 3 links to download source files for the image:

<img width="440" alt="Screenshot 2024-01-30 at 4 36 59 PM" src="https://github.com/Qiskit/documentation/assets/14852634/2d26fb75-2ade-4772-9021-e076a40204fe">

This is how we render that RST in our platform after running the HTML through our pipeline:

<img width="445" alt="Screenshot 2024-01-30 at 4 37 40 PM" src="https://github.com/Qiskit/documentation/assets/14852634/637c0a80-f765-4b53-9462-b508d61dabcb">

The 3 URLs all are 404s. Instead, we should remove the caption entirely. The 3 URLs are not useful and we don't plan to support them—you can download the image directly. That would only leave the text `Fig. 1`, which we decided isn't a priority. 

--

This impacts the Runtime 0.18 docs in https://github.com/Qiskit/documentation/pull/729, along with some historical docs if you run `rg '_downloads' -l`, e.g.:

https://github.com/Qiskit/documentation/blob/75aaa317a204068fb940cc2776b0ce5935471cb2/docs/api/qiskit/0.42/qiskit.transpiler.passes.DynamicalDecoupling.md?plain=1#L69-L75

--

Example HTML from Sphinx for a single figure:

```html
<figure class="align-default" id="id1">
    <img alt="../_images/fake_provider-1_00.png" class="plot-directive" src="../_images/fake_provider-1_00.png" />
    <figcaption>
        <p>
            <span class="caption-number">Fig. 1 </span>
            <span class="caption-text">
                (
                <a class="reference download internal" download="" href="../_downloads/a640acbc08577560dc62a3c02c6ca2ac/fake_provider-1_00.png">
                    <code class="xref download docutils literal notranslate"><span class="pre">png</span></code>
                </a>
                ,
[...]

PR #848, opened 2024-02-19
remove matplotlib figure captions

kevinsung:
Fixes https://github.com/Qiskit/documentation/issues/732


scripts/lib/api/processHtml.test.ts
```diff
@@ -191,6 +192,98 @@ test("removeColonSpans()", () => {
   doc.expectHtml(`<dt class="field-odd">Parameters</dt>`);
 });
 
+test("removeMatplotlibFigCaptions()", () => {
```
Eric-Arellano, line 195: It'd be helpful to use `describe()` to group the two tests together - you can find some examples in this file. Then, in the `describe()` block, define a variable for `inputHtml` with the multiline string. That will allow you to avoid repeating the same HTML three times.

scripts/lib/api/processHtml.ts
```diff
@@ -122,6 +122,13 @@ export function removePermalinks($main: Cheerio<any>): void {
   }
 }
 
+export function removeMatplotlibFigCaptions($main: Cheerio<any>): void {
```
Eric-Arellano, line 125: This function needs to be called inside `processHtml`. I recommend putting it after `removeDownloadSourceCode($main);`

Eric-Arellano (commented):
Huzzah! After making the requested changes, you'll need to run `npm run regen-apis`. This takes ~40 minutes and means you can't make changes to the repo because it messes up your Git state, so I recommend doing it during meetings. It runs in the background. Or let me know if you want me to run it for you.

scripts/lib/api/processHtml.test.ts
kevinsung: I've grouped the tests. I didn't define the variable for the inputHtml because the HTML is actually different every time. The second test edits it so that it doesn't match the selector. Still, even for that test the HTML gets modified. Specifically,
```html
<img alt="../_images/fake_provider-1_00.png" class="plot-directive" src="../_images/fake_provider-1_00.png" />
```
gets changed to
```html
<img alt="../_images/fake_provider-1_00.png" class="plot-directive" src="../_images/fake_provider-1_00.png">
```
(note the missing backslash at the end). I don't know why that happens.

scripts/lib/api/processHtml.ts
kevinsung: Done.

kevinsung:
@Eric-Arellano I've run `npm run regen-apis`

kevinsung:
Apparently that introduced some broken links. Do you know how to fix it?

Eric-Arellano:
Oops, I introduced them in https://github.com/Qiskit/documentation/pull/850. Our link checker didn't run like it should have. I'll fix.

PR #849, opened 2024-02-19
Add tests showing that `sphinx.ext.linkcode` can be handled

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/517. Turns out that we already do the right thing! That's because both `sphinx.ext.viewcode` and `sphinx.ext.linkcode` use the same HTML structure.

This PR adds tests and better comments to make that clear.

The key insight is that we don't need to change the URLs at all, unlike using `sphinx.ext.viewcode`. We assume the source repositories are giving us valid URLs.


frankharkins (approved):
That's fortunate, nice!

PR #850, opened 2024-02-20
Add objects.inv for historical API docs

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/829.

PR #857, opened 2024-02-20
Fix link checker for historical objects.inv files

Eric-Arellano:
The API checks didn't run in https://github.com/Qiskit/documentation/pull/850 because our GitHub Action didn't include `public/api` in its globs. So, we accidentally broke the link checker.

This PR adds the new mechanism of `IGNORED_FILES`. The `objects.inv` has way too many errors in 0.14 - 0.16 to be worth trying to fix or blocklist specific failures.

Finally, it removes ignores that are now stale thanks to https://github.com/Qiskit/documentation/pull/848.


arnaucasau (approved):
Looks good to me! Thanks!

PR #862, opened 2024-02-21
Link checker: One release note per FileBatch

arnaucasau:
Part of #755 

This PR changes how we are checking the links of the Qiskit release notes. The link checker will check every file in a different batch allowing us to load all the necessary files for each release note.


scripts/commands/checkLinks.ts
```diff
@@ -141,6 +147,8 @@ async function determineCurrentDocsFileBatch(
     // Ignore dev version
     "!docs/api/{qiskit,qiskit-ibm-provider,qiskit-ibm-runtime}/dev/*",
     "!public/api/{qiskit,qiskit-ibm-provider,qiskit-ibm-runtime}/dev/*",
+    // Ignore Qiskit release notes
+    "!docs/api/qiskit/release-notes/*",
   ];
   const toLoad = [
     // The 0.46 docs are used by release notes for APIs that were removed in 1.0.
```
Eric-Arellano, line 154: I think you'll temporarily need to load these files for _all_ Qiskit release notes. We can clean that up after https://github.com/Qiskit/documentation/pull/865

For now, you could make this list a file constant and pass it as an arg for Qiskit with `extraFilesToLoad`.

scripts/commands/checkLinks.ts
```diff
@@ -203,19 +218,44 @@ async function determineDevFileBatches(): Promise<FileBatch[]> {
 
 async function determineHistoricalFileBatches(
   projectName: string,
-  toLoad: string[] = [],
+  extraFilesToLoad: string[],
```
Eric-Arellano, line 221: Nit: `extraGlobsToLoad` is more clear.

scripts/commands/checkLinks.ts
```diff
@@ -203,19 +218,44 @@ async function determineDevFileBatches(): Promise<FileBatch[]> {
...
   ).filter((file) => file.isDirectory() && file.name.match(/[0-9].*/));
 
   const result = [];
   for (const folder of historicalFolders) {
-    const fileBatch = await FileBatch.fromGlobs(
-      [
+    const toCheck: string[] = [];
+    const toLoad = [...extraFilesToLoad];
+
+    if (checkHistoricalApiDocs) {
```
Eric-Arellano, line 238: If `checkSeparateReleaseNotes` is set but not `checkHistoricalApiDocs`, then you need to add ``docs/api/${projectName}/${folder.name}/*`,` in `toLoad`.

Eric-Arellano (commented):
Nice work! This is complex logic but your code is easy to follow.

scripts/commands/checkLinks.ts
arnaucasau: Good catch! I added them to the `toLoad` directly inside the release notes' if statement. Given that is going to be super provisional, we can remove it easily that way without modifying the current version. To avoid any link breaking I had to add 1.0 and other links like 0.45 besides the 0.46 version.

Eric-Arellano (approved):
Thanks!

PR #864, opened 2024-02-21
Update drawer output images in mdx files to match Qiskit 1.0 (part one)

### LINKED ISSUE #833: Re-draw images for MDX files with Qiskit 1.0
~In most cases, code in the docs has been updated to use `style="iqp"` (note: still need to fix several code blocks that do not have this update!).~ (Not necessary in 1.0)

In order to bring the images in line with IBM Quantum colors etc., we need to regenerate the images in the docs so that the output style matches the updated code. 

This means re-running notebooks for ipynb files, and for .mdx files it means running the code and import the output image file.

abbycross:
Closes #833


docs/build/Untitled.ipynb
Eric-Arellano, line 1: Bad change

Eric-Arellano (approved):
Thanks for doing this! Can you please delete the old image if it's no longer being used?

abbycross:
> Thanks for doing this! Can you please delete the old image if it's no longer being used?

of course -- will delete once I've verified it works ok. That will be my pattern for each image.

public/images/transpile/defaults-and-configuration-options/original-circuit-qiskit-1.0.png
Eric-Arellano, line 1: This new image is a lot smaller. Is that intentional?

abbycross: It was downloaded directly from jupyter lab rather than a screenshot - I'm waiting to view it in preview

abbycross: I can see it now -- looks better than the larger image, which was a bit blurry.

abbycross:
I will break this work up into chunks, so renaming the PR to "part 1"

Issue #493, opened 2023-12-07
API docs script should relativize internal links


Eric-Arellano:
If API docs use https://docs.quantum.ibm.com/start/configure-qiskit-local, then we should change this to `/start/configure-qiskit-local`. That is, we should chop off the prefix.

We actually already did this for `qiskit.org`:

https://github.com/Qiskit/documentation/blob/bbccec5f9a20b59c414272422e4d65b827a2e7f7/scripts/lib/api/Pkg.ts#L386-L404

We need to keep that code for historical API docs. But we should expand it to handle `https://docs.quantum.ibm.com` and `https://docs.quantum-computing.ibm.com`. We should also apply the logic to Qiskit, whereas right now it only is used for Runtime and Provider.

PR #869, opened 2024-02-22
Refactor how we relativize links to qiskit.org

Eric-Arellano:
This is prework for https://github.com/Qiskit/documentation/issues/493 to relativize links to docs.quantum.ibm.com. 

This refactors the code that we'll be modifying:

* Always applies the link transformation, whereas before it wasn't used on Qiskit.
* renames `updateUrl` to `normalizeUrl`
* renames `transformLink` to `relativizeLink`
* adds tests for `relativizeLink`

### DIFF
--- a/scripts/lib/api/objectsInv.ts
+++ b/scripts/lib/api/objectsInv.ts
@@ -129,7 +129,7 @@ export class ObjectsInv {
     return uri;
   }
 
-  updateUris(transformLink: Function): void {
+  updateUris(transformLink: (uri: string) => string): void {
     for (const entry of this.entries) {
       entry.uri = entry.uri.replace(/\.html/, "");
       entry.uri = transformLink(entry.uri);
--- a/scripts/lib/api/updateLinks.test.ts
+++ b/scripts/lib/api/updateLinks.test.ts
@@ -12,9 +12,8 @@
 
 import { describe, expect, test } from "@jest/globals";
 import { ObjectsInv } from "./objectsInv";
-import { last } from "lodash";
 
-import { updateLinks, updateUrl } from "./updateLinks";
+import { updateLinks, normalizeUrl, relativizeLink } from "./updateLinks";
 import { HtmlToMdResultWithUrl } from "./HtmlToMdResult";
 
 describe("updateLinks", () => {
@@ -29,6 +28,8 @@ describe("updateLinks", () => {
 [link5](../apidocs/qiskit_ibm_runtime.RuntimeJob#qiskit_ibm_runtime.RuntimeJob)
 [link6](qiskit_ibm_runtime.RuntimeJob)
 [link7](#qiskit_ibm_runtime.RuntimeJob.job)
+[link8](https://qiskit.org/documentation/apidoc/algorithms.html)
+[link9](https://qiskit.org/documentation/stubs/qiskit.circuit.BreakLoopOp.html#assemble)
           `,
         meta: {
           apiType: "class",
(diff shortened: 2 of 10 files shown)


scripts/lib/api/updateLinks.test.ts
```diff
@@ -127,136 +130,48 @@ describe("updateLinks", () => {
       ]
     `);
   });
-
-  test("update links using a transform function", async () => {
```
Eric-Arellano, line 131: I removed this test since you can't give a custom transform function anymore.

scripts/lib/api/updateLinks.test.ts
```diff
@@ -127,136 +130,48 @@ describe("updateLinks", () => {
...
-
-    const objectsInvEntries = [
-      "stubs/qiskit.algorithms.Eigensolver.html#$",
-      "stubs/qiskit.algorithms.Eigensolver.html#$",
-      "stubs/qiskit.algorithms.EigensolverResult.html#$",
-      "stubs/qiskit.algorithms.EstimationProblem.html#$",
-      "stubs/qiskit.algorithms.EvolutionProblem.html#$",
-      "stubs/qiskit.algorithms.EvolutionProblem.html#$",
-      "stubs/qiskit.algorithms.FasterAmplitudeEstimationResult.html#$",
-      "stubs/qiskit.algorithms.FasterAmplitudeEstimationResult.html#$",
```
Eric-Arellano, line 159: @frankharkins should I add any of these entries to the single test? I wasn't sure if this is testing something unique.

scripts/lib/api/updateLinks.test.ts
```diff
@@ -127,136 +130,48 @@ describe("updateLinks", () => {
...
-        "qiskit.algorithms.EvolutionProblem#$",
-        "qiskit.algorithms.EvolutionProblem#$",
-        "qiskit.algorithms.FasterAmplitudeEstimationResult#$",
-        "qiskit.algorithms.FasterAmplitudeEstimationResult#$",
-      ]
-   `);
-  });
 });
 
-describe("updateUrl", () => {
```
Eric-Arellano, line 221: No changes made to this test - only removed the `describe` so it's dedented.

scripts/lib/api/updateLinks.ts
```diff
@@ -95,10 +99,29 @@ export function updateUrl(
   return url;
 }
 
+export function relativizeLink(link: Link): Link | undefined {
```
Eric-Arellano, line 102: No changes made to this function other than a small change I call out below.

scripts/lib/api/updateLinks.ts
```diff
@@ -95,10 +99,29 @@ export function updateUrl(
...
+    return;
+  }
+  let [url, anchor] = link.url.split("#");
+  url = removePrefix(url, prefix);
+  url = removeSuffix(url, ".html");
+  if (anchor && anchor !== url) {
+    url = `${url}#${anchor}`;
+  }
+
+  const newText = link.url === link.text ? url : undefined;
```
Eric-Arellano, line 118: We used to have a boolean variable called `updateText` for the condition. I inlined it.

scripts/lib/api/objectsInv.ts
```diff
@@ -129,7 +129,7 @@ export class ObjectsInv {
     return uri;
   }
 
-  updateUris(transformLink: Function): void {
+  updateUris(transformLink: (uri: string) => string): void {
```
frankharkins, line 132: Cool! I remember wondering if there was a better way when I wrote that

Eric-Arellano:
I skipped the extended API check because it doesn't even run on the historical API docs we changed.

scripts/lib/api/updateLinks.test.ts
frankharkins: I'm not sure why each line is duplicated, and I don't think it's testing anything unique.

If we remove this, do we have a way of checking `objects.inv` is updated in `updateLinks`?

Eric-Arellano: Yeah, the existing test alrealdy has a bunch of stuff for `objects.inv`. Only not `qiskit.algorithms`. I figure you put that there because the custom `transformLink` logic specifically looked at `algorithms`. I think this is all fine.

PR #870, opened 2024-02-22
Relativize API links to docs.quantum.ibm.com

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/493. This change makes our link checker work better because it can quickly validate the relative links and also validate anchors, whereas external links are much slower.


scripts/lib/api/updateLinks.test.ts
```diff
@@ -212,7 +212,15 @@ describe("relativizeLink()", () => {
...
     ],
-  ])("transform qiskit.org Qiskit API links", (input, expected) => {
+  ])("relativize qiskit.org Qiskit API links", (input, expected) => {
+    expect(relativizeLink({ url: input })).toEqual({ url: expected });
+  });
+
+  test.each([
+    ["https://docs.quantum.ibm.com/run", "/run"],
+    ["https://docs.quantum.ibm.com/api/qiskit/utils", "/api/qiskit/utils"],
+    ["https://docs.quantum-computing.ibm.com/run", "/run"],
```
frankharkins, line 222:
```suggestion
    ["https://docs.quantum-computing.ibm.com/run", "/run"],
    ["https://docs.quantum-computing.ibm.com", "/"],
```

Probably an edge case, but should we handle this?

Eric-Arellano: Good thinking. I think it's fine though. The script would do nothing for `https://docs.quantum-computing.ibm.com` (good), whereas it would transform `https://docs.quantum-computing.ibm.com/` to `/`. But that's not a valid synthetic file, so the link checker would error:

https://github.com/Qiskit/documentation/blob/566b553a243ac9801890157950350ea498735137/scripts/commands/checkLinks.ts#L25-L28

I can add a test that `https://docs.quantum-computing.ibm.com` without the `/` is ignored.

frankharkins: I guess I meant "Should we relativize `https://docs.quantum-computing.ibm.com`"? But it's probably rare enough that I think it's fine either way.

frankharkins (approved):
Logic and API docs diff look good 🙂

Eric-Arellano:
Review appreciated on the last commit: https://github.com/Qiskit/documentation/pull/870/commits/9ec2203f5aa0e82cf8124509a2b2c23b56fadd7d.

scripts/commands/checkLinks.ts
```diff
@@ -179,6 +189,7 @@ async function determineCurrentDocsFileBatch(
       .join(".");
 
     toCheck.push(`docs/api/qiskit/release-notes/${currentVersion}.md`);
+    toLoad.push("docs/api/qiskit/release-notes/0.{45,46}.md");
```
arnaucasau, line 192: We need the 0.44 release notes to fix the last broken link. I added the comment to clarify that we only need them as long as qiskit 1.0 is the latest version. After that, we'll need to move it to the historical `toLoad`
```suggestion
    // Necessary files for docs/api/qiskit/release-notes/1.0.md
    toLoad.push("docs/api/qiskit/release-notes/0.{44,45,46}.md");
```

Issue #827, opened 2024-02-16
Better configure cspell by using dictionaries


Eric-Arellano:
See https://github.ibm.com/IBM-Q-Software/iqp-channel-docs/pull/1131 for inspiration.

When applying this change, see if it would have caught the problems we missed in https://github.com/Qiskit/documentation/pull/872#issuecomment-1960165963.

PR #872, opened 2024-02-22
fix spelling issues found by spellcheck

abbycross:


Eric-Arellano:
Hm, I don't know why our spellchecker didn't fail on these! It should be checking these files. How did you find the problems?

Eric-Arellano (approved):
Thanks! Our link checker here should have caught this. Kaelyn will see if https://github.com/Qiskit/documentation/issues/827 would have caught this problem.

PR #865, opened 2024-02-21
Update legacy release notes links

arnaucasau:
Part of #755

This PR changes the legacy release notes links to point to the qiskit 0.45 version folder.

### How the change was done

I created a little script to transform every link in a release note file and re-write it in the same destination. I named it `updateLinks.ts`.

```ts
import { zxMain } from "../lib/zx";
import { globby } from "globby";
import { pathExists, getRoot } from "../lib/fs";
import { readFile, writeFile } from "fs/promises";
import transformLinks from "transform-markdown-links";

zxMain(async () => {
  const allVersions = (
    await globby("docs/api/qiskit/release-notes/[!index]*")
  ).map((releaseNotesPath) =>
    releaseNotesPath.split("/").pop()!.split(".").slice(0, -1).join("."),
  );

  for (const releaseNotesVersion of allVersions) {
    const source = `${getRoot()}/docs/api/qiskit/release-notes/${releaseNotesVersion}.md`;
    if (await pathExists(source)) {
      updateLinksFile('qiskit', releaseNotesVersion, source, source);
    }
  }
});

async function updateLinksFile(pkgName: string, versionWithoutPatch: string, source: string, dest: string) {
  let markdown = await readFile(source, { encoding: "utf8" });

  const regexAbsolutePath = new RegExp(
    "/api/" + pkgName + "/(?!release-notes)(?![0-9])",
  );

  markdown = transformLinks(markdown, (link, _) =>
    link.replace(regexAbsolutePath, `/api/${pkgName}/0.45/`),
  );

  await writeFile(dest, markdown);
}


```

To execute the script using `npm run update-legacy-links` you can modify the `package.json` file by adding the following line in the `scripts` section:

```ts
"update-legacy-links": "node -r esbuild-register scripts/commands/updateLinks.ts"
```

PR #881, opened 2024-02-23
Link checker: Qiskit legacy release notes in a single FileBatch

arnaucasau:
The legacy release notes of qiskit were generated from qiskit 0.45 and we need to check them loading those docs. This PR groups together all the legacy release notes in a FileBatch to avoid loading the same docs multiple times.

We define a legacy version to be prior to qiskit 0.45. In order to compare which version is newer or older than 0.45, this PR introduces a new function named `compareReleaseNotes`. We can't use simple comparisons like +version < 0.45 as we do in the other scripts because the comparison between version 0.5 and 0.45 will be incorrect.

### DIFF
--- a/scripts/commands/checkLinks.ts
+++ b/scripts/commands/checkLinks.ts
@@ -318,7 +318,7 @@ async function determineQiskitLegacyReleaseNotes(): Promise<FileBatch> {
 
   return await FileBatch.fromGlobs(
     toCheck,
-    [`docs/api/qiskit/0.44/*`],
+    [`docs/api/qiskit/0.45/*`],
     `qiskit legacy release notes`,
   );
 }
--- a/scripts/commands/checkLinks.ts
+++ b/scripts/commands/checkLinks.ts
@@ -19,7 +19,7 @@ import { hideBin } from "yargs/helpers";
 import { pathExists } from "../lib/fs";
 import { File } from "../lib/links/LinkChecker";
 import { FileBatch } from "../lib/links/FileBatch";
-import { QISKIT_MISSING_VERSION_MAPPING } from "../lib/qiskitMetapackage";
+import { compareReleaseNotes } from "../lib/api/releaseNotes";
 
 // While these files don't exist in this repository, the link
 // checker should assume that they exist in production.
@@ -147,6 +147,10 @@ async function determineFileBatches(args: Arguments): Promise<FileBatch[]> {
 
   result.push(...provider, ...runtime, ...qiskit);
 
+  if (args.qiskitReleaseNotes) {
+    result.push(await determineQiskitLegacyReleaseNotes());
+  }
+
   return result;
 }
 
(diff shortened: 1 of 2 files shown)


scripts/commands/checkLinks.ts
```diff
@@ -19,7 +19,7 @@ import { hideBin } from "yargs/helpers";
 import { pathExists } from "../lib/fs";
 import { File } from "../lib/links/LinkChecker";
 import { FileBatch } from "../lib/links/FileBatch";
-import { QISKIT_MISSING_VERSION_MAPPING } from "../lib/qiskitMetapackage";
```
Eric-Arellano, line 22: This constant can be deleted now, right? It's still good to have the dedicated file though.

scripts/commands/checkLinks.ts
```diff
@@ -295,4 +295,32 @@ async function determineHistoricalFileBatches(
...
+async function determineQiskitLegacyReleaseNotes(): Promise<FileBatch> {
+  const result: FileBatch[] = [];
+
+  const legacyVersions = (
+    await globby("docs/api/qiskit/release-notes/[!index]*")
+  )
+    .map((releaseNotesPath) =>
+      releaseNotesPath.split("/").pop()!.split(".").slice(0, -1).join("."),
+    )
+    .filter((version) => compareReleaseNotes(version, "0.45") < 0);
```
Eric-Arellano, line 307: We know there will never be another version between Qiskit 0.46 and 1.0. So, how about simplifying this to check if the version is `<1 && != 0.45 && != 0.46`? Good job writing generic code with `compareReleaseNotes`! But I think we can simplify this.

Note that the sorting doesn't matter either - it all gets checked in the same `FileBatch` either way.

scripts/commands/checkLinks.ts
```diff
@@ -295,4 +295,32 @@ async function determineHistoricalFileBatches(
...
+
+  const toCheck = legacyVersionsSorted.map(
+    (legacyVersion) => `docs/api/qiskit/release-notes/${legacyVersion}.md`,
+  );
+
+  console.log(legacyVersionsSorted);
+
+  return await FileBatch.fromGlobs(
+    toCheck,
+    [`docs/api/qiskit/0.45/*`],
```
Eric-Arellano, line 321: Does this pass for current code? I think you will want to move the code you currently have marked as TODO to instead apply here.

Eric-Arellano (commented):
Thanks for diving further deeper into the complexity of this all 😅

scripts/commands/checkLinks.ts
arnaucasau: Yes, good catch! I have added it there and I will remove it in my other [PR](https://github.com/Qiskit/documentation/pull/865) :+1:

scripts/commands/checkLinks.ts
```diff
@@ -295,4 +295,39 @@ async function determineHistoricalFileBatches(
...
+      (version) => +version < 1 && version != "0.45" && version != "0.46",
+    );
+
+  const toCheck = legacyVersions.map(
+    (legacyVersion) => `docs/api/qiskit/release-notes/${legacyVersion}.md`,
+  );
+
+  return await FileBatch.fromGlobs(
+    toCheck,
+    // Temporary - remove after https://github.com/Qiskit/documentation/pull/865 is merged
```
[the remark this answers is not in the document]
arnaucasau: Yeah, I was double checking everything. The link we need to load is "docs/api/qiskit/release-notes/index.md`

Eric-Arellano (approved):
🙇

Eric-Arellano:
PR description is out of date btw

Issue #876, opened 2024-02-23
Improve link checker API arguments to allow being more precise


Eric-Arellano:
Our link checker is crucial to avoiding 404s, but it's somewhat slow because we have so many API docs. The external link checker is also flakey due to rate limiting (https://github.com/Qiskit/documentation/issues/823), which is annoying because it results in red CI runs.

Before we start link checking historical Qiskit versions via https://github.com/Qiskit/documentation/issues/823, we should make some changes.

* Remove `--qiskit-release-notes` but only after https://github.com/Qiskit/documentation/issues/823 is done.
* Add an `--internal` arg with default to `true` so that you can set `--no-internal` to say to only check external links.
* Have some way to specify to check historical API docs but only the most recent, like the last 2-3 of a release? Something like `--all-historical-apis` vs `--latest-historical-apis`? This is so that we can e.g. check 0.45 and 0.46 (1.0 is latest), but skip 0.19.

With those changes, I think we'd want this for our API checks workflow.

```bash
# Internal links
--current-apis --dev-apis --all-historical-apis

# External links
--current-apis --dev-apis --latest-historical-apis --no-internal --external
```

I'm unsure if we would ever want the external link checker to run for `--all-historical-apis`? Maybe as a weekly cron job, or even only as a manual workflow_dispatch? The issue with this check is rate limiting https://github.com/Qiskit/documentation/issues/823.

PR #882, opened 2024-02-23
Split out `InternalLink` and `ExternalLink` classes

Eric-Arellano:
We decided as part of https://github.com/Qiskit/documentation/issues/876 that we should have distinct programs for checking internal vs external links, since they are such distinct workflows.

This PR is a refactor to prepare for that change. The link checker program still behaves the same as before; this is only a refactor.

### DIFF
--- a/scripts/lib/links/ExternalLink.ts
+++ b/scripts/lib/links/ExternalLink.ts
@@ -1,0 +1,58 @@
+// This code is a Qiskit project.
+//
+// (C) Copyright IBM 2024.
+//
+// This code is licensed under the Apache License, Version 2.0. You may
+// obtain a copy of this license in the LICENSE file in the root directory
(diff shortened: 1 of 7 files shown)


scripts/lib/links/ExternalLink.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * Returns an error message if link failed.
+   */
+  async check(): Promise<string | undefined> {
+    let error: string = "";
+    try {
+      const response = await fetch(this.value, {
+        headers: { "User-Agent": "qiskit-documentation-broken-links-finder" },
+      });
```
frankharkins, line 41: Not technically part of this PR, but should we be using `method: "HEAD"` here to avoid downloading the body? Might make things faster and reduce rate limiting if we're not doing that already.

frankharkins (approved):
Thanks! This feels much neater

PR #884, opened 2024-02-26
Determine internal vs external links at parse time

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/876 to split the internal and external link checkers into distinct programs. To do this, it's useful to split out internal vs. external links at parse time, whereas before we did it inside `FileBatch`. The external link checker won't use `FileBatch`.

This is only a refactor and doesn't change the program, other than now using `HEAD` for external link requests rather than `GET`. We also now use a `Set` for anchors so that checking if an anchor is included is faster.

### DIFF
--- a/scripts/commands/checkLinks.ts
+++ b/scripts/commands/checkLinks.ts
@@ -83,8 +83,10 @@ async function main() {
 
   const fileBatches = await determineFileBatches(args);
   const otherFiles = [
-    ...(await globby("public/**/*")).map((fp) => new File(fp, [])),
-    ...SYNTHETIC_FILES.map((fp) => new File(fp, [], true)),
+    ...(await globby("public/{images,videos}/**/*")).map(
+      (fp) => new File(fp, new Set()),
+    ),
+    ...SYNTHETIC_FILES.map((fp) => new File(fp, new Set(), true)),
   ];
 
   let allGood = true;
--- a/scripts/lib/links/FileBatch.ts
+++ b/scripts/lib/links/FileBatch.ts
@@ -137,9 +137,9 @@ export function addLinksToMap(
   links: Set<string>,
   linksToOriginFiles: Map<string, string[]>,
 ): void {
+  if (IGNORED_FILES.has(filePath)) return;
   links.forEach((link) => {
     if (
-      IGNORED_FILES.has(filePath) ||
       ALWAYS_IGNORED_URLS.has(link) ||
       FILES_TO_IGNORES[filePath]?.includes(link)
     ) {
--- a/scripts/lib/links/FileBatch.ts
+++ b/scripts/lib/links/FileBatch.ts
@@ -66,43 +66,36 @@ export class FileBatch {
    *   2. A list of InternalLink objects to validate.
    *   3. A list of ExternalLink objects to validate.
    */
-  async load(): Promise<[File[], InternalLink[], ExternalLink[]]> {
+  async load(
+    loadExternalLinks: boolean,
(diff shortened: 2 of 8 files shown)


scripts/commands/checkLinks.ts
```diff
@@ -83,8 +83,10 @@ async function main() {
 
   const fileBatches = await determineFileBatches(args);
   const otherFiles = [
-    ...(await globby("public/**/*")).map((fp) => new File(fp, [])),
-    ...SYNTHETIC_FILES.map((fp) => new File(fp, [], true)),
+    ...(await globby("public/{images,videos}/**/*")).map(
```
Eric-Arellano, line 86: This would include the `objects.inv` files, which was wrong.

scripts/lib/links/FileBatch.ts
```diff
@@ -66,43 +66,36 @@ export class FileBatch {
...
       files.push(new File(filePath, parsed.anchors));
     }
 
-    const linksToOriginFiles = new Map<string, string[]>();
+    const internalLinksToOriginFiles = new Map<string, string[]>();
+    const externalLinksToOriginFiles = new Map<string, string[]>();
     for (const filePath of this.toCheck) {
       const parsed = await parseFile(filePath);
       files.push(new File(filePath, parsed.anchors));
-      if (!IGNORED_FILES.has(filePath)) {
```
Eric-Arellano, line 80: All three ignores are moved into `addLinksToMap`.

scripts/lib/links/FileBatch.ts
```diff
@@ -66,43 +66,36 @@ export class FileBatch {
...
-      );
-
-      if (originFiles.length > 0) {
-        if (linkPath.startsWith("http")) {
-          externalLinks.push(new ExternalLink(linkPath, originFiles));
-        } else {
-          internalLinks.push(new InternalLink(linkPath, originFiles));
-        }
+      addLinksToMap(filePath, parsed.internalLinks, internalLinksToOriginFiles);
+      if (loadExternalLinks) {
```
Eric-Arellano, line 84: We don't waste our time storing external links if not necessary. That should slightly decrease memory usage and improve performance.

scripts/lib/links/FileBatch.ts
```diff
@@ -111,19 +104,21 @@ export class FileBatch {
...
 
     const results = internalLinkList.map((link) => link.check(existingFiles));
 
-    if (externalLinks) {
-      // For loop reduces the risk of rate-limiting.
-      for (let link of externalLinkList) {
-        results.push(await link.check());
-      }
+    // For loop reduces the risk of rate-limiting.
+    for (let link of externalLinkList) {
```
Eric-Arellano, line 120: `externalLinkList` will be empty if we're not checking external links.

arnaucasau (approved):
Thank you Eric! Tested locally and all looks good :)

scripts/lib/links/FileBatch.ts
```diff
@@ -139,10 +134,18 @@ export class FileBatch {
...
 ): void {
   links.forEach((link) => {
+    if (
+      IGNORED_FILES.has(filePath) ||
+      ALWAYS_IGNORED_URLS.has(link) ||
+      FILES_TO_IGNORES[filePath]?.includes(link)
+    ) {
+      return;
+    }
+
```
frankharkins: Very nice

frankharkins (approved):
Looks good, thanks!

PR #889, opened 2024-02-26
Fix logic for checking historical Qiskit docs

Eric-Arellano:
Fixes a bug in https://github.com/Qiskit/documentation/pull/881. It handled `--qiskit-release-notes` correctly, but it incorrectly would skip `--historical-apis` for Qiskit versions before 0.45.

This also loads fewer files for the migration guides.


arnaucasau (approved):
Good catch! Thank you for fixing this Eric! Tested locally, and now it checks all historical versions

PR #890, opened 2024-02-26
enforce 50 <= description length <= 160

### LINKED ISSUE #327: Enforce that metadata descriptions are not more than 160 characters
Relates to https://github.com/Qiskit/documentation/issues/131. It's bad for SEO to exceed 160 characters.

We should also check that the title != description, which is a common practice. And probably a minimum of 50 characters for the description.

Skip APIs since those are generated.

We should include general tips for how to write good SEO, such as the tips from https://github.com/Qiskit/documentation/issues/131. We could either link to resources like the GitHub issue or directly write the tips.

--

This should simply modify `checkMetadata.ts`. Similar to our check for missing metadata, we should batch all failing files into a single list and single error message.

kevinsung:
Fixes #327

### DIFF
--- a/docs/start/index.mdx
+++ b/docs/start/index.mdx
@@ -1,6 +1,6 @@
 ---
 title: Introduction
-description: Introduction to IBM Quantum documentation
+description: Introduction to IBM Quantum, Qiskit, and Qiskit Runtime.
 
 in_page_toc_max_heading_level: 2
 
--- a/docs/start/index.mdx
+++ b/docs/start/index.mdx
@@ -1,6 +1,6 @@
 ---
 title: Introduction
-description: Introduction to IBM Quantum, Qiskit, and Qiskit Runtime.
+description: Introduction to the documentation for IBM Quantum, Qiskit, and Qiskit Runtime.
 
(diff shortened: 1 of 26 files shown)


psschwei:
:+1: for docs/run/quantum-serverless.mdx

Eric-Arellano (approved):
Thanks! We'll want to make an exception for API docs but this can be fixed in my follow up work to start running the metadata checker on API docs.

The new descriptions look solid to me. Maybe wait for sign off from @javabster because she did the first pass at rewriting these descriptions last time.

abbycross (approved):
👍

docs/start/index.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: Introduction
-description: Introduction to IBM Quantum documentation
+description: Introduction to IBM Quantum, Qiskit, and Qiskit Runtime.
```
javabster, line 3: I think this one should still mention the word documentation right? The page itself mostly explains what the docs platform is more than anything else

kevinsung: Done.

Eric-Arellano (approved):

PR #895, opened 2024-02-27
Deduplicate headers for overloaded methods

### LINKED ISSUE #167: Duplicate headers for overloaded methods

https://docs.quantum-computing.ibm.com/api/qiskit/qiskit.circuit.QuantumCircuit

### Please describe the UI problem.

For example, look at the right sidebar:
image

Each Example and Notes section should have the most indentation, but they have the least. Also, there are duplicates like `if_test` that show up with the same indentation.

arnaucasau:
Closes #167

The script was creating a title for every `<dt>` tag inside a `<dl>` tag. This is almost always correct given that we only have a `<dt>` tag for each method. The problem with overload methods is that we have a `<dt>` for each `@typing.overload` definition.

For example, the `quantumcircuit.py` file overloads the definition of for_loop:

https://github.com/Qiskit/qiskit/blob/244940a10c17f1dc3d74e2665a3ae869dc3d7291/qiskit/circuit/quantumcircuit.py#L4862-L4884

This is an example of the HTML generated

```html
<dl class="py method">
    <dt class="sig sig-object py" id="qiskit.circuit.QuantumCircuit.for_loop">
    <span class="sig-name descname"><span class="pre">for_loop</span>

    <dt class="sig sig-object py">
    <span class="sig-name descname"><span class="pre">for_loop</span>
    
    <dd><p>Create a <code class="docutils literal notranslate"><span class="pre">for</span></code> loop on this circuit.</p>
```
To solve the problem this PR adds an extra condition to check whether the `<dt>` tag has an id or not. Only the first `<dt>` tag will have an id, so we can use this fact to create only one `<h3>` for each method.

### DIFF
--- a/docs/api/qiskit/0.33/qiskit.circuit.QuantumCircuit.md
+++ b/docs/api/qiskit/0.33/qiskit.circuit.QuantumCircuit.md
@@ -1259,13 +1259,9 @@ The circuit index of an [`AncillaQubit`](qiskit.circuit.AncillaQubit "qiskit.cir
 
 ### for\_loop
 
-##### for\_loop
-
 <span id="qiskit.circuit.QuantumCircuit.for_loop" />
 
 `QuantumCircuit.for_loop(indexset: Iterable[int], loop_parameter: Optional[qiskit.circuit.parameter.Parameter], body: None, qubits: None, clbits: None, *, label: Optional[str]) → qiskit.circuit.controlflow.for_loop.ForLoopContext`[GitHub](https://github.com/qiskit/qiskit/tree/stable/0.19/qiskit/circuit/quantumcircuit.py "view source code")
-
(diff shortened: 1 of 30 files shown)


Eric-Arellano (commented):
Excellent! Great work. I like the solution.

Thoughts on adding a test for these? I know it's not super obvious how to do so because we punted on adding tests for `processMembersAndSetMeta`.

arnaucasau:
I think it's a good idea to add tests. What do you think about adding them as a follow-up together with more cases for the `processMembersAndSetMeta` function? I'm going to test the generation with runtime and provider to make sure it doesn't break anything

frankharkins (approved):
Neat solution and nice comment, thanks!

Eric-Arellano (approved):

PR #899, opened 2024-02-27
Fix API generation regression

### LINKED ISSUE #224: API generation regression in properties
From https://github.com/Qiskit/documentation/pull/222#discussion_r1367492777:

This is a regression.

Now:

Screenshot 2023-10-20 at 5 19 17 PM

Before:

Screenshot 2023-10-20 at 5 19 30 PM

qiskit.org:

Screenshot 2023-10-20 at 5 19 57 PM

arnaucasau:
Closes #224

This PR fixes an issue with the API generation script that causes regressions in attributes. For each attribute, we show on the documentation website, its type (if exists) and its default value (if exists).

An example of an attribute could be `instance: str | None = None` which has the type `str | None` and the default value `None`. This is its representation on the website:

Captura desde 2024-02-27 19-12-42

To find the attribute and the default value, the script searches for the index of the colon (type separator) and the index of the equal sign (default value separator). The problem is that those two values are optional, and we could have, for example, a default value involving a string containing a colon symbol that would be detected as the type separator when it's not.

To fix this, we could only search for the type separator before the default value. We don't have any problem with the equal sign because it cannot be a type of attribute.

The PR has regenerated all versions of qiskit, runtime, and provider.


docs/api/qiskit/0.29/qiskit.providers.basicaer.QasmSimulatorPy.md
```diff
@@ -231,17 +231,13 @@ the status of the backend.
...
...kend_version': '2.1.0', 'basis_gates': ['u1', 'u2', 'u3', 'rz', 'sx', 'x', 'cx', 'id', 'unitary'], 'conditional': True, 'coupling_map': None, 'description': 'A python simulator for qasm experiments', 'gates': [{'name': 'u1', 'parameters': ['lambda'], 'qasm_def': 'gate u1(lambda) q { U(0,0,lambda) q; }'}, {'name': 'u2', 'parameters': ['phi', 'lambda'], 'qasm_def': 'gate u2(phi,lambda) q { U(pi/2,phi,lambda) q; }'}, {'name': 'u3', 'parameters': ['theta', 'phi', 'lambda'], 'qasm_def': 'gate u3(theta,phi,lambda) q { U(theta,phi,lambda) q; }'}, {'name': 'rz', 'parameters': ['phi'], 'qasm_def': 'gate rz(phi) q { U(0,0,phi) q; }'}, {'name': 'sx', 'parameters': [], 'qasm_def': 'gate sx(phi) q { U(pi/2,7*pi/2,pi/2) q; }'}, {'name': 'x', 'parameters': [], 'qasm_def': 'gate x q { U(pi,7*pi/2,pi/2) q; }'}, {'name': 'cx', 'parameters': [], 'qasm_def': 'gate cx c,t { CX c,t; }'}, {'name': 'id', 'parameters': [], 'qasm_def': 'gate id a { U(0,0,0) a; }'}, {'name': 'unitary', 'parameters': ['matrix'], 'qasm_def': 'unitary(matrix) q1, q2,...'}], 'local': True, 'max_shots': 65536, 'memory': True, 'n_qubits': 24, 'open_pulse': False, 'simulator': True, 'url': 'https://github.com/Qiskit/qiskit-terra'}`
```
Eric-Arellano, line 236: I was wondering if we should go back to the qiskit.org style of the default value being on the same line as the attribute. But this example makes it clear that that won't work. The default value and type hint might be way too long.

frankharkins (approved):

docs/api/qiskit/0.43/qiskit.transpiler.StagedPassManager.md
```diff
@@ -216,9 +216,7 @@ Expanded Pass manager stages including `pre_` and `post_` phases.
 
 <span id="qiskit.transpiler.StagedPassManager.invalid_stage_regex" />
 
-### invalid\_stage\_regex = re.compile('\\\s|\\\\+|\\\\-|\\\\\*|\\\\/|\\\\\\\\|\\\\%|\\\\\<|\\\\>|\\\\@|\\\\!|\\\\\~|\\\\^|\\\\&|\\\\
```
Eric-Arellano, line 219: !!!!

Eric-Arellano (approved):
Nice work!

This would also be great to test, but totally fine to do in a follow up.

Issue #886, opened 2024-02-26
Only run pages render check on changed files


Eric-Arellano:
Relates to https://github.com/Qiskit/documentation/pull/792. That was on the right track, but still not optimal.

--

Whether a file renders is entirely independent from other files. So there is no dependency with other files; this is unlike internal links, where changing one file might break others that link to it. So, if a page had rendered before, the only way it will stop working is if we made a breaking change to the file, or the app itself is broken.

Therefore, in most cases, it is unnecessary in both our normal CI job and our API job that we check every page. This makes our CI unnecessarily slow.

https://github.com/Qiskit/documentation/blob/74b544a9f4a955baa43d4c82b9d8b78b39b021c4/.github/workflows/main.yml#L46-L47

 https://github.com/Qiskit/documentation/blob/74b544a9f4a955baa43d4c82b9d8b78b39b021c4/.github/workflows/api-checks.yml#L65-L71

It'd be faster to only check changed files. And maybe get rid of the api-checks.yml check and move it into main.yml directly.

> or the app itself is broken.

This is a legit risk, still. For example, in the past, we changed how math expressions rendered which broke some pages. But this should be caught by a weekly cron job, rather than running in every PR. This cron job should check that all pages render.

Note that iqp-channel-docs already validates that non-API docs and current API docs can render. It only ignores historical API docs, dev docs, and translations.

Issue #823, opened 2024-02-15
Figure out link checker rate limiting in CI


Eric-Arellano:
For example, we got rate limited by GitHub on https://github.com/Qiskit/documentation/pull/822.

I'm wondering if we can have some type of persistent cache across PR runs. Don't check links within N days. See internal PR 1082 for how this was done in closed source.

--

~Outside of an externally-managed cache, we should definitely at least have an in-memory cache. Note that we don't store what external links have been checked between `FileBatch`es, so we check the same URL multiple times if e.g. Qiskit 0.44 and 0.45 have the same link.~

~This internal cache won't completely solve the problem: we may still make too many requests to the same host but with different pages, like GitHub source code links. But it helps with certain websites.~

This was implemented by https://github.com/Qiskit/documentation/pull/916. We also now only check links once a week in a cron job, so this issue is less urgent.

frankharkins:
I think you could have an action run a slow, extended check on `main` every month or so, then PRs could use that cache, at least according to this [SO answer](https://stackoverflow.com/a/66632107).

Eric-Arellano:
One technique is to set the auth for GitHub like this:

```
 headers.append('Authorization', `token ${getRequiredEnv('GITHUB_TOKEN')}`);
```

Won't help with other hosts, but GitHub is the most common where this happens.

Eric-Arellano:
This hasn't happened in a long time, in large part thanks to using the GitHub token https://github.com/Qiskit/documentation/pull/1392.

PR #372, opened 2023-11-17
Link checker: skip folders, validation of release notes & new CI tests

arnaucasau:
Closes #368. 

It adds the following features:

- A method to skip complete folders using glob patterns.
- A daily cron job to validate all the internal and external links.
- Validation of the release notes by default

The CI test has been changed to validate only internal links. The motivation behind this is we want the CI to be fast, and also, the validation of external links can make it fail due to fetch errors for external reasons that have nothing to do with the correctness of the link. Regardless, we still want to validate external links, and a new daily test has been added to check all the external links.

PR #913, opened 2024-02-27
[wip] Pages render check only runs on changed files

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/886. As explained there, it's safe to only run this check on changed files because whether a page renders is independent from all the other pages. 

We use a weekly cron job, though, to check all pages. That reduces our risk from things like the docs preview Docker image changing on us and resulting in some pages now failing to render.

--

This PR also stops checking external links in API docs builds and instead moves the check to the new weekly cron job. Relates to https://github.com/Qiskit/documentation/issues/876.

Per https://github.com/Qiskit/documentation/pull/372, it's annoying to check external links in CI because it can be flaky (https://github.com/Qiskit/documentation/issues/823).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,7 +30,7 @@ jobs:
         run: npm run check:metadata
       - name: Spellcheck
         run: npm run check:spelling
-      - name: Link checker
+      - name: Internal link checker
         run: npm run check:links
       - name: Formatting
         run: npm run check:fmt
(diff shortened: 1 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -39,9 +39,19 @@ jobs:
...
       - name: Start local Docker preview
+        if: steps.changed-docs-files.outputs.any_changed == 'true'
         run: |
           ./start &
           sleep 20
       - name: Check that pages render
-        run: npm run check-pages-render
+        if: steps.changed-docs-files.outputs.any_changed == 'true'
+        run: |
+          echo "${{ steps.changed-docs-files.outputs.all_changed_files }}" > changed.txt
```
frankharkins, line 56: Nice solution

frankharkins (approved):
Code looks good and works well 🚀

PR #916, opened 2024-02-28
Split internal and external link checker into separate programs

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/876.

This split brings several improvements:

1. Simpler internal link checker code.
2. The external link checker now deduplicates links across all files, which avoids repeating requests.
3. The external link checker logs its progress every 20 links.
4. It's possible to check external links without checking internal links.

We run the external link checker in our weekly cron job over all files, other than translations and historical Qiskit versions. That change was added in https://github.com/Qiskit/documentation/pull/913.


README.md
```diff
@@ -178,27 +178,35 @@ about it.
...
 
+To check external links:
+
+```bash
+# Specify the files you want after `--`
+npm run check-external-links -- docs/run/index.md
+
+# You can also use globs
+npm run check-external-links -- 'docs/run/*'
+```
```
arnaucasau, line 208: Do you think it would be useful to show how to pass more than one globs expression or file as arguments? Otherwise, someone could think that it only accepts one expression

`npm run check-external-links -- 'docs/run/*' 'docs/start/*'`

package.json
```diff
@@ -5,11 +5,12 @@
...
   "scripts": {
-    "check": "npm run check:metadata && npm run check:spelling && npm run check:links && npm run check:fmt",
+    "check": "npm run check:metadata && npm run check:spelling && npm run check-internal-links && npm run check:fmt",
     "check:metadata": "node -r esbuild-register scripts/commands/checkMetadata.ts",
-    "check:links": "node -r esbuild-register scripts/commands/checkLinks.ts",
     "check:spelling": "cspell --relative --no-progress docs/**/*.md* docs/api/**/*.md* --config cspell/cSpell.json",
     "check:fmt": "prettier --check .",
+    "check-internal-links": "node -r esbuild-register scripts/commands/checkInternalLinks.ts",
+    "check-external-links": "node -r esbuild-register scripts/commands/checkExternalLinks.ts",
     "check-pages-render": "node -r esbuild-register scripts/commands/checkPagesRender.ts",
```
arnaucasau, line 14: Nit: What do you think about changing these three commands to follow the `check:` convention? I think it would be good given that they start with the word `check`, and we also have `check:metadata`, `check:spelling,` and `check:fmt`

scripts/commands/checkExternalLinks.ts
```diff
@@ -0,0 +1,78 @@
...
+
+interface Arguments {
+  [x: string]: unknown;
+  globs?: string[];
+}
+
+const readArgs = (): Arguments => {
+  return yargs(hideBin(process.argv))
+    .version(false)
+    .command("$0 [globs..]", "")
```
arnaucasau, line 29: Nice command! I didn't know this was possible, TIL

arnaucasau (approved):
Awesome work with the link checker Eric! I left two non-blocking comments. Tested the link checker locally and works well!

scripts/commands/checkExternalLinks.ts
Eric-Arellano: Same. The search phrase was "yargs variadic positional arguments"

PR #929, opened 2024-02-28
[do not merge] Preview Qiskit PR 11904

Eric-Arellano:
A preview of https://github.com/Qiskit/qiskit/pull/11904, which would use a new API structure that doesn't rely on `autosummary`.

### To preview locally

Run `./start` then go to http://localhost:3000/api/qiskit/dev/circuit

### To regenerate these docs

1. Download the Azure artifact from CI, or a local Sphinx build is fine too. Unpack the Azure artifact and then re-compress the contents of the folder (like `stubs/`) into a zip file called `1.1-dev-pr-11904.zip`
2. Upload to https://ibm.ent.box.com/folder/246867920686
3. `npm run gen-api -- -p qiskit -v 1.1.0-dev --dev` on this branch

### Screenshots

Screenshot 2024-02-28 at 4 36 16 PM

Screenshot 2024-02-28 at 4 36 37 PM

Screenshot 2024-02-28 at 4 36 50 PM


jakelishman:
Thanks very much for this, Eric.

The right-hand sidebars look like they've got a bug in them - they're showing methods in the local ToC but not the classes of those methods.  I don't know how the generation is going, but Sphinx will have populated its ToC entries with both the classes and the methods.

Fwiw, this multi-class structure isn't actually new - it's already live in the module-level docs of `qiskit.circuit.singleton`, `qiskit.qasm2` and `qiskit.circuit.classical`, which pre-dated the move of the documentation from qiskit.org to docs.quantum.ibm.com, so if the sidebar isn't behaving correctly here, it's probably also not behaving correctly on those pages.

Eric-Arellano:
Closing because the PR was merged and is now live on the dev docs at https://docs.quantum.ibm.com/api/qiskit/dev/circuit. We'll keep improving the API docs so this renders better.

Issue #845, opened 2024-02-19
Refactor `htmlToMarkdown.ts`


Eric-Arellano:
Split out of https://github.com/Qiskit/documentation/issues/223 and relates to https://github.com/Qiskit/documentation/issues/715.

To do: Split out the huge function into smaller helper functions & write more focused unit tests for those. Right now our unit tests are a bit hard to read and update since they have to mock so much.

PR #934, opened 2024-02-29
Refactor htmlToMd.ts

arnaucasau:
Part of #845

This is the first refactor of `htmlToMd.ts`. It brings the following changes:

- New function named `buildMathExpression` to build math expressions whether they are inlined or not
- New function named `removeEmphasisSpaces` to remove the initial and trailing spaces in an emphasis node
- The Unified plugin and the creation of the handlers for it were separated into two functions. The unified plugin was inlined into `generateMarkdownFile`

### DIFF
--- a/scripts/lib/api/htmlToMd.ts
+++ b/scripts/lib/api/htmlToMd.ts
@@ -77,7 +77,6 @@ async function generateMarkdownFile(
         }
         parent.children.splice(index + 1, nextIndex - (index + 1));
 
-        // remove initial and trailing spaces from emphasis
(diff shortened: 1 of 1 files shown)


scripts/lib/api/htmlToMd.ts
```diff
@@ -54,6 +54,40 @@ async function generateMarkdownFile(
   mainHtml: string,
   meta: Metadata,
 ): Promise<string> {
+  const handlers = prepareHandlers(meta);
+  const mdFile = await unified()
```
arnaucasau, line 58: Same logic except for the `removeEmphasisSpaces` calls

PR #940, opened 2024-02-29
Fix accidental deletion of GitHub source links for methods

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/517.

Turns out that we would accidentally delete all the GitHub links for methods because we were using `closest().remove()` rather than `.closest().first().remove()`. That means that we'd only have the link for the class definition and not individual methods, which defeats the point of precise source code links from https://github.com/Qiskit/documentation/issues/517.

## Historical API docs

However, the behavior to not show methods is actually quite sensible for our API docs, where we were using `sphinx.ext.viewcode` rather than `sphinx.ext.linkcode`. That meant that we could only link to the overall file, rather than specific line numbers. So, it's not very useful to keep linking to the overall file for every method since they will each have the same value and the class definition already has the value. So, we check if the link is for a method and proactively remove it if so.

This change to proactively remove method links changes some historical API doc versions, but not others. Why? It only removes the links if we were doing method inlining, where the method had a dedicated HTML page originally. That's because the `.remove()` line wouldn't impat the distinct HTML pages.


scripts/lib/api/processHtml.test.ts
```diff
@@ -432,14 +432,34 @@ describe("prepareGitHubLink()", () => {
     const doc = Doc.load(
       `<span class="pre">None</span><span class="sig-paren">)</span><a class="reference external" href="https://github.com/Qiskit/qiskit/blob/stable/1.0/qiskit/utils/deprecation.py#L24-L101"><span class="viewcode-link"><span class="pre">[source]</span></span></a><a class="headerlink" href="#qiskit_ibm_runtime.IBMBackend" title="Link to this definition">#</a>`,
     );
-    const result = prepareGitHubLink(doc.$, doc.$main);
+    const result = prepareGitHubLink(doc.$main, false);
     expect(result).toEqual(
-      `<a href="https://github.com/Qiskit/qiskit/blob/stable/1.0/qiskit/utils/deprecation.py#L24-L101" title="view source code">GitHub</a>`,
+      ` <a href="https://github.com/Qiskit/qiskit/blob/stable/1.0/qiskit/utils/deprecation.py#L24-L101" title="view source code">GitHub</a>`,
```
frankharkins: Is the extra space important? If so maybe a quick comment so we don't delete it at some point in the future

frankharkins (approved):
Thank you!

scripts/lib/api/processHtml.test.ts
Eric-Arellano: It is https://github.com/Qiskit/documentation/pull/915. I moved where the space is added as a refactor of #915.

PR #943, opened 2024-03-01
Refactor `htmlToMd.ts`: Split `prepareHandlers` function

arnaucasau:
Part of #845

This PR refactors the `htmlToMd.ts` script. It splits the `prepareHandlers` function into some helper functions to simplify the logic and improve the readability.

### DIFF
--- a/scripts/lib/api/htmlToMd.ts
+++ b/scripts/lib/api/htmlToMd.ts
@@ -16,13 +16,13 @@ import rehypeRemark from "rehype-remark";
 import remarkStringify from "remark-stringify";
 import remarkGfm from "remark-gfm";
 import { last, first, without, initial, tail } from "lodash";
-import { defaultHandlers, Handle, toMdast, all } from "hast-util-to-mdast";
+import { defaultHandlers, Handle, toMdast, all, H } from "hast-util-to-mdast";
 import { toText } from "hast-util-to-text";
 import remarkMath from "remark-math";
 import remarkMdx from "remark-mdx";
 import { MdxJsxFlowElement } from "mdast-util-mdx-jsx";
 import { visit } from "unist-util-visit";
-import { Emphasis, Root } from "mdast";
+import { Emphasis, Root, Content } from "mdast";
 
 import { processHtml } from "./processHtml";
 import { HtmlToMdResult } from "./HtmlToMdResult";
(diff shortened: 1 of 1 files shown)


scripts/lib/api/htmlToMd.ts
```diff
@@ -224,6 +180,73 @@ function removeEmphasisSpaces(
...
+    children: [
+      {
+        type: "paragraph",
+        children: otherChildren.map((node: any) => toMdast(node, { handlers })),
+      },
+    ],
+  });
+}
+
+function buildDt(
```
arnaucasau, line 229: Same logic as before but encapsulated in a helper function

scripts/lib/api/htmlToMd.ts
```diff
@@ -102,14 +102,14 @@ function prepareHandlers(meta: Metadata): Record<string, Handle> {
...
+        (node.properties.className?.includes("target") ||
+          node.children.length === 0)
+      ) {
         return [buildSpanId(node.properties.id), ...all(h, node)];
       }
 
-      if (node.properties.id && node.children.length === 0) {
-        return buildSpanId(node.properties.id);
-      }
-
```
arnaucasau, line 112: The difference between these two ifs was the call to all(h, node). This merge should be safe given that the `all` method transforms the children of a hast parent to mdast or returns an empty array if no children are found.

scripts/lib/api/htmlToMd.ts
```diff
@@ -224,6 +180,73 @@ function removeEmphasisSpaces(
...
+  return nodeList.find(
+    (child: any) => child.properties.className?.includes(propertyName),
+  );
+}
+
+function buildDiv(
+  childrenList: any[],
+  nodeClasses: string[],
+  handlers: Record<string, Handle>,
+  divType: "admonition" | "deprecated",
```
arnaucasau, line 193: Type defined to avoid calling the function with different divTypes. The logic of the function is the same but using a new helper function named findNodeWithProperty and merging the titleNode, title, and otherChildren definitions outside the ifs.

Eric-Arellano: After looking at this function for a while, I don't think I find it more readable because there is too much divergence between the two types:

* the `titleNode` search
* computing the `type`
* whether to call `toText` on the `title`, but I think that's actually unnecessary on line 211?

I think you could make this helper function worth it, though. Some suggestions:

* compute the `type` outside the helper function and pass it in
* explicitly pass in the title node to look for
* see if you can get rid of the `toText` in the return statement so that you can have one single `return`
* I'm trying to think of a more clear name than `buildDiv`, which isn't very precise. The best I'm thinking of is to rename `buildAdmonition` to be something more specific like `buildAdmonitionBase` and then change this to be `buildAdmonitionFromDiv`?

Issue #932, opened 2024-02-29
Module API pages should not include `h1` in their page table of contents


Eric-Arellano:
We hardcode that API docs always should show `h1 - h3` in the right table of contents, whereas the default is to only show `h2 - h3`.

https://github.com/Qiskit/documentation/blob/c7bbfb1b4e86b318c3512c555112b84d8bccf3cc/scripts/lib/api/addFrontMatter.ts#L34

We do this because it's useful for API class pages, where it's nice to have the class name show up in the right table of contents. That's useful as a way to quickly jump to the class definition because it is non-trivial, like including the constructor and class docstring.

<img width="241" alt="Screenshot 2024-02-29 at 11 08 07 AM" src="https://github.com/Qiskit/documentation/assets/14852634/8139e2d4-021c-4f7c-adc0-8752877375ae">

However, that is a bad default for module pages like https://docs.quantum.ibm.com/api/qiskit/exceptions where the page title is not very informative.

<img width="282" alt="Screenshot 2024-02-29 at 11 10 01 AM" src="https://github.com/Qiskit/documentation/assets/14852634/44a4869b-f648-4844-9d44-043164846d12">

Instead, for module pages, we should keep the default of h2 being the min level to show. We should also show h4 because these module pages often have nested information that is worth rendering, although we need to preview a few pages to see what this looks like and confirm it's a good heuristic.

Eric-Arellano:
All module pages with h4s:

```
❯ rg '#### ' -l | xargs rg -l 'python_api_type: module' | sort
docs/api/qiskit-ibm-provider/0.7/qiskit_ibm_provider.transpiler.passes.scheduling.md
docs/api/qiskit-ibm-provider/0.8/qiskit_ibm_provider.transpiler.passes.scheduling.md
docs/api/qiskit-ibm-provider/0.9/qiskit_ibm_provider.transpiler.passes.scheduling.md
docs/api/qiskit-ibm-provider/qiskit_ibm_provider.transpiler.passes.scheduling.md
docs/api/qiskit-ibm-runtime/0.18/qiskit_ibm_runtime.transpiler.passes.scheduling.md
docs/api/qiskit-ibm-runtime/0.19/qiskit_ibm_runtime.transpiler.passes.scheduling.md
docs/api/qiskit-ibm-runtime/dev/qiskit_ibm_runtime.transpiler.passes.scheduling.md
[...]

Issue #942, opened 2024-03-01
Initial audit of header hierarchy for all module API pages


Eric-Arellano:
Like https://github.com/Qiskit/qiskit/pull/11923, but for all the other modules. We want to look for a  few things:

* We don't jump from h1 to h3, which is an accessibility violation. When this happens, we should an h2 to the API doc
* There aren't any unintended heading mistakes like how in https://github.com/Qiskit/qiskit/pull/11923 we were making something a lower header level without realizing.

Note that we should never use more than one h1 on module pages; don't try to improve the header hierarchy that way.

Module pages can be identified by searching for `python_api_type: module`. This shouldn't be a problem on function and class API pages, which are much more auto-generated than human-written.

--

- [x] qiskit.assembler
- [x] qiskit.circuit
- [x] qiskit.circuit.classical
- [x] qiskit.circuit.classicalfunction
- [x] qiskit.circuit.library
- [x] qiskit.circuit.singleton
- [x] qiskit.compiler
- [x] qiskit.converters
- [x] qiskit.dagcircuit
- [x] qiskit.exceptions
- [x] qiskit.passmanager
- [x] qiskit.primitives
- [x] qiskit.providers
- [x] qiskit.providers.basic_provider
- [x] qiskit.providers.fake_provider
- [x] qiskit.providers.models
- [x] qiskit.pulse
- [x] qiskit.qasm2
- [x] qiskit.qasm3
- [x] qiskit.qobj
- [x] qiskit.qpy
- [x] qiskit.quantum_info
- [x] qiskit.result
- [x] qiskit.scheduler
- [x] qiskit.synthesis
- [x] qiskit.synthesis.unitary.aqc
- [x] qiskit.transpiler
- [x] qiskit.transpiler.passes
- [x] qiskit.transpiler.passes.synthesis.plugin
- [x] qiskit.transpiler.preset_passmanagers
- [x] qiskit.transpiler.preset_passmanagers.plugin
- [x] qiskit.utils
- [x] qiskit.visualization

PR #949, opened 2024-03-01
Don't show h1s in module page table of contents

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/932.

We decided:

1. We are going to keep showing h1s for class and function pages. It's useful because they have non-trivial code at the top level, like their docstring and constructor. It's nice to have the quick way to access the value.
2. For now at least, we won't show h4s in the module pages. There are 8 module pages in the current API docs that use h4s. I think showing h4s improved ~3 of them, and made ~5 worse. So, for now, let's not include them. This can be revisited, especially after https://github.com/Qiskit/documentation/issues/942. 

This regens all APIs but reverts dev docs, which have unrelated changes and will be regenerated in a dedicated PR.


Eric-Arellano:
Any opinion on whether we should simply not set the line `in_page_toc_min_heading_level` for modules since 2 is already the default? It makes zero difference for the docs app / for users. I figured always setting `in_page_toc_min_heading_level` was simpler code, and it also maybe makes more explicit to devs the difference between class pages & module pages.

arnaucasau (approved):
Thanks!

PR #951, opened 2024-03-04
Check for orphan pages

### LINKED ISSUE #388: Add automated check for orphan docs pages
An orphan page doesn't show up in `_toc.json`, so it's not apparent to users how to navigate to it. There are some legit orphan pages like `api/qiskit/release-notes/index.md`, but usually it's an oversight from us. So we should check for this automatically.

## Command line interface

I think we should have arguments like `checkMetadata.ts` has, where you can specify `--apis`, e.g. `npm run check:orphan-pages -- --apis`. It will also always check non-API docs.

For now, translations seem irrelevant since they should have the same structure as English docs. We can revisit this in the future when we work more on translations.

## How to add a new script

Add the file to `scripts/commands` like `checkOrphanPages.ts`. Look at `checkPagesRender.ts` as an example with how it uses `zxMain`.

Update `package.json` with a new command like `check:orphan-pages`, and have `check` run it too.

Update `.github/workflows/main.yml` to run `check:orphan-pages` and update `api-checks.yml` to run `check:orphan-pages -- --apis`.

Update the README with instructions; look at other sections for inspiration.

## Helpful libraries/APIs

Use `globby()` to discover the file paths of everything.

You need to read in the `_toc.json` files similar to this:

https://github.com/Qiskit/documentation/blob/abde486fd2a903fac5a46f5627623df6637b6c3f/scripts/commands/checkMetadata.ts#L49-L50

## Necessary: ignore list mechanism

Have a `Set<string>` where we hardcode a list of allowable orphan pages. If the file is in the ignore list, then skip checking it.

## Recommendation: check one folder at a time

We have _a ton_ of docs. Loading everything all at once in memory will be slow. Instead, process one folder at a time so that we reduce how much is loaded in memory.

Depending on how fast the overall script is, you'll either want to eagerly `console.error()` the failing folder right after processing it, or collate all the errors and write one single final error message.

kaelynj:
Close #388 

I ended up not using `zxMain` and set a few different options for choosing which api docs to look through (similar to `checkPagesRender.ts`).  @Eric-Arellano Let me know if there's anything that needs fixed or should be refactored!


package.json
```diff
@@ -5,13 +5,14 @@
   "author": "Qiskit Development Team",
   "license": "Apache-2.0",
   "scripts": {
-    "check": "npm run check:metadata && npm run check:spelling && npm run check:internal-links && npm run check:fmt",
+    "check": "npm run check:metadata && npm run check:orphan-pages && npm run check:spelling && npm run check:internal-links && npm run check:fmt",
```
Eric-Arellano, line 8: Nit: we want to order these by what's most likely to fail for a _technical writer_:

```suggestion
    "check": "npm run check:metadata && npm run check:spelling && npm run check:internal-links && npm run check:orphan-pages && npm run check:fmt",
```

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,170 @@
...
+
+// The steps to this script are:
+//  ✅ 1. Enter each folder within docs
+//  ✅ 2. Create a list of files in the _toc
+//  ✅ 3. Collect the list of files in directory
+//  ✅ 4. Check if each file in directory is in _toc
+
+// Needed features:
+//  ✅ - Check api docs if desired
+//   - Include command in .github/workflows/main.yml
```
Eric-Arellano, line 34: Leftover? It's usually better to have self-documenting code that describes the script flow, meaning that your `main` function is easy to read thanks to helper functions and friends.

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,170 @@
...
+      type: "boolean",
+      default: false,
+      description: "Check the links in the /dev API docs.",
+    })
+    .option("historical-apis", {
+      type: "boolean",
+      default: false,
+      description:
+        "Check the links in the historical API docs, e.g. `api/qiskit/0.44`. " +
+        "Warning: this is slow.",
```
Eric-Arellano, line 53: Do you know how slow? You can use `time npm run check:orphan-pages -- --historical-apis`.

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,170 @@
...
+    })
+    .parseSync();
+};
+
+async function main() {
+  const args = readArgs();
+
+  // DETERMINE THE TOC FILES TO CHECK AGAINST
```
Eric-Arellano, line 61: No need for comments like this. The code is self documenting. Ditto the other comments.

Also, with comments, you want to use Sentence case, rather than all caps.

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,170 @@
...
+    // LOAD JSON CONTENTS OF TOC FILE
+    console.log("Checking toc in:", tocFile);
+    const jsonFileContents = await fs.readFile(tocFile, "utf-8");
+    const children = JSON.parse(jsonFileContents).children;
+    const files: [string] = [];
+
+    // COLLECT ALL URLS REFERENCED IN TOC FILE
+    const fileContents = await collectTocFileContents(children);
+    const flatFileContents = flattenArray(fileContents);
+    flatFileContents[0] = flatFileContents[0] + "/index";
```
Eric-Arellano, line 78: This would be a good helper function: `getTocUrls(filePath: string): Set<string>`. It will make this `main` function easier to read.

The helper functions you already have are good. Keep those. This new helper function will still call those.

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,170 @@
...
+  // SET PASS/FAIL STATUS
+  let allGood = true;
+  const orphanPages = [];
+
+  for (tocFile of tocFiles) {
+    // LOAD JSON CONTENTS OF TOC FILE
+    console.log("Checking toc in:", tocFile);
+    const jsonFileContents = await fs.readFile(tocFile, "utf-8");
+    const children = JSON.parse(jsonFileContents).children;
+    const files: [string] = [];
```
Eric-Arellano, line 73: Better is a `Set<string>`. That's because you're going to repeatedly call `.includes()` on it to see if a file is included in the collection. In an array, that's O(n). In a set, that's O(1).

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,170 @@
...
+  const fileList = await globby(fileGlob);
+  const processedFileList = fileList.map((fileName) =>
+    fileName.replace("docs", "").replace(".mdx", "").replace(".ipynb", ""),
+  );
+  return processedFileList;
+}
+
+// Collect directories to look through
+async function determineTocFiles(args: Arguments): Promise<string[]> {
+  const globs = ["docs/**/_toc.json", "!./docs/api/**"];
```
Eric-Arellano, line 116:
```suggestion
  const globs = ["docs/**/_toc.json", "!docs/api/**"];
```

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+    // Just add url if there's no children
+    else {
+      urls.push(child.url);
+    }
+  }
+  return urls;
+}
+
+function flattenArray(urlArray: [string[]]): [string[]] {
```
Eric-Arellano, line 151: Nice use of `.reduce()`! Simpler is to use Lodash to do this for us: https://lodash.com/docs/#flattenDeep. You can remove this function and replace its call with Lodash.

Lodash is a great library of utilities like this.

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,170 @@
...
+    if (Array.isArray(childElement)) {
+      flattenedArray.push(...flattenArray(childElement));
+    } else {
+      flattenedArray.push(childElement);
+    }
+    return flattenedArray;
+  }, []);
+}
+
+function isIterable(obj) {
```
Eric-Arellano, line 162: Appears unused.

Eric-Arellano (commented):
Thanks, Kaelyn!

scripts/commands/checkOrphanPages.ts
kaelynj: Not actually slow, I just copied over from `checkPagesRender.ts` and forgot to update the description.

kaelynj:
@Eric-Arellano, just implemented your suggestions.  Let me know if there's anything else before it's ready for approval

scripts/commands/checkOrphanPages.ts
kaelynj: Thanks!  Recursion always ends up being trickier than I expect for me, but glad there's a much easier library to call from.

.github/workflows/main.yml
```diff
@@ -30,6 +30,8 @@ jobs:
         run: npm run check:metadata
       - name: Spellcheck
         run: npm run check:spelling
+      - name: Check orphaned pages
+        run: npm run check:orphan-pages
```
Eric-Arellano, line 34: Nit: please put this after the internal link checker. We want to order these by the likelihood of failure so that you get faster CI feedback.

README.md
```diff
@@ -209,6 +209,24 @@ npm run check:external-links -- docs/run/index.md docs/run/circuit-execution.mdx
...
+```bash
+# Only check non-API docs
+npm run check:orphan-pages
+
+# You can also add any of the below arguments to check API docs
+npm run check:orphan-pages -- --current-apis --dev-apis --historical-apis
+
+# Or, run all the checks.  However this will skip the API docs
+npm run check
+
```
Eric-Arellano, line 227: Nit:

```suggestion
```

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,147 @@
...
+interface Arguments {
+  [x: string]: unknown;
+  currentApis: boolean;
+  devApis: boolean;
+  historicalApis: boolean;
+}
+
+const ALLOWED_ORPHANS = new Set([
+  "/api/qiskit/qiskit.primitives.BaseEstimator",
+  "/api/qiskit/qiskit.primitives.BaseSampler",
```
Eric-Arellano, line 29: CI is failing because `/run/circuit-execution` is not found.

Maybe call the constant `ALLOWED_ORPHAN_URLS` to make clear that they're not file paths, like not `docs/run/circuit-execution.ipynb`.

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,147 @@
...
+  const args = readArgs();
+
+  const tocFiles = await determineTocFiles(args);
+
+  let allGood = true;
+  const orphanPages = [];
+
+  for (tocFile of tocFiles) {
+    tocUrls = await getTocUrls(tocFile);
+    const dir = tocFile.replace("_toc.json", "");
```
Eric-Arellano, line 64: More clear is to use `path.dirname(tocFile)`. You'll need to `import path from "path";`.

This code works, but it's good to use `path` to be more declarative rather than imperative. Meaning, your code says _what_ it's doing, rather than _how_ it's doing it.

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,147 @@
...
+async function main() {
+  const args = readArgs();
+
+  const tocFiles = await determineTocFiles(args);
+
+  let allGood = true;
+  const orphanPages = [];
+
+  for (tocFile of tocFiles) {
+    tocUrls = await getTocUrls(tocFile);
```
Eric-Arellano, line 63:
```suggestion
    const tocUrls = await getTocUrls(tocFile);
```

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,147 @@
...
+    const dir = tocFile.replace("_toc.json", "");
+    const dirFiles = await collectDirFiles(dir);
```
Eric-Arellano, line 65: `file` is a misnomer because file normally refers to a path like `docs/run/index.mdx`. Instead, these are URLs. But the key difference from `tocUrls` is that these are the URLs that are valid, or that actually exist. So you could call this something like `existingUrls` or `actualUrls`. You'd want to rename the helper function too.

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,147 @@
...
+    for (file of dirFiles) {
+      if (!tocUrls.has(file) && !ALLOWED_ORPHANS.has(file)) {
+        allGood = false;
+        orphanPages.push(file);
+      }
+    }
```
Eric-Arellano, line 71: You can make this code more declarative and concise by using `.filter()`. 

```suggestion
    orphanPages.push(...dirFiles.filter((file) => !tocUrls.has(file) && !ALLLOWED_ORPHANS.has(file))
```

--

You can remove the `allGood` variable. Instead, on line 73, check if `orphanPages.length > 0`.

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,147 @@
...
+  }
+  if (!allGood) {
+    console.error(
+      "\n There's some orphaned pages!  These files need a home: \n",
```
Eric-Arellano, line 75: Grammar nit:

```suggestion
      "\n There are some orphaned pages! These files need a home: \n",
```

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,147 @@
...
+      orphanPages,
```
Eric-Arellano, line 76:
```suggestion
      orphanPages.join("\n"),
```

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,147 @@
...
+    );
+    process.exit(1);
+  }
+  console.log("\nNo orphan pages found ✅\n");
+}
+
+async function getTocUrls(filePath: string): Set<string> {
+  console.log("Checking toc in:", tocFile);
```
Eric-Arellano, line 84: Nit: I recommend moving this log back to the `main` function so that the control flow is more clear. This log is referring to the entire process of checking a folder, whereas this function is solely about reading the single toc.json as step 1. Since the log refers to the entire process, it's better to live at that abstraction level rather than hidden inside a single step.

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,147 @@
...
+  const jsonFileContents = await fs.readFile(tocFile, "utf-8");
+  const children = JSON.parse(jsonFileContents).children;
+  const files: [string] = [];
```
Eric-Arellano, line 87: Unused

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  const fileContents = await collectTocFileContents(children);
+  const flatFileContents = flattenDeep(fileContents);
+  const indexPageUrl = flatFileContents[0] + "/index";
+  flatFileContents.push(indexPageUrl);
```
Eric-Arellano, line 92: 1. you can use string interpolation for more concise code
2. imo it's fine to inline this since the code isn't very complex, but feel free to disagree

```suggestion
  flatFileContents.push(`${flatFileContents[0]}/index`);
```

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,147 @@
...
+  const tocUrlSet = new Set(flatFileContents);
+
+  return tocUrlSet;
```
Eric-Arellano, line 95: No need for an intermediate variable, which adds extra noise

```suggestion
  return new Set(flatFileContents);
```

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+async function collectDirFiles(directory: string): Promise<string[]> {
+  const fileGlob = [
+    directory + "/*.md",
+    directory + "/*.mdx",
+    directory + "/*.ipynb",
+  ];
+  const fileList = await globby(fileGlob);
```
Eric-Arellano, line 104:
```suggestion
  const fileList = await globby([`${directory}/*.{md,mdx,ipynb}`]);
```

scripts/commands/checkOrphanPages.ts
```diff
@@ -0,0 +1,147 @@
...
+  const processedFileList = fileList.map((fileName) =>
+    fileName
+      .replace("docs", "")
+      .replace(".mdx", "")
+      .replace(".ipynb", "")
+      .replace(".md", ""),
+  );
+  return processedFileList;
```
Eric-Arellano, line 112: Nice use of map!

```suggestion
  return fileList.map((fileName) =>
    fileName
      .replace("docs", "")
      .replace(".mdx", "")
      .replace(".ipynb", "")
      .replace(".md", ""),
  );
```

Eric-Arellano (commented):
Nice! This is looking really good.

Eric-Arellano (approved):

Issue #715, opened 2024-01-26
Add integration test for API docs based on qiskit-sphinx-theme


Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/223.

The designers have found it really useful to see how our app renders the example docs from qiskit-sphinx-theme because they have a comparison to qiskit.org. See https://github.com/Qiskit/documentation/pull/706.

Whereas that PR #706 tried adding the example docs as a legit project, that's not necessary; the HTML doesn't ever change. Instead, we should save the HTML input as testdata and have integration tests for how `htmlToMd.ts` handles it. We don't need to test the whole pipeline imo, only the Sphinx conversion. This should use snapshot testing (not inline).

When adding this, we should audit which tests can be deleted in `htmlToMd.test.ts` - those already are more or less integration tests, but some have been replaced by `processHtml.test.ts` and imo those tests are too noisy.

## Bonus: allow previewing the built docs

It'd be really neat if we could look at the example docs in the preview with `./start`. To do that, you'd volume mount `./start` so that the files are loaded into something like the folder `docs/example-api-docs`.

PR #952, opened 2024-03-04
Split out `conversionPipeline.ts` from `updateApiDocs.ts`

Eric-Arellano:
This is a refactor to unblock the new integration test in https://github.com/Qiskit/documentation/issues/715. We need to test the conversion pipeline, so its code needs to live in `lib` not `commands`.

I renamed the function `convertHtmlToMarkdown` to instead `runConversionPipeline` and moved it to `conversionPipeline.ts`. I didn't want the phrase "html to markdown" because because we already have `htmlToMd.ts`; this pipeline does way more than convert HTML to markdown, such as also copying image files and inlining methods into classes.

This PR also refactors the rest of `updateApiDocs.ts` via helper functions. A follow up will refactor `conversionPipeline.ts`.

### DIFF
--- a/scripts/commands/updateApiDocs.ts
+++ b/scripts/commands/updateApiDocs.ts
@@ -10,37 +10,15 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
-import { join, parse, relative } from "path";
-import { readFile, writeFile } from "fs/promises";
-
-import { globby } from "globby";
-import { uniq, uniqBy } from "lodash";
 import { mkdirp } from "mkdirp";
 import yargs from "yargs/yargs";
 import { hideBin } from "yargs/helpers";
-import transformLinks from "transform-markdown-links";
 
-import { ObjectsInv } from "../lib/api/objectsInv";
-import { sphinxHtmlToMarkdown } from "../lib/api/htmlToMd";
-import { saveImages } from "../lib/api/saveImages";
-import { generateToc } from "../lib/api/generateToc";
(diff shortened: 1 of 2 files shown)


scripts/commands/updateApiDocs.ts
```diff
@@ -89,17 +67,37 @@ const readArgs = (): Arguments => {
 zxMain(async () => {
   const args = readArgs();
 
-  // Determine the minor version, e.g. 0.44.0 -> 0.44
-  const versionMatch = args.version.match(/^(\d+\.\d+)/);
-  if (versionMatch === null) {
+  if (args.historical && args.dev) {
     throw new Error(
-      `Invalid --version. Expected the format 0.44.0, but received ${args.version}`,
+      `${args.package} ${args.version} cannot be historical and dev at the same time. Please remove at least only one of these two arguments: --historical, --dev.`,
```
Eric-Arellano, line 72: The diff is weird on this. It's the same check as before, only moved up.

scripts/lib/api/conversionPipeline.ts
```diff
@@ -0,0 +1,184 @@
...
+import { Pkg } from "./Pkg";
+import { pathExists, getRoot } from "../fs";
+import {
+  addNewReleaseNotes,
+  generateReleaseNotesIndex,
+  updateHistoricalTocFiles,
+  writeReleaseNoteForVersion,
+} from "./releaseNotes";
+
+export async function runConversionPipeline(
```
Eric-Arellano, line 41: No changes made to this function other than renaming it.

PR #953, opened 2024-03-04
Refactor release notes logic in `conversionPipeline.ts` 

Eric-Arellano:
Uses helper functions and better groups code to make it easier to read. Should be the same behavior.

### DIFF
--- a/scripts/lib/api/conversionPipeline.ts
+++ b/scripts/lib/api/conversionPipeline.ts
@@ -22,7 +22,7 @@ import { ObjectsInv } from "./objectsInv";
 import { sphinxHtmlToMarkdown } from "./htmlToMd";
 import { saveImages } from "./saveImages";
 import { generateToc } from "./generateToc";
-import { HtmlToMdResult } from "./HtmlToMdResult";
+import { HtmlToMdResultWithUrl } from "./HtmlToMdResult";
 import { mergeClassMembers } from "./mergeClassMembers";
 import flattenFolders from "./flattenFolders";
 import { updateLinks } from "./updateLinks";
@@ -58,7 +58,7 @@ export async function runConversionPipeline(
     },
   );
 
-  let results: Array<HtmlToMdResult & { url: string }> = [];
+  let results: Array<HtmlToMdResultWithUrl> = [];
   for (const file of files) {
     const html = await readFile(join(htmlPath, file), "utf-8");
     const result = await sphinxHtmlToMarkdown({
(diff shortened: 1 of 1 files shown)


scripts/lib/api/conversionPipeline.ts
```diff
@@ -78,14 +78,6 @@ export async function runConversionPipeline(
...
     let url = `/${relative(`${getRoot()}/docs`, dir)}/${name}`;
     results.push({ ...result, url });
-
-    if (
-      pkg.isLatest() &&
-      pkg.hasSeparateReleaseNotes &&
-      file.endsWith("release_notes.html")
-    ) {
-      addNewReleaseNotes(pkg);
-    }
```
Eric-Arellano: This wasn't necessary in the for loop. We only need to call it before we call `generateReleaseNotesIndex`. So I moved it to live near that.

scripts/lib/api/conversionPipeline.ts
```diff
@@ -153,17 +118,7 @@ export async function runConversionPipeline(
     JSON.stringify(toc, null, 2) + "\n",
   );
 
-  // Add the new release entry to the _toc.json for all historical API versions.
-  // We don't need to add any entries in projects with a single release notes file.
-  if (pkg.isLatest() && pkg.hasSeparateReleaseNotes) {
```
Eric-Arellano: Note that this condition is the same as 162 and also 82-86 from up above.

scripts/lib/api/conversionPipeline.ts
```diff
@@ -0,0 +1,194 @@
...
+  await dedupeHtmlIdsFromResults(results);
+  addFrontMatter(results, pkg);
+
+  await maybeObjectsInv?.write(pkg.outputDir("public"));
+  for (const result of results) {
+    let path = urlToPath(result.url);
+
+    if (path.endsWith("release-notes.md")) {
+      const shouldUse = await handleReleaseNotesFile(result, pkg);
+      if (!shouldUse) continue;
```
arnaucasau, line 108: Nit: What about changing the `shouldUse` name for something like `shouldWrite` or `writeSingleFile`? I think it will make it clear that we are checking whether to write a single release note file or not in the `writeFile` call of line 111

arnaucasau (approved):
Thanks Eric! Looks good to me. I like how you were able to merge different code that has as condition `pkg.isLatest() && pkg.hasSeparateReleaseNotes`. The diff on GitHub will be reduced after merging #952.

The behavior of the script is the same. I tested it by regenerating the following versions:

- Qiskit 1.0.1
- Qiskit 0.45.3
- Qiskit 0.33.1
- All versions of runtime
- All versions of provider

PR #954, opened 2024-03-04
Split `processMembersAndSetMeta` function into helper functions

arnaucasau:
## Summary

This PR splits the `processMembersAndSetMeta` function from the `processHtml.ts` script into several helper functions to process each `apiType`.

## Changes

These changes are done to the `map()` used to create the `replacement` variable (line 291). The map used to have all the logic to process each child depending on its `apiType`. Now the map starts by early returning in the case we are not processing a `<dt>` tag and calling to a new helper function to process an individual child.

Before the `map()` was setting the metadata, the refactor moved the code outside of it, given that the metadata doesn't depend on the elements iterated.

The new helper function is called `processMember` and starts by preparing the GitHub source link, removing the `apiType` from the `em.property` selector, and calling other new helper functions to process each of our available `apiType`.

All the process `apiType` functions have the same code as before but `processAttribute`, which has been refactored to find the name, type, and default value of an attribute  in a more concise way

### DIFF
--- a/scripts/lib/api/processHtml.ts
+++ b/scripts/lib/api/processHtml.ts
@@ -279,24 +279,23 @@ export function processMembersAndSetMeta(
     }
 
     const $dl = $(dl);
+    const id = $dl.find("dt").attr("id") || "";
+    const apiType = getApiType($dl);
+    const priorApiType = meta.apiType;
+
+    if (!priorApiType) {
+      meta.apiType = apiType;
+      meta.apiName = id;
+    }
+
     const replacement = $dl
       .children()
       .toArray()
       .map((child) => {
         const $child = $(child);
-        const id = $dl.find("dt").attr("id") || "";
-        const apiType = getApiType($dl);
-
         if (child.name !== "dt" || !apiType) {
           return `<div>${$child.html()}</div>`;
         }
-
-        const priorApiType = meta.apiType;
-        if (!priorApiType) {
-          meta.apiType = apiType;
-          meta.apiName = id;
-        }
-
         return processMember($, $main, $child, $dl, priorApiType, apiType, id);
       })
       .join("\n");
--- a/scripts/lib/api/processHtml.ts
+++ b/scripts/lib/api/processHtml.ts
@@ -293,10 +293,24 @@ export function processMembersAndSetMeta(
       .toArray()
       .map((child) => {
         const $child = $(child);
+        const githubSourceLink = prepareGitHubLink(
+          $child,
+          apiType === "method",
+        );
+
         if (child.name !== "dt" || !apiType) {
           return `<div>${$child.html()}</div>`;
         }
-        return processMember($, $main, $child, $dl, priorApiType, apiType, id);
+        return processMember(
+          $,
+          $main,
+          $child,
+          $dl,
+          priorApiType,
+          apiType,
+          id,
+          githubSourceLink,
+        );
       })
       .join("\n");
 
@@ -312,9 +326,8 @@ function processMember(
   priorApiType: string | undefined,
   apiType: string,
   id: string,
+  githubSourceLink: string,
 ) {
-  const githubSourceLink = prepareGitHubLink($child, apiType === "method");
-
   findByText($, $main, "em.property", apiType).remove();
 
   if (apiType == "class") {
--- a/scripts/lib/api/processHtml.ts
+++ b/scripts/lib/api/processHtml.ts
@@ -314,98 +314,138 @@ function processMember(
   apiType: string,
   id: string,
 ) {
-  const github = prepareGitHubLink($child, apiType === "method");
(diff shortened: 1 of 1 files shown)


scripts/lib/api/processHtml.ts
```diff
@@ -299,107 +297,155 @@ export function processMembersAndSetMeta(
...
+  if (colonIndex == -1) {
+    colonIndex = text.length;
+  }
+
+  // The attributes have the following shape: name [: type] [= value]
+  const name = text.slice(0, Math.min(colonIndex, equalIndex)).trim();
+  const type = text
+    .slice(Math.min(colonIndex + 1, equalIndex), equalIndex)
+    .trim();
+  const value = text.slice(equalIndex, text.length).trim();
```
arnaucasau, line 421: Different logic than before

scripts/lib/api/processHtml.ts
```diff
@@ -299,107 +297,155 @@ export function processMembersAndSetMeta(
...
-              $(`<h3>${getLastPartFromFullIdentifier(id)}</h3>`).insertBefore(
-                $dl,
-              );
-            }
-          }
-
-          findByText($, $main, "em.property", "method").remove();
-          return `<span class="target" id='${id}'/><p><code>${$child.html()}</code>${github}</p>`;
-        }
+  findByText($, $main, "em.property", apiType).remove();
```
arnaucasau, line 319: This call was common in all the `apiType` helper functions and I extracted it in the `processMember` function instead.

scripts/lib/api/processHtml.ts
```diff
@@ -287,8 +287,6 @@ export function processMembersAndSetMeta(
         const id = $dl.find("dt").attr("id") || "";
         const apiType = getApiType($dl);
 
-        const github = prepareGitHubLink($child, apiType === "method");
-
         if (child.name !== "dt" || !apiType) {
           return `<div>${$child.html()}</div>`;
         }
```
arnaucasau, line 292: Reversing the order of the `prepareGitHubLinks` call, and the if statement adds a bunch of new GitHub source links that we didn't have before. I'm not sure why yet, but the change seems desirable

arnaucasau (commented):
The diff on GitHub might seem a little weird because of the new helper functions

scripts/lib/api/processHtml.ts
Eric-Arellano: Ah. `prepareGitHubLink` calls `remove()` on the closest `<a>`. So it's removing the link for a child element when it shouldn't be. This would actually be good to have in a dedicated PR since it's a behavior change, and so that we can document it better in the PR description etc. No need to run `npm run regen-apis`. Instead, create a new branch based on this one, then interactive rebase to drop your first two commits and add a new one with the behavior change and no refactors.

scripts/lib/api/processHtml.ts
Eric-Arellano: I was going to ask if you're confident it's safe to do this because you're changing the sequence: often this call was at the end, not beginning. But looks like the API generation behaves the same, so I presume it's fine.

scripts/lib/api/processHtml.ts
```diff
@@ -279,24 +279,23 @@ export function processMembersAndSetMeta(
     }
 
     const $dl = $(dl);
+    const id = $dl.find("dt").attr("id") || "";
```
Eric-Arellano, line 282: Excellent commit. This is really helpful to better understand what is child-specific vs from the parent.

Eric-Arellano (commented):
Great improvement!

scripts/lib/api/processHtml.ts
arnaucasau: Yes. At the beginning, I was wondering whether we were able to move it at the start of the function and it seems the script behaves the same. That line of code only removes the `apiType` from the beginning of each header, so it shouldn't be important for any other step

scripts/lib/api/processHtml.ts
arnaucasau: Ahh good spot! What do you think about making that change after merging this PR in a follow-up? By calling `prepareGitHubLink` inside of each child's `process` function, we end up with the same behavior as before plus we fixed some links

Eric-Arellano: That's fine with me to be in a follow up, but then you should revert the docs regen in this PR. It's important that `npm run regen-apis` doesn't result in changes when run on `main`, since that messes up other people making docs improvements.

arnaucasau: Done! This PR only refactors the code without changing the order of `prepareGitHubLinks`. In a follow-up, I'll add the change and the new links in the docs.

Eric-Arellano (approved):
Thanks!

PR #957, opened 2024-03-04
Split up conversionPipeline.ts into helper functions

Eric-Arellano:
Refactor for https://github.com/Qiskit/documentation/issues/715. Functionality is the same and regenerating provider didn't change.

### DIFF
--- a/scripts/lib/api/conversionPipeline.ts
+++ b/scripts/lib/api/conversionPipeline.ts
@@ -43,6 +43,33 @@ export async function runConversionPipeline(
   markdownPath: string,
   pkg: Pkg,
 ) {
+  const [files, maybeObjectsInv] = await determineFilePaths(htmlPath, pkg);
+  let initialResults = await convertFilesToMarkdown(
+    pkg,
+    htmlPath,
+    markdownPath,
+    files,
+  );
+
+  await createNecessaryDirs(initialResults);
+  await copyImages(pkg, htmlPath, initialResults);
+
+  const results = await postProcessResults(
+    pkg,
+    maybeObjectsInv,
+    initialResults,
+  );
+  await writeMarkdownResults(pkg, results);
+  await maybeObjectsInv?.write(pkg.outputDir("public"));
+  await writeTocFile(pkg, markdownPath, results);
+  await writeVersionFile(pkg, markdownPath);
+  await maybeUpdateReleaseNotesFolder(pkg, markdownPath);
+}
+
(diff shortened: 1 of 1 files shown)


scripts/lib/api/conversionPipeline.ts
```diff
@@ -43,6 +43,33 @@ export async function runConversionPipeline(
...
 ) {
+  const [files, maybeObjectsInv] = await determineFilePaths(htmlPath, pkg);
+  let initialResults = await convertFilesToMarkdown(
+    pkg,
+    htmlPath,
+    markdownPath,
+    files,
+  );
+
+  await createNecessaryDirs(initialResults);
```
Eric-Arellano, line 54: I don't think this is actually necessary. But I didn't want to make semantic changes in this PR. I'll try to remove it in a follow up.

scripts/lib/api/conversionPipeline.ts
```diff
@@ -43,6 +43,33 @@ export async function runConversionPipeline(
...
+  await copyImages(pkg, htmlPath, initialResults);
```
Eric-Arellano, line 55: The copy used to happen at the very end. But the copy doesn't depend on anything else, so I moved it higher.

I think technically `copyImages` needs to happen before `postProcessResults` because `flattenFolders` will result in losing image paths. I don't like that dependency; it's not clear. I plan to fix this in a follow up by changing `flattenFolders` to correctly preserve image metadata.

scripts/lib/api/conversionPipeline.ts
```diff
@@ -79,27 +114,53 @@ export async function runConversionPipeline(
...
   }
+}
+
+async function copyImages(
+  pkg: Pkg,
+  htmlPath: string,
+  results: HtmlToMdResultWithUrl[],
+): Promise<void> {
+  // Some historical versions don't have the `_images` folder in the artifact store in Box (https://ibm.ent.box.com/folder/246867452622)
+  if (pkg.isHistorical() && !(await pathExists(`${htmlPath}/_images`))) return;
```
Eric-Arellano, line 137: Early returning here is new. We used to have this conditional after `const allImages` and before `await saveImages`. No need to compute `allImages` first.

frankharkins (approved):
Thanks!

Each change seems reasonable and I tried regenerating the current Qiskit API docs and it showed no changes (except to `objects.inv`, but I believe that's just due to the compression).

PR #962, opened 2024-03-05
Add .gitattributes and guide for diffing objects.inv

frankharkins:
This PR adds instructions on setting up Git to use [`sphobjinv`](https://github.com/bskinn/sphobjinv) to display diffs for `objects.inv`.

Since `objects.inv` is compressed, we can't review changes through `git diff`. Git _does_ tell you if the file has changed, but this isn't that helpful as the compressed file can be different even if the uncompressed contents are the same.

Git will default back to binary diff if it doesn't find a rule in `.gitconfig` so adding `.gitattributes` shouldn't disrupt contributors that don't follow the guide.

Hopefully this information will help keep more eyes on the file and spot any unintended changes.

Here's an example of it in action (downgrading qiskit to 0.46):
Screenshot 2024-03-05 at 14 20 04

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+objects.inv diff=objects_inv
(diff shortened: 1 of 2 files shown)


Eric-Arellano (approved):
Nice! Thanks.

PR #964, opened 2024-03-05
Fix incorrect removal of GitHub source links by `prepareGitHubLink`

arnaucasau:
The `prepareGitHubLink` function calls `remove() `on the closest `<a>`. This can be a problem when used before the following early return:
```ts
if (child.name !== "dt" || !apiType) {
    return `<div>${$child.html()}</div>`;
}
```
Calling the function in a case where we should enter into the early return leads us to remove the link for a child element when we shouldn't. To fix this, this PR moves the `prepareGitHubLink` call to be performed inside the `processMember` function.

The PR also regenerates all versions of qiskit, runtime, and provider to add some links we were incorrectly removing.


Eric-Arellano (approved):
Great

PR #965, opened 2024-03-05
Improve Qiskit release notes ToC settings

### LINKED ISSUE #941: Improve page table of contents settings for Qiskit release notes
We set that release notes should only show h2 headers:

https://github.com/Qiskit/documentation/blob/8ca657f0c030a2de72f44ca33bd2ce34e79f2736/scripts/lib/api/addFrontMatter.ts#L41-L52

(Note that `in_page_toc_min_heading_level` is set to 2 by default)

That setting to only show h2 works great for Runtime and Provider. Their release notes are very small in each release, and they have a lot of releases all on the same page.

But this is a bad setting for Qiskit, which has a dedicated release note per version:

* For Qiskit <=0.44 with its metapackage, we should be setting `in_page_toc_max_heading_level` to 4
* For Qiskit 0.45+, we should be using the default of `in_page_toc_max_heading_level` being 3, rather than setting it to 2.
    - Why is the setting only being applied for 0.45.md and not 0.46 and 1.0?

arnaucasau:
Closes #941

This PR sets the `in_page_toc_max_heading_level` to 4 for legacy versions of qiskit (< 0.45) and to 3 for 0.45+ versions. It also changes the description of all the release notes to match the one we are using in the API generation script: `Changes made in Qiskit {version}`

Given that we don't regenerate legacy release notes using the script, we can set the `in_page_toc_max_heading_level` to 3 for all qiskit versions and 2 for the rest of the APIs.

The changes to the latest release note were done by removing the file and regenerating the 1.0.1 version. The rest of the files we manually updated.

### DIFF
--- a/docs/api/qiskit/release-notes/0.10.md
+++ b/docs/api/qiskit/release-notes/0.10.md
@@ -1,6 +1,7 @@
 ---
 title: Qiskit 0.10 release notes
-description: New features and bug fixes
+description: Changes made in Qiskit 0.10
+in_page_toc_max_heading_level: 4
 ---
 
 # Qiskit 0.10 release notes
--- a/docs/api/qiskit/release-notes/0.11.md
+++ b/docs/api/qiskit/release-notes/0.11.md
@@ -1,6 +1,7 @@
 ---
 title: Qiskit 0.11 release notes
-description: New features and bug fixes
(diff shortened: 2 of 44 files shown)


Eric-Arellano (approved):
Nice. Thanks

Issue #841, opened 2024-02-19
[ NEW PAGE ] Visualising results


javabster:
Note: on hold until we figure out where in docs to put this and overall patterns structure

content to include:

```[tasklist]
### Tasks
- [ ] marginal counts
- [ ] move stuff from [Visualize a circuit](https://docs.quantum-computing.ibm.com/build/circuit-visualization) page that talks about result visualisation (just leave circuit visualisation in that other page)
```

frankharkins:
Closed in #977 (continuation of Nick's PR)

PR #971, opened 2024-03-05
split visualizations page into three

nbronn:
The circuit visualizations are most valuable in the build stage of the circuit. This splits the results visualization into a separate file for the analysis/post-processing stage. Additionally, it makes sense to split the quantum information visualizations into the verify (or learning) step.


nbronn:
Fixes Issue #841

frankharkins (commented):
Thanks @nbronn!

Should the `analyze` folder live under `docs/`?

It also looks like you're missing an `Aer` import.

frankharkins:
Closing in favour of #977

PR #977, opened 2024-03-06
Split visualization page into three

frankharkins:
Continuation of #971

> The circuit visualizations are most valuable in the build stage of the circuit. This splits the results visualization into a separate file for the analysis/post-processing stage. Additionally, it makes sense to split the quantum information visualizations into the verify (or learning) step.

### Preview quick links

- [analyze/visualize-results](https://qiskit-docs-preview-pr-977.1799mxdls7qz.us-south.codeengine.appdomain.cloud/analyze/visualize-results)
- [build-new/circuit-visualization](https://qiskit-docs-preview-pr-977.1799mxdls7qz.us-south.codeengine.appdomain.cloud/build-new/circuit-visualization)
- [build-new/plot-quantum-states](https://qiskit-docs-preview-pr-977.1799mxdls7qz.us-south.codeengine.appdomain.cloud/build-new/plot-quantum-states)


docs/analyze/visualizing-results.ipynb
javabster, line 1: I think we tend to use "visualize" instead of "visualising" for titles

Eric-Arellano: > visualize
> visualising

The USA has corrupted you @javabster 😔

frankharkins: What about "Vizualiseing"?

frankharkins: Thoughts on the other two? Should they be `visualize-circuits` and `visualize-quantum-info`?

docs/build-new/quantum-info-visualizations.ipynb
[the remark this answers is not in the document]
javabster: Could you expand on the definitions of the state city, hinton, pauli vector, qsphere and mulitivector? these graphs are typically the least understood by our users and would be good to give a more thorough explanation rather than just redirecting them to the api reference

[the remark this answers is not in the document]
javabster: I think the bloch vector paragraph should come before the multi-qubit explanations, as it feels more natural to first learn the single qubit visualisation before the mutliqubit one. Also I think there is more we could demonstrate here than the current example (i think you can even pass a circuit as well? maybe that would be a good example for the circuit vis page...)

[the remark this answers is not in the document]
javabster: In general I think this page will be super useful for users, but currently it feels like it's not providing enough context/examples to be as useful as it could be, compared to the other visualisation pages that feel more fleshed out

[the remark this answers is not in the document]
frankharkins: Thanks! I agree with the other points and have incorporated them. RE Bloch plot:

> I think the bloch vector paragraph should come before the multi-qubit explanations, as it feels more natural to first learn the single qubit visualisation before the mutliqubit one. Also I think there is more we could demonstrate here than the current example (i think you can even pass a circuit as well? maybe that would be a good example for the circuit vis page...)

The `plot_bloch_vector` is a bit weird as it doesn't accept a quantum state like the others do. Instead, it needs 3D coordinates. If you want to plot a single-qubit state, you're actually still best off using `plot_bloch_multivector`.

But I do agree with your point in the sense that a user might not understand that from skimming the page. Reflecting on it, I think it might be better to remove that section and just dedicate the page to state visualizations. I don't really understand the use-case for the `plot_bloch_vector` function.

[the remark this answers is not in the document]
frankharkins: I've gone ahead and done that in 250a0ae80590220d6c9627f8b1e227b0fbcc7bc1, can always revert if you want to include that function.

docs/analyze/visualize-results.ipynb
```diff
@@ -0,0 +1,166 @@
...
+   "cell_type": "markdown",
+   "id": "7373f30c-a034-4abf-9dd9-71d74c8b296e",
+   "metadata": {},
+   "source": [
+    "## Plot histogram <a name='histogram'></a>\n",
+    "\n",
+    "The `plot_histogram` function visualizes the result of sampling a quantum circuit on a system or simulator.\n",
+    "\n",
+    "<Admonition title=\"Using the output from functions\" type=\"tip\">\n",
+    "This function returns a `matplotlib.Figure` object. When the last line of a code cell outputs these objects, Jupyter notebooks display them below the cell. If you call these functions in some other environments or in scripts then you will need to explicitly show or save the outputs.\n",
```
abbycross, line 21:
```suggestion
    "This function returns a `matplotlib.Figure` object. When the last line of a code cell outputs these objects, Jupyter notebooks display them below the cell. If you call these functions in some other environments or in scripts, you will need to explicitly show or save the outputs.\n",
```

docs/analyze/visualize-results.ipynb
```diff
@@ -0,0 +1,166 @@
...
+   "cell_type": "markdown",
+   "id": "57a1f479-3a5c-4eb4-a550-48c12d8f5c17",
+   "metadata": {},
+   "source": [
+    "### Options when plotting a histogram\n",
+    "\n",
+    "Use the following options for `plot_histogram` to adjust the output graph.\n",
+    "\n",
+    "* `legend`: Provides a label for the executions. It takes a list of strings used to label each execution's results. This is mostly useful when plotting multiple execution results in the same histogram\n",
+    "* `sort`: Adjusts the order the bars in the histogram are rendered. It can be set to either ascending order with `asc` or descending order with `desc`\n",
```
abbycross, line 102:
```suggestion
    "* `sort`: Adjusts the order of the bars in the histogram. It can be set to either ascending order with `asc` or descending order with `desc`\n",
```

docs/analyze/visualize-results.ipynb
```diff
@@ -0,0 +1,166 @@
...
+    "* `number_to_keep`: Takes an integer for the number of terms to show. The rest are grouped together in a single bar called \"rest\"\n",
+    "* `color`: Adjusts the color of the bars; takes a string or a list of strings for the colors to use for the bars for each execution.\n",
```
abbycross, line 104:
```suggestion
    "* `color`: Adjusts the color of the bars; takes a string or a list of strings for the colors to use for the bars for each execution\n",
```

docs/build-new/plot-quantum-states.ipynb
```diff
@@ -0,0 +1,508 @@
...
+   "source": [
+    "# Plot quantum states"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e0242a86-db2b-4cdd-835e-4aa8a5349fb8",
+   "metadata": {},
+   "source": [
+    "In many situations – such as learning or debugging – it's helpful to visualize the state of a quantum computer. Here we assume you already have a particular state from simulation or state tomography. These visualizations require resources exponential in the number of qubits, so it's only possible to view the states of small quantum systems.\n",
```
abbycross, line 16: Feels a bit awkward - possible to reword?

> These visualizations require resources exponential in the number of qubits,

abbycross:
Oh dear - I see we have both `statevector` and `state vector` all throughout our docs, when I do a search! So I don't even know which one to suggest you change to in this PR.... 👀

docs/build-new/circuit-visualization.ipynb
```diff
@@ -0,0 +1,485 @@
...
+   "cell_type": "markdown",
+   "id": "fc41270b-3518-40af-8921-dfa9666fc8e0",
+   "metadata": {},
+   "source": [
+    "### Customize the output\n",
+    "\n",
+    "Depending on the output, there are also options to customize the circuit diagram.\n",
+    "\n",
+    "#### Disable plot barriers and reverse bit order\n",
+    "The first two options are shared among all three backends. They allow you to configure both the bit orders and whether or not you draw barriers. These can be set by the `reverse_bits` kwarg and `plot_barriers` kwarg, respectively. The following examples will work with any output renderer; `mpl` is used here for brevity."
```
abbycross, line 207:
```suggestion
    "The first two options are shared among all three backends. They allow you to configure both the bit orders and whether or not you draw barriers. These can be set by the `reverse_bits` kwarg and `plot_barriers` kwarg, respectively. The following examples work with any output renderer; `mpl` is used here for brevity."
```

docs/build-new/circuit-visualization.ipynb
```diff
@@ -0,0 +1,485 @@
...
+   "id": "18b0bfbe-b586-4182-9a0d-b858655a9240",
+   "metadata": {},
+   "source": [
+    "### Renderer-specific customizations\n",
+    "\n",
+    "Some available customizing options are specific to a renderer.\n",
+    "\n",
+    "The `fold` argument sets a maximum width for the output. In the `text` renderer, this sets the length of the lines of the diagram before it is wrapped to the next line.  When using the 'mpl' renderer, this is the number of (visual) layers before folding to the next line.\n",
+    "\n",
+    "The `mpl` renderer has the `style` kwarg, which changes the colors and outlines. See the [API documentation](https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.QuantumCircuit#draw) for more details.\n",
```
abbycross, line 322:
```suggestion
    "The `mpl` renderer has the `style` kwarg, which changes the colors and outlines. See the [API documentation](/api/qiskit/qiskit.circuit.QuantumCircuit#draw) for more details.\n",
```

docs/build-new/circuit-visualization.ipynb
```diff
@@ -0,0 +1,485 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "495b560a-d835-448b-aa01-5025a7a0689e",
+   "metadata": {},
+   "source": [
+    "### Standalone circuit-drawing function\n",
+    "\n",
+    "If you have an application where you prefer to draw a circuit with a self-contained function instead of as a method of a circuit object, you can directly use the `circuit_drawer()` function, which is part of the public stable interface from `qiskit.visualization`. The function behaves identically to the `circuit.draw()` method, except that it takes in a circuit object as required argument."
```
abbycross, line 420:
```suggestion
    "If you have an application where you prefer to draw a circuit with a self-contained function instead of as a method of a circuit object, you can directly use the `circuit_drawer()` function, which is part of the public stable interface from `qiskit.visualization`. The function behaves identically to the `circuit.draw()` method, except that it takes in a circuit object as a required argument."
```

abbycross (approved):
Nice job and I love the use of the tabs for the different visualization types! Super cool. I had some lightweight suggestions and also pushed some changes from my local branch - hope that's ok; feel free to revert anything you don't like!

docs/build-new/plot-quantum-states.ipynb
frankharkins: Could leave that bit out and just say

> It's only possible to view the states of small quantum systems.

?

abbycross: That works! 👍

docs/analyze/visualize-results.ipynb
Eric-Arellano, line 1: It'd be helpful to talk about how to visualize results for the Estimator primitive. In Hello World, we use `plt.errorbar`, but I think @kaelynj and Kevin were saying another type of graph would be more useful.

frankharkins: Good call, I've had a go at that in [Plotting estimator results](https://qiskit-docs-preview-pr-977.1799mxdls7qz.us-south.codeengine.appdomain.cloud/analyze/visualize-results#plotting-estimator-results). I'm gonna get some more reviewers though as my stats knowledge is quite shaky.

frankharkins:
@abbycross 

> Oh dear - I see we have both `statevector` and `state vector` all throughout our docs, when I do a search! So I don't even know which one to suggest you change to in this PR.... 👀

I think the "correct" term is [state vector](https://en.wikipedia.org/wiki/State_vector), but statevector is a common compouding that happens quite a lot within Qiskit (e.g. we called the class `Statevector` rather than `StateVector`).

IIRC we discussed this at some point with Jennifer and Becky and decided to double down on the compounding and go with "statevector", but I'm not 100% sure about that and can't find the conversation. I've changed all instances in this PR to `statevector` for now but we can easily switch back if needed.

abbycross:
> @abbycross
> 
> > Oh dear - I see we have both `statevector` and `state vector` all throughout our docs, when I do a search! So I don't even know which one to suggest you change to in this PR.... 👀
> 
> I think the "correct" term is [state vector](https://en.wikipedia.org/wiki/State_vector), but statevector is a common compouding that happens quite a lot within Qiskit (e.g. we called the class `Statevector` rather than `StateVector`).
> 
> IIRC we discussed this at some point with Jennifer and Becky and decided to double down on the compounding and go with "statevector", but I'm not 100% sure about that and can't find the conversation. I've changed all instances in this PR to `statevector` for now but we can easily switch back if needed.

I remember we had a conversation too but couldn't find where we landed with it! I'm good to leave it compounded.

PR #983, opened 2024-03-06
Describe broadcasting rules

### LINKED ISSUE #981: Add examples / picture about EstimatorV2 broadcasting
### URL, if applicable

_No response_

Day 1 after the release and we already get questions about how `EstimatorV2` broadcasting works 🙂. We should probably give more information than just linking to the numpy array broadcasting page like it does today. There is a section in the EstimatorV2 [RFC](https://github.com/ihincks/qiskit-rfcs/blob/4ad2e9006b167652ae52040e51af61a2809fe789/0015-estimator-interface.md#arrays-and-broadcasting-) that has nice examples and a picture. 

I will write a draft of the proposed content

beckykd:
closes #981


jyu00 (commented):
For what it's worth, my presentation at practitioner's forum also talked about broadcasting if you want to take a look: https://learning.quantum.ibm.com/tutorial/updates-to-qiskit-runtime#numpy-array-broadcasting

beckykd:
> For what it's worth, my presentation at practitioner's forum also talked about broadcasting if you want to take a look: https://learning.quantum.ibm.com/tutorial/updates-to-qiskit-runtime#numpy-array-broadcasting

I'm getting a "file not found" error.

docs/run/primitives.mdx
```diff
@@ -90,6 +90,53 @@ Example:
...
+
+`EstimatorV2` returns one expectation value estimate for each element of the broadcasted shape.  
+
+Here are some examples of common patterns expressed in terms of array broadcasting.  Their accompanying visual representation is shown in the figure that follows:
+
+```python
+# Broadcast single observable
+parameter_values = np.array(1.0)
+parameter_values.shape == ()
+observables = ObservablesArray(["III", "XXX", Pauli("YYY"), "ZZZ", Pauli("XYZ")])
```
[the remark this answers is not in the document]
beckykd: > Unfortunately `ObservablesArray` is not a public class, so I'd avoid using it.

I put in SparsePauliOp("Z" * 5) instead.  Would that work?

docs/run/primitives.mdx
```diff
@@ -90,6 +90,81 @@ Example:
...
+Here are some examples of common patterns expressed in terms of array broadcasting.  Their accompanying visual representation is shown in the figure that follows:
+
+```python
+# Broadcast single observable
+parameter_values = np.array(1.0)
+parameter_values.shape == ()
+observables = observable = SparsePauliOp("Z" * 5)
+
+observables.shape == (5,)
+>> task_result.shape == (5,)
```
jyu00, line 140: This changes parameter_values to be (5,) and observables (), to match the picture. 

```suggestion
parameter_values = np.random.uniform(size=(5,))  # this has shape (5,)
observables = SparsePauliOp("ZZZ")  # this has shape ()
>> pub result has shape (5,)
```

docs/run/primitives.mdx
```diff
@@ -90,6 +90,81 @@ Example:
...
+
+# Zip
+parameter_values = BindingsArray(np.random.uniform(size=(5,)))
+parameter_values.shape == (5,)
+observables = SparsePauliOp("Z" * 5)
+observables.shape == (5,)
+>> task_result.shape == (5,)
```
jyu00, line 147: `BindingsArray` also didn't make into the release, so we can just use numpy array.

```suggestion
parameter_values = np.random.uniform(size=(5,))  # this has shape (5,)
observables = [SparsePauliOp(pauli) for pauli in ["III", "XXX", "YYY", "ZZZ", "XYZ"]]  # this has shape (5,)
>> pub result has shape (5,)
```

docs/run/primitives.mdx
```diff
@@ -90,6 +90,81 @@ Example:
...
+
+# Outer/Product
+parameter_values = BindingsArray(np.random.uniform(size=(1, 6)))
+parameter_values.shape == (1, 6)
+observables = SparsePauliOp("Z" * 4)
+observables.shape == (4, 1)
+>> task_result.shape == (4,6)
```
jyu00, line 154:
```suggestion
parameter_values = np.random.uniform(size=(1, 6))  # this has shape (1, 6)
observables = [[SparsePauliOp(pauli)] for pauli in ["III", "XXX", "YYY", "ZZZ"]]  # this has shape (4, 1)
>> pub result has shape (4, 6)
```

docs/run/primitives.mdx
```diff
@@ -90,6 +90,81 @@ Example:
...
+
+# Standard nd generalization
+parameter_values = BindingsArray(np.random.uniform(size=(3, 6)))
+parameter_values.shape == (3, 6)
+observables = SparsePauliOp("Z" * 6)
+observables.shape == (2, 3, 1)
+>> task_result.shape == (3, 6, 2)
```
jyu00, line 161:
```suggestion
parameter_values = np.random.uniform(size=(3, 6))  # this has shape (3, 6)
observables = [
    [[SparsePauliOp(['XII'])], [SparsePauliOp(['IXI'])], [SparsePauliOp(['IIX'])]],
    [[SparsePauliOp(['ZII'])], [SparsePauliOp(['IZI'])], [SparsePauliOp(['IIZ'])]]
]  # this has shape (2, 3, 1)
>> pub result has shape (2, 3, 6)
```

ElePT (commented):
+1 to Jessie's suggestions, I left an additional comment that I think would be useful to have, and I also see 2 missing pieces of information:

1. These rules only talk about circuits and observables, what about the parameters? Shouldn't we also include examples with different parameter array sizes?

2. What about the sampler rules? Even if the mechanics are the same, I think that the sampler should be at least mentioned. Once the parameter mechanics are explained in the estimator it's quite natural to just say "the sampler is the same but without observables"

jyu00:
> +1 to Jessie's suggestions, I left an additional comment that I think would be useful to have, and I also see 2 missing pieces of information:
> 
> 1. These rules only talk about circuits and observables, what about the parameters? Shouldn't we also include examples with different parameter array sizes?

> 2. What about the sampler rules? Even if the mechanics are the same, I think that the sampler should be at least mentioned. Once the parameter mechanics are explained in the estimator it's quite natural to just say "the sampler is the same but without observables"

@ElePT, these rules actually apply to observables and parameter arrays in a PUB. Each pub only has 1 circuit, and each element in the broadcasted observables + parameter arrays is then applied to that circuit (and hence the result has the same size/shape as the broadcasted arrays). Sampler is not mentioned because there are no special rules - i.e. each parameter value set is applied to the circuits. But we should definitely clarify if it's not obvious.

ElePT:
> @ElePT, these rules actually apply to observables and parameter arrays in a PUB. Each pub only has 1 circuit, and each element in the broadcasted observables + parameter arrays is then applied to that circuit (and hence the result has the same size/shape as the broadcasted arrays). Sampler is not mentioned because there are no special rules - i.e. each parameter value set is applied to the circuits. But we should definitely clarify if it's not obvious.

I guess that the point I was trying to make is that I don't think it's obvious to everyone, even if it's not that difficult to extrapolate, I think one example for the sampler and one for the estimator with parameters (even if it is to show that the rules are the same) will go a long way.

beckykd:
> each parameter value set is applied to the circuits.

> > @ElePT, these rules actually apply to observables and parameter arrays in a PUB. Each pub only has 1 circuit, and each element in the broadcasted observables + parameter arrays is then applied to that circuit (and hence the result has the same size/shape as the broadcasted arrays). Sampler is not mentioned because there are no special rules - i.e. each parameter value set is applied to the circuits. But we should definitely clarify if it's not obvious.
> 
> I guess that the point I was trying to make is that I don't think it's obvious to everyone, even if it's not that difficult to extrapolate, I think one example for the sampler and one for the estimator with parameters (even if it is to show that the rules are the same) will go a long way.

But I think that the broadcasting rules apply only to how we combine multiple arrays.  Since you can't have multiple arrays with Sampler, they don't apply.

docs/run/primitives.mdx
```diff
@@ -80,8 +80,8 @@ estimator.run([circuit1, circuit2, ...],[observable1, observable2, ...],[param_v
 
 #### Estimator V2
 
-* Takes one parameter: PUBs in the format (`<circuit>`, `<observables>`, `<parameter values>`, `<precision>`)
-* Numpy [broadcasting rules](https://numpy.org/doc/stable/user/basics.broadcasting.html) are used when combining observables and parameter values.
+* Takes one parameter: PUBs in the format (`<circuit>`, `<observables>`, `<parameter values>`, `<precision>`) (Note that there is only one circuit and optional `precision` value per PUB, but `parameter values` and `observables` can be lists or a single value.)
```
jyu00, line 83: parameter values are also optional. Now that I'm reading this line, it has some other problems. Estimator V2 `run()` actually takes `pubs` (which is an array of PUB) _and_ `precision` (same goes to Sampler which takes `pubs` and `shots`). 

Would it be clearer if we just say
"
The `run()` method takes an array of PUBs. Each PUB is  in the format (`<single circuit>`, `<one or more observables>`, `<optional one or more parameter values>`, `<optional precision>`)
"

docs/run/primitives.mdx
```diff
@@ -80,8 +80,8 @@ estimator.run([circuit1, circuit2, ...],[observable1, observable2, ...],[param_v
...
+* Combines elements from multiple arrays by following NumPy broadcasting rules as described below.
```
jyu00, line 84:
```suggestion
* Combines elements from observables and parameter values by following NumPy broadcasting rules as described below. 
```

jyu00 (approved):
I think the wording on pubs needs to be changed, but the rest LGTM!

docs/run/primitives.mdx
```diff
@@ -90,6 +90,91 @@ Example:
 estimator.run([(circuit1, observable1, param_values1),(circuit2, observable2, param_values2) ])
```
ElePT, line 90: Tiny suggestion:

```suggestion
estimator.run([(circuit1, observable1, param_values1),(circuit2, observable2, param_values2)])
```

ElePT: I cannot leave an inline suggestion because the code wasn't changed but I was checking the rendered output and the example for EstimatorV1 seems to get cut out, it might make sense to divide it into two lines:

Screenshot 2024-03-13 at 10 27 54

docs/run/primitives.mdx
```diff
@@ -80,8 +80,8 @@ estimator.run([circuit1, circuit2, ...],[observable1, observable2, ...],[param_v
 
 #### Estimator V2
 
-* Takes one parameter: PUBs in the format (`<circuit>`, `<observables>`, `<parameter values>`, `<precision>`)
-* Numpy [broadcasting rules](https://numpy.org/doc/stable/user/basics.broadcasting.html) are used when combining observables and parameter values.
+* The `run()` method takes an array of PUBs. Each PUB is in the format (`<single circuit>`, `<one or more observables>`, `<optional one or more parameter values>`, `<optional precision>`)
```
ElePT, line 83: I think that the clarification made for the sampler section about num.circuits/parameters also fits well here (repetition doesn't hurt). It's also good to note that each item of "parameter_values" might contain one or more parameters. What I mean here is that for a circuit with 2 parameters, a pub with 1 circuit, 1 observable and 1 "parameter value" would look like this:

`estimator.run([(circuit, observable, [0.5, 0.5])])`

While 2 "parameter values" would look like this:

`estimator.run([(circuit, observable, [[0.5, 0.5], [0.6, 0.7]])])`

So when we commonly talk about parameter values, we are not talking about individual parameters, we are talking about "sets of parameters" or "arrays of parameters", and the number of parameters in a set depends on the number of parameters in the circuit. Anyway, this is all to preface this small suggestion:

```suggestion
* The `run()` method takes an array of PUBs. Each PUB is in the format (`<single circuit>`, `<one or more observables>`, `<optional one or more parameter values>`, `<optional precision>`). Note that there is only one circuit per PUB, there can be multiple items in `observables` and `parameter values`, and each item of `parameter values` can be a list or a single parameter.
```

docs/run/primitives.mdx
```diff
@@ -90,6 +90,91 @@ Example:
...
+```
+
+Examples of array pairs that do not broadcast:
+
+```text
+A1     (1d array):  5
+A2     (1d array):  3 
+
+A1     (2d array):      2 x 1
+A2     (3d array):  6 x 5 x 4 #This would work except for the middle dimensions (2 compared to 5).
```
ElePT, line 127: Could this be easier to understand?

```suggestion
A2     (3d array):  6 x 5 x 4 # This would work if the middle dimension were 2, but it is 5.
```

docs/run/primitives.mdx
```diff
@@ -90,6 +90,91 @@ Example:
...
+>> pub result has shape (5,)
+
+# Zip
+parameter_values = np.random.uniform(size=(5,))  # this has shape (5,)
+observables = [SparsePauliOp(pauli) for pauli in ["III", "XXX", "YYY", "ZZZ", "XYZ"]]  # this has shape (5,)
+>> pub result has shape (5,)
+
+# Outer/Product
+parameter_values = np.random.uniform(size=(1, 6))  # this has shape (1, 6)
+observables = [[SparsePauliOp(pauli)] for pauli in ["III", "XXX", "YYY", "ZZZ"]]  # this has shape (4, 1)
```
ElePT, line 147: Again, some of these comments get cut out in the rendered version:

Screenshot 2024-03-13 at 10 37 32
```suggestion
# the observables have shape (5,)
observables = [SparsePauliOp(pauli) for pauli in ["III", "XXX", "YYY", "ZZZ", "XYZ"]]
>> pub result has shape (5,)

# Outer/Product
parameter_values = np.random.uniform(size=(1, 6))  # this has shape (1, 6)
# the observables have shape (4, 1)
observables = [[SparsePauliOp(pauli)] for pauli in ["III", "XXX", "YYY", "ZZZ"]]  
```

docs/run/primitives.mdx
```diff
@@ -105,8 +190,8 @@ sampler.run([circuit1, circuit2, ...],[observable1, observable2, ...],[param_val
 
 #### Sampler V2
 
-* Takes one parameter: PUBs in the format  (`<circuit>`, `<parameter values>`, `<shots>`)
-* Elements from each are aggregated.
+* Takes one parameter: PUBs in the format  (`<circuit>`, `<parameter values>`, `<shots>`) (Note that there is only one circuit per PUB and an optional value for `shots`, but `parameter values` can be a list or a single parameter)
+* Elements from each are aggregated. For example, each parameter value in the PUB is applied to the PUB's circuit.
```
ElePT, line 194: Again, slight rewording regarding parameter values:

```suggestion
* Takes one parameter: PUBs in the format  (`<single circuit>`, `<optional one or more parameter values>`, `<optional shots>`). Note that there is only one circuit per PUB and an optional value for `shots`, but there can be multiple items of `parameter values`, and each item can be either an array or a single parameter (depending on the chosen circuit).
* Elements from each are aggregated. For example, each array of parameter values in the PUB is applied to the PUB's circuit.
```

ElePT (commented):
I gave it a final review looking and the rendered version and found a few tiny formatting issues, I also suggested a couple of small wording changes, but other than that I think that the content is there! LGTM

docs/run/primitives.mdx
beckykd: You can scroll the code examples, but I deleted "this has" from each comment, so I think they fit now.

PR #989, opened 2024-03-08
Refactor `htmlToMd.test.ts`

arnaucasau:
### Summary

Part of  #845

This PR refactors the `htmlToMd.test.ts` test file.

### Changes

- Three tests were removed
  1. `merge contiguous emphasis` was merged with `remove spaces from emphasis boundaries`. They were using the same HTML but only adding spaces
  2. `convert method and attributes to titles`. This test was incomplete and it's implicitly checked at `convert inline methods`.
  3. `convert class method with a problematic output`. The test wasn't checking anything related to the script behavior. The problematic output is due to a bad HTML that should be fixed in the original repo.

- Input HTML trimmed for some tests. We were using long inputs to test certain behaviors that don't need it. For example, we were testing the transformation of a table with 11 entries when the same test could use 2 or 3.

- Some tests have been organized into sections to easily find them scanning the file. These sections can be found at the end of the test file leaving the uncategorized tests at the beginning.

- We were using two helper functions for the tests named `toMd` and `toMdWithMeta` with the only difference that in the first one, we returned the markdown attribute of the transformation instead of the whole object. I merged these two functions with an extra argument to specify when we need to return the metadata.

### DIFF
--- a/scripts/lib/api/htmlToMd.test.ts
+++ b/scripts/lib/api/htmlToMd.test.ts
@@ -21,188 +21,51 @@ const DEFAULT_ARGS = {
   releaseNotesTitle: "My Quantum release notes",
 };
 
+async function toMd(html: string, withMetadata: boolean = false) {
+  const result = await sphinxHtmlToMarkdown({
+    fileName: "stubs/qiskit_ibm_runtime.Sampler.html",
+    html,
+    ...DEFAULT_ARGS,
+  });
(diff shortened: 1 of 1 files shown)


Eric-Arellano (approved):
Great work! I appreciate the well-written PR description also. This will make it so much easier to understand the tests and make changes over time.

Issue #74, opened 2023-10-04
Docker preview is extremely slow and outdated


Eric-Arellano:
It's taking several seconds to load the docs preview, both with GitHub deploys and with local previews via `./start`.

![image](https://github.com/Qiskit/documentation/assets/14852634/6b351e23-4421-4df3-80a8-6dba1890b07d)

Given that localhost is slow too, it seems likely this is an issue with the Docker image itself, rather than our CI infrastructure.

Perhaps https://github.com/Qiskit/documentation/issues/53 will speed things up by changing to a new Docker image.

Eric-Arellano:
The new Docker image in https://github.com/Qiskit/documentation/pull/150 is substantially faster. This is almost certainly because the current Docker image we're using from `icr.io/quantum-computing/iqp-channel-docs-dev` was built in dev mode, not prod, so it's slow.

As explained in #150, we want to continue using the password protected image from `icr.io/quantum-computing/iqp-channel-docs-dev` in PR builds and staging. So we will need to redeploy that image to not be in dev mode.

Eric-Arellano:
We're trying to test out a new image in https://github.com/Qiskit/documentation/pull/279, but the deploys are failing with these. It's unclear what's going on and we can't get access to the logs (we're asking someone for them)

![image_720](https://github.com/Qiskit/documentation/assets/14852634/9b36531e-fc0a-47eb-a2c8-deea5599ef0f)

Eric-Arellano:
Switching from the normal docs app to the preview image in https://github.com/Qiskit/documentation/pull/1045 did not help. We still get this error from IBM Cloud

```
Revision failed
Unable to start the app revision. Make sure that the configured port matches the port that the application code uses to listen to receive HTTP requests and try again.
```

PR #990, opened 2024-03-08
Upgrade to new staging Docker image

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/74.

We were finally able to get a Docker image working with Code Engine. Turns out the problem was upgrading from Next.js 13.4 to 13.5 because Code Engine does not play nicely with 13.4.13 or newer. 

The Docker image was so slow because we were under-allocating resources. We double the resources. It's still somewhat slow, but less laggy than before. The next step would be to double again, which costs more money, so I was thinking to start with this for now?


frankharkins (approved):
It is just a preview so I don't mind it being slow. This PR preview takes ~4s to load a page for me which is fast enough to see how things look.

Is there a way to monitor how much things cost? I'd be interested to see whether this has a linear effect on cost or whether the initial upload or spin up costs contribute significantly.

Eric-Arellano:
> Is there a way to monitor how much things cost? I'd be interested to see whether this doubles the cost or whether the initial upload or spin up costs contribute significantly.

Yeah, we can see it in Cloud. I'm going to work on https://github.com/Qiskit/documentation/issues/1034 this week as a way to cut costs.

PR #1005, opened 2024-03-11
Split `generateToc` into helper functions

arnaucasau:
This PR refactors the `generateToc.ts` script splitting out the `generateToc` function into some helper functions. It also removes the if condition to check if the length of the modules is greater than 0 because it's not necessary.

The PR added a test to check the latter, and it was tested regenerating every qiskit and provider versions.

### DIFF
--- a/scripts/lib/api/generateToc.ts
+++ b/scripts/lib/api/generateToc.ts
@@ -13,7 +13,7 @@
 import { Dictionary, isEmpty, keyBy, keys, orderBy } from "lodash";
 
 import { getLastPartFromFullIdentifier } from "../stringUtils";
-import { Metadata } from "./Metadata";
+import { HtmlToMdResultWithUrl } from "./HtmlToMdResult";
 import { Pkg } from "./Pkg";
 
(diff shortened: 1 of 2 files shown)


scripts/lib/api/generateToc.ts
```diff
@@ -39,7 +39,32 @@ export function generateToc(
   pkg: Pkg,
   results: Array<{ meta: Metadata; url: string }>,
```
Eric-Arellano, line 40: Consider either defining a type `MetadataWithUrl` or using `HtmlToMdResultWithUrl` from `HtmlToMdResult.ts`. The downside of `HtmlToMdResultWithUrl` is it makes your test setup more complex.

Eric-Arellano (approved):
Nice improvement!

scripts/lib/api/generateToc.ts
arnaucasau: Good idea! I used `HtmlToMdResultWithUrl` to match the type we were already passing from `conversionPipeline.ts`

PR #1006, opened 2024-03-11
Refactor `mergeClassMembers` unified plugin

arnaucasau:
Small PR to add some missing types to variables and merge two unified plugins into one. Both unified plugins were using different phases of the whole process (parse, run, and stringify).

### DIFF
--- a/scripts/lib/api/mergeClassMembers.ts
+++ b/scripts/lib/api/mergeClassMembers.ts
@@ -62,7 +62,7 @@ export async function mergeClassMembers(
           .use(remarkGfm)
           .use(remarkMath)
           .use(() => {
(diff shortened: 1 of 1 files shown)


scripts/lib/api/mergeClassMembers.ts
```diff
@@ -127,21 +127,18 @@ async function parseMarkdownIncreasingHeading(
   md: string,
   depthIncrease: number,
 ): Promise<Root> {
-  const root = await unified()
```
arnaucasau, line 130: I removed the `await` because vs code says: `'await' has no effect on the type of this expression`.

Eric-Arellano (approved):
Nice!

Issue #885, opened 2024-02-26
Refactor "Configure transpilation" section to use preset pass managers instead of transpile function


kevinsung:
### Feedback

They are also relevant to `generate_preset_pass_manager`. We should remove the "Use the transpile function" heading or otherwise reorganize the pages. One possibility is to change the heading to "Configuring transpilation" or something like that.

https://github.com/Qiskit/documentation/blob/abde486fd2a903fac5a46f5627623df6637b6c3f/docs/transpile/_toc.json#L26-L45

abbycross:
@javabster how do you feel about renaming this section "Configure transpilation" instead of "Use the transpile function". as Kevin suggests?

javabster:
I think this is more than just renaming the section, we'd need to go through each page with a fine toothed comb and replace any use/mention of `transpile()` with a passmanager equivalent, and make sure that the info is the same. But in general I think it would be a good idea (now that the summit deadline pressure is lifted) to go back and refactor these pages to use passmanagers, and then just dissolve the subsection. This would probably also requiring checking all pages in the transpile section, as I know a lot of the other pages reference these ones. Also the introduction would probably need a bit of a refactor too 😅 

But yes, it would be a big piece of refactoring work but I think it's worth it given we are moving away from `transpile()` wherever possible

PR #1007, opened 2024-03-11
Rewrite "optimization level" documentation

kevinsung:
Part of #885 

- Convert from MDX to Jupyter notebook
- Rewrite introduction to be more concise
- Use preset pass manager instead of `transpile`
- Use Eagle device and ECR gates instead of retired Falcon device and CX gates


javabster (approved):
LGTM 🚀

PR #1009, opened 2024-03-11
add Q-CTRL exception to ISA circuit notifications

abbycross:
Adding this text wherever the notification about ISA circuits appears:

```

> **Exception**: When you initialize the Qiskit Runtime Service with the Q-CTRL channel strategy (example below), abstract circuits are still supported.
> ``` python
> service = QiskitRuntimeService(channel="ibm_cloud", channel_strategy="q-ctrl")
> ```
> 

```

Files where the notification appears:

- [x] docs/run/primitives-examples.mdx
- [x] docs/run/primitives-get-started.mdx
- [x] docs/run/configure-error-mitigation.mdx
- [x] docs/start/hello-world.ipynb
- [x] docs/run/estimate-job-run-time.mdx
- [x] docs/run/advanced-runtime-options.mdx
- [x] docs/run/configure-runtime-compilation.mdx
- [x] docs/run/primitives.mdx


docs/run/configure-error-mitigation.mdx
```diff
@@ -42,6 +42,14 @@ expectation values.
 
 <Admonition type="caution" title="Important">
 To ensure faster and more efficient results, as of 1 March 2024, circuits and observables need to be transformed to only use instructions supported by the system (referred to as *instruction set architecture (ISA)* circuits and observables) before being submitted to the Qiskit Runtime primitives.  See the [transpilation documentation](../transpile) for instructions to transform circuits.  Due to this change, the primitives will no longer perform layout or routing operations.  Consequently, transpilation options referring to those tasks will no longer have any effect. Users can still request that the primitives do no optimization of input circuits by specifying `optimization_level=0`.
+
+
+**Exception**: The Q-CTRL primitives continue to support abstract circuits, as in the following code:
```
Eric-Arellano, line 47: I think we got feedback in the past that we shouldn't bold as much because it looks similar to an `h3`. Maybe use italics instead?

abbycross: Happy to use italics here. However, I'm curious about this feedback, especially as we use bold often in the docs (usually as part of a block of text, or with a colon following it) and wonder if in those contexts it would not be mistaken for h3, and can be left alone?

docs/run/configure-error-mitigation.mdx
```diff
@@ -42,6 +42,15 @@ expectation values.
 
 <Admonition type="caution" title="Important">
 To ensure faster and more efficient results, as of 1 March 2024, circuits and observables need to be transformed to only use instructions supported by the system (referred to as *instruction set architecture (ISA)* circuits and observables) before being submitted to the Qiskit Runtime primitives.  See the [transpilation documentation](../transpile) for instructions to transform circuits.  Due to this change, the primitives will no longer perform layout or routing operations.  Consequently, transpilation options referring to those tasks will no longer have any effect. Users can still request that the primitives do no optimization of input circuits by specifying `optimization_level=0`.
+
+
+
+> *Exception*: When you initialize the Qiskit Runtime Service with the Q-CTRL channel strategy (example below), abstract circuits are still supported.
```
Eric-Arellano, line 48: Using `>` like this won't include the code block. Also you probably want to use the same styling for all the files.

abbycross: whoops, that was a copy-paste error!

Eric-Arellano (approved):

PR #1011, opened 2024-03-11
Remove extra tab

beckykd:


docs/api/migration-guides/v2-primitives.mdx
```diff
@@ -408,25 +408,6 @@ print(f">> dynamical decoupling sequence to use: {sampler.options.dynamical_deco
...
-options.resilience_level = 2 
-sampler = Sampler(backend=backend, options=options)
-
-# Setting options after primitive initialization.
-# This does bulk update.
-sampler.set_options(shots=2000, optimization_level=0)
-  ```  
-  </TabItem>
-
   <TabItem value="SamplerV1" label="Sampler (V1)">
```
Eric-Arellano, line 411: Why is the V1 tab so much shorter than the V2 tab? Only double checking this is the correct content.

Eric-Arellano (approved):

docs/api/migration-guides/v2-primitives.mdx
beckykd: The V2 tab calls the service and defines the backend, but it doesn't really need to so I deleted those lines.

abbycross (approved):

Issue #998, opened 2024-03-08
Change page render check to automatically support new folders


Eric-Arellano:
Right now, we hardcode the non-API docs to check:

https://github.com/Qiskit/documentation/blob/51667c31b61729d6ac78c209b5168d64086e3425/scripts/commands/checkPagesRender.ts#L162-L166

That breaks when adding new folders like `docs/analyze`. Instead, we should probably take the same approach as our link checker:

https://github.com/Qiskit/documentation/blob/51667c31b61729d6ac78c209b5168d64086e3425/scripts/commands/checkInternalLinks.ts#L157-L168

PR #1012, opened 2024-03-11
Future proof pages render check for new Patterns folders

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/998. Before this PR, new folders like `docs/analyze` would not be included in `npm run check:pages-render -- --non-api`.

### DIFF
--- a/scripts/commands/checkPagesRender.ts
+++ b/scripts/commands/checkPagesRender.ts
@@ -159,32 +159,27 @@ async function determineFilePaths(args: Arguments): Promise<string[]> {
   }
 
   if (args.nonApi) {
-    globs.push(
-      "docs/support.mdx",
-      "docs/{build,run,start,transpile,verify}/*.{ipynb,md,mdx}",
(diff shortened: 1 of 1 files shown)


scripts/commands/checkPagesRender.ts
```diff
@@ -159,32 +159,27 @@ async function determineFilePaths(args: Arguments): Promise<string[]> {
...
-  if (args.qiskitReleaseNotes) {
-    globs.push("docs/api/qiskit/release-notes/*.{ipynb,md,mdx}");
-  }
-  if (args.translations) {
-    globs.push("translations/**/*.{ipynb,md,mdx}");
+
+  for (const [isIncluded, glob] of [
+    [
+      args.currentApis,
+      "docs/api/{qiskit,qiskit-ibm-provider,qiskit-ibm-runtime}/*.{md,mdx}",
```
Eric-Arellano, line 168: Note that I removed `.ipynb` from the API docs. We don't use Jupyter notebooks there.

kevinsung (approved):

Issue #1008, opened 2024-03-11
Set up Function component for API docs redesign


Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/479. We're going to start with adding an MDX component for `Function`s before adding the more complex class/method components, since functions don't have to deal as much with nesting. Functions are always top-level.

- [x] Teach our API generation script to generate `<Function>`
- [x] MDX will add a `span id` for the full path:

https://github.com/Qiskit/documentation/blob/14717912a767e0b2e89b0c3e0e217d217d7e32bd/docs/api/qiskit/utils.md?plain=1#L25

- [x] MDX will have a left colored bar, https://github.com/Qiskit/documentation/issues/735
- [ ] https://github.com/Qiskit/documentation/issues/948, meaning we no longer generate a header, but there is still an `id` set with the short function name, a copy anchor link button, and the header shows up in the page ToC

It is not yet in scope to worry about indentation, https://github.com/Qiskit/documentation/issues/736. That is lower priority. 

### Component name

Should it be `<Function>` or `<ApiFunction>`?

PR #1013, opened 2024-03-11
Rename API files to use MDX extension

Eric-Arellano:
API docs use Markdown components, and are about to use a lot more of them with the new API components like https://github.com/Qiskit/documentation/issues/1008. It's confusing to pretend they're MD files. 

This PR makes things consistent, which allows us to simplify some of our scripts to only deal with MDX and IPYNB extensions.

The API docs were manually changed with this Bash snippet:

```bash
for file in docs/api/**/*.md; do
  if [ -f "$file" ]; then
    mv "$file" "${file%.md}.mdx"
  fi
done
```


arnaucasau (approved):
Thanks!

PR #1017, opened 2024-03-12
Add info for Cloud users to access support

abbycross:
Requested by @fvarchon to clarify for PayGO users how to contact support.

### DIFF
--- a/docs/support.mdx
+++ b/docs/support.mdx
@@ -10,6 +10,7 @@ description: How to find answers to questions or problems you encounter while us
 
 Members of the IBM Quantum&trade; Network can reach out to IBM Quantum Support if they have technical difficulties, questions, or concerns. Ask your administrator or IBM&reg; representative for IBM Quantum Support contact information.
 
+Pay-As-You-Go users can open a case in the IBM Cloud&reg; Support Center by following [this link](https://cloud.ibm.com/unifiedsupport/cases/form), and choosing 'Qiskit Runtime' from the Topic field.
 
 ## Qiskit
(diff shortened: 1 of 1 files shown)


docs/support.mdx
```diff
@@ -10,6 +10,7 @@ description: How to find answers to questions or problems you encounter while us
 
 Members of the IBM Quantum&trade; Network can reach out to IBM Quantum Support if they have technical difficulties, questions, or concerns. Ask your administrator or IBM&reg; representative for IBM Quantum Support contact information.
 
+Pay-As-You-Go users can open a case in the IBM Cloud&reg; Support Center by following [this link](https://cloud.ibm.com/unifiedsupport/cases/form), and choosing 'Qiskit Runtime' from the Topic field.
```
Eric-Arellano, line 13: Is it correct to use dashes for Pay-As-You-Go? Will that mess up translations?

abbycross: The hyphens group the phrase so that it describes "users" -- so makes it better for translation (and for reading in English) :)

abbycross: I would like to clarify with @fvarchon, however, that PayGO users identify themselves as such -- will Qiskit Runtime on the Cloud users think of themselves as PayGO?

abbycross: Update: Valerie Chiang confirms that Pay-as-you-go is the correct term; it is used at Summit and within IBM Quantum with users.

Matt-Stypulkoski (approved):

PR #1021, opened 2024-03-12
Add API entries to allowed urls

### LINKED ISSUE #1014: Orphan check fails for API docs
`npm run check:orphan-pages -- ---historical-apis --current-apis --dev-apis` fails because it needs more entries in the ignore list.

kaelynj:
Closes #1014 

In the interest of readability, the list of allowed urls includes pages that don't exist.  There is a small risk of a false positive if there are any new pages that get added to the api docs of the same name which should actually be checked against their toc instead of ignored.

### DIFF
--- a/scripts/commands/checkOrphanPages.ts
+++ b/scripts/commands/checkOrphanPages.ts
@@ -178,26 +178,28 @@ function apiDocsIgnores(): string[] {
     "0.46/",
   ];
 
-  return versions.flatMap((vers) => [
-    `/api/qiskit-ibm-runtime/${vers}index`,
-    `/api/qiskit-ibm-provider/${vers}index`,
-    `/api/qiskit-ibm-runtime/${vers}qiskit_ibm_runtime.Estimator`,
-    `/api/qiskit-ibm-runtime/${vers}qiskit_ibm_runtime.Sampler`,
-    `/api/qiskit/${vers}aer`,
-    `/api/qiskit/${vers}aqua`,
-    `/api/qiskit/${vers}ibmq-provider`,
-    `/api/qiskit/${vers}ibmq_jupyter`,
-    `/api/qiskit/${vers}ibmq_visualization`,
-    `/api/qiskit/${vers}qiskit.aqua.aqua_globals`,
-    `/api/qiskit/${vers}qiskit.optimization.INFINITY`,
-    `/api/qiskit/${vers}qiskit.quantum_info.two_qubit_cnot_decompose`,
-    `/api/qiskit/${vers}qiskit.utils.algorithm_globals`,
-    `/api/qiskit/${vers}parallel`,
-    `/api/qiskit/${vers}transpiler_builtin_plugins`,
+  return [
+    ...versions.flatMap((vers) => [
+      `/api/qiskit-ibm-runtime/${vers}index`,
+      `/api/qiskit-ibm-provider/${vers}index`,
+      `/api/qiskit-ibm-runtime/${vers}qiskit_ibm_runtime.Estimator`,
+      `/api/qiskit-ibm-runtime/${vers}qiskit_ibm_runtime.Sampler`,
+      `/api/qiskit/${vers}aer`,
+      `/api/qiskit/${vers}aqua`,
+      `/api/qiskit/${vers}ibmq-provider`,
+      `/api/qiskit/${vers}ibmq_jupyter`,
+      `/api/qiskit/${vers}ibmq_visualization`,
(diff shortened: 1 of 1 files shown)


scripts/commands/checkOrphanPages.ts
```diff
@@ -74,6 +75,7 @@ async function main() {
       ),
     );
   }
+  console.log("\n Allowed Orphans are: \n", ALLOWED_ORPHAN_URLS);
```
Eric-Arellano, line 78: Leftover? This is noisy 

```suggestion
```

scripts/commands/checkOrphanPages.ts
```diff
@@ -121,6 +123,7 @@ async function determineTocFiles(args: Arguments): Promise<string[]> {
     globs.push(
       "docs/api/{qiskit,qiskit-ibm-provider,qiskit-ibm-runtime}/[0-9]*/_toc.json",
     );
+    ``;
```
Eric-Arellano, line 126: Tip before opening up PRs to review your PR proactively to check for issues like this

```suggestion
```

scripts/commands/checkOrphanPages.ts
```diff
@@ -139,4 +142,67 @@ function collectTocFileContents(children: TocEntry[]): string[] {
...
+    "0.41/",
+    "0.42/",
+    "0.43/",
+    "0.44/",
+    "0.45/",
+    "0.46/",
+  ];
+
+  return flattenDeep([
+    versions.map((vers) => [
```
Eric-Arellano, line 184: Instead, use `.flatMap()` here. Then you can remove `flattenDeep` :)

scripts/commands/checkOrphanPages.ts
```diff
@@ -139,4 +142,67 @@ function collectTocFileContents(children: TocEntry[]): string[] {
...
+      `/api/qiskit/${vers}qiskit.optimization.INFINITY`,
+      `/api/qiskit/${vers}qiskit.quantum_info.two_qubit_cnot_decompose`,
+      `/api/qiskit/${vers}qiskit.utils.algorithm_globals`,
+      `/api/qiskit/${vers}parallel`,
+      `/api/qiskit/${vers}transpiler_builtin_plugins`,
+    ]),
+    `/api/qiskit/0.19/index`,
+    `/api/qiskit/dev/qiskit.primitives.BaseEstimator`,
+    `/api/qiskit/dev/qiskit.primitives.BaseSampler`,
+    `/api/qiskit/dev/qiskit.primitives.BaseSampler`,
```
Eric-Arellano, line 204: This doesn't fail on the current API docs? Also BaseSampler is repeated

Eric-Arellano (commented):
Thanks!

scripts/commands/checkOrphanPages.ts
kaelynj: Nope, but it does fail on the dev version

scripts/commands/checkOrphanPages.ts
kaelynj: Fixed!

scripts/commands/checkOrphanPages.ts
kaelynj: Duly noted, I appreciate your patience in reviewing these :)

scripts/commands/checkOrphanPages.ts
kaelynj: Good suggestion, thanks!

scripts/commands/checkOrphanPages.ts
```diff
@@ -180,28 +178,25 @@ function apiDocsIgnores(): string[] {
...
+    `/api/qiskit/${vers}ibmq_visualization`,
+    `/api/qiskit/${vers}qiskit.aqua.aqua_globals`,
+    `/api/qiskit/${vers}qiskit.optimization.INFINITY`,
+    `/api/qiskit/${vers}qiskit.quantum_info.two_qubit_cnot_decompose`,
+    `/api/qiskit/${vers}qiskit.utils.algorithm_globals`,
+    `/api/qiskit/${vers}parallel`,
+    `/api/qiskit/${vers}transpiler_builtin_plugins`,
     `/api/qiskit/0.19/index`,
     `/api/qiskit/dev/qiskit.primitives.BaseEstimator`,
     `/api/qiskit/dev/qiskit.primitives.BaseSampler`,
```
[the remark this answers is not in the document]
kaelynj: Fixed now!

Eric-Arellano (approved):

Issue #948, opened 2024-03-01
Remove API docs headers for functions, exceptions, and methods


Eric-Arellano:
For functions, methods, and exceptions, we manually create h3 headers:

<img width="793" alt="Screenshot 2024-03-01 at 4 33 18 PM" src="https://github.com/Qiskit/documentation/assets/14852634/162799d6-26a6-4dc3-9c29-42521752e6f5">

We do this so that:

1. it shows up in the page ToC, since that is computed based on headers
2. there's the little icon to make an anchor link

<img width="154" alt="Screenshot 2024-03-01 at 4 34 45 PM" src="https://github.com/Qiskit/documentation/assets/14852634/fb911cc9-ee36-4c8e-87be-a7d48012fe8c">

But we don't have to have a header for those two benefits; there are other ways we could get that. We can teach our page ToC to understand code objects, and we can add the little anchor icon to the code object directly.

Fixing this will remove duplication and get our design closer to qiskit-sphinx-theme, which we liked:

<img width="725" alt="Screenshot 2024-03-01 at 4 36 36 PM" src="https://github.com/Qiskit/documentation/assets/14852634/a6ca6f47-ad7a-4414-8a30-f012b4735702">

To implement this, we'll want MDX components like `<Function>`, which relates to the implementation for https://github.com/Qiskit/documentation/issues/735.

The trickiest part will be the page ToC. We need to make sure we consider the surrounding context like if the parent heading level.

Eric-Arellano:
We decided to not do this:

1. We're making h3s larger soon and think it actually looks good when combined with the API docs redesign

<img width="689" alt="Screenshot 2024-04-19 at 4 48 31 PM" src="https://github.com/Qiskit/documentation/assets/14852634/3e237406-5548-41b8-ac3d-03d5f154ad57">

2. Figuring out the right page ToC is tricky, since it's based on h1-h6
3. Figuring out the search index is tricky, since it differentiates between headers and non-headers
4. It makes semantic sense to have the headers. It's helpful for screen readers and SEO

PR #1026, opened 2024-03-13
API generation script uses MDX components

arnaucasau:
Part of #1008

This PR teaches our API generation script to generate MDX components that will define the styling of the classes, properties, attributes, methods, functions, and exceptions.

The script will generate custom tags to wrap all the content and store useful information as props. This is an example of git diff for an Attribute:
```diff
- ### dtm
- Return the system time resolution of output signals :returns: The output signal timestep in seconds. :rtype: dtm
+  ### dtm
+  <Attribute id="qiskit_ibm_provider.IBMBackend.dtm" name="dtm">
+    Return the system time resolution of output signals :returns: The output signal timestep in seconds. :rtype: dtm
+  </Attribute>
```
The script continues generating headers for every apiType that needs it, but this will be removed in a follow-up.

### Details

The PR adds a new script named `generateMdxComponents.ts` that will generate all the components. The main function is `processMdxComponent` where we prepare the necessary props for each component and then return an open tag and a closed tag ready for its use.

To process the new tags, new handlers have been added to `htmlToMd.ts`.


scripts/lib/api/processHtml.ts
```diff
@@ -432,20 +432,11 @@ function processAttribute(
 
 function processFunctionOrException(
```
Eric-Arellano, line 433: This also changes Exceptions. You should split those out into two separate helper functions.

scripts/lib/api/processHtml.ts
```diff
@@ -432,20 +432,11 @@ function processAttribute(
...
-  const descriptionHtml = `<span class="target" id="${id}"/><p><code>${$child.html()}</code>${githubSourceLink}</p>`;
-
-  const pageHeading = $dl.siblings("h1").text();
-  if (id.endsWith(pageHeading) && pageHeading != "") {
-    // Page is already dedicated to apiType; no heading needed
-    return descriptionHtml;
-  }
-
-  const apiName = id.split(".").slice(-1)[0];
-  return `<h3>${apiName}</h3>${descriptionHtml}`;
```
Eric-Arellano, line 448: We still need the heading for now, and we always need the short name so that it shows up in the right page ToC, even after https://github.com/Qiskit/documentation/issues/948. 

Is this going to be computed in the docs app? Consider keeping the calculation here in this script. I think it makes sense to have the computation of all the relevant attributes live in this script at build-time, rather than dynamically by the app at runtime. That avoids wasted computation when the app is running, and keeps the docs app simpler that it doesn't need to know how to calculate a function name given its full import path.

It probably makes sense to set the short name via a prop and the docs app will for now create the h3 for you, if it's set. That might allow us to avoid having to regenerate the MDX files again once we implement https://github.com/Qiskit/documentation/issues/948. It'd be nice to only have to change the MDX this first time.

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+ *
+ * This function works the same regardless of whether the Sphinx build used `sphinx.ext.viewcode`
+ * or `sphinx.ext.linkcode` because they have the same HTML structure.
+ *
+ * If the link corresponds to a method, we only return a link if it has line numbers included,
+ * which implies that the link came from `sphinx.ext.linkcode` rather than `sphinx.ext.viewcode`.
+ * That's because the owning class will already have a link to the relevant file; it's
+ * noisy and not helpful to repeat the same link without line numbers for the individual methods.
+ */
+export function prepareGitHubLink(
```
arnaucasau, line 303: The only change on this function is that now only returns the link that will be used as a prop

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+  }
+  const href = originalLink.attr("href")!;
+  originalLink.first().remove();
+  return !isMethod || href.includes(".py#") ? href : "";
+}
+
+/**
+ * Find the element that both matches the `selector` and whose content is the same as `text`
+ */
+function findByText(
```
arnaucasau, line 319: Same function as before

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+  }
+
+  const html = `<span class="target" id=''/><p><code>${signatureHtml}</code></p>`;
+  const file = await unified()
+    .use(rehypeParse)
+    .use(rehypeRemark)
+    .use(remarkStringify)
+    .process(html);
+
+  return String(file)
```
arnaucasau, line 356: The replaces handle the following cases:

- There were signatures ending with an end line
- The signatures have backticks that need to be removed at the beginning and at the end
- Transform the single quotes for their ASCII code to escape them

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+  const githubSourceLink = prepareGitHubLink(
+    $firstSignature,
+    apiType === "method",
+  );
+
+  const componentProps = prepareProps(
+    $,
+    $firstSignature,
+    $dl,
+    priorApiType,
```
arnaucasau, line 53: This argument needs to be different for overload signatures (`apiType`) than for the first call (`priorApiType`)

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+        $dl,
+        apiType,
+        apiType,
+        prepareGitHubLink($overloadedSignature, apiType === "method"),
+        id,
+      ),
+    )
+    .flatMap((prop) => (prop ? prop : []));
+
+  if (componentProps) {
```
arnaucasau, line 73: Properties and attributes can return without any props if there's no signature

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+function prepareProps(
+  $: CheerioAPI,
+  $child: Cheerio<Element>,
+  $dl: Cheerio<any>,
+  priorApiType: ApiType | undefined,
+  apiType: ApiType,
+  githubSourceLink: string,
+  id: string,
+): componentProps | undefined {
+  const preparePropsPerApiType: Record<
```
arnaucasau, line 96: Dispatch table to avoid repeating the same code only changing the `prepareProps` call

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+    ...props,
+    name: id.split(".").slice(-1)[0],
+  };
+}
+
+// ------------------------------------------------------------------
+// Generate MDX components
+// ------------------------------------------------------------------
+
+export async function generateMdxComponent(
```
arnaucasau, line 261: Single function to generate the component open tag. If any prop is empty will be removed by the unified plugin on `htmlToMd.ts`

scripts/lib/api/htmlToMd.ts
```diff
@@ -123,6 +123,24 @@ function prepareHandlers(meta: Metadata): Record<string, Handle> {
 
       return defaultHandlers.div(h, node);
     },
+    class(h, node: any): any {
+      return buildMdxComponent(h, node);
+    },
+    property(h, node: any): any {
+      return buildMdxComponent(h, node);
+    },
```
Eric-Arellano, line 131: Can we combine property and attribute? From an end-user perspective, I think they show up the same. I see that in the conversion pipeline, they are different; but maybe we can have them use the same component still?

scripts/lib/api/htmlToMd.ts
```diff
@@ -301,3 +319,76 @@ function buildMathExpression(node: any, type: "math" | "inlineMath"): any {
   }
   return { type: type, value };
 }
+
+function buildMdxComponent(h: H, node: any): any {
```
Eric-Arellano, line 323: I recommend `buildApiComponent`.

scripts/lib/api/htmlToMd.ts
```diff
@@ -301,3 +319,76 @@ function buildMathExpression(node: any, type: "math" | "inlineMath"): any {
...
+  const apiType = node.tagName.charAt(0).toUpperCase() + node.tagName.slice(1);
```
Eric-Arellano, line 324: Nit: consider creating a `capitalize` function in `stringUtils.tsx` since this function already has a lot going on.

Eric-Arellano: Also `componentName` might be more clear than `apiType`. You want to make clear this is referring to the specific component from MDX.

scripts/lib/api/htmlToMd.ts
```diff
@@ -301,3 +319,76 @@ function buildMathExpression(node: any, type: "math" | "inlineMath"): any {
...
+      {
+        type: "mdxJsxAttribute",
+        name: "id",
+        value: node.properties.id,
+      },
+    ],
+    children: all(h, node),
+  };
+
+  addSimpleAttributeToHastTree(hastTree, "name", node.properties.name);
```
Eric-Arellano, line 339: Nit: consider calling this `maybeAddAttribute`

1. the `maybe` thing expresses this depends on if the attribute is defined
2. imo you can cut `toHastTree` because the first arg already makes that clear.

scripts/lib/api/htmlToMd.ts
```diff
@@ -301,3 +319,76 @@ function buildMathExpression(node: any, type: "math" | "inlineMath"): any {
...
+  if (value && value != "undefined") {
+    hastTree.attributes.push({
+      type: "mdxJsxAttribute",
+      name,
+      value,
+    });
+  }
+}
+
+function addExpressionAttributeToHastTree(
```
Eric-Arellano, line 379: `maybeAddExpressionAttribute`

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+import { CheerioAPI, Cheerio, Element } from "cheerio";
+import { unified } from "unified";
+import rehypeParse from "rehype-parse";
+import rehypeRemark from "rehype-remark";
+import remarkStringify from "remark-stringify";
+
+import { ApiType } from "./Metadata";
+import { getLastPartFromFullIdentifier } from "../stringUtils";
+
+export type componentProps = {
```
Eric-Arellano, line 22: Nit: types usually are UpperCase.

scripts/lib/api/generateMdxComponents.ts
Eric-Arellano: Thoughts on it returning either `string | undefined`? The surrounding logic is more complex now, so I think the more explicit modeling might be worth it.

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+  $main: Cheerio<any>,
+  selector: string,
+  text: string,
+): Cheerio<any> {
+  return $main.find(selector).filter((i, el) => $(el).text().trim() === text);
+}
+
+export function mergeProps(
+  componentProps: componentProps,
+  props: componentProps[],
```
Eric-Arellano, line 330: This name is a bit confusing. Maybe call it `extraSignatures: ComponentProps[]`? And call the function `addExtraSignatures` or `mergeExtraSignatures`?

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+): void {
+  componentProps.extraSignatures = [];
+  for (const prop of props) {
+    if (props.length == 0 || !prop.signature) {
```
Eric-Arellano, line 334: Why check `props.length == 0`? In that case, we won't enter the for loop.

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+      continue;
+    }
+    componentProps.extraSignatures.push(prop.signature);
```
Eric-Arellano, line 337:
```typescript
componentProps.extraSignatures.push(
  ...extraSignatures
     .map((sigProps) => sigProps.signature)
     .filter((sig) => sig)
)
```

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
+// This code is a Qiskit project.
```
Eric-Arellano, line 1: How about calling this file `generateApiComponents.ts`?

scripts/lib/api/generateMdxComponents.ts
Eric-Arellano: Nice. This would be good to reflect in the code:

1. Consider calling the function `createOpeningTag`
2. It'd be good to have a comment above the JSX explaining that about removing empty props.

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+// ------------------------------------------------------------------
+// Generate MDX components
+// ------------------------------------------------------------------
+
+export async function generateMdxComponent(
+  apiType: ApiType,
+  props: componentProps,
+): Promise<string> {
+  const type = props.attributeType
+    ? props.attributeType.replaceAll("'", "&#x27;")
```
Eric-Arellano, line 266: `&#x27;` is used in a few places. It'd be good to define a constant for that.

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+    : undefined;
+  const value = props.attributeValue
+    ? props.attributeValue.replaceAll("'", "&#x27;")
```
Eric-Arellano, line 269: Nit: you can use `props.attributeValue?.replaceAll("'", "&#x27;")`

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+        $,
+        $overloadedSignature,
+        $dl,
+        apiType,
+        apiType,
+        prepareGitHubLink($overloadedSignature, apiType === "method"),
+        id,
+      ),
+    )
+    .flatMap((prop) => (prop ? prop : []));
```
Eric-Arellano, line 71: I think you can call flatMap directly on line 60? I'm not following what this is doing.

scripts/lib/api/generateMdxComponents.ts
Eric-Arellano: What does that mean? Will they not get a `<Attribute>` component?

scripts/lib/api/generateMdxComponents.ts
```diff
@@ -0,0 +1,361 @@
...
+  apiType: ApiType,
+  props: componentProps,
+): Promise<string> {
+  const type = props.attributeType
+    ? props.attributeType.replaceAll("'", "&#x27;")
+    : undefined;
+  const value = props.attributeValue
+    ? props.attributeValue.replaceAll("'", "&#x27;")
+    : undefined;
+  const signature = await htmlSignatureToMd(props.signature!);
```
Eric-Arellano, line 271: Consider calling it `props.rawSignature` to signify that it still needs to be processed. That reduces the risk of someone not realizing they need to do that.

Eric-Arellano (commented):
👏

scripts/lib/api/generateMdxComponents.ts
arnaucasau: Oops :eyes:

scripts/lib/api/generateMdxComponents.ts
arnaucasau: Yes. There's a case in attributes and properties where you could have empty props. I haven't seen any case, but our script allowed it before, so I maintained the option here to not have any unexpected behavior. If the attribute/property doesn't have an id, I think we shouldn't treat it as a component, wdyt?

scripts/lib/api/htmlToMd.ts
arnaucasau: Good idea! As we discussed, I combined class-exception, attribute-property, and function-method

scripts/lib/api/generateMdxComponents.ts
arnaucasau: That's right. I changed for `extraRawSignatures` following the same advice as in `rawSignatures` :+1:

scripts/lib/api/generateMdxComponents.ts
arnaucasau: That was my first idea, but the filter doesn't work very well with undefined or nulls. The return value is still `string | undefined` afterward.

Nonetheless, I like the idea, and I think we can make it more compact with flatMap:

```ts
componentProps.extraRawSignatures = [
  ...extraRawSignatures.flatMap((sigProps) => sigProps.rawSignature ?? []),
];
```

scripts/lib/api/generateMdxComponents.ts
arnaucasau: Yeah, good catch!

scripts/lib/api/generateMdxComponents.ts
Eric-Arellano: I'm a big fan of removing edge cases when possible. If we haven't seen the edge case before with over 60 docs versions, we're unlikely to see it now. So it would be helpful to fail eagerly. Run over our current docs with this eager error to verify we indeed never have the edge case.

Keep in mind that the original script might be overly generalized in some places because we didn't always have a perfect understanding of Sphinx, so we might have defensively added allowances that aren't actually necessary. Or, you've seen cases where there are edge cases we didn't know about.

scripts/lib/api/generateApiComponents.ts
```diff
@@ -0,0 +1,378 @@
...
+import rehypeRemark from "rehype-remark";
+import remarkStringify from "remark-stringify";
+
+import { ApiType } from "./Metadata";
+import {
+  getLastPartFromFullIdentifier,
+  APOSTROPHE_HEX_CODE,
+} from "../stringUtils";
+
+export type COMPONENT_PROPS = {
```
Eric-Arellano, line 25: Types are UpperCase, not ALL_CAPS. ALL_CAPS is only for constants.

scripts/lib/api/mergeClassMembers.ts
```diff
@@ -63,20 +64,33 @@ export async function mergeClassMembers(
...
-                  "Methods Defined Here",
-                  methods,
-                );
+              // The attribute and method's section can be found under the class component
+              const mdxClassElements = tree.children.filter(
+                (node): node is MdxJsxFlowElement =>
+                  node.type == "mdxJsxFlowElement" && node.name == "Class",
+              );
+
+              for (const tree of mdxClassElements) {
```
arnaucasau, line 73: Probably we can remove this first loop as we only have one `Class` component. I'll test it and if that's the case I'll simplify it :+1:

scripts/lib/api/generateApiComponents.ts
```diff
@@ -28,6 +32,225 @@ export type ComponentProps = {
...
+  );
+
+  const extraProps = signatures.flatMap(
+    ($overloadedSignature) =>
+      prepareProps($, $main, $overloadedSignature, $dl, apiType, apiType, id) ??
+      [],
+  );
+
+  const tagName = APITYPE_TO_TAG[apiType];
+  addExtraSignatures(componentProps, extraProps);
```
Eric-Arellano, line 72: Nit: I recommend moving `addExtraSignatures` to live right next to `extraProps` because those two things are related. `tagName` can be moved to the top of the function because it's independent of the rest.

scripts/lib/api/generateApiComponents.ts
```diff
@@ -52,11 +275,7 @@ export async function createOpeningTag(
   const signature = await htmlSignatureToMd(props.rawSignature!);
   const extraSignatures: string[] = [];
   for (const sig of props.extraRawSignatures ?? []) {
-    extraSignatures.push(
-      `${APOSTROPHE_HEX_CODE}${await htmlSignatureToMd(
-        sig!,
-      )}${APOSTROPHE_HEX_CODE}`,
-    );
+    extraSignatures.push(`"${await htmlSignatureToMd(sig!)}"`);
```
Eric-Arellano, line 278:
```suggestion
    extraSignatures.push(await htmlSignatureToMd(sig!));
```

Eric-Arellano (approved):
Nice! I'm confused by `processHtml.test.ts`: why the tags from `generateApiComponents.ts` aren't showing up? I'm trying to figure out if the unit tests we have are still useful.

scripts/lib/api/generateApiComponents.ts
arnaucasau: Notice that we have quotes surrounding the extra signature. We can't remove them because we need the signatures as a string inside the `extraSignatures` array. The result is something like `<Function ... extraSignatures = {["sig 1", "sig 2"]}>` where each signature has the quotes

arnaucasau:
> I'm confused by `processHtml.test.ts`

Nice catch!! We weren't using an `async` test and the `processMembersAndSetMeta` call didn't have an `await`. Thanks for the feedback Eric!

Eric-Arellano (approved):
Nice! `processHtml.test.ts` looks great now.

scripts/lib/api/mergeClassMembers.ts
```diff
@@ -62,17 +63,30 @@ export async function mergeClassMembers(
...
-            return async (tree: Root) => {
-              for (const node of tree.children) {
+            return async (root: Root) => {
+              // The attribute and method's section can be found under the class component
+              const mdxClassElement = root.children
+                .filter(
+                  (node): node is MdxJsxFlowElement =>
+                    node.type == "mdxJsxFlowElement" && node.name == "Class",
+                )
+                .shift();
```
Eric-Arellano, line 73:
```suggestion
                .find(
                  (node): node is MdxJsxFlowElement =>
                    node.type == "mdxJsxFlowElement" && node.name == "Class",
                );
```

scripts/lib/api/mergeClassMembers.ts
```diff
@@ -62,17 +63,30 @@ export async function mergeClassMembers(
...
+
+              for (const node of mdxClassElement?.children ?? []) {
                 await replaceMembersAfterTitle(
-                  tree,
+                  mdxClassElement!,
```
Eric-Arellano, line 77: You can make the types and code more clear by early returning if `mdxClassElement` is undefined rather than entering this for loop

Eric-Arellano (approved):
🚀

Eric-Arellano (approved):

PR #1035, opened 2024-03-14
Improvements to "Commonly used parameters for transpilation"

kevinsung:
Part of https://github.com/Qiskit/documentation/issues/885

- Don't use `transpile` function
- Rewrite the "Initial layout" section to add code examples and correct information about what arguments are allowed.


docs/transpile/common-parameters.ipynb
javabster, line 1: for the "Random number generator seed" title I think we should keep the original "seed transpiler" wording so it's consistent with the other titles (i.e. the titles match the actual name of the argument they are describing)

javabster, line 1: "In the example below, we construct an initial layout for the FakeSherbrooke mock backend by creating a [Layout](/api/qiskit/qiskit.transpiler.Layout) object"

`FakeSherbrooke` not FakeSherbrooke

javabster: "You can visit https://quantum.ibm.com/services/resources to access device graph diagrams with physical qubits labeled."

I think there may also be a function in qiskit that lets you do this? Might be a nice addition

javabster (commented):
Overall liking the improvements to this piece, just left a couple of minor comments :)

docs/transpile/common-parameters.ipynb
kevinsung: I think the titles should either be
1. The name of the concept (random number generator seed)
2. The exact name of the argument (`seed_transpiler`)

What do you think? If you want to go with (2) then I'll change the other titles to match.

kevinsung: > `FakeSherbrooke` not FakeSherbrooke

Done.

> I think there may also be a function in qiskit that lets you do this? Might be a nice addition

For 127 qubits the output has overlapping text and is pretty ugly (see image below). I don't think we should include it. I could just refer them to the function though. What do you think?

image

javabster: hmm I prefer 2. but I know that we've got a css bug with code blocks in headers so maybe we should stick with 1 then if 2 looks terrible 🤔

javabster: is the overlapping text fixed if you make the matplotlib axis bigger?

kevinsung: Yes, thanks for the suggestion. Done!

kevinsung: As we discussed in person, I went with (1).

javabster (approved):
LGTM 🚀

PR #1038, opened 2024-03-14
Fix ISA note 

beckykd:
Sampler V2 doesn't allow optimization level. Also, some instances still mentioned skip_transpilation, which is not supported in V2.


abbycross (approved):
👍

Issue #1043, opened 2024-03-18
Migrate away from IBM Code Engine for PR previews


Eric-Arellano:
We're using IBM Code Engine for staging and PR previews via https://github.com/Qiskit/documentation/blob/main/.github/workflows/code-engine-preview.yml.

We'll lose account access on June 14, 2024. So we need to migrate to an alternative mechanism to deploy the app.

PR #1045, opened 2024-03-18
[wip] Use preview Docker image for staging and PR previews

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/74. 

We have not been able to figure out why the Docker image will not launch with IBM Code Engine when using the Staging Docker image rather than the Preview Docker image. We can't get logs because our logging plan expired for the IBM Cloud account, so it's hard to debug.

The main downside of using the Preview app is that it doesn't have a home page and a top nav bar, so you have to explicitly navigate to the URL for the folder you want. Once you're in the folder, the left table of contents still works.

Everything else renders the same between Staging and Preview. Not having search is irrelevant because that never worked.

The benefits of this change are:

* Closing https://github.com/Qiskit/documentation/issues/74, which unblocks us using SVGs in notebooks and also fixes how slow the preview is
* Simpler infrastructure, including in the closed source repo because we only have to publish one Docker image

Given that we're going to need to migrate away from Code Engine anyways in 3 months (https://github.com/Qiskit/documentation/issues/1043), the benefits of this change seem worth the downside of you having to explicitly navigate to the folder you want.


Eric-Arellano:
Superseded by https://github.com/Qiskit/documentation/pull/990

PR #1047, opened 2024-03-18
Run notebooks that submit jobs in cron job

### LINKED ISSUE #917: Set up cron job to test notebooks that submit jobs
Action item from https://github.com/Qiskit/documentation/issues/133.

We want to test that the notebooks that submit jobs to actual hardware work correctly. We don't run them in normal CI because they are so slow and cost resources.

https://github.com/Qiskit/documentation/blob/8158062c29fc0157d61a3bfc99e61cc658b0eb6a/scripts/nb-tester/test-notebook.py#L35-L37

We still shouldn't them in PRs, but it would be good to test them in a cron job, at least every month and possibly every 2-3 weeks. The `nb-tester` script already supports that:

https://github.com/Qiskit/documentation/blob/8158062c29fc0157d61a3bfc99e61cc658b0eb6a/scripts/nb-tester/test-notebook.py#L232-L240

We also already have the Qiskit Runtime account credentials set up in CI. So this ticket is only setting up the GitHub workflow to test all notebooks that have jobs every 2-4 weeks. Bonus points if we create an issue on failures (relates to https://github.com/Qiskit/documentation/issues/547).

I am a little concerned if we'll exceed GitHub Actions' timeout. We should investigate that when working on this.

frankharkins:
- Added a new flag `--only-submit-jobs` to only execute notebooks that submit jobs. Another option is to just test all the notebooks.
- Added a new action to run this once per week

Also neatens up the script a little.

***

Closes #917


scripts/nb-tester/test-notebook.py
```diff
@@ -165,19 +173,17 @@ def _execute_notebook(filepath: Path, options: ExecuteOptions) -> nbformat.Noteb
     return nb
 
 
-def find_notebooks(*, submit_jobs: bool = False) -> list[Path]:
```
frankharkins, line 168: This is already caught in `filter_paths`

.github/workflows/notebook-weekly-test.yml
```diff
@@ -0,0 +1,81 @@
...
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+name: Test notebooks that submit jobs
+on:
+  schedule:
+    - cron: "0 0 * * 1" # At 00:00 on Monday
```
Eric-Arellano, line 16: We were thinking of every other week or even every month (probably too infrequent). Simply to conserve resources. The downside is it taking longer to catch problems.

Maybe we go with every two weeks?

.github/workflows/notebook-weekly-test.yml
```diff
@@ -0,0 +1,81 @@
...
+        run: |
+          sudo apt-get update
+          sudo apt-get install texlive-pictures texlive-latex-extra poppler-utils graphviz
+
+      - name: Install Python packages
+        # This is to save our account in the next step. Note that the
+        # package will be re-installed during the "Run tox" step.
+        run: pip install qiskit-ibm-runtime tox
+
+      - name: Save IBM Quantum account
```
Eric-Arellano, line 39: Thoughts on DRYing this and preceding steps by creating a custom action locally? https://docs.github.com/en/actions/creating-actions

Not a deal breaker, but might be interesting to learn and helps keep the code maintainable.

.github/workflows/notebook-weekly-test.yml
```diff
@@ -0,0 +1,81 @@
...
+        shell: python
+        run: |
+          # This saves the result for qiskit-ibm-provider too
```
Eric-Arellano, line 42: What does this mean?

.github/workflows/notebook-weekly-test.yml
```diff
@@ -0,0 +1,81 @@
...
+            Please [check the logs](https://github.com/Qiskit/documentation/actions/runs/${{ github.run_id }}/).
+            > [!NOTE]
+            > This issue was created by a GitHub action.
+            `
+            github.rest.issues.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              title: "Weekly notebook test failed",
+              body: message,
+              assignees: ["frankharkins", "javabster",]
```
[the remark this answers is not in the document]
Eric-Arellano: Maybe Kevin also?

scripts/nb-tester/test-notebook.py
```diff
@@ -165,19 +173,17 @@ def _execute_notebook(filepath: Path, options: ExecuteOptions) -> nbformat.Noteb
...
+def find_notebooks() -> list[Path]:
     """
     Get paths to all notebooks in NOTEBOOKS_GLOB that are not excluded by
     NOTEBOOKS_EXCLUDE
     """
     all_notebooks = Path(".").rglob(NOTEBOOKS_GLOB)
     excluded_notebooks = NOTEBOOKS_EXCLUDE
```
Eric-Arellano, line 182: Can inline this variable

Eric-Arellano (commented):
Nice!

.github/workflows/notebook-weekly-test.yml
frankharkins: Sounds good, done :)

.github/workflows/notebook-weekly-test.yml
frankharkins: Good call, done in 0bc5f57 and I think it's much neater. Makes the individual actions smaller too.

.github/workflows/notebook-weekly-test.yml
frankharkins: This sets up the account for both `qiskit-ibm-runtime` and `qiskit-ibm-provider`. Originally I thought we'd need to do both. I commented as it surprised me, but I'll remove it as I can't imagine it being that helpful to someone in the future.

.github/workflows/notebook-test.yml
```diff
@@ -46,35 +42,11 @@ jobs:
...
-
-      - name: Cache tox environment
-        uses: actions/cache@v3
+      - name: Setup environment
+        uses: ./.github/workflows/reusable/setup-notebook-testing
         with:
-          path: ".tox"
-          key: ${{ hashFiles('scripts/nb-tester/requirements.txt') }}
+          install-linux-deps: ${{ steps.linux-changed-files.outputs.any_changed }}
+          ibm-quantum-token: ${{ secrets.IBM_QUANTUM_TEST_TOKEN }}
```
Eric-Arellano, line 49: Instead, I think you're supposed to pass it via `secrets`. But maybe not because you're using an action over worfklow?

https://github.com/Qiskit/documentation/blob/9ff00250401453df6a164185e9038f57f6579d42/.github/workflows/preview-docs.yml#L36-37

.github/workflows/notebook-weekly-test.yml
```diff
@@ -0,0 +1,55 @@
...
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+name: Test notebooks that submit jobs
+on:
+  schedule:
+    - cron: "0 0 1,15 * *" # At 00:00 on day-of-month 1 and 15.
+  workflow_dispatch:
+  pull_request:
```
Eric-Arellano, line 18: Leftover?

.github/workflows/reusable/setup-notebook-testing/action.yml
```diff
@@ -0,0 +1,55 @@
+# This code is a Qiskit project.
```
Eric-Arellano, line 1: I think the convention is to put this in `.github/actions/` rather than `.github/workflows`

scripts/nb-tester/test-notebook.py
```diff
@@ -236,19 +241,27 @@ def create_argument_parser() -> argparse.ArgumentParser:
...
     return parser
 
 
 if __name__ == "__main__":
     args = create_argument_parser().parse_args()
-
-    paths = map(Path, args.filenames or find_notebooks(submit_jobs=args.submit_jobs))
-    filtered_paths = filter_paths(paths, submit_jobs=args.submit_jobs)
+    paths = map(Path, args.filenames or find_notebooks())
+    filtered_paths = filter_paths(paths, submit_jobs=args.submit_jobs, only_submit_jobs=args.only_submit_jobs)
```
Eric-Arellano, line 255: Consider having this also take `ExecuteOptions`.

.github/workflows/notebook-weekly-test.yml
frankharkins: Sorry; I wasn't quite ready to request the next review. The test is failing but I can't tell why yet. Unfortunately it takes a couple of hours to run so debugging is slow 😢

frankharkins: Although now I'm thinking maybe testing in an action is not worth it if a single notebook takes that long?

Eric-Arellano: What else would we do? Regular reminder for a human to test? What about one CI workflow per notebook that needs it?

.github/workflows/notebook-test.yml
frankharkins: According to [SO](https://stackoverflow.com/a/70111134) (I can't find an official source) actions must use inputs

.github/workflows/notebook-weekly-test.yml
frankharkins: I accepted defeat too quickly. I've given the job-submitting workflows internal access and they're going through immediately this morning. Queue times are usually shorter in the mornings (my time) so hopefully these two factors are enough.

Let's continue with the cron job and see how reliable it is.

scripts/nb-tester/test-notebook.py
```diff
@@ -193,7 +192,13 @@ def cancel_trailing_jobs(start_time: datetime) -> bool:
...
     # QiskitRuntimeService().jobs() includes qiskit-ibm-provider jobs too
     service = QiskitRuntimeService()
-    jobs = [j for j in service.jobs(created_after=start_time) if not j.in_final_state()]
+
+    def _is_not_finished(job):
+        # Force runtime to update job status
+        job.status()
+        return not job.in_final_state()
+
+    jobs = list(filter(_is_not_finished, service.jobs(created_after=start_time)))
```
frankharkins, line 201: Workaround for https://github.com/Qiskit/qiskit-ibm-runtime/issues/1547

scripts/nb-tester/test-notebook.py
frankharkins: Good call. I realised `ExecuteOptions` is now basically the same as the args object so I'm just using that instead.

.github/workflows/reusable/setup-notebook-testing/action.yml
frankharkins: Done, thanks!

scripts/nb-tester/test-notebook.py
Eric-Arellano: It'd be good to link to this in a comment so that we don't keep it forever.

Eric-Arellano (approved):
Thanks! Looks good, pending confirmation from the team about frequency of the cron job. Thanks for making the GH Action.

scripts/nb-tester/test-notebook.py
frankharkins:
```suggestion
    def _is_not_finished(job):
        # Force runtime to update job status
        # Workaround for Qiskit/qiskit-ibm-runtime#1547
        job.status()
        return not job.in_final_state()

    jobs = list(filter(_is_not_finished, service.jobs(created_after=start_time)))
```

PR #1049, opened 2024-03-18
Add default posarg to tox.ini

frankharkins:
The linting instructions are actually lies as they you should run `tox -e fix` but this doesn't work unless you pass the files as a posarg. This PR bumps the version of Squeaky to one that supports recursive searching and sets a default posarg of `docs/` so writers can just run `tox -e fix` to clean everything.


Eric-Arellano (approved):
Nice. Thanks!

PR #1018, opened 2024-03-12
convert "Represent quantum computers" from MDX to Jupyter notebook

kevinsung:
part of #169

PR #1059, opened 2024-03-19
Improve "Represent quantum computers" page

### LINKED ISSUE #885: Refactor "Configure transpilation" section to use preset pass managers instead of transpile function

kevinsung:
Closes #885 

- Use preset pass managers instead of `transpile` function
- Rewrite prose and simplify code
- Add back section on device error rates that was removed in #1018


javabster (approved):
LGTM 🚀

PR #1069, opened 2024-03-21
Generate Runtime 0.22 docs

Eric-Arellano:
This is our first time regenerating the Runtime docs with towncrier as the release notes generator rather than Reno.


Eric-Arellano:
Double checked the API docs with Kevin Tian over Slack.

PR #1074, opened 2024-03-22
Add `generateApiComponents.ts` and tests

arnaucasau:
Part of https://github.com/Qiskit/documentation/issues/1008

This PR is a precursor of #1026, and it creates a new script named `generateApiComponents.ts` with some tests. This script will have all the logic to generate the MDX components, but, in the meantime, this PR adds some functions that can be tested independently by just receiving, as parameters, a list of props or HTML code.

It also copies over the functions `prepareGitHubLink` and `findByText` from `processHtml.ts`

### DIFF
--- a/scripts/lib/api/generateApiComponents.test.ts
+++ b/scripts/lib/api/generateApiComponents.test.ts
@@ -1,0 +1,206 @@
+// This code is a Qiskit project.
+//
+// (C) Copyright IBM 2024.
+//
+// This code is licensed under the Apache License, Version 2.0. You may
+// obtain a copy of this license in the LICENSE file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+import { expect, test } from "@jest/globals";
+import { CheerioAPI, Cheerio, load as cheerioLoad } from "cheerio";
+
+import {
+  ComponentProps,
+  prepareGitHubLink,
+  htmlSignatureToMd,
+  addExtraSignatures,
+  createOpeningTag,
+} from "./generateApiComponents";
+
+class Doc {
+  readonly $: CheerioAPI;
+  readonly $main: Cheerio<any>;
+
+  constructor($: CheerioAPI, $main: Cheerio<any>) {
+    this.$ = $;
+    this.$main = $main;
+  }
+
+  static load(html: string): Doc {
+    const $ = cheerioLoad(`<div role="main">${html}</div>`);
+    return new Doc($, $("[role='main']"));
+  }
+
+  html(): string {
+    return this.$main.html()!.trim();
+  }
+
+  expectHtml(expected: string): void {
+    expect(this.html()).toEqual(expected.trim());
+  }
+}
+
+test("htmlSignatureToMd", async () => {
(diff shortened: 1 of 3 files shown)


scripts/lib/stringUtils.ts
```diff
@@ -12,6 +12,8 @@
 
 import { last, split } from "lodash";
```
Eric-Arellano, line 14: You could add `capitalize` in this PR

scripts/lib/api/generateApiComponents.ts
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * This function works the same regardless of whether the Sphinx build used `sphinx.ext.viewcode`
+ * or `sphinx.ext.linkcode` because they have the same HTML structure.
+ *
+ * If the link corresponds to a method, we only return a link if it has line numbers included,
+ * which implies that the link came from `sphinx.ext.linkcode` rather than `sphinx.ext.viewcode`.
+ * That's because the owning class will already have a link to the relevant file; it's
+ * noisy and not helpful to repeat the same link without line numbers for the individual methods.
+ */
+export function prepareGitHubLink(
```
Eric-Arellano, line 82: Not a big deal if this is going to be annoying to implement, but you can switch processHtml.md to use this now, along with `findByText`. That way we don't duplicate. And you could move the tests too if it's too hard.

scripts/lib/api/generateApiComponents.ts
```diff
@@ -0,0 +1,138 @@
...
+  return $main.find(selector).filter((i, el) => $(el).text().trim() === text);
+}
+
+export function addExtraSignatures(
+  componentProps: ComponentProps,
+  extraRawSignatures: ComponentProps[],
+): void {
+  componentProps.extraRawSignatures = [
+    ...extraRawSignatures.flatMap((sigProps) => sigProps.rawSignature ?? []),
+  ];
```
Eric-Arellano, line 113: Using `flatMap` surprises me because `rawSignature` is a string. Should this be

```
componentProps.extraRawSignatures = Array.from(
    extraRawSignatures.map((x) => x.signature).filter((sig) => sig !== undefined)
)
```

Also can `extraRawSignatures` have a more precise type like `string[]` or `Array<string | undefined>`? Seems weird to have all the ComponentProps if we're going to discard them.

scripts/lib/api/generateApiComponents.ts
```diff
@@ -0,0 +1,138 @@
...
+ * Converts a given HTML into markdown
+ */
+export async function htmlSignatureToMd(
+  signatureHtml: string,
+): Promise<string> {
+  if (!signatureHtml) {
+    return "";
+  }
+
+  const html = `<span class="target" id=''/><p><code>${signatureHtml}</code></p>`;
```
Eric-Arellano, line 126: Why is `id` an empty string?

scripts/lib/api/generateApiComponents.ts
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Creates the opening tag of the API components. The function sets all possible
+ * props values even if they are empty or undefined. All the props without value
+ * will be removed when generating the markdown file in `htmlToMd.ts`.
+ */
+export async function createOpeningTag(
+  tagName: string,
+  props: ComponentProps,
+): Promise<string> {
+  const type = props.attributeTypeHint?.replaceAll("'", APOSTROPHE_HEX_CODE);
```
Eric-Arellano, line 44: I'd call this `attributeTypeHint` and the below `attributeValue` to make clear this isn't relevant to other components.

scripts/lib/api/generateApiComponents.test.ts
```diff
@@ -0,0 +1,224 @@
...
+
+import {
+  ComponentProps,
+  prepareGitHubLink,
+  htmlSignatureToMd,
+  addExtraSignatures,
+  createOpeningTag,
+} from "./generateApiComponents";
+
+class Doc {
```
Eric-Arellano, line 24: Should probably deduplicate with `processHtml`. I think there can be issues with test files importing other test files. If so, you can define a new `lib/testUtils.ts`. Maybe call this type `CheerioDoc`.

scripts/lib/api/generateApiComponents.test.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    const tag = await createOpeningTag("Function", componentProps);
+    expect(tag).toEqual(`<Function 
+    id='qiskit_ibm_runtime.Estimator.run'
+    name='run'
+    attributeTypeHint='undefined'
+    attributeValue='undefined'
+    github='undefined'
+    signature='Estimator.run(circuits, observables, parameter_values=None, **kwargs)'
+    extraSignatures='[&#x27;Estimator.run(circuits, observables, parameter_values=None, **kwargs)&#x27;, &#x27;Estimator.run(circuits, observables, parameter_values=None, **kwargs)&#x27;]'
```
Eric-Arellano, line 127: Probably should de-magicify `&#x27;`

Eric-Arellano (commented):
Nice!

scripts/lib/api/generateApiComponents.ts
arnaucasau: It's empty because it's not needed, but I was able to remove the span and the paragraph too :+1: . Apparently, we only need to add the code tag. Without the code tag, the signature has an extra apostrophe at the beginning and at the end

scripts/lib/api/generateApiComponents.ts
arnaucasau: In reality, `rawSignature` is a `string` but could be `undefined` given that we have API types without it. The problem with using the `filter` is that we get the type `(string | undefined)[]` instead of `string[] | undefined` as in the `flatMap`.

> Also can `extraRawSignatures` have a more precise type 

This is a good idea. I think we could remove the `flatMap` or the `map` + `filter` entirely, but I would like to do it in #1026 as a follow-up because we can simplify the main function by only storing the overloaded signature and not all props. The function `addExtraSignatures` will make all the work as a black box

scripts/lib/stringUtils.ts
arnaucasau: Done!

scripts/lib/api/generateApiComponents.ts
Eric-Arellano: Actually, it probably would be good to do this if isn't too much of a pain because it may take a few days to land the follow up PR since it's blocked by the Docker image being updated for staging.

Eric-Arellano (approved):

PR #1084, opened 2024-03-25
Elaborate on the use of a `Target`

### LINKED ISSUE #1071: Elaborate on the `Target` object
Edit subsection on the `Target` object on the _Representing Quantum Computers_ page. Make more explicit what the `Target` object is and what it's used for.

kaelynj:
Closes #1071


docs/transpile/representing_quantum_computers.ipynb
```diff
@@ -5,9 +5,19 @@
    "id": "817782c5-8d62-492b-9633-6d57307e5d4b",
    "metadata": {},
    "source": [
-    "# Represent quantum computers\n",
+    "# Representing quantum computers for the transpiler\n",
```
frankharkins, line 8: Nice change. We had a page in the old textbook called "Representing quantum computers" that was actually about the states of the computer rather than transpiler properties.

docs/transpile/representing_quantum_computers.ipynb
```diff
@@ -74,7 +84,7 @@
    "source": [
     "### Transpile the circuit to `backend` target\n",
     "\n",
-    "This example uses the default `optimization_level=1` to transpile to the `backend` `target`, which providers all the information to the transpiler that is necessary to convert circuit to one that will run on the system."
+    "This example uses the default `optimization_level=1` to transpile to the `backend`'s `target`, which providers all the information to the transpiler that is necessary to convert circuit to one that will run on the system.  (Note that if a `target` is specified in the `transpile` function, this will override the target from the `backend`.)"
```
frankharkins, line 87:
```suggestion
    "This example uses the default `optimization_level=1` to transpile to the `backend`'s `target`, which provides all the information needed to convert circuit to one that will run on the system. (Note that if a `target` is specified in the `transpile` function, this will override the target from the `backend`.)"
```

docs/transpile/representing_quantum_computers.ipynb
```diff
@@ -102,21 +112,27 @@
...
     "\n",
     "The qubits from our abstract circuits are called _virtual qubits_ and those on the coupling map are _physical qubits_. The transpiler provides a mapping between virtual and physical qubits. One of the first steps in transpilation, the _routing_ stage, performs this mapping.\n",
     "\n",
     "<Admonition type=\"note\">\n",
-    "Although the routing stage is intertwined with the _layout_ stage, which selects the actual qubits, we will consider them as separate stages for simplicity. The combination of routing and layout is called _qubit mapping_.  Learn more about these stages in the [Transpiler stages](transpiler-stages) topic.\n",
+    "Although the routing stage is intertwined with the _layout_ stage -- which selects the actual qubits -- by default, we will consider them as separate stages for simplicity. The combination of routing and layout is called _qubit mapping_.  Learn more about these stages in the [Transpiler stages](transpiler-stages) topic.\n",
```
frankharkins, line 135:
```suggestion
    "Although the routing stage is intertwined with the _layout_ stage — which selects the actual qubits — by default, we will consider them as separate stages for simplicity. The combination of routing and layout is called _qubit mapping_.  Learn more about these stages in the [Transpiler stages](transpiler-stages) topic.\n",
```

frankharkins (approved):
Nice improvement, thanks! I have a couple of minor suggestions.

docs/transpile/representing_quantum_computers.ipynb
abbycross: Is it possible to still remove the -ing from "Representing"? Stylistically we're trying to keep consistency with directly active words (no -ing) but if there's a case to keep it, we of course can leave it as-is.

docs/transpile/representing_quantum_computers.ipynb
```diff
@@ -5,9 +5,15 @@
...
+    "This page overviews how the constraints of a `backend` are represented and what the transpiler does with that information. In order to construct an optimized circuit that can run on a specific system, the transpiler needs certain details about the system.  This information is found in two places: the `backend` object you plan to submit jobs to, and the backend's `Target` attribute.\n",
+    "\n",
+    "- The [`Target`](../api/qiskit/qiskit.transpiler.Target) contains all the relevant constraints of a device, such as its native basis gates, qubit connectivity, and pulse or timing information.\n",
+    "- The [`Backend`](../api/qiskit/qiskit.providers.BackendV2) typically possesses a `Target` by default and contains additional information, such as the most recently calibrated error rates or the [`InstructionScheduleMap`](../api/qiskit/qiskit.pulse.InstructionScheduleMap).\n",
+    "\n",
+    "\n",
+    "The precision with which the transpiler produces the most appropriate circuit for specific hardware depends on how much information the `Target` or `Backend` has about its constraints.\n",
```
abbycross, line 16: @kaelynj pls review what I changed this line to in my recent commit -- I was trying to make it smoother and more translatable but I am not sure if I changed the overall meaning?...

abbycross:
I notice that sometimes `Target` and `Backend` are capitalized and sometimes they aren't - do they always refer to the same thing, in which case they can all be consistent -- or does it signify a difference in function?

docs/transpile/representing_quantum_computers.ipynb
kaelynj: The -ing makes sense to me mostly because using just "Represent" in the title reads as a verb and implies that we're going to do some action in order to represent quantum computers to the transpiler, rather than a passive description of how quantum computers are represented for the transpiler.  But I do understand the desire to keep everything stylistically consistent.

kaelynj:
> I notice that sometimes `Target` and `Backend` are capitalized and sometimes they aren't - do they always refer to the same thing, in which case they can all be consistent -- or does it signify a difference in function?

Nope, just edited the page to make sure it was consistently capitalized.

docs/transpile/representing_quantum_computers.ipynb
kaelynj: Looks good to me!

abbycross:
Just noticed we should update the _toc.json file so that the page title matches the table of contents. (Right now the toc still says "Represent quantum computers" rather than "Representing quantum computers for the transpiler")

abbycross:
Oops - also i see that we have two "Default configuration" sections in a row on this page :-)

PR #783, opened 2024-02-07
Sessions migration guide

beckykd:
Closes #595  
Closes #1041 
Closes #582 
Closes #780
Closes #582
Closes #1156 
Closes #577 
Closes #576 

- [x] Add text about job splitting, such as "If you split your workload into multiple jobs and run them in Batch mode, you can get results from individual jobs. You can, for example, decide to cancel the rest of the jobs if the earlier job results don't meet your expectations. If one of the jobs fail, you can also re-submit just that one instead of re-running the entire workload."

There are several new and existing topics that are impacted:

- This documentation section has several topics about execution modes: https://qiskit-docs-preview-pr-783.1799mxdls7qz.us-south.codeengine.appdomain.cloud/run/execution-modes
- This migration guide is new: https://qiskit-docs-preview-pr-783.1799mxdls7qz.us-south.codeengine.appdomain.cloud/api/migration-guides/sessions

PR #1085, opened 2024-03-25
Changes for batch mode

### LINKED ISSUE #1081: Updates to batch mode pages for march release

beckykd:
closes #1081 

View changes here: https://qiskit-docs-preview-pr-1085.1799mxdls7qz.us-south.codeengine.appdomain.cloud/run/run-jobs-batch

### DIFF
--- a/docs/run/run-jobs-batch.mdx
+++ b/docs/run/run-jobs-batch.mdx
@@ -9,6 +9,9 @@ description: How to run quantum computing jobs in batch mode using Qiskit Runtim
 Batch mode can shorten processing time if all jobs can be provided at the outset. If you want to submit iterative jobs, use [sessions](sessions) instead.  Using batch mode has these benefits:
 
 -  The jobs' classical computation, such as compilation, is run in parallel. Thus, running multiple jobs in a batch is significantly faster than running them serially. 
+    <Admonition>
+    At most five jobs can be processed in parallel, which might include other users' jobs.
+    </Admonition>
 -  There is minimal delay between jobs, which can help avoid drift.
 -  You don't get exclusive access to the quantum system.  Therefore, if another user's job is already running on the system, your jobs have to wait.  Additionally, system calibration jobs could run between the batched jobs. 
 
--- a/docs/run/run-jobs-batch.mdx
+++ b/docs/run/run-jobs-batch.mdx
@@ -10,7 +10,7 @@ Batch mode can shorten processing time if all jobs can be provided at the outset
 
 -  The jobs' classical computation, such as compilation, is run in parallel. Thus, running multiple jobs in a batch is significantly faster than running them serially. 
 -  There is minimal delay between jobs, which can help avoid drift.
--  You don't get exclusive access to the quantum system.  Therefore, if another user's job is already using the system, your jobs have to wait.  Additionally, system calibration jobs could run between the batched jobs. 
+-  You don't get exclusive access to the quantum system.  Therefore, if another user's job is already running on the system, your jobs have to wait.  Additionally, system calibration jobs could run between the batched jobs. 
 
 <Admonition type="note">
     When batching, jobs are not guaranteed to run in the order they are submitted.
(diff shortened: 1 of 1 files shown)


docs/run/run-jobs-batch.mdx
```diff
@@ -9,7 +9,11 @@ description: How to run quantum computing jobs in batch mode using Qiskit Runtim
 Batch mode can shorten processing time if all jobs can be provided at the outset. If you want to submit iterative jobs, use [sessions](sessions) instead.  Using batch mode has these benefits:
 
 -  The jobs' classical computation, such as compilation, is run in parallel. Thus, running multiple jobs in a batch is significantly faster than running them serially. 
--  There is no delay between jobs, which can help avoid drift.
+    <Admonition>
+    At most five jobs can be processed in parallel, which might include other users' jobs.
+    </Admonition>
```
jyu00: Perhaps we can combine this with the last sentence and say something like

"While your batch jobs will run as closely together as possible, they don't get exclusive access to the backend. Therefore, your batch jobs may run in parallel with other users' jobs if there is enough processing capacity on the system. Additionally, system calibration jobs could run between the batched jobs."

docs/run/run-jobs-batch.mdx
```diff
@@ -9,7 +9,11 @@ description: How to run quantum computing jobs in batch mode using Qiskit Runtim
...
+-  There is minimal delay between jobs, which can help avoid drift.
+-  You don't get exclusive access to the quantum system.  Therefore, if another user's job is already running on the system, your jobs have to wait.  Additionally, system calibration jobs could run between the batched jobs.
```
jyu00: It's weird to list this under "Using batch mode has these benefits:". I suggest we combine this in the admonition (see my comment above).

docs/run/run-jobs-batch.mdx
```diff
@@ -8,11 +8,11 @@ description: How to run quantum computing jobs in batch mode using Qiskit Runtim
 
 Batch mode can shorten processing time if all jobs can be provided at the outset. If you want to submit iterative jobs, use [sessions](sessions) instead.  Using batch mode has these benefits:
 
--  The jobs' classical computation, such as compilation, is run in parallel. Thus, running multiple jobs in a batch is significantly faster than running them serially. 
--  There is no delay between jobs, which can help avoid drift.
+-  The jobs' classical computation, such as compilation, is run in parallel (up to five jobs can be run simultaneously). Thus, running multiple jobs in a batch is significantly faster than running them serially.
```
jyu00, line 11: Let's not mention the number 5, since that's subject to change.

docs/run/run-jobs-batch.mdx
```diff
@@ -8,11 +8,11 @@ description: How to run quantum computing jobs in batch mode using Qiskit Runtim
 
 Batch mode can shorten processing time if all jobs can be provided at the outset. If you want to submit iterative jobs, use [sessions](sessions) instead.  Using batch mode has these benefits:
 
--  The jobs' classical computation, such as compilation, is run in parallel. Thus, running multiple jobs in a batch is significantly faster than running them serially. 
--  There is no delay between jobs, which can help avoid drift.
+-  The jobs' classical computation, such as compilation, is run in parallel (up to five jobs can be run simultaneously). Thus, running multiple jobs in a batch is significantly faster than running them serially.  
+-  There is minimal delay between jobs, which can help avoid drift.
```
jyu00, line 13: After thinking about this some more, we can also add

- If you split your workload into multiple jobs and run them in Batch mode, you can get results from individual jobs. You can, for example, decide to cancel the rest of the jobs if the earlier job results don't meet your expectations. If one of the jobs fail, you can also re-submit just that one instead of re-running the entire workload.

jyu00: Actually, I'm not sure if we should mention job splitting now, or as part of the best practices. I don't want people to run tiny jobs after seeing this statement since there is an overhead associated with it.

beckykd: So you're thinking we should put it in the Sessions work instead ? https://github.com/Qiskit/documentation/pull/783 @jyu00

jyu00: Yeah let's just wait for that.

jyu00 (approved):

PR #1104, opened 2024-03-27
Link checker: use the non-greedy quantifier to parse anchors

arnaucasau:
This small PR changes the regex used in the link checker to parse the anchors for a non-greedy version of it. After regenerating the API docs with the mdx components, our link checker wasn't able to detect the id from them that we previously had as `<span>` tags.

The problem was how the script was searching for the ids. In the the following screenshots, we can see the change introduced here.

**Before**

Captura desde 2024-03-27 17-27-06

**After**
Captura desde 2024-03-27 17-39-51

More information on the non-greedy quantifier here: https://www.ibm.com/docs/en/netcoolomnibus/8.1?topic=library-minimal-non-greedy-quantifiers

To see that it fixed the issue, this change was applied to ##1094, #1101, and #1102

### DIFF
--- a/scripts/lib/links/extractLinks.test.ts
+++ b/scripts/lib/links/extractLinks.test.ts
@@ -62,6 +62,14 @@ test("parseAnchors()", () => {
   ##### Header 2
 
   ## \`code-header\`
+
+  <Function id="mdx.component.testId" name="testId" signature="testId">
+    Convert to dictionary.
+
+    **Return type**
+
+    \`Dict\`
+  </Function>
   `);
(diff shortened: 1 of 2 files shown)


frankharkins (approved):
Nice, thanks!

Good explanation too.

Eric-Arellano (commented):
Thanks! Can you please add a test for this?

arnaucasau:
Thank you both for the review! I added an mdx component to the test we already had :+1:

Eric-Arellano (approved):

PR #1108, opened 2024-03-29
Document platform support changes for Qiskit 1.1

mtreinish:
In Qiskit/qiskit#12102 we're promoting the macOS arm64 environment to tier 1 now that github actions exposes native arm64 macOS workers. This commit updates the documentation accordingly. This should not be merged until Qiskit 1.1.0 is released (~or 1.0.3 if we decided to backport Qiskit/qiskit#12102~ this isn't an option because there are issues with arm64 mac in numpy that need pretty extensive changes to workaround which will only be in 1.1.0).


docs/start/install.mdx
```diff
@@ -198,7 +198,7 @@ Additionally, Qiskit only supports the CPython implementation of the Python lang
   - Linux x86_64 (distributions compatible with the [manylinux 2014](https://www.python.org/dev/peps/pep-0599/) packaging specification).
   - macOS x86_64 (10.12 or later)
   - macOS ARM64 (10.15 or newer)
-  - Windows 64 bit
+  - Windows 64 bit (Windows 10 and newer supported)
```
beckykd, line 201:
```suggestion
  - Windows 64 bit (Windows 10 and later supported)
```

docs/start/install.mdx
```diff
@@ -197,7 +197,8 @@ Additionally, Qiskit only supports the CPython implementation of the Python lang
 
   - Linux x86_64 (distributions compatible with the [manylinux 2014](https://www.python.org/dev/peps/pep-0599/) packaging specification).
   - macOS x86_64 (10.12 or later)
-  - Windows 64 bit
+  - macOS ARM64 (10.15 or newer)
```
mtreinish, line 200: I just checked the wheels and the arm64 wheels are only compatible with macos 11.0 or newer.
```suggestion
  - macOS ARM64 (11.0 or newer)
```

docs/start/install.mdx
```diff
@@ -197,7 +197,8 @@ Additionally, Qiskit only supports the CPython implementation of the Python lang
 
   - Linux x86_64 (distributions compatible with the [manylinux 2014](https://www.python.org/dev/peps/pep-0599/) packaging specification).
   - macOS x86_64 (10.12 or later)
-  - Windows 64 bit
+  - macOS ARM64 (11.0 or newer)
+  - Windows 64 bit (Windows 10 and later supported)
```
abbycross, line 201:
```suggestion
  - Windows 64 bit (Windows 10 and later supported)
```
I think 64-bit should be hyphenated, unless it's standard in this OS context??

docs/start/install.mdx
```diff
@@ -221,9 +222,8 @@ Additionally, Qiskit only supports the CPython implementation of the Python lang
 
   - Linux ppc64le (distributions compatible with the [manylinux 2014](https://www.python.org/dev/peps/pep-0599/)packaging specification)
   - Linux s390x (distributions compatible with the [manylinux 2014](https://www.python.org/dev/peps/pep-0599/) packaging specification)
-  - macOS ARM64 (10.15 or newer)
   - Linux i686 (distributions compatible with the [manylinux 2014](https://www.python.org/dev/peps/pep-0599/) packaging specification)
-  - Windows 32 bit
+  - Windows 10 32 bit
```
abbycross, line 226: Same comment as above: can we do 32-bit hyphenated?

docs/start/install.mdx
mtreinish: It looks like they're hyphenated in the windows official documentation: https://support.microsoft.com/en-us/windows/32-bit-and-64-bit-windows-frequently-asked-questions-c6ca9541-8dce-4d48-0415-94a3faa2e13d so let's make the change

```suggestion
  - Windows 64-bit (Windows 10 and later supported)
```

docs/start/install.mdx
mtreinish:
```suggestion
  - Windows 10 32-bit
```

PR #1122, opened 2024-04-03
Use simpler argument style for V2 Sampler

Eric-Arellano:
`SamplerV2.run()` has a flexible argument style. It always needs a list, but each element can either be a single `QuantumCircuit` or it can be a tuple like `(my_circuit,)` or `(my_circuit, None, 123)`.

These two lines are functionally equivalent:

```python
sampler.run([circuit1, circuit2])
```

```python
sampler.run([(circuit1,), (circuit2,)])
```

But the first one is simpler and easier to understand. So, this PR consistently switches to that style. When we do need the tuple-syntax, we now put each tuple on a dedicated line so that it's easier to see each distinct tuple.


frankharkins (approved):
Nice, thanks!

PR #1123, opened 2024-04-03
Remove the math block's indentation from the API docs

arnaucasau:
Turns out the math blocks (latex expressions between double dollars `$$`) can't be indented more than once without making the pages not render. After introducing the MDX components for classes, functions, and attributes on #1026, we can't guarantee only one level of indentation, and this PR adds a new function to remove the indentation as a postprocess of all the MDX files.

Example of a math expression that was not rendering before:
Captura desde 2024-04-03 23-35-05


scripts/lib/api/removeMathBlocksIndentation.test.ts
```diff
@@ -0,0 +1,81 @@
+// This code is a Qiskit project.
+//
+// (C) Copyright IBM 2023.
```
Eric-Arellano: Should be 2024

scripts/lib/api/removeMathBlocksIndentation.ts
```diff
@@ -0,0 +1,33 @@
...
+import { HtmlToMdResultWithUrl } from "./HtmlToMdResult";
+
+/**
+ * The math blocks (latex expressions between double dollars $$) can't be
+ * indented more than one time. After introducing the MDX components for
+ * classes, functions, and attributes, we can't guarantee only one indentation,
+ * and this function removes them all
+ */
+function removeMathBlocksIndentation(results: HtmlToMdResultWithUrl[]): void {
+  for (const result of results) {
```
Eric-Arellano: Nit, can use `results.forEach()`

scripts/lib/api/removeMathBlocksIndentation.ts
```diff
@@ -0,0 +1,33 @@
...
+    // Captures all the lines in the math blocks, including the indentation
+    // before the $$ characters
+    const mathBlockRegex = /^[^#\r\n](\s*)\$\$([\s\S]*?)\$\$/gm;
+    const indentRegex = /^ +/gm;
+    result.markdown = result.markdown.replace(mathBlockRegex, function (match) {
+      return match.replace(indentRegex, "");
+    });
```
Eric-Arellano: You can use an inline function, `(match) => match.replace(indentRegex, "")`

scripts/lib/api/removeMathBlocksIndentation.test.ts
```diff
@@ -0,0 +1,81 @@
...
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+import { expect, test } from "@jest/globals";
+
+import removeMathBlocksIndentation from "./removeMathBlocksIndentation";
+
+test("dedupeHtmlIds()", async () => {
```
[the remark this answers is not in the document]
arnaucasau: Oops :+1:

Eric-Arellano (approved):

PR #1127, opened 2024-04-04
Upgrade Aer to 0.14

Eric-Arellano:
Also explain what to install to get the notebooks working locally. Tested on macOS Apple Silicon. The Linux requirements come from CI but weren't tested locally.


frankharkins (approved):
Thanks for adding the instructions

Issue #1135, opened 2024-04-05
move tutorial notebooks into this repo


javabster:
The following tutorials should be migrated from their existing repos to this repo:

- [x] [Variational Quantum Eigensolver](https://learning.quantum.ibm.com/tutorial/variational-quantum-eigensolver) (currently in the [qiskit-ibm-runtime repo](https://github.com/Qiskit/qiskit-ibm-runtime))
- [x] [Quantum Approximate Optimization Algorithm](https://learning.quantum.ibm.com/tutorial/quantum-approximate-optimization-algorithm) (currently in the [qiskit-ibm-runtime repo](https://github.com/Qiskit/qiskit-ibm-runtime))
- [x] [CHSH Inequality](https://learning.quantum.ibm.com/tutorial/chsh-inequality) (currently in the [qiskit-ibm-runtime repo](https://github.com/Qiskit/qiskit-ibm-runtime))
- [x] [Grover's algorithm](https://learning.quantum.ibm.com/tutorial/grovers-algorithm) (currently in the [qiskit-ibm-runtime repo](https://github.com/Qiskit/qiskit-ibm-runtime))
- [x] [Build repetition codes](https://learning.quantum.ibm.com/tutorial/build-repetition-codes) (currently in inner source repo)
- [x] [Repeat until success](https://learning.quantum.ibm.com/tutorial/repeat-until-success) (currently in inner source repo)
- [x] [Explore gates and circuits with the Quantum Composer](https://learning.quantum.ibm.com/tutorial/explore-gates-and-circuits-with-the-quantum-composer) (currently in inner source repo)
- [x] [Submit transpiled circuits](https://learning.quantum.ibm.com/tutorial/submit-transpiled-circuits) (currently in the [qiskit-ibm-runtime repo](https://github.com/Qiskit/qiskit-ibm-runtime))

As before we should migrate these notebooks over with the commit history where possible, and ensure the original authors are included as authors here. Please work with @jyu00 and @kt474 for the runtime repo tutorials

PR #91, opened 2023-10-06
Migrate open source docs

Eric-Arellano:
First part of https://github.com/Qiskit/documentation/issues/22 and https://github.com/Qiskit/documentation/issues/23. This copies over the open source files we plan to use in this repository. They are not actually used yet and still require some work like converting RST to MDX, but it gets the files and their Git history here to start the process.

Qiskit docs are placed in the folder `qiskit-docs/` and Runtime docs in `runtime-docs`. These are only for unprocessed docs. Once a document is ready, we will move it to the official `docs/` folder and have it go live.

This uses `git-filter-repo` to preserve all the Git history and author attribution from the respective open source repositories. It uses the below Python script and this [zipped folder](https://github.com/Qiskit/documentation/files/12835669/migrate-open-source.zip), using an approach very similar to @jakelishman's original work in https://github.com/Qiskit/qiskit/pull/10611.

This PR has no additional commits. It solely copies over the other repositories into this one.

```
# Runtime files
docs/faqs/max_execution_time.rst
docs/faqs/open_source_vs_ibm_cloud_primitives.rst
```
```
# Qiskit files

# start/
docs/migration_guides/algorithms_migration.rst
docs/migration_guides/opflow_migration.rst
docs/migration_guides/qi_migration.rst

# build/
docs/tutorials/circuits/01_circuit_basics.ipynb
docs/tutorials/circuits/1_getting_started_with_qiskit.ipynb
docs/tutorials/circuits/2_plotting_data_in_qiskit.ipynb
docs/tutorials/circuits_advanced/01_advanced_circuits.ipynb
docs/tutorials/circuits_advanced/02_operators_overview.ipynb
docs/tutorials/circuits_advanced/05_pulse_gates.ipynb
docs/tutorials/circuits_advanced/06_building_pulse_schedules.ipynb
docs/tutorials/circuits_advanced/07_pulse_scheduler.ipynb
docs/explanation/endianness.rst

# transpile/
docs/tutorials/circuits_advanced/04_transpiler_passes_and_passmanager.ipynb

# test/
docs/how_to/use_estimator.rst 
docs/how_to/use_sampler.rst

# run/
docs/tutorials/circuits_advanced/08_gathering_system_information.ipynb

# other/
docs/faq.rst

# Tutorial renames from the original qiskit-tutorials repo
docs/tutorials/terra/fundamentals/1_getting_started_with_qiskit.ipynb
docs/tutorials/terra/fundamentals/2_plotting_data_in_qiskit.ipynb
[...]

PR #1149, opened 2024-04-09
Migrate runtime tutorials

frankharkins:
Adds the runtime tutorial notebooks as per #1135. I couldn't use the same approach as #91 as the file history is broken in the runtime repo, but since there were few enough files I used `git log --follow` to find the original PR and all authors that touched the files.


Eric-Arellano (commented):
Thanks! Sgtm to manually do co-authors.

About the file organization:

1. What other top-level folders are you envisioning beyond runtime?
2. Why have the folders with single files, rather than flattening that?

frankharkins:
> What other top-level folders are you envisioning beyond runtime?

Good question. I'm not sure; I just named the folder based on their origin but I think it would also be OK to remove that layer too.

> Why have the folders with single files, rather than flattening that?

If we want to include separate images we need to upload a folder of the following form. We don't have any images yet but I thought it would be easier to start with this form.

```
tutorial-name
├── images/
└── notebook.ipynb
```

Eric-Arellano:
Can the file be called `notebook.ipynb`? Avoid the repetition.

Eric-Arellano (approved):

PR #1153, opened 2024-04-10
API generation script: fix attributes headings

### LINKED ISSUE #1120: Bad attribute in `qiskit.algorithms.eigensolvers.VQD`
In the `qiskit.algorithms.eigensolvers.VQD` page for historical versions of qiskit, we have a weird attribute with a text description as a name:

Captura desde 2024-04-03 15-21-14

This attribute has an entry in the ToC as well:

Captura desde 2024-04-03 15-21-32

Link to the page https://docs.quantum.ibm.com/api/qiskit/0.46/qiskit.algorithms.eigensolvers.VQD#one-for-every-k-th-eigenvalue-can-either-be-a-qiskit-optimizer-or-a-callable

arnaucasau:
Closes #1120 

This PR changes the script to use the `name` extracted from the `signature` of the attributes instead of using the `id` to create the `<h3>`s. This behavior was changed in #1026, but It broke a couple of headings.

It also fixes two attributes of `qiskit.algorithms.eigensolvers.VQD` manually modifying the artifacts stored in [Box](https://ibm.ent.box.com/folder/246867452622?s=tqmvw6ldivzkhh934nhufan4rg3dregr) for versions 0.39, 0.40, 0.41, 0.42, 0.43, 0.44, 0.45 and 0.46. See more at #1120


docs/api/qiskit/0.45/qiskit.algorithms.eigensolvers.VQD.mdx
```diff
@@ -91,16 +85,6 @@ python_api_name: qiskit.algorithms.eigensolvers.VQD
...
 
-  ###
-
-  <Attribute name="initial point">
-    An optional initial point (i.e. initial parameter values) or a list of initial points (one for every k-th eigenvalue) for the optimizer. If `None` then VQD will look to the ansatz for a preferred point and if not will simply compute a random one.
-
-    **Type**
-
-    Sequence\[[float](https://docs.python.org/3/library/functions.html#float "(in Python v3.12)")] | Sequence\[Sequence\[[float](https://docs.python.org/3/library/functions.html#float "(in Python v3.12)")]] | None
-  </Attribute>
```
frankharkins, line 102: Should this have been removed completely? Just going off the class description but it seems like it should still be listed

docs/api/qiskit/0.19/qiskit.validation.BaseSchema.mdx
```diff
@@ -26,7 +26,7 @@ python_api_name: qiskit.validation.BaseSchema
     type
   </Attribute>
 
-  ###
+  ### constructor is passed all named parameters from deserialization.
 
   <Attribute name="constructor is passed all named parameters from deserialization." />
```
frankharkins, line 32: This section also feels like an error. I'm not too worried about it as it's an obscure class from an ancient version of Qiskit, but if we can work out why it happened then it'll make the script more robust

frankharkins (changes requested):
Thanks! This is a net improvement but I noticed a couple of quirks

docs/api/qiskit/0.19/qiskit.validation.BaseSchema.mdx
arnaucasau: Agreed! I noticed it too, but that's the same heading we had before changing to MDX components as you can see [here](https://github.com/Qiskit/documentation/blob/0406030c7e13a828bc5bd164b9ccb7aac29f4b06/docs/api/qiskit/0.19/qiskit.validation.BaseSchema.mdx?plain=1#L32). My plan was to investigate it in a follow-up, I think it could be similar to [#1120](https://github.com/Qiskit/documentation/issues/1120) with the `model\_cls` attributes just above of it (it seems like it's missing description matching the heading)

docs/api/qiskit/0.45/qiskit.algorithms.eigensolvers.VQD.mdx
arnaucasau: Yes, this is a consequence of this [PR](https://github.com/Qiskit/qiskit/pull/12162). The attribute in the original file was missing a hyphen (`initial point` instead of `initial_point`). If we add the hyphen, sphinx complained that the attribute is duplicated and we decided to remove it entirely [here](https://github.com/Qiskit/qiskit/pull/12162#discussion_r1557834962)

docs/api/qiskit/0.19/qiskit.validation.BaseSchema.mdx
frankharkins: Ok I see, that's fine then; thanks

frankharkins (approved):
Nice work, thanks!

docs/api/qiskit/0.19/qiskit.validation.BaseSchema.mdx
Eric-Arellano: Might be helpful to open an issue with the problem so we don't forget it

docs/api/qiskit/dev/qpy.mdx
```diff
@@ -172,7 +172,7 @@ These functions will raise a custom subclass of [`QiskitError`](exceptions#qiski
   Set the error message.
 </Class>
 
-### QPY\_VERSION
+### qiskit.qpy.QPY\_VERSION
```
Eric-Arellano, line 175: Do you know what these were before the docs redesign? I'm not certain it makes sense to use the full path here because it makes the header longer. I'm surprised the name is the full path and not only `QPY_VERSION`. It's not necessarily bad, I'm only curious why `name` was set to that

Eric-Arellano (approved):
Thanks! Can you please clarify in the PR that you regenerated the HTML to fix the bad attribute? This PR is a mix of using the new HTML + your new code change

docs/api/qiskit/dev/qpy.mdx
arnaucasau: Yes, they had the long `name` as well. [Here](https://github.com/Qiskit/documentation/blob/0406030c7e13a828bc5bd164b9ccb7aac29f4b06/docs/api/qiskit/dev/qpy.mdx?plain=1#L183) is the file before the docs redesign. I can investigate more about why it was set that way if you want :)

PR #1154, opened 2024-04-10
Add tutorial deployment process

frankharkins:
Adds a workflow to deploy tutorials from `main`, and documentation on how to do it. This PR also switches to using environment secrets rather than actions secrets, which should be more secure as we can limit who can run them and which branches they can run on. See https://github.com/Qiskit/documentation/actions/runs/8647825547/job/23710189504 for an example of the action using the staging environment.


.github/workflows/deploy-tutorials.yml
```diff
@@ -0,0 +1,38 @@
...
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+name: Deploy tutorials
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        type: environment
+        description: Where to upload the tutorials. Must start with "Learning platform".
```
Eric-Arellano, line 19: Any reason to not explicitly say the two options?

```suggestion
        description: Where to upload the tutorials. Either "Learning platform (staging)" or "Learning platform (production)".
```

Eric-Arellano (approved):
Thanks!

PR #1159, opened 2024-04-11
Upload artifacts from cron job

frankharkins:
We run notebooks that submit jobs twice a month to check they don't raise errors, but we currently throw away the results. This PR uploads the executed notebooks as an artifact so we can inspect the outputs and update the files in this repo with more recent output.

See https://github.com/Qiskit/documentation/actions/runs/8645293171 for an example of this working.

### DIFF
--- a/.github/workflows/notebook-test-cron.yml
+++ b/.github/workflows/notebook-test-cron.yml
@@ -29,8 +29,8 @@ jobs:
           instance: "client-enablement/documentation/qiskit-documenta"
 
       - name: Execute notebooks
-        timeout-minutes: 10
-        run: tox -- --write
+        timeout-minutes: 330
+        run: tox -- --write --only-submit-jobs
 
       - name: Detect changed notebooks
         id: changed-notebooks
--- a/.github/workflows/notebook-test-cron.yml
+++ b/.github/workflows/notebook-test-cron.yml
@@ -29,8 +29,8 @@ jobs:
           instance: "client-enablement/documentation/qiskit-documenta"
 
       - name: Execute notebooks
-        timeout-minutes: 330
-        run: tox -- --write --only-submit-jobs
+        timeout-minutes: 10
+        run: tox -- --write
 
       - name: Detect changed notebooks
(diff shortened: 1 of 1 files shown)


.github/workflows/notebook-test-cron.yml
```diff
@@ -29,7 +29,19 @@ jobs:
           instance: "client-enablement/documentation/qiskit-documenta"
 
       - name: Execute notebooks
-        run: tox -- --only-submit-jobs
+        timeout-minutes: 330
```
frankharkins, line 32: This is 5hr 30mins, which is 30mins short of the maximum 6hrs. This means if jobs take too long out we'll still have some time to finish the action and upload any notebooks that did execute.

Eric-Arellano: Upload should be quite fast. I recommend 350, along with a comment explaining thew workflow max is 360 minutes so we leave 10 minutes for the other steps. Or 345 works, but I doubt the other steps take 15 minutes.

.github/workflows/notebook-test-cron.yml
```diff
@@ -29,7 +29,23 @@ jobs:
...
+        run: tox -- --write --only-submit-jobs
+
+      - name: Detect changed notebooks
+        id: changed-notebooks
+        if: "!cancelled()"
+        run: |
+          echo "CHANGED_NOTEBOOKS<<EOF" >> $GITHUB_OUTPUT
+          echo "$(git diff --name-only)" >> $GITHUB_OUTPUT
```
Eric-Arellano, line 40: Nit: this should work the same:

```suggestion
          git diff --name-only >> $GITHUB_OUTPUT
```

.github/workflows/notebook-test-cron.yml
frankharkins: That's true. Looking at the job, the other steps take ~1m total. I'll still leave 10mins just in case.

```suggestion
        timeout-minutes: 350
```

PR #1099, opened 2024-03-27
Add "Save and retrieve Qiskit objects" page

frankharkins:
Continuation of #956

[Link to preview](https://qiskit-docs-preview-pr-1099.1799mxdls7qz.us-south.codeengine.appdomain.cloud/analyze/save-and-retrieve).

PR #1160, opened 2024-04-11
Fit new notebooks into existing content

frankharkins:
PRs #977 and #1099 added some new pages in anticipation of a restructure which has since been postponed. This PR fits those pages into our current content structure.


docs/run/_toc.json
```diff
@@ -75,13 +75,21 @@
         {
           "title": "Maximum execution time",
           "url": "/run/max-execution-time"
+        },
+        {
+          "title": "Save and retrieve Qiskit objects",
+          "url": "/run/save-and-retrieve"
```
Eric-Arellano, line 81: This URL is a bit vague. How about using the full name `save-and-retrieve-qiskit-objects`?

Eric-Arellano: Although looking over the file, maybe we should rename to "Save progress to disk"? "Qiskit objects" is a bit vague, and the "retrieve" part is implied imo

scripts/nb-tester/test-notebook.py
```diff
@@ -32,7 +32,7 @@
 ]
 NOTEBOOKS_THAT_SUBMIT_JOBS = [
     "docs/start/hello-world.ipynb",
-    "docs/analyze/saving-and-retrieving.ipynb",
```
Eric-Arellano, line 35: Fwiw, this seems like it always belonged in "execute on hardware", not postprocess/analyze

Eric-Arellano (commented):
Nice! The notebooks look good and are appropriate in the new sections.

docs/run/_toc.json
frankharkins: That's true, what about "Save circuits and jobs"? It includes instructions on getting results from IBM Quantum so it's not all technically "to disk".

Also, as a user, would you be surprised to find circuts under "Manage jobs"? I wonder if I should split the page into two and have the jobs on this page and the circuits somewhere else...

Eric-Arellano: I like splitting it up! "Save circuits to file" (or disk), and "Retrieve jobs from IBM Quantum" as distinct files. That better matches the workflow we give people.

frankharkins:
Good point, I've made #1194 to track that.

kaelynj (approved):
> Good point, I've made #1194 to track that.

Great! Looks good to ship :100:

docs/run/_toc.json
```diff
@@ -75,13 +75,21 @@
...
+          "url": "/run/save-jobs"
         }
       ]
     },
     {
       "title": "Quantum Serverless workloads",
       "url": "/run/quantum-serverless"
     },
+    {
+      "title": "Visualize results",
```
javabster, line 90: I think this should go above quantum serverless in the side bar

docs/run/save-jobs.ipynb
Eric-Arellano, line 1: How does output work for this file? It doesn't seem like it'd be deterministic when we rerun the notebook, or if a local user runs with a different account?

scripts/nb-tester/test-notebook.py
```diff
@@ -32,7 +32,7 @@
...
+    "docs/run/save-and-retrieve.ipynb",
```
Eric-Arellano, line 35: Outdated

docs/run/save-jobs.ipynb
frankharkins: The output will be deterministic when running using my account (which we also use in our actions). Users won't have access to this job but I think that's ok, plus there's no way to demonstrate retrieving a job without picking one from my account.

Issue #1174, opened 2024-04-12
Wildcard: allow running our notebooks with simulators?


Eric-Arellano:
We are using more hardware rather than simulators in our notebooks. This is meant to help users by showing real-world workflows, where the emphasis is running on actual hardware rather than simulators. 

We don't want to complicate the notebooks by showing both simulators and hardware.

However, that poses a challenge for our docs writing pipeline:

1. We cannot easily test the jobs in every CI run (https://github.com/Qiskit/documentation/issues/1173)
2. It is harder for us to get sample output for the notebooks to show to the user

This issue is to explore the idea of allowing us to dynamically rewrite our notebooks to use simulators rather than hardware when using our tool `nb-tester`. It would, for example, rewrite the backend to use `FakeManilaV2`. We'd need to figure out the rewriting of the code, along with reckoning with simulators not working when the qubit count is too high.

frankharkins:
@beckykd had the idea of overwriting the `least_busy` function to always return `FakeManilaV2`. I'm trying it out now with a patched fork and it seems to work pretty well bar a few quirks we can probably iron out.

1. `docs/start/hello-world.ipynb` fails in the final cell with

   ```pytb
   AttributeError: 'DataBin' object has no attribute 'ensemble_standard_error'
   ```
   Because the fake backend doesn't set that attribute. We could just avoid using this attribute as they're only used for error bars in the plot.
2. `tutorials/submitting-transpiled-circuits/notebook.ipynb` fails when using `Batch`. It's possible this notebook is just broken anyway as there were some changes to batch mode recently.

   ```pytb
   AttributeError: 'QiskitRuntimeLocalService' object has no attribute 'backend'
   ```
3. This doesn't work for `tutorials/build-repitition-codes/notebook.ipynb` and `tutorials/build-repitition-codes/notebook.ipynb` as they use `qiskit-ibm-provider`. We probably want to switch them to runtime anyway.

Eric-Arellano:
@frankharkins it may be desirable for us to always try to have the notebooks work without issue with local simulators, and even communicate to users that they can swap out `.least_busy()` with `FakeManilaV2`. That allows us to have hardware-first docs, but that are still accessible. So, when we have issues like the above, that means we should update the docs to work with simulators. The one fundamental barrier I see is if we need to use utility-scale # of qubits?

[...]

Issue #1173, opened 2024-04-12
Test the safe parts of notebooks that run on hardware in every CI run


Eric-Arellano:
We cannot easily test notebooks that run on hardware in CI for every PR because a) there is often a queue and b) it costs actual resources. So, we currently skip those notebooks in PR runs:

https://github.com/Qiskit/documentation/blob/8642507fc2c5fc4982e79f16f414ba66fc629708/scripts/nb-tester/test-notebook.py#L33-L43

We then run those notebooks every two weeks in a cron job with GH Actions.

However, we can actually run part of the notebooks safely! We can run every code cell up until the point when we call `Estimator.run()` and `Sampler.run()`, or do some other slow thing with Qiskit Runtime. All the code up to that point is safe to run.

We can add a Jupyter notebook tag like `submits-job` to the code cells that use hardware. Then teach `nb-tester` to only run up to that code block, but stop there. We wouldn't run any subsequent code blocks since they may depend on the tagged code block.

Eric-Arellano:
@frankharkins think we should close this as not worth the complexity? In hello world, we would only be able to test about 25% of the notebook with this strategy, which isn't that big of a win. 

Removing tutorials changed the value, imo. This is what would be impacted:

https://github.com/Qiskit/documentation/blob/632cb2cc4c83432212d30f4e2b8262cba2831d3a/scripts/config/notebook-testing.toml#L48-L54

I lean towards closing this as not worth it. The cron job does a good job as is.

frankharkins:
Yeah I agree

johannesgreiner:
an idea is to have the notebooks modified such that our test_* backends will be used for pipeline testing wherever possible

PR #1158, opened 2024-04-10
Update hello-world for 1.0

kaelynj:
Closes #699 

Currently waiting for the job to finish in order to plot the last set of data.  However, all the other content should be there and updated to support 1.0.

PR #1184, opened 2024-04-16
Mock hardware when testing notebooks

frankharkins:
This uses the approach discussed in #1174 to test notebooks that submit jobs to IBM Quantum.

The script behaviour has changed a bit:
* We now have a list of `notebooks_no_mock`, these could include notebooks with large circuits or that demonstrate features not available on the fake backend. Future work could partially test these using the approach in [#1173](https://github.com/Qiskit/documentation/issues/1173). 
* By default, it runs all notebooks except `notebooks_no_mock`, with the patched package.
* `--submit-jobs` will run **all** notebooks, without the patched package.
* `--only-unmockable` will run only `notebooks_no_mock`, without the patched package.

The following notebooks do **not** work with this patch, but some of them might not have worked anyway.
* `tutorials/variational-quantum-eigensolver/notebook.ipynb`
* `tutorials/submitting-transpiled-circuits/notebook.ipynb`
* `tutorials/build-repitition-codes/notebook.ipynb`


scripts/nb-tester/test-notebook.py
```diff
@@ -30,18 +30,21 @@
...
-    "tutorials/chsh-inequality/notebook.ipynb",
-    "tutorials/grovers-algorithm/notebook.ipynb",
-    "tutorials/quantum-approximate-optimization-algorithm/notebook.ipynb",
-    "tutorials/repeat-until-success/notebook.ipynb",
-    "tutorials/submitting-transpiled-circuits/notebook.ipynb",
-    "tutorials/variational-quantum-eigensolver/notebook.ipynb",
+NOTEBOOKS_THAT_CANT_BE_MOCKED = [
+    # Notebooks that don't work with the fake backend
+    # (e.g. due to large circuits)
 ]
```
Eric-Arellano, line 36: Nit because the comment describes the list, not element 0 of the list

```suggestion
# Notebooks that don't work with the fake backend, e.g. due to large circuits
NOTEBOOKS_THAT_CANT_BE_MOCKED = []
```

Also should this have entries in this PR?

scripts/nb-tester/test-notebook.py
```diff
@@ -269,6 +293,8 @@ def create_argument_parser() -> argparse.ArgumentParser:
     # Execute notebooks
     start_time = datetime.now()
     print("Executing notebooks:")
+    if not args.submit_jobs and not args.only_unmockable:
+        print("ℹ️  Note: Using patched qiskit-ibm-runtime; least_busy will return FakeManilaV2")
```
Eric-Arellano, line 297:
```suggestion
        print("ℹ️  Note: Using patched qiskit-ibm-runtime; .least_busy() method will return FakeWashingtonV2")
```

scripts/nb-tester/test-notebook.py
```diff
@@ -168,6 +189,9 @@ def _execute_notebook(filepath: Path, args: argparse.Namespace) -> nbformat.Note
     if not args.write:
         return nb
```
Eric-Arellano, line 190: Should this happen after remove_mocking_cell()?

scripts/nb-tester/test-notebook.py
```diff
@@ -168,6 +189,9 @@ def _execute_notebook(filepath: Path, args: argparse.Namespace) -> nbformat.Note
...
 
+    if not submit_jobs:
+        remove_mocking_cell(nb)
```
Eric-Arellano, line 193: One option is to use a context manager, like a `with` statement. While you're in the `with` block, it will patch the notebook. Once you exit it, we remove the patch. It might make the code more readable? 

You'd still keep the dedicated functions you have, only define a new context manager like https://www.pythonmorsels.com/creating-a-context-manager/

scripts/nb-tester/test-notebook.py
frankharkins: No because we don't care about the final state of the notebook if we're not writing the result to disk

scripts/nb-tester/test-notebook.py
frankharkins: I think "no" because there aren't any notebooks that "can't" be mocked, they are just broken. I would add them to `NOTEBOOKS_EXCLUDE` but I believe they're all about to be reviewed and edited.

I added this empty list in anticipation of the 100-qubit job in #1158.

scripts/nb-tester/test-notebook.py
frankharkins: Good idea, added in ddb05c97d334ac789288e9efac91c3995a033b64. Thanks!

scripts/nb-tester/test-notebook.py
```diff
@@ -50,7 +51,7 @@ def filter_paths(paths: list[Path], args: argparse.Namespace) -> Iterator[Path]:
     """
     Filter out any paths we don't want to run, printing messages.
     """
-    submit_jobs = args.submit_jobs or args.only_submit_jobs
+    submit_jobs = args.submit_jobs or args.only_unmockable
```
Eric-Arellano, line 54: It's confusing to call the variable `submit_jobs` since it looks like `args.submit_jobs`. What about `use_hardware`?

scripts/nb-tester/test-notebook.py
```diff
@@ -123,6 +124,38 @@ def extract_warnings(notebook: nbformat.NotebookNode) -> list[NotebookWarning]:
     return notebook_warnings
 
 
+class PatchRuntime:
```
Eric-Arellano, line 127: I forgot about `@contextlib.contextmanager`, which makes this code less verbose. https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager 

It would look like this, roughly

```python
@contextmanager
def patch_runtime(nb: nbformat.NotebookNode, *, should_patch: bool):
    if should_patch:
        self.nb.cells.insert(0, nbformat.v4.new_code_cell(source=MOCKING_CODE))
    try:
        yield
    finally:
         if not should_patch:
              return
        # your exit code
```

```python
with patch_runtime(nb, should_patch=not submit_jobs):
    processor.preprocess(nb)
```

Feel free to ignore - this is fine as is.

Eric-Arellano (approved):
Great work, Frank!

It would be good to start running this change in CI. That's fine if you prefer it to be a follow up. For the broken notebooks, we can exclude them with an ignore list and then open up an issue to fix them

scripts/nb-tester/test-notebook.py
frankharkins: What do you think of 90a348b238a54e4a4ac18e42ccd7b808cf1344a2 as a slightly different solution? It sets `args.submit_jobs` to `True` if `only_unmockable` is set, which lets us remove a few of these lines.

scripts/nb-tester/test-notebook.py
frankharkins: Nice, thanks! Incorporated in 5e391acdfeafe549b2e10b42cc602f5a072eaeb3. FYI the try/finally seems to suppress exceptions but it works well otherwise

frankharkins:
> It would be good to start running this change in CI.

This will take effect as soon as the PR is merged; is that ok? RE ignore list: We only run changed notebooks on PR so the broken notebooks shouldn't block anyone who's not working on them.

Eric-Arellano:
The code looks good, but let's wait to merge this until we decide how to deal with Becky's concern that some of our output will leak that we're running on simulators

image

Maybe we should not apply this patch for certain allow-listed notebooks, like the ones about how to interact with the hardware? Or we allow-list which notebooks should be mocked? Always using simulators unconditionally seems like it can cause issues.

scripts/nb-tester/test-notebook.py
Eric-Arellano: I like that. Maybe worth editing the argparse docstring to say

> Setting this option implicitly sets `--submit-jobs`.

frankharkins:
This feature should only be used to test code and catch bugs, not to generate notebook outputs. It does give you the _option_ to write the simulator results to the notebook, but I don't think it's that dangerous. We could write a test for it if needed.

Eric-Arellano:
> This feature should only be used to test code and catch bugs, not to generate notebook outputs.

But we instruct people to get output from the PR CI build:

https://github.com/Qiskit/documentation/blob/124e14359dd1da55fa34ff2ffe2d0c06ab0155a0/README.md?plain=1#L127-L131

And when they run locally with `tox`, it will be using the patch. We have two conflicting goals:

1. Test in every PR that the notebook has not broke
2. Provide a way to get the output, ideally using hardware. Do this either locally or from a CI artifact

Perhaps we should consider something like this?

* Make `--patch` an argument to the script, so by default we still run on hardware. 
* In CI for PRs, test that notebooks work with `--patch`. At least changed notebooks, maybe all notebooks if it isn't too slow
* In CI for PRs, _also_ run changed notebooks on hardware if they aren't too slow. That's so that you can get the output of the notebook from CI

frankharkins:
> We have two conflicting goals:
> 1. Test in every PR that the notebook has not broke
> 2. Provide a way to get the output, ideally using hardware. Do this either locally or from a CI artifact

This PR (mostly) solves 1. It doesn't regress on 2 as we don't run job-submitting notebooks in CI anyway. Writers can still get hardware outputs locally using `--submit-jobs --write`, same as before.

I'll update the README with the caveat about getting a notebook from CI. Maybe I can emit a warning whenever the patched `least_busy` is called so we don't accidentally commit the results to CI.

> - Make --patch an argument to the script, so by default we still run on hardware.
> - In CI for PRs, test that notebooks work with --patch. At least changed notebooks, maybe all notebooks if it isn't too slow
> - In CI for PRs, also run changed notebooks on hardware if they aren't too slow. That's so that you can get the output of the notebook from CI

The first two bullets are basically what this PR does, just the argument to **not** patch is `--submit-jobs`. The last bullet is sadly not feasible IMO, it's really unlikely a job will run in a period of time acceptable for PR CI, they'll also use up a lot of device time if someone pushes commits regularly.

frankharkins:
I'm hesitant to make submitting jobs the default as it'd be easy for a writer to unknowingly send lots of jobs to IBM Quantum, and if they kill the script while waiting, those jobs would continue to run.

Eric-Arellano:
>  It doesn't regress on 2 as we don't run job-submitting notebooks in CI anyway.

Non-job-submitting notebooks will now run with the patched simulator. For example:

https://github.com/Qiskit/documentation/blob/76725900afde5ceacb73de770e6a5759cf137d1e/docs/run/get-backend-information.ipynb#L163

I wasn't very clear in differentiating between submitting jobs vs using Runtime but not for a job. It seems we have a few categories of notebooks:

* Don't submit a job but use Runtime, like `get-backend-information.ipynb`. Should probably not be patched.
* Submit a job that can be simulated. Should be patched for the sake of testing, but for output it sounds like we want it to be from actual hardware
* Submit a job that cannot be simulated. We cannot patch, so we can only verify via submitting a job. Output needs to come from hardware too.

This PR will regress with the first category of notebooks like `get-backend-information.ipynb`. Before, they ran with real hardware in CI and you could download their notebook to get the output. Now, they will use the patch.

--

What if we had our CI reflect these 3 categories?

* Notebooks that don't submit a job: do not patch them, and upload their output to CI. So you get validation & output.
* Notebooks that submit a simulatable job: patch them, but do not upload their output to CI. You get validation. For output, you must run it locally using `--submit-jobs`
* Notebooks that submit a non-simulatable job: ignore them in PR builds. Run in cron job and upload those artifacts

In the script itself, notebook authors would need to categorize which of the 3 buckets it falls into. 

Wdyt?

frankharkins:
This is now working (including cells with `backend.name`; see https://github.com/Qiskit/documentation/pull/1184/commits/c89e972bc03f09cf4c609421fbf8c2f53cea8461 🤦).

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -119,12 +131,34 @@ def extract_warnings(notebook: nbformat.NotebookNode) -> list[NotebookWarning]:
...
 
-async def execute_notebook(path: Path, args: argparse.Namespace) -> bool:
+async def execute_notebook(path: Path, args: argparse.Namespace, config: Config) -> bool:
     """
-    Wrapper function for `_execute_notebook` to print status
+    Wrapper function for `_execute_notebook` to print status and write result
     """
-    print(f"▶️ Executing {path}")
+    def _is_patched():
+        return not args.submit_jobs and matches(path, config.notebooks_that_submit_jobs)
```
Eric-Arellano, line 157: why is this a function rather than a variable `is_patched`?

Eric-Arellano (commented):
Nice! Is my understanding correct?

- If the notebook is not listed in the TOML file, we don't patch it. We do write the output
- If the notebook is in `submit-jobs`:
     - if you use `--submit-jobs`, we use Runtime and `--write` will work
     - if you don't use `--submit-jobs`, we use simulators and `--write` will be ignored
- If the notebook is in no_mock:
    - always skipped unless you use `--submit-jobs` or `--only-unmockable`

Sgtm. This README should probably be updated. Maybe mention how we use simulation in CI to test notebooks that submit jobs, but we do not upload their output?

> When you make a pull request with a changed notebook, you can get a version of that notebook that was executed in a uniform environment from CI. To do this, click "Show all checks" in the info box at the bottom of the pull request page on GitHub, then choose "Details" for the "Test notebooks" job. From the job page, click "Summary", then download "Executed notebooks".

--

Also PR description could use an update

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
frankharkins: Good point

README.md
```diff
@@ -116,20 +116,23 @@ You also need to install a few system dependencies: TeX, Poppler, and graphviz.
...
-> [`scripts/nb-tester/notebooks.toml`](scripts/nb-tester/notebooks.toml). This
-> is not needed if the notebook only retrieves information.
->
+> When testing notebooks, we avoid sending jobs to IBM Quantum by patching
+> `least_busy` to return a fake backend. Try to make sure your notebook works
+> with this patch. To use the patch, add your notebook to the list
+> `notebooks_that_submit_jobs` in
+> [`scripts/nb-tester/notebooks.toml`](scripts/nb-tester/notebooks.toml). If
+> your notebook can't be tested with this patch, you must add it to the list
+> `notebooks_no_mock`.
```
Eric-Arellano, line 125:
```suggestion
> If your notebook submits hardware jobs to Qiskit Runtime, you must add it to
> [`scripts/nb-tester/notebooks.toml`](scripts/nb-tester/notebooks.toml). If it
> can be run with simulators, i.e., the circuit is not too large, add it to `notebooks_that_submit_jobs`.
> Otherwise, add it to `notebooks_no_mock`.
```

README.md
```diff
@@ -116,20 +116,23 @@ You also need to install a few system dependencies: TeX, Poppler, and graphviz.
...
-When you make a pull request with a changed notebook, you can get a version of
-that notebook that was executed in a uniform environment from CI. To do this,
-click "Show all checks" in the info box at the bottom of the pull request page
-on GitHub, then choose "Details" for the "Test notebooks" job. From the job
-page, click "Summary", then download "Executed notebooks".
+When you make a pull request changing a notebook that doesn't submit jobs, you
+can get a version of that notebook that was executed in a uniform environment
+from CI. To do this, click "Show all checks" in the info box at the bottom of
+the pull request page on GitHub, then choose "Details" for the "Test notebooks"
+job. From the job page, click "Summary", then download "Executed notebooks".
```
Eric-Arellano, line 135:
```suggestion
job. From the job page, click "Summary", then download "Executed notebooks".
Otherwise, if your notebook does submit jobs, you need to run it locally with
`tox -- --write --submit-jobs <path/to/notebook.ipynb>`.
```

Eric-Arellano (approved):
Great work!

.github/workflows/notebook-test-cron.yml
```diff
@@ -30,7 +30,7 @@ jobs:
 
       - name: Execute notebooks
         timeout-minutes: 350
-        run: tox -- --write --only-submit-jobs
+        run: tox -- --write --only-unmockable
```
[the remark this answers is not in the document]
frankharkins: I've changed the option back to `--only-submit-jobs`, which will run all job-submitting notebooks without the mock backend.

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -67,15 +79,15 @@ def filter_paths(paths: list[Path], args: argparse.Namespace, config: Config) ->
...
-        if not submit_jobs and matches(path, config.notebooks_that_submit_jobs):
+        if not args.submit_jobs and matches(path, config.notebooks_no_mock):
             print(
-                f"ℹ️ Skipping {path} as it submits jobs; use the --submit-jobs flag to run it."
+                f"ℹ️ Skipping {path} as it doesn't work with mock hardware; use the --submit-jobs flag to run it."
             )
             continue
 
-        if args.only_submit_jobs and not matches(path, config.notebooks_that_submit_jobs):
+        if args.only_submit_jobs and not matches(path, config.notebooks_that_submit_jobs+config.notebooks_no_mock):
```
Eric-Arellano, line 88: Nit: better to not concatenate the two collections repeatedly inside the for loop. Instead, define a variable outside of the for loop

Eric-Arellano (approved):

Issue #1167, opened 2024-04-12
review/refactor AdaptVQE tutorial (HIGH PRIORITY)


javabster:
[ CURRENTLY UNDER PRODUCT REVIEW IN QTE REPO]

This is a new tutorial that Meltem on Nate's team has been working on, please work with her on this issue.

- [ ] #1193
- [ ] Address any feedback from product review in QTE repo
- [ ] Flesh out some of the text descriptions
- [ ] refactor to use real hardware from the get go (atm the real hardware bit is just in one cell right at the end)
- [ ] is there a way to scale about to more qubits? Or explain at a high level how to get there?

javabster:
focus for first release: get it working with real hardware from the get go. overall improving pedagogy would be good, but can be improved incrementally after first release

PR #1187, opened 2024-04-16
Add Adapt VQE to tutorials

iheuze:
Closes https://github.com/Qiskit/documentation/issues/1167

- I added in some more explanations to the text descriptions.
- I commented the code a bit more.
- I added in refactored code for real hardware in step 3.

But I also came into some issues.

I only have the open plan and this one tutorial exceeds an entire month's worth of usage limits so I could not actually run it on the hardware
At the start when defining the molecule, mol.energy_elec and mol.energy_tot don't work for me and I can't find source code where they would work in the same way as energy_nuc.

Maybe I am just missing something but could be worth checking.
Any suggestions or insights would be appreciated. Thanks!


javabster:
hi @iheuze thank you for your interest in this piece but it has already been assigned to another docs maintainer. In future could you please ask to be assigned an issue before starting work on it

PR #1197, opened 2024-04-18
Indicate IBM Provider is deprecated in left navbar

Eric-Arellano:
We're also changing other parts of the UI like the API dropdown in the top nav bar. This will change the left ToC here:

Screenshot 2024-04-18 at 10 49 23 AM

This PR also removes the version number from the H1 for the release notes for Runtime and Provider to no longer have the version number in them. That didn't make sense to do because they're the same release notes for all versions.


arnaucasau (approved):
Looks good, thanks!

PR #1181, opened 2024-04-16
Execute notebooks in parallel

frankharkins:
Executes notebooks in parallel. This should help when executing many notebooks that submit jobs as they'll all submit their jobs immediately and begin queueing in parallel.

### Screenshots
<details><summary>Standard run</summary>
Screenshot 2024-04-16 at 13 15 20
</details>
<details><summary>Standard run (with errors)</summary>
Screenshot 2024-04-16 at 13 07 59
</details>
<details><summary>Running notebooks that submit jobs</summary>
Screenshot 2024-04-16 at 13 10 44
Screenshot 2024-04-16 at 13 09 41
</details>

PR #1201, opened 2024-04-18
Use asyncio to parallelize notebook testing

Eric-Arellano:
An alternative to https://github.com/Qiskit/documentation/pull/1181. This should help when executing many notebooks that submit jobs as they'll all submit their jobs immediately and begin queueing in parallel.

### DIFF
--- a/scripts/nb-tester/test-notebook.py
+++ b/scripts/nb-tester/test-notebook.py
@@ -82,11 +82,11 @@ class NotebookWarning:
     cell_index: int
     msg: str
 
-    def report(self):
+    def format(self) -> str:
         """
-        Format warning and print it
+        Format warning to pretty string
         """
-        message = f"Warning detected in cell {self.cell_index}:\n"
+        message = f"Cell {self.cell_index}:\n"
         for line in self.msg.splitlines():
             message += (
                 textwrap.fill(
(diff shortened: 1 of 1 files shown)


scripts/nb-tester/test-notebook.py
```diff
@@ -122,17 +122,17 @@ def extract_warnings(notebook: nbformat.NotebookNode) -> list[NotebookWarning]:
...
     possible_exceptions = (
         nbconvert.preprocessors.CellExecutionError,
         nbclient.exceptions.CellTimeoutError,
     )
     try:
-        nb = _execute_notebook(path, args)
+        nb = await _execute_notebook(path, args)
     except possible_exceptions as err:
         print("\r❌\n")
         print(err)
```
frankharkins, line 138: These print statements assumed synchronicity, for async we need to remove the carriage return and make them make sense in isolation. Same for the warnings and success messages.

```suggestion
        print(f"❌ Problem in {path}\n", err)
```
Is this the printing problem you mentioned? Or are you not seeing output at all?

Eric-Arellano: I wasn't seeing anything print after the `await`, like no results at all. Only the starting line. I did not debug at all why that is and whether the notebook finishes running or not. Although I did get an exception raised with the save-jobs notebook 🤔

frankharkins: FYI the problem was in line 167 which was passing the function return to `to_thread` rather than the function itself. See 92534d56a9f455273f03847e3bfbbcb3827174cb.

frankharkins (approved):

Eric-Arellano (commented):
Nice! Thanks

arnaucasau (approved):
Nice! All looks good to me, I also tested the script locally, and worked well!

PR #1202, opened 2024-04-18
Upgrade Runtime and Provider for notebook testing

Eric-Arellano:
Notebooks were tested locally with `tox`. Everything succeeded except for `save-jobs.ipynb` because it doesn't work with other users' accounts.

### DIFF
--- a/scripts/nb-tester/test-notebook.py
+++ b/scripts/nb-tester/test-notebook.py
@@ -12,7 +12,6 @@
 
 import argparse
 import sys
-import warnings
 import textwrap
 from dataclasses import dataclass
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


scripts/nb-tester/test-notebook.py
```diff
@@ -189,8 +186,8 @@ def find_notebooks() -> list[Path]:
 
 
 def cancel_trailing_jobs(start_time: datetime) -> bool:
-    """
-    Cancel any runtime jobs created after `start_time`.
+    """Cancel any runtime jobs created after `start_time`.
+
```
frankharkins, line 190: Is this from a linter or just personal preference? Personally, I prefer the original formatting but I don't mind too much as long as we pick one and stick to it.

frankharkins (approved):
Thanks!

scripts/nb-tester/test-notebook.py
Eric-Arellano: Personal preference so it's less verbose. What I cared about was adding a blank line between the first two sentences because Python docstring convention is to have only one sentence summary in the first paragraph. 

I'll revert the change to be consistent with what we had before. Either style is fine.

PR #1203, opened 2024-04-19
Reorganise "Save jobs" page

frankharkins:
This page originally submitted a simple job using the cloud simulator as a demonstration. Since the simulators are being deprecated, #1160 removed the job-submitting step and hardcoded the `job_id` into the notebook. This was a bad idea because it means the notebook breaks if not run with my IBM Quantum account.

This PR reorganises the page to get `job_id` programatically first. This lets us avoid hardcoding the `job_id` in the notebook, which means we can run the notebook with any IBM account that has submitted at least one job.

### DIFF
--- a/docs/run/save-jobs.ipynb
+++ b/docs/run/save-jobs.ipynb
@@ -19,7 +19,7 @@
     "\n",
     "IBM Quantum automatically stores results from every job for you to retrieve at a later date. Use this feature to continue quantum programs across kernel restarts and review past results. You can get the ID of a job programmatically through its `job_id` method, or you can see all your submitted jobs and their IDs through the [Jobs dashboard](https://quantum.ibm.com/jobs).\n",
     "\n",
-    "Once you have the job ID, use the `QiskitRuntimeService.jobs` method to retrieve it."
+    "To find a job programatically, use the [`QiskitRuntimeService.jobs`](/api/qiskit-ibm-runtime/qiskit_ibm_runtime.QiskitRuntimeService#jobs) method. The following cell finds any jobs submitted in the last day. The `created_after` argument must be a [`datetime.datetime`](https://docs.python.org/3.8/library/datetime.html#datetime.datetime) object."
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


Eric-Arellano (approved):
Looks good! Are we confident that CI will always have jobs run in the past day, the next time we go to regenerate this notebook? Would it instead be safer to do something like past 3 days or past week? It won't fail if no jobs were run in the past day, but the output would be empty

frankharkins:
> Are we confident that CI will always have jobs run in the past day...

It think it's likely it might not, but IMO empty output is OK. Open to other suggestions though.

Eric-Arellano:
> Open to other suggestions though.

You could change it to the past week. Or probably even better: ask for all jobs but use a slice to only show up to 3, `jobs[:3]`

```python
>>> [1,2,3,4,5][:3]
[1, 2, 3]
>>> [1][:3]
[1]
```

frankharkins:
I quite like the `datetime` example since it's something a user would realistically want to do, and a bit harder to work out on your own. Maybe we could extend the time period to a year but only display three of them?

frankharkins:
As a compromise, I went with the last 90 days as `timedelta` doesn't seem to accept `years` and 3 months should be more than enough if we're running the cron job every two weeks.

docs/run/save-jobs.ipynb
```diff
@@ -19,7 +19,7 @@
...
     "\n",
-    "To find a job programatically, use the [`QiskitRuntimeService.jobs`](/api/qiskit-ibm-runtime/qiskit_ibm_runtime.QiskitRuntimeService#jobs) method. The following cell finds any jobs submitted in the last day. The `created_after` argument must be a [`datetime.datetime`](https://docs.python.org/3.8/library/datetime.html#datetime.datetime) object."
+    "To find a job programatically, use the [`QiskitRuntimeService.jobs`](/api/qiskit-ibm-runtime/qiskit_ibm_runtime.QiskitRuntimeService#jobs) method. By default, this returns the most recent jobs, but you can also filter jobs by backend name, creation date, and more. The following cell finds any jobs submitted in the last year. The `created_after` argument must be a [`datetime.datetime`](https://docs.python.org/3.8/library/datetime.html#datetime.datetime) object."
```
Eric-Arellano, line 22:
```suggestion
    "To find a job programatically, use the [`QiskitRuntimeService.jobs`](/api/qiskit-ibm-runtime/qiskit_ibm_runtime.QiskitRuntimeService#jobs) method. By default, this returns the most recent jobs, but you can also filter jobs by backend name, creation date, and more. The following cell finds any jobs submitted in the last three months. The `created_after` argument must be a [`datetime.datetime`](https://docs.python.org/3.8/library/datetime.html#datetime.datetime) object."
```

Eric-Arellano (approved):

docs/run/save-jobs.ipynb
frankharkins: oops thanks

PR #1204, opened 2024-04-19
Add the `isDedicatedPage` prop to the Class component

arnaucasau:
This PR adds the `isDedicatedPage` prop we use for exceptions and functions to the class component. This change will help us decide when we need a blue left bar in inlined classes.


Eric-Arellano (approved):
Nice 🚀

PR #1207, opened 2024-04-19
Change how we specify Python versions for testing

frankharkins:
We currently test everything with Python 3.11, but this is a problem as not all writers have it available on their system. We also want to make sure our docs work with as many versions of Python as possible.

This PR allows writers to test their docs with any version of Python, and sets CI to test with Python 3.8 (the minimum version that Qiskit supports).


Eric-Arellano (approved):
Thanks!

PR #1208, opened 2024-04-19
Make notebook tester into package

frankharkins:
Makes the notebook tester into an installable package so we can re-use it in other repos.

Main changes are:
* Adding `pyproject.toml` and moving script to `qiskit_docs_notebook_tester/__init__.py`
* Moving the lists of notebooks to a config file
* Splitting the requirements needed for running the code in the notebooks form those neeed by the testing script itself

The interface via `tox` should be unaffected.


scripts/nb-tester/notebooks.toml
```diff
@@ -0,0 +1,20 @@
+# Used to find all notebooks in the repo
+all-notebooks = "[!.]*/**/*.ipynb"
+
+# Always exclude notebooks matching the following patterns
+notebooks-exclude = [
+    "docs/api/**",
```
Eric-Arellano, line 6: These shouldn't exist for actual APIs. This would also unintentionally ignore `/api/migration-guides`. 

```suggestion
```

You could make this the default option so that it doesn't have to be set at all in TOML. I imagine it will be the same value for closed source?

scripts/nb-tester/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[build-system]
+requires = [ "hatchling",]
+build-backend = "hatchling.build"
+
+[project]
+name = "qiskit-docs-notebook-tester"
+version = "0.0.1"
+description = "Tool for the Qiskit docs team to test their notebooks"
+requires-python = ">=3.8"
```
Eric-Arellano, line 9: We use Python 3.9 syntax, which imo is fine for our tutorials. See https://github.com/Qiskit/documentation/pull/1210

But you'll need to require 3.11+ if you want to use `tomllib` since it wasn't in the std lib. You could instead use `tomli` or `tomlkit`. (`toml` is not maintained.) `tomli` looks good to me: https://pypi.org/project/tomli/. Technically this would only impact people running `nb-tester`, but if we're not running our notebooks with 3.9 in CI then it's likely we'll unintentionally use code that breaks with 3.9 in our notebooks.

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -24,27 +25,16 @@
...
-    "tutorials/build-repitition-codes/notebook.ipynb",
-    "tutorials/chsh-inequality/notebook.ipynb",
-    "tutorials/grovers-algorithm/notebook.ipynb",
-    "tutorials/quantum-approximate-optimization-algorithm/notebook.ipynb",
-    "tutorials/repeat-until-success/notebook.ipynb",
-    "tutorials/submitting-transpiled-circuits/notebook.ipynb",
-    "tutorials/variational-quantum-eigensolver/notebook.ipynb",
-]
-
+def get_globs(path: str) -> dict[str, str | list[str]]:
```
Eric-Arellano, line 28: Nit: more appropriate name would be `read_config`. It's not quite globs, which I would expect to be a `list[str]`.

Consider used `TypedDict` to make this interface more clear: https://docs.python.org/3/library/typing.html#typing.TypedDict. I recommend it - it makes the code much more obvious.

At that point, I like to use this idiom:

```python
from __future__ import annotations

from dataclasses import dataclass

@dataclass
class Config:
    opt1: str
    opt2: bool

     @classmethod
     def read(cls, path: read) -> Config:
         return cls(...)
```

`config = Config.read()` reads quite nicely.

Eric-Arellano: Added benefits of the dataclass:

1. You can error if required keys are not set
2. Avoid `KeyError` when accessing the config. Right now, `globs["notebooks-that-submit-jobs"])` would be a cryptic error if it was not set properly in the TOML file.

tox.ini
```diff
@@ -4,12 +4,14 @@ env_list = py311
...
-commands = python scripts/nb-tester/test-notebook.py {posargs}
+commands = test-docs-notebooks {posargs} --config-path scripts/nb-tester/notebooks.toml
 
-[testenv:lint]
-commands = squeaky --check --no-advice {posargs:docs tutorials}
-
-[testenv:fix]
-commands = squeaky {posargs:docs tutorials}
+[testenv:{lint,fix}]
+deps = squeaky==0.7.0
```
Eric-Arellano, line 14: Excellent change so that the install is faster

tox.ini
```diff
@@ -4,12 +4,14 @@ env_list = py311
...
+commands = 
+  lint: squeaky --check --no-advice {posargs:docs tutorials}
+  fix:  squeaky {posargs:docs tutorials}
```
Eric-Arellano, line 17: TIL! Cool.

scripts/nb-tester/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[build-system]
+requires = [ "hatchling",]
+build-backend = "hatchling.build"
+
+[project]
+name = "qiskit-docs-notebook-tester"
```
Eric-Arellano, line 6: Nit: `ibm-quantum-notebook-tester` is more appropriate.

Eric-Arellano (commented):
Nice!

scripts/nb-tester/pyproject.toml
frankharkins: I'm not sure, the package is specific to our team and I think this name makes it clear where the package comes from (e.g. if someone's inspecting their environment).

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
frankharkins: Nice, thanks!

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -10,10 +10,14 @@
...
 # been altered from the originals.
 
+from __future__ import annotations
+
+
 import argparse
 import asyncio
 import sys
 import textwrap
+import tomllib
```
Eric-Arellano, line 20: This will cause a `ModuleNotFoundError`, which is a subclass of `ImportError`. Instead, the Python idiom is:

```python
try:
    import tomllib
except ImportError:
    import tomli as tomllib
```

Fun fact: I first got involved in Python dev tooling by doing Python 2 -> 3 migrations 🐍

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -22,30 +26,27 @@
...
+    all_notebooks: str
+    notebooks_exclude: list[str]
+    notebooks_that_submit_jobs: list[str]
 
+    @classmethod
+    def read(cls, path: str) -> Config:
+        """
+        Load the globs from the TOML file
+        """
+        return cls(**tomllib.loads(Path(path).read_text()))
```
Eric-Arellano, line 44: Warning this will be a cryptic error if the three options are not set:

```
>>> from dataclasses import dataclass
>>> @dataclass
... class Foo:
...   a: str
...
>>> Foo(**{})
  File "<stdin>", line 1, in <module>
TypeError: __init__() missing 1 required positional argument: 'a'
```

Imo that's fine though because this is only going to be used in our two repos that we tightly control. I don't expect content writers to be changing the keys in the TOML file. So we can keep it simple.

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -207,8 +207,8 @@ def cancel_trailing_jobs(start_time: datetime) -> bool:
 
     print(
         f"⚠️ Cancelling {len(jobs)} job(s) created after {start_time}.\n"
-        "Add any notebooks that submit jobs to NOTEBOOKS_EXCLUDE in "
-        "`scripts/nb-tester/test-notebook.py`."
+        "Add any notebooks that submit jobs to `notebooks-that-submit-jobs` in "
```
Eric-Arellano, line 210: Style nit: consider having the arg be `config_path: str`, then the caller passes `arg.config_path` directly. It keeps the scope of this function smaller and easier to anticipate in `_main()` what the function does.

Eric-Arellano (commented):
Thanks! Readme needs to be updated with new instructions for updating the TOML file when you have a notebook that submits jobs

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
frankharkins: Is there any advantage to the try/except over just using `tomli` all the time?

Eric-Arellano: No, it's simpler to always use `tomli`. But I wouldn't rename it to be `tomllib` because that's confusing - they're not 1:1 libraries.

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
frankharkins: I've added a more helpful error message for that case

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -41,7 +40,13 @@ def read(cls, path: str) -> Config:
...
         Load the globs from the TOML file
         """
-        return cls(**tomllib.loads(Path(path).read_text()))
+        try:
+            return cls(**tomli.loads(Path(path).read_text()))
+        except TypeError as err:
+            raise ValueError(
+                f"Couldn't read config from {path}; check it exists and the"
+                " entries are correct."
+            ) from err
```
Eric-Arellano, line 49: nice use of `from err`!

PR #1218, opened 2024-04-22
Remove tabs from "Plot quantum states"

frankharkins:
This page wrapped code blocks in `<TabItem>` tags to include them in the tab component. This looks nice but unfortunately relied on some implementation details of how the notebook is converted to the web page, which is something we want to avoid.

This PR removes the tabs and uses headings instead to unblock publishing the page while we work on a more robust solution.


Eric-Arellano (approved):
Thanks!

PR #1221, opened 2024-04-23
Remove code block inside `<details>` tag

frankharkins:
I missed this in #1218; removing for now while we work on a solution.


Eric-Arellano (approved):
Thanks!

PR #1222, opened 2024-04-23
Fix tutorial deployment process

frankharkins:
I thought I'd tested this in #1154 but I actually only tested the uploader test workflow. I'd forgotten to pass the environment from the input to the job.

Something which wasn't my fault is that some tutorial UUIDs seem to have changed, I'm not sure why this is but am investigating.

With the new UUIDs, the workflow is working ([see run](https://github.com/Qiskit/documentation/actions/runs/8798350887/job/24145159885)).

### DIFF
--- a/.github/workflows/deploy-tutorials.yml
+++ b/.github/workflows/deploy-tutorials.yml
@@ -22,7 +22,7 @@ jobs:
   deploy:
     name: Deploy tutorials to learning platform
     runs-on: ubuntu-latest
-    environment: ${{ github.events.inputs.environment }}
+    environment: ${{ inputs.environment }}
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


Eric-Arellano (approved):
Thanks! Spooky with the UUIDs. Does changing the metadata of a tutorial like its title impact anything? If you don't already, you could create a new fake tutorial to test things out.

frankharkins:
That's a good idea. I tried changing the title of the tutorial we use to test the uploader but that didn't do anything :/

frankharkins:
Spoke to the team and they confirmed UUIDs shouldn't change. I'm not sure what went wrong here (possibly still human error on my part), but I think it's OK to merge for now and we can handle any future problems as they occur.

PR #1224, opened 2024-04-23
Rename tutorial notebooks to be unique

Eric-Arellano:
We can name the notebook file what we want. So, we were using `notebook.ipynb` to not repeat the tutorial name both in its parent folder and the notebook file.

But repeating `notebook.ipynb` has been annoying. It makes it harder in your IDE to know which file is which. It also makes the CI artifact upload less useful because the artifact only has the file names without the parent folder name.


abbycross (approved):
👍

PR #1225, opened 2024-04-23
Switch to new Docker image for staging/PR builds

Eric-Arellano:
We use a new approach in closed source that sets `ENV HOSTNAME 0.0.0.0` in the Dockerfile to work around https://github.com/vercel/next.js/issues/53171 breaking IBM Code Engine with IPv6.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,9 +33,6 @@
 
 FROM icr.io/quantum-computing/iqp-channel-docs-dev:hostname-experiment
 
-# Debug https://github.com/vercel/next.js/issues/53171
-ENV HOSTNAME 0.0.0.0
-
 COPY docs/ /home/node/app/docs
 COPY translations/ /home/node/app/docs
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -31,7 +31,7 @@
 #   2. ❯ docker run --rm -p 3000:3000 -t qiskit-docs-preview
 #   3. Open up http://localhost:3000
 
-FROM icr.io/quantum-computing/iqp-channel-docs-dev:next13.4
+FROM icr.io/quantum-computing/iqp-channel-docs-dev:hostname-experiment
```
frankharkins, line 34: Is this a permanent name, or just while we test it out?

frankharkins (approved):

Dockerfile
Eric-Arellano: I was using it to experiment, then was too lazy to change it. I can change it if we want to `latest`?

Issue #1139, opened 2024-04-05
Review/refactor "submit user transpiled circuits" tutorials


javabster:
https://learning.quantum.ibm.com/tutorial/submit-transpiled-circuits

- [ ] review if this tutorial should be a tutorial or a docs page
- [ ] improve overall quality of content (including update to qiskit 1.0 and primitives v2)

frankharkins:
Piggybacking on this to report that the current notebook does not work. Here's the problem cell and its traceback.

```python
# Need to add measurements to the circuit
qc.measure_all()

circuits = []
for optimization_level in [0, 3]:
    t_qc = transpile(qc, backend, optimization_level=optimization_level, seed_transpiler=0)
    print(f"CNOTs (optimization_level={optimization_level}): ", t_qc.count_ops()["cx"])
    circuits.append(t_qc)
```
```pytb
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[5], line 7
      5 for optimization_level in [0, 3]:
      6     t_qc = transpile(qc, backend, optimization_level=optimization_level, seed_transpiler=0)
----> 7     print(f"CNOTs (optimization_level={optimization_level}): ", t_qc.count_ops()["cx"])
      8     circuits.append(t_qc)

KeyError: 'cx'
```

beckykd:
Also note that this tutorial says that the runtime still transpiles circuits but it doesn't.

beckykd:
Accidentally started updating this with #1286 , so now there is a PR for this work: #1540

Issue #1137, opened 2024-04-05
Review/refactor "build repetition codes" tutorial


javabster:
once this tutorial has been moved to this repo we need to do the following:

- [ ] work with Roy Elkabetz  to update the content (ibm provider is being deprecated so that needs changing)
- [ ] decide if this works better as a learning platform tutorial, or as a page in docs

PR #1226, opened 2024-04-24
Refresh outputs in all notebooks

frankharkins:
I ran the notebook testing script yesterday to test it was submitting jobs correctly, and noticed the jobs were being executed immediately. I decided to take advantage of this and leave the script running. These are the notebooks with the refreshed outputs.

Two notebooks did not execute correctly:
- `docs/run/save-jobs.ipynb` (#1227)
- `tutorials/submitting-transpiled-circuits/submitting-transpiled-circuits.ipynb` (#1139)
- `docs/run/get-backend-information.ipynb` (#1228)

And `tutorials/variational-quantum-eigensolver/vqe.ipynb` had not completed when I cancelled the script this morning (#1229).


Eric-Arellano (approved):
Thanks for doing this! I opened https://github.com/Qiskit/documentation/issues/1239 so that we don't need this as often in the future.

Before merging, please double check if the learning platform supports SVGs. Or merge this and confirm it works in staging, but be ready to revert the tutorials/ changes quickly need be.

frankharkins:
CI is also failing because the repetition codes notebook uses `qiskit-ibm-provider`, which doesn't work with our testing patch. I think it's ok to merge as the failure is not due to this PR and the notebook should be migrated to `qiskit-ibm-runtime` as part of #1137.

```
❌ Problem in tutorials/build-repetition-codes/build-repetition-codes.ipynb:
A cell timed out while it was being executed, after 300 seconds.
The message was: Cell execution timed out.
Here is a preview of the cell contents:
-------------------
job_no_correction = backend.run(transpiled_no_correction_circuit, shots=1000, dynamic=True)
result_no_correction = job_no_correction.result()
-------------------

Checking for trailing jobs...
⚠️ Cancelling 1 job(s) created after 2024-05-09 12:31:15.823406.
```

Eric-Arellano:
Instead, I recommend we do the blocklist mechanism: https://github.com/Qiskit/documentation/pull/1332#issuecomment-2102638367. We should avoid ever merging broken CI.

Eric-Arellano (approved):

PR #1232, opened 2024-04-24
Remove simulator from "Get backend information"

### LINKED ISSUE #1228: "Get backend information" uses cloud simulator

https://docs.quantum.ibm.com/run/get-backend-information

Here's the problem cell:

```python
service.backend("ibmq_qasm_simulator")
```

### Select all that apply

- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

### Describe the fix.

Switch to a different backend. While this should work for open users at the moment, the simulators will be deprecated soon and this notebook will break.

frankharkins:
This notebook explicitly requested a simulator backend, which doesn't work with our testing provider. They'll also be removed soon, so the best fix is to use a real device instead.

I've also changed the instance to the open provider so users see the same list of backends.

---

Fixes #1228.

### DIFF
--- a/docs/run/get-backend-information.ipynb
+++ b/docs/run/get-backend-information.ipynb
@@ -52,7 +52,7 @@
    "source": [
     "# Initialize your account\n",
     "from qiskit_ibm_runtime import QiskitRuntimeService\n",
-    "service = QiskitRuntimeService()\n",
+    "service = QiskitRuntimeService(instance=\"ibm-q/open/main\")\n",
     "\n",
     "service.backends()"
    ]
@@ -83,7 +83,7 @@
     }
    ],
    "source": [
-    "service.backend(\"ibmq_qasm_simulator\")"
+    "service.backend(\"ibm_kyoto\")"
    ]
   },
   {
--- a/docs/run/get-backend-information.ipynb
+++ b/docs/run/get-backend-information.ipynb
@@ -127,7 +127,7 @@
    "id": "6de23dcc-c2ba-4338-b1e8-4086ed653240",
    "metadata": {},
    "source": [
-    "Use these keyword arguments to filter by any attribute in backend configuration or status. A similar method is [`QiskitRuntimeService.least_busy()`](../api/qiskit-ibm-runtime/qiskit_ibm_runtime.QiskitRuntimeService#least_busy), which takes the same filters as `backends()` but returns the backend that matches the filters and has the least number of jobs pending in the queue:"
+    "Use these keyword arguments to filter by any attribute in backend configuration ([JSON schema](https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_configuration_schema.json)) or status ([JSON schema](https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_status_schema.json)). A similar method is [`QiskitRuntimeService.least_busy()`](../api/qiskit-ibm-runtime/qiskit_ibm_runtime.QiskitRuntimeService#least_busy), which takes the same filters as `backends()` but returns the backend that matches the filters and has the least number of jobs pending in the queue:"
    ]
   },
   {
--- a/docs/run/get-backend-information.ipynb
+++ b/docs/run/get-backend-information.ipynb
@@ -91,7 +91,7 @@
     "\n",
     "You can also filter the available backends by their properties. For more general filters, you can make advanced functions using a lambda function. Refer to the [API documentation](../api/qiskit-ibm-runtime/qiskit_ibm_runtime.QiskitRuntimeService#backends) for more details.\n",
     "\n",
-    "Let’s try getting only backends that fit these criteria:\n",
+    "The following code only returns backends that fit these criteria:\n",
     "\n",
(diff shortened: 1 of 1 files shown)


docs/run/get-backend-information.ipynb
```diff
@@ -52,7 +48,7 @@
    "source": [
     "# Initialize your account\n",
     "from qiskit_ibm_runtime import QiskitRuntimeService\n",
-    "service = QiskitRuntimeService()\n",
+    "service = QiskitRuntimeService(instance=\"ibm-q/open/main\")\n",
```
Eric-Arellano, line 51: Will this line work for general users who are running this notebook at home?

docs/run/get-backend-information.ipynb
```diff
@@ -151,7 +139,7 @@
     {
      "data": {
       "text/plain": [
-       "<IBMBackend('ibm_cairo')>"
+       "<IBMBackend('simulator_statevector')>"
```
Eric-Arellano, line 142: Womp womp. Should we set `simulator=False`?

frankharkins: IMO this is ok as it comes from `least_busy`, so the code will still work when the simulators are removed.

docs/run/get-backend-information.ipynb
frankharkins: I believe so, but I'll double-check with my personal account.

beckykd:
I also wanted to update this file to expand on the Filter backends section.  You can use it to add other things that are specified in backend(), like service.least_busy(operational=True, min_num_qubits=100, dynamic_reprate_enabled=True)  I didn't know what a Lambda function was, but it was really easy to add in that last value and saved me a lot of headaches.   The tricky bit was that the setting is `backend.dynamic_reprate_enabled=True`,  but you don't include `backend` in that service call.  Hopefully with your programming knowledge, you can make that more clear :)

docs/run/get-backend-information.ipynb
frankharkins: Yep; just confirmed

docs/run/get-backend-information.ipynb
```diff
@@ -95,7 +91,7 @@
     "\n",
     "You can also filter the available backends by their properties. For more general filters, you can make advanced functions using a lambda function. Refer to the [API documentation](../api/qiskit-ibm-runtime/qiskit_ibm_runtime.QiskitRuntimeService#backends) for more details.\n",
```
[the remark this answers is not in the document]
frankharkins: A lambda function is special syntax for writing a one-line function. The only real advantage is you can define them inline, so they're often used in cases like this where you need a simple, one-time-use function. Here's how you'd filter using a standard function.

```python
def has_dynamic_reprate(backend):
    return backend.has_dynamic_reprate == True

service.backends(filter=has_dynamic_reprate)
```

If we rewrite it using a lambda function, we can fit it all on one line.

```python
service.backends(filter=lambda backend: backend.has_dynamic_reprate==True)
```

The sentence is a bit clunky. I think we can remove the mention of "lambda" as it's not really necessary, just a style choice. What do you think of this?

```suggestion
    "You can also filter the available backends by their properties. For more general filters, use a function that accepts a backend object and returns `True` if it meets your criteria. Refer to the [API documentation](../api/qiskit-ibm-runtime/qiskit_ibm_runtime.QiskitRuntimeService#backends) for more details.\n",
```

docs/run/get-backend-information.ipynb
```diff
@@ -139,7 +127,7 @@
    "id": "6de23dcc-c2ba-4338-b1e8-4086ed653240",
    "metadata": {},
    "source": [
-    "A similar method is [`QiskitRuntimeService.least_busy()`](../api/qiskit-ibm-runtime/qiskit_ibm_runtime.QiskitRuntimeService#least_busy), which takes the same filters as `backends()` but returns the backend that matches the filters and has the least number of jobs pending in the queue:"
+    "Use these keyword arguments to filter by any attribute in backend configuration or status. A similar method is [`QiskitRuntimeService.least_busy()`](../api/qiskit-ibm-runtime/qiskit_ibm_runtime.QiskitRuntimeService#least_busy), which takes the same filters as `backends()` but returns the backend that matches the filters and has the least number of jobs pending in the queue:"
```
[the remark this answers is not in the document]
frankharkins: Good point, I can't find anywhere in our docs that explains what values this can take. The best I can find is the JSON schema on GitHub, but it's not very reader-friendly.

```suggestion
    "Use these keyword arguments to filter by any attribute in backend configuration ([JSON schema](https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_configuration_schema.json)) or status ([JSON schema](https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_status_schema.json)). A similar method is [`QiskitRuntimeService.least_busy()`](../api/qiskit-ibm-runtime/qiskit_ibm_runtime.QiskitRuntimeService#least_busy), which takes the same filters as `backends()` but returns the backend that matches the filters and has the least number of jobs pending in the queue:"
```

beckykd (approved):

Issue #1216, opened 2024-04-19
Attributes: redesign how we present the default value and type hint


Eric-Arellano:
TODO

Eric-Arellano:
Closed by @arnaucasau. After #1233, the `=` will be removed.

<img width="358" alt="Screenshot 2024-04-29 at 12 29 47 PM" src="https://github.com/Qiskit/documentation/assets/14852634/681f030e-12af-4c75-a50a-e95a83979fc1">

PR #1233, opened 2024-04-24
Remove the extra `=` character from the attribute's default value prop

arnaucasau:
Part of #1216
The attributes will be redesigned, and we won't need to store the `=` character as part of the default value prop. This character can be added from the react component if it's needed again in the future, similar to how we are already using the type hints.


scripts/lib/api/generateApiComponents.ts
```diff
@@ -230,7 +230,7 @@ function prepareAttributeProps(
   const attributeTypeHint = text
     .slice(Math.min(colonIndex + 1, equalIndex), equalIndex)
     .trim();
-  const attributeValue = text.slice(equalIndex, text.length).trim();
+  const attributeValue = text.slice(equalIndex + 1, text.length).trim();
```
Eric-Arellano: Comment would be helpful:

> We skip the first character to leave off the `=`.

But should this be +2 to skip the blank space?

arnaucasau: > But should this be +2 to skip the blank space?

Technically, yes. The catch is that we need to use a `trim()` to it anyway, so it has no difference. I think it's good to use +1 and only remove the `=,` just in case any attribute doesn't have the space for any weird circumstance like the other bugs in the historical versions.

> Comment would be helpful

I'll add a general comment for `attributeValue` and `attributeTypeHint` :+1:

Eric-Arellano (approved):

Issue #1212, opened 2024-04-19
Runtime and Provider APIs: flatten submodules in left navigation


Eric-Arellano:
Before:

![image](https://github.com/Qiskit/documentation/assets/14852634/298d69ca-2e97-410d-a058-747773d6a718)

After:

![image](https://github.com/Qiskit/documentation/assets/14852634/3703bba1-511d-4fdc-8414-aaba035d2313)

Qiskit is different and should be handled in https://github.com/Qiskit/documentation/issues/1211.

### Provider modules

<details><summary>code to find modules</summary>

```
from pathlib import Path
import json

project_folder = "qiskit-ibm-provider"
module_name = "qiskit_ibm_provider"

raw = Path(f"docs/api/{project_folder}/_toc.json").read_text()
content = json.loads(raw)

all_modules = set()

def add_modules(entry) -> None:
    if entry.get("title").startswith(module_name):
        all_modules.add(entry["title"])
    if "children" in entry:
        for child in entry["children"]:
            add_modules(child)


add_modules(content)
print("\n".join(sorted(all_modules)))
```

</details>

qiskit_ibm_provider
qiskit_ibm_provider.job
qiskit_ibm_provider.jupyter
qiskit_ibm_provider.transpiler
qiskit_ibm_provider.transpiler.passes
qiskit_ibm_provider.transpiler.passes.basis
qiskit_ibm_provider.transpiler.passes.scheduling
qiskit_ibm_provider.utils
qiskit_ibm_provider.visualization

### Runtime modules
qiskit_ibm_runtime.transpiler is basically empty and we may want to remove it by modifying Sphinx itself. But we need to be careful to not break URLs without care.

qiskit_ibm_runtime
qiskit_ibm_runtime.fake_provider
qiskit_ibm_runtime.options
qiskit_ibm_runtime.transpiler
qiskit_ibm_runtime.transpiler.passes
qiskit_ibm_runtime.transpiler.passes.basis
qiskit_ibm_runtime.transpiler.passes.scheduling

jyu00:
This looks good, but I don't think there is much value in updating qiskit-ibm-provider, since it's deprecated. Unless it's easier to do them both at once.

Eric-Arellano:
> Unless it's easier to do them both at once.

Yeh, easier to be consistent. Thanks!

Issue #1240, opened 2024-04-24
Revisit the transpiler module pages for Runtime API docs


Eric-Arellano:
The module pages aren't particularly useful.

https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/transpiler:

<img width="563" alt="Screenshot 2024-04-24 at 3 34 13 PM" src="https://github.com/Qiskit/documentation/assets/14852634/94f2f194-8a63-4c69-b730-af95387a5718">

https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.transpiler.passes

<img width="585" alt="Screenshot 2024-04-24 at 3 34 31 PM" src="https://github.com/Qiskit/documentation/assets/14852634/1fef94c9-3a68-4660-8667-5d15516d5c9b">

https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.transpiler.passes.basis: 
<img width="531" alt="Screenshot 2024-04-24 at 3 34 40 PM" src="https://github.com/Qiskit/documentation/assets/14852634/058ffc4b-358b-4ab2-8f38-1b5198c02bca">

https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.transpiler.passes.scheduling is really useful though.

frankharkins:
What's the solution? To add more information to the docstrings? I wonder if the reason scheduling had so much information is because it's not covered in the rest of the docs, whereas the transpiler is.

Eric-Arellano:
Or consider if we should remove any of the pages. In particular, I don't understand why https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.transpiler.passes.basis is documented

Issue #1211, opened 2024-04-19
Qiskit API: group modules into logical sections


Eric-Arellano:
For example

![image](https://github.com/Qiskit/documentation/assets/14852634/b3af19bd-fcbe-49a2-8609-68a6f943ac57)

![image](https://github.com/Qiskit/documentation/assets/14852634/e3446ffa-1bf3-4213-afd8-afd2eb464d97)

The main changes:

1. group related modules into new categories
2. flatten submodules to reduce overall nesting

## Qiskit 1.0 modules

<details><summary>code to find modules</summary>

```python
from pathlib import Path
import json

raw = Path("docs/api/qiskit/_toc.json").read_text()
content = json.loads(raw)

all_modules = set()

def add_modules(entry) -> None:
    if entry.get("title").startswith("qiskit."):
        all_modules.add(entry["title"])
    if "children" in entry:
        for child in entry["children"]:
            add_modules(child)


add_modules(content)
print("\n".join(sorted(all_modules)))
```

</details>

<details><summary>module list</summary>

qiskit.assembler
qiskit.circuit
qiskit.circuit.classical
qiskit.circuit.classicalfunction
qiskit.circuit.library
qiskit.circuit.singleton
qiskit.compiler
qiskit.converters
qiskit.dagcircuit
qiskit.exceptions
qiskit.passmanager
qiskit.primitives
qiskit.providers
qiskit.providers.basic_provider
qiskit.providers.fake_provider
qiskit.providers.models
qiskit.pulse
qiskit.qasm2
qiskit.qasm3
qiskit.qobj
qiskit.qpy
qiskit.quantum_info
qiskit.result
qiskit.scheduler
qiskit.synthesis
qiskit.synthesis.unitary.aqc
qiskit.transpiler
qiskit.transpiler.passes
qiskit.transpiler.passes.synthesis.plugin
qiskit.transpiler.preset_passmanagers
qiskit.transpiler.preset_passmanagers.plugin
qiskit.utils
qiskit.visualization
</details>

javabster:
I'm sure the qiskit devs will have a better suggestion, but how about something like this for the grouping? I tried to group things into sections according to what users might be most likely to look for:

Circuit & Operator Construction

- qiskit.circuit
- qiskit.circuit.classical
- qiskit.circuit.classicalfunction
- qiskit.circuit.library
- qiskit.circuit.singleton
- qiskit.quantum_info

Transpilation
- qiskit.dagcircuit
[...]

PR #1241, opened 2024-04-24
Flatten left table of contents for Runtime and Provider

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/1212. As explained there, this is because:

1. Qiskit SDK is getting a bigger revamp in https://github.com/Qiskit/documentation/issues/1211 and it will also flatten submodules, so we want to be consistent.
2. The module list is so short that it's useful to have the navbar be flatter. It makes things less crowded and more discoverable.

Note that some of the Runtime provider pages aren't very useful...I opened https://github.com/Qiskit/documentation/issues/1240 to track that.

For now, this is implemented as the `Pkg` constructor setting a boolean argument for whether modules should be nested or not. I'm coming up with an alternative—but similar factoring—while working on https://github.com/Qiskit/documentation/issues/1211. This PR is ~prework.

Screenshot 2024-04-24 at 4 02 34 PM

Screenshot 2024-04-24 at 4 02 10 PM

### DIFF
--- a/scripts/lib/api/generateToc.ts
+++ b/scripts/lib/api/generateToc.ts
@@ -29,7 +29,12 @@ type Toc = {
   collapsed: boolean;
 };
 
-function nestModule(id: string): boolean {
+export function nestModule(pkg: Pkg, id: string): boolean {
+  // Most packages don't nest submodules because there module list is so small,
+  // so it's more useful to show them all and have less nesting.
+  if (!pkg.nestModulesInToc) {
+    return false;
+  }
   // For example, nest `qiskit.algorithms.submodule`, but
   // not `qiskit.algorithms` which should be top-level.
   return id.split(".").length > 2;
@@ -43,6 +48,7 @@ export function generateToc(pkg: Pkg, results: HtmlToMdResultWithUrl[]): Toc {
 
   addItemsToModules(items, tocModulesByTitle, tocModuleTitles);
   const nestedTocModules = getNestedTocModulesSorted(
+    pkg,
     tocModules,
     tocModulesByTitle,
     tocModuleTitles,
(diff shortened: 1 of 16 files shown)


Eric-Arellano:
@arnaucasau @frankharkins this got sign off from Abby M, Kevin T, and Jessie, so this review is ensuring the implementation is good.

scripts/lib/api/generateToc.ts
```diff
@@ -111,12 +112,18 @@ function addItemsToModules(
 }
 
 function getNestedTocModulesSorted(
+  pkg: Pkg,
   tocModules: TocEntry[],
   tocModulesByTitle: Dictionary<TocEntry>,
   tocModuleTitles: string[],
 ): TocEntry[] {
-  const nestedTocModules: TocEntry[] = [];
+  // Most packages don't nest submodules because there module list is so small,
```
Eric-Arellano, line 120:
```suggestion
  // Most packages don't nest submodules because their module list is so small,
```

frankharkins (approved):
Thanks! Code makes sense and preview looks good.

Something we might want to think about (out of scope of this PR) is how we handle these long names on the sidebar. One option is to remove the `qiskit_ibm_runtime.` from each module name, but design might have a better solution.

Screenshot 2024-04-25 at 12 10 55

Eric-Arellano:
> One option is to remove the qiskit_ibm_runtime. from each module name, but design might have a better solution.

I don't think we could do that though because the top level module `qiskit_ibm_runtime` has a lot of imports and we couldn't use the empty string 🤔 I think it'd be confusing if we had `qiskit_ibm_runtime` then `fake_provider` - people might think you could do `import fake_provider`.

frankharkins:
Ok, let's see if design have a better idea

Issue #1242, opened 2024-04-25
Design: ideas for long module names in left nav bar of API docs?


Eric-Arellano:
We've gotten feedback the module names are quite long, which makes the left navbar look cramped:

<img width="253" alt="Screenshot 2024-04-24 at 4 02 34 PM" src="https://github.com/Qiskit/documentation/assets/14852634/2bbbaa7f-e187-4170-b2c5-02273e340b07">

<img width="257" alt="Screenshot 2024-04-24 at 4 02 10 PM" src="https://github.com/Qiskit/documentation/assets/14852634/53816ee2-f7ee-4d1d-90a5-64d7795fd776">

Is there anything design wise we can do to improve the situation? I don't think we can remove `qiskit_ibm_runtime` entirely because that top entry has files underneath it, like you can directly import `qiskit_ibm_runtime` -- if we removed `qiskit_ibm_runtime`, it would be an empty string. 

We could maybe do something like this?

```
qiskit_ibm_runtime >
...fake_provider >
...options >
...transpiler.passes >
```

PR #1250, opened 2024-04-26
Truncate left ToC entries for Runtime and Provider

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/1242. We decided with design that the simplest solution to the visual clutter is to truncate.

Before:

Screenshot 2024-04-26 at 5 04 55 PM

After:

Screenshot 2024-04-26 at 5 05 51 PM

## Does not add a hover

We were hoping to allow you to hover to see the full text, like you do with links via `title`, where we show you the full module path. 

But I could not see an easy way to do this because this is plain text rather than a hover. We'd need to add a new mechanism to `_toc.json` where you can set something like `hoverTitle` for each entry, and then we'd need to add a new mechanism to the docs app. So, I didn't add this for now. 

I think it's obvious enough, especially because we show you the full import path at the top of the module overview page:
Screenshot 2024-04-26 at 5 06 14 PM


frankharkins (approved):
Thanks! Diff is very neat

PR #1254, opened 2024-04-29
Advanced Runtime topic edits

beckykd:


abbycross (approved):

docs/run/max-execution-time.mdx
```diff
@@ -16,11 +16,7 @@ The maximum execution time for a job is the smaller of these values:
 - The value set for max_execution_time
 - The system-determined job timeout value
 
-<Admonition type="note">
-  As of 7 August 2023, the `max_execution_time` value is based on _quantum time_ instead of wall clock time. Quantum time is the duration a quantum system is committed to fulfilling a user request.
-  
-  Simulator jobs continue to use wall clock time because they do not have quantum time.
-</Admonition>
+The `max_execution_time` value is based on _quantum time_, not wall clock time. Quantum time is the duration a quantum system is committed to fulfilling a user request. However, simulator jobs use wall clock time because they do not have quantum time.
```
jyu00, line 19: I think we should change this wording to avoid confusion with session usage, which is also time a system is committed to a user's workload. Perhaps we can use the "quantum time is the time spent by the QPU complex to process the job".

docs/run/max-execution-time.mdx
```diff
@@ -48,10 +44,6 @@ The system calculates an appropriate job timeout value based on the input circui
 
 For example, if you specify `max_execution_time=5000` (approximately 83 minutes), but the system determines it should not take more than 5 minutes (300 seconds) to execute the job, then the job is canceled after 5 minutes.
 
-## Session maximum execution time
```
jyu00, line 51: This doesn't need to be removed. A session still has a `max_time`, and the behavior remains the same.

PR #1258, opened 2024-04-29
Remove confusing API index pages in Qiskit 0.19

Eric-Arellano:
## Problem

This is how Qiskit 0.19 looks:

Screenshot 2024-04-29 at 5 00 14 PM

The three `qiskit` entries are:

https://docs.quantum.ibm.com/api/qiskit/0.19
https://docs.quantum.ibm.com/api/qiskit/0.19/ibmq-provider
https://docs.quantum.ibm.com/api/qiskit/0.19/aqua

0.19 is the only version like that, e.g. Qiskit 0.24 doesn't have duplicates:

Screenshot 2024-04-29 at 5 00 36 PM

In 0.19 docs, the first two `qiskit` entries are index pages for Qiskit Aqua and Qiskit IBMQ Provider. Those are index pages that simply list all modules, rather than module overview pages that describe the modules `qiskit.aqua` and `qiskit.providers.ibmq`.

Having the three duplicate `qiskit` entries is going to be a problem with https://github.com/Qiskit/documentation/issues/1211 when we group modules.  

## Solution

I simply deleted the two index files from the HTML and regenerated the docs. If we like this PR, I will upload that HTML change to Box.

I went with deletion because it's the simplest fix. We don't have these index pages in every version after 0.19, and it works fine. https://github.com/Qiskit/documentation/issues/1211 will also help with module discovery through better grouping.

We could instead rewrite the HTML for the two problematic files, but I figured it's fine to do the simpler approach that's consistent with later versions.


frankharkins (approved):
Thanks! I agree with this solution. Weird that these only appear in 0.19

PR #1259, opened 2024-04-30
Refactor generateToc.ts to not use dictionary type

Eric-Arellano:
Prework for https://github.com/Qiskit/documentation/pull/1256. It was confusing how we pass both `tocModules: TocEntry[]` and `tocModulesByTitle: Dictionary<TocEntry>`. It wasn't clear how those values relate, and `Dictionary` is an under-documented API by Lodash.

Also adds `hasPrefix`, which we will need in https://github.com/Qiskit/documentation/pull/1256.

### DIFF
--- a/scripts/lib/api/generateToc.ts
+++ b/scripts/lib/api/generateToc.ts
@@ -10,7 +10,7 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
-import { Dictionary, isEmpty, keyBy, keys, orderBy } from "lodash";
+import { isEmpty, orderBy } from "lodash";
 
 import { getLastPartFromFullIdentifier } from "../stringUtils";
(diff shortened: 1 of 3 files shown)


scripts/lib/api/generateToc.ts
```diff
@@ -100,7 +102,7 @@ function addItemsToModules(
     );
 
     if (itemModuleTitle) {
-      const itemModule = tocModulesByTitle[itemModuleTitle];
+      const itemModule = tocModulesByTitle.get(itemModuleTitle) as TocEntry;
```
Eric-Arellano, line 105: We can be confident `.get()` works because `itemModuleTitle` uses `tocModuleTitles`, which is all the `.keys()` of `tocModulesByTitle`.

frankharkins (approved):
Thanks! My initial confusion was working out why we were passing the same data in different forms. It makes sense now I know its for optimization, but this type hint helps too.

PR #1260, opened 2024-04-30
Update feedback issue template to point straight to new feedback form

abbycross:
In conversation with @JulesMurphy - remove the current "Impressions and feedback" form in the issue templates, and instead link out directly to the new feedback form at https://ibmxm.iad1.qualtrics.com/jfe/form/SV_7Uq9FCMjZPyTsFM?Q_PopulateResponse={%22QID20%22:%223%22}


Eric-Arellano (approved):
Thanks!

Issue #1249, opened 2024-04-26
Expose index page in left ToC for Runtime and Provider


Eric-Arellano:
https://docs.quantum.ibm.com/api/qiskit-ibm-runtime and https://docs.quantum.ibm.com/api/qiskit-ibm-provider are orphan pages, meaning they don't show up in our left ToC. That is unlike Qiskit https://docs.quantum.ibm.com/api/qiskit, which is the index page. 

Note that we only started surfacing the index pages for provider and runtime as part of https://github.com/Qiskit/documentation/pull/561. Runtime and Provider are interesting because they have an index page that lists all the modules, then a page about the top-level modules qiskit_ibm_runtime and qiskit_ibm_provider. This gets set up in 

https://github.com/Qiskit/documentation/blob/2a2c40a9045106c954c49bbd873d71148e02120c/scripts/lib/api/specialCaseResults.ts#L27-L28

Qiskit is unlike Provider and Runtime, as it does have its index page https://docs.quantum.ibm.com/api/qiskit in the left sidebar. Qiskit does not have a page for all the imports of the top level module `qiskit`, it only has this index page. Note how `qiskit/index` has module metadata:

https://github.com/Qiskit/documentation/blob/2a2c40a9045106c954c49bbd873d71148e02120c/docs/api/qiskit/index.mdx?plain=1#L5-L6

But Runtime and Provider do not:

https://github.com/Qiskit/documentation/blob/2a2c40a9045106c954c49bbd873d71148e02120c/docs/api/qiskit-ibm-runtime/index.mdx?plain=1#L1-L5

Qiskit should stay the same, other than renaming "qiskit" to something like "Index" or "API index" as part of https://github.com/Qiskit/documentation/issues/1211. Meanwhile, Runtime and Provider need to add their index pages to the left ToC using the same name as Qiskit uses.

PR #1267, opened 2024-05-01
Refactor generateToc tests to use normal project structure

Eric-Arellano:
This is to unblock https://github.com/Qiskit/documentation/issues/1249. We now use the project folder that `Pkg.mock().outputDir()` would use.

This also removes inline snapshots. They can't be updated via `npm test -- -u` due to Prettier, so it's more helpful to use a normal `expect().toEqual()`.

### DIFF
--- a/scripts/lib/api/generateToc.test.ts
+++ b/scripts/lib/api/generateToc.test.ts
@@ -28,50 +28,50 @@ describe("generateToc", () => {
       {
         meta: {
           apiType: "module",
-          apiName: "qiskit_ibm_runtime",
+          apiName: "my_quantum_project",
         },
-        url: "/docs/runtime",
+        url: "/api/my-quantum-project",
         ...DEFAULT_ARGS,
       },
       {
         meta: {
           apiType: "module",
-          apiName: "qiskit_ibm_runtime.options",
+          apiName: "my_quantum_project.options",
         },
-        url: "/docs/options",
+        url: "/api/my-quantum-project/options",
         ...DEFAULT_ARGS,
       },
       {
         meta: {
           apiType: "module",
-          apiName: "qiskit_ibm_runtime.options.submodule",
+          apiName: "my_quantum_project.options.submodule",
         },
-        url: "/docs/qiskit_ibm_runtime.options.submodule",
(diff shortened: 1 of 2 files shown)


scripts/lib/api/generateToc.test.ts
```diff
@@ -28,145 +28,143 @@ describe("generateToc", () => {
...
-          apiName: "qiskit_ibm_runtime.single",
+          apiName: "my_quantum_project.single",
         },
-        url: "/docs/single",
+        url: "/api/my-quantum-project/single",
         ...DEFAULT_ARGS,
       },
     ]);
 
-    expect(toc).toMatchInlineSnapshot(`
```
Eric-Arellano, line 95: Only the page names and URLs changed in this diff

scripts/lib/api/generateToc.test.ts
```diff
@@ -181,42 +179,55 @@ describe("generateToc", () => {
...
     const topLevelEntries: TocGroupingEntry[] = [
-      { moduleId: "my_project", title: "API index", kind: "module" },
+      { moduleId: "my_quantum_project", title: "API index", kind: "module" },
       { name: "Group 2", kind: "section" },
       { name: "Group 1", kind: "section" },
+      {
+        moduleId: "my_quantum_project.another",
+        title: "dedicated module",
+        kind: "module",
+      },
```
Eric-Arellano, line 189: This is a new test because once we have https://github.com/Qiskit/documentation/issues/1249, we couldn't be confident that putting the API index page in the title came from our grouping code rather than the new `ensureIndexPage()`.

scripts/lib/api/generateToc.test.ts
```diff
@@ -258,98 +273,94 @@ describe("generateToc", () => {
...
+            apiName: "my_quantum_project",
           },
-          url: "/docs/runtime/qiskit_ibm_runtime",
+          url: "/api/my-quantum-project",
           isReleaseNotes: true,
         },
       ],
     );
 
-    expect(toc).toMatchInlineSnapshot(`
```
Eric-Arellano, line 278: No changes made other than URLs and page titles

frankharkins (approved):
Thanks!

Issue #1274, opened 2024-05-01
Attribute name and header not showing


Eric-Arellano:
Given this MDX:

https://github.com/Qiskit/documentation/blob/c387562e8ca3dac26a653cb96773022bd78410bf/docs/api/qiskit/dev/circuit.mdx?plain=1#L323-L329

We render it like this:

<img width="265" alt="Screenshot 2024-05-01 at 2 14 08 PM" src="https://github.com/Qiskit/documentation/assets/14852634/5b2728fa-5d00-4493-9036-b8afb3b12504">

This makes it unclear what the docs are referring to. This is the corresponding Sphinx:

https://github.com/Qiskit/qiskit/blob/a41690d6075ac9438ff01b2346a35aa21dc3a568/qiskit/circuit/__init__.py#L367-L368

We need to do something like show a header with the attribute name.

PR #1282, opened 2024-05-02
Add missing properties headers

arnaucasau:
This PR adds headers to all properties in the API docs that didn't have a dedicated page to match the design of the other components.

Closes https://github.com/Qiskit/documentation/issues/1274


Eric-Arellano (approved):
Huh, so this only impacted 4 pages? Weird. To confirm, you checked Provider and Runtime too?

PR #1283, opened 2024-05-02
[WIP] Fix header hierarchy

### LINKED ISSUE #1215: inline classes: fix header hierarchy
Part of https://github.com/Qiskit/documentation/issues/199.

Right now, we hardcode methods/attributes/exceptions to have an h3 header no matter what. For classes, we don't add any header. That does not work though when you for example a function documented on a module page inside an h3 section already.

Instead, the header should take into account the parent header. Top-level functions / classes / exceptions should be one heading level lower than the parent. If the parent is the h1, then classes/exceptions should probably be h2 and functions h3.

Methods and Attributes should have a lower heading level than their class. Right now, we add the `Methods` and `Attributes` headers. That would mean the class should be `-1` from the parent header, `Methods`/`Attributes` headers should be `-2`, and Function/Attribute components should be `-2`. We might want to remove the `Methods`/`Attributes` headers for inline classes, though, to reduce the amount of heading levels.

arnaucasau:
Closes #1215

This PR is a **work in progress** and aims to fix the header hierarchy for the API docs as explained in https://github.com/Qiskit/documentation/issues/1215

### DIFF
--- a/scripts/lib/api/generateApiComponents.ts
+++ b/scripts/lib/api/generateApiComponents.ts
@@ -88,17 +88,17 @@ function prepareProps(
   id: string,
 ): ComponentProps {
   const preparePropsPerApiType: Record<string, () => ComponentProps> = {
-    class: () => prepareClassProps($child, $dl, githubSourceLink, id),
+    class: () =>
+      prepareClassOrExceptionProps($, $main, $child, $dl, githubSourceLink, id),
     property: () =>
       preparePropertyProps($child, $dl, priorApiType, githubSourceLink, id),
     method: () =>
-      prepareMethodProps($, $child, $dl, priorApiType, githubSourceLink, id),
+      prepareMethodProps($, $main, $child, $dl, priorApiType, githubSourceLink, id),
     attribute: () =>
-      prepareAttributeProps($, $child, $dl, priorApiType, githubSourceLink, id),
-    function: () =>
-      prepareFunctionOrExceptionProps($, $child, $dl, id, githubSourceLink),
+      prepareAttributeProps($, $main, $child, $dl, priorApiType, githubSourceLink, id),
+    function: () => prepareFunctionProps($, $main, $child, $dl, id, githubSourceLink),
     exception: () =>
-      prepareFunctionOrExceptionProps($, $child, $dl, id, githubSourceLink),
+      prepareClassOrExceptionProps($, $main, $child, $dl, githubSourceLink, id),
   };
 
   const githubSourceLink = prepareGitHubLink($child, apiType === "method");
@@ -111,7 +111,9 @@ function prepareProps(
   return preparePropsPerApiType[apiType]();
 }
 
-function prepareClassProps(
+function prepareClassOrExceptionProps(
+  $: CheerioAPI,
+  $main: Cheerio<any>,
   $child: Cheerio<any>,
   $dl: Cheerio<any>,
   githubSourceLink: string | undefined,
@@ -131,6 +133,10 @@ function prepareClassProps(
       isDedicatedPage: true,
     };
   }
+  const headerLevel = getHeaderLevel($, $main, $dl);
+  $(
+    `<h${headerLevel} class="class-header">${getLastPartFromFullIdentifier(id)}</h${headerLevel}>`,
+  ).insertBefore($dl);
   return props;
 }
 
@@ -162,6 +168,7 @@ function preparePropertyProps(
 
 function prepareMethodProps(
   $: CheerioAPI,
+  $main: Cheerio<any>,
   $child: Cheerio<any>,
   $dl: Cheerio<any>,
   priorApiType: string | undefined,
@@ -183,7 +190,10 @@ function prepareMethodProps(
         isDedicatedPage: true,
       };
     } else if ($child.attr("id")) {
-      $(`<h3>${getLastPartFromFullIdentifier(id)}</h3>`).insertBefore($dl);
+      const headerLevel = getHeaderLevel($, $main, $dl);
+      $(
+        `<h${headerLevel} class="method-header">${getLastPartFromFullIdentifier(id)}</h${headerLevel}>`,
+      ).insertBefore($dl);
     }
   }
   return props;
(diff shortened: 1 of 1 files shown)


arnaucasau:
All docs have been regenerated in another PR to verify everything is correct before adding them to this one. This is the [preview](https://qiskit-docs-preview-pr-1289.1799mxdls7qz.us-south.codeengine.appdomain.cloud/)

scripts/lib/api/generateApiComponents.ts
```diff
@@ -87,17 +87,42 @@ function prepareProps(
...
+      prepareAttributeProps(
+        $,
+        $main,
+        $child,
+        $dl,
+        priorApiType,
+        githubSourceLink,
+        id,
+      ),
     method: () =>
```
arnaucasau, line 102: I was thinking that, probably, we can merge `prepareMethodProps` and `prepareFunctionProps`, but I think it's better to explore this in a follow-up that will have as a goal to simplify this script.

scripts/lib/api/generateApiComponents.ts
```diff
@@ -350,3 +399,44 @@ export async function htmlSignatureToMd(
     .replace(/^`/, "")
     .replace(/`$/, "");
 }
+
+function getHeaderLevel($: CheerioAPI, $main: Cheerio<any>, $dl: Cheerio<any>) {
```
Eric-Arellano, line 403:
```suggestion
function getHeaderLevel($: CheerioAPI, $main: Cheerio<any>, $dl: Cheerio<any>): number {
```

scripts/lib/api/generateApiComponents.ts
```diff
@@ -350,3 +399,44 @@ export async function htmlSignatureToMd(
...
+  const priorHeaderLevel = getPriorHeaderLevel($, $main, $dl);
+  if (priorHeaderLevel) {
+    if (+priorHeaderLevel == 6) {
+      throw new Error("API component cannot set inexisting header: <h7>");
```
Eric-Arellano, line 407: Grammar nit:

```suggestion
      throw new Error("API component cannot set non-existent header: <h7>");
```

scripts/lib/api/generateApiComponents.ts
```diff
@@ -350,3 +399,44 @@ export async function htmlSignatureToMd(
...
+    }
+
+    return Math.max(3, +priorHeaderLevel + 1);
```
Eric-Arellano, line 410: It'd be good to use a variable to set the `3` number since you use it twice, e.g. `minLevel = 3`. For the comment, suggested tweak:

// We don't allow the header to be h1 or h2 because it's too large design-wise for API components.
// We try to ensure that the API docs are set up so there is always at least an h2 above the API component,
// but this is not always the case, especially with historical API docs. That means that we sometimes
// jump from h1 to h3. That's bad, but the tradeoff we're making to avoid using h2 for API components.

scripts/lib/api/generateApiComponents.ts
```diff
@@ -350,3 +399,44 @@ export async function htmlSignatureToMd(
...
+  $: CheerioAPI,
+  $main: Cheerio<any>,
+  $dl: Cheerio<any>,
+): string | undefined {
+  const siblings = $dl.siblings();
+  for (const sibling of siblings) {
+    const $sibling = $(sibling);
+    if ($sibling.data("header-type")) {
+      continue;
+    }
```
Eric-Arellano, line 428: A comment explaining this would be helpful

scripts/lib/api/generateApiComponents.ts
```diff
@@ -350,3 +399,44 @@ export async function htmlSignatureToMd(
...
+    const tagName = $sibling.get(0)?.tagName;
+    if (tagName?.match(/h[1-6]/)) {
+      return tagName.substring(1);
+    }
+  }
+
+  // If there's no header among the siblings, we look for the prior inline class in some
+  // parent node previously set
```
Eric-Arellano, line 436:
```suggestion
  // If there's no header among the siblings, we look if the component belongs inside a class.
  // Otherwise, we return undefined.
```

scripts/lib/api/generateApiComponents.ts
```diff
@@ -350,3 +399,44 @@ export async function htmlSignatureToMd(
...
+  return $main
+    .find("[data-header-type=class-header]")
```
Eric-Arellano, line 438: Why is this safe? Imagine this HTML

```html
<Function id="func" />

<Class>
   <Function id="method" />
</Class>
```

Because you're searching `$main`, wouldn't the first `Function` end up matching with the class after it?

scripts/lib/api/htmlToMd.test.ts
```diff
@@ -987,3 +989,113 @@ By default this is sys.stdout.</p></li>
...
+
+    ##### method4
+
+    <Function id="qiskit.test.myClass2.myClass3.method4" signature="abstract method4()">
+      Method 4
+    </Function>
+  </Class>
+</Class>
+
+### method5
```
Eric-Arellano, line 1095: This would be a function, not a method

Could you please add an Exception as well? It'd be good for it to be under an h3 header to test that we set it to h4

scripts/lib/api/generateApiComponents.ts
arnaucasau: Good question! That wouldn't happen because the header for the class after it wouldn't have been created yet. To search for the previous headers, I leverage the fact that we set each header in the same order they appear on the website.

If the case were in reverse, like:
```mdx
<Class>
   <Function id="method" />
</Class>

<Function id="func" />
```

wouldn't be a problem either because we would find a header among the siblings before getting to this last return. All methods that are not in an inline class have a sibling header already in the HTML.

Eric-Arellano: Ah, makes sense. That would be great to comment

scripts/lib/api/htmlToMd.test.ts
arnaucasau: Done! I added two exceptions inside the classes with different ordering to test that we always set the correct heading level. I also added an attribute in between the classes :+1:

scripts/lib/api/generateApiComponents.ts
arnaucasau: That's a cool idea! As discussed in our 1:1, I'll implement this and try to simplify the script in a follow-up :+1:

Issue #1281, opened 2024-05-02
Remove `name` attribute from MDX API components


Eric-Arellano:
Now that we decided to not remove headers for API docs (https://github.com/Qiskit/documentation/issues/948), we no longer use the `name` attribute for anything. We should remove it to simplify the codebase.

PR #1284, opened 2024-05-02
Remove `name` attribute from API components

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/1281.


arnaucasau (approved):
Awesome, thank you Eric! We also need to remove the unnecessary code on IQP, but I can do it :)

Eric-Arellano:
> We also need to remove the unnecessary code on IQP, but I can do it :)

Agreed. I was thinking of doing it once we finish https://github.com/Qiskit/documentation/issues/1252

PR #1291, opened 2024-05-03
[WIP] Remove signature for the Attribute component

### LINKED ISSUE #1217: Attributes with incomplete signatures for some versions of qiskit
There are some attributes in the qiskit API docs with incomplete signatures. We can find two examples in qiskit 0.19 at [BaseSchema](https://docs.quantum.ibm.com/api/qiskit/0.19/qiskit.validation.BaseSchema#attributes) and [FlowController](https://docs.quantum.ibm.com/api/qiskit/0.19/qiskit.transpiler.FlowController#attributes).

For the former, we can see how the attribute `TYPE_MAPPING` and `opts` don't have a correct default value:

Captura desde 2024-04-19 22-59-31

These attributes seem to correspond to the `Schema` class of the marshmallow library (search for `TYPE_MAPPING` on this [page](https://marshmallow.readthedocs.io/en/stable/_modules/marshmallow/schema.html#)). In the next screenshot we can see how the signature would look like without processing it with a unified plugin like in the API generation script and instead using `$child.text()`:

Captura desde 2024-04-18 14-33-02

In the previous screenshot we can see also one attribute from the `FlowController` class which in some versions like 0.40, has that exact [value](https://docs.quantum.ibm.com/api/qiskit/0.40/qiskit.transpiler.FlowController#attributes):

Captura desde 2024-04-19 23-25-42

We should figure out why sometimes these signatures are incomplete for some versions. One solution to explore would be to remove the [unified plugin](https://github.com/Qiskit/documentation/blob/f480d9c33e11a8be62fde525385eff181da872cb/scripts/lib/api/generateApiComponents.ts#L363) used in `generateApiComponents.ts` and replace it with `$child.text()`, making sure we espace the apostrophes and double quotes symbols of the signature.

arnaucasau:
This PR removes the signature prop from the Attribute component and instead, sets attributeTypeHint and attributeValue. This change simplifies the script that was using different props for attributes embedded in the class page and the rest.

As a side effect, this PR also fixes an issue with some default values not showing correctly. These values were treated as signatures and wrongly converted into HTML by a unified plugin.

Closes #1217 #1252


Eric-Arellano (approved):
👏 did you try regenerating the other APIs to confirm this does not unintentionally change them?

PR #1294, opened 2024-05-03
Add script to migrate to new `guides/` structure

Eric-Arellano:
This is the first step in https://github.com/Qiskit/documentation/issues/1297.

We're going to be making last-minute changes for the docs reorg, such as finalizing URLs at the last minute. So, it's useful to have a program that can immediately regenerate the entire migration for us at any time. This program will eventually generate two commits:

1. Move files to guides/
2. Update non-API internal links, like `/run/my-page` to `/guides/my-page`

Thanks to this script being checked in to version control beforehand, we will have less code to review in the final migration PR.

This program is not going to be permanent. We decided in https://github.com/Qiskit/documentation/issues/1299 that we will not generate the Workflow index pages longterm, only in this initial PR. Instead, we'll add a validation check. So this program is only temporary while we work on the migration.

## Follow up improvements

This PR is only the initial foundation of the migration program. Remaining tasks:

* Finalize the page URLs
* Finalize the page organization in the left ToC and Patterns Index pages
* Finalize the content for the new pages
* Delete the existing pages and save the change as a Git commit.
* Get CI to be green when running the migration program, such as the link checker.
* https://github.com/Qiskit/documentation/issues/1299
* https://github.com/Qiskit/documentation/issues/1460


javabster:
- ~/guides/index -> /guides/intro-to-qiskit~
- some pages giving 404s: guides/hello-world, circuit-library etc.
- guides/patterns-index -> guides/intro-to-patterns
- guides/pattern -> remove "index"
- "synthesizing unitary operators" -> "synthesize unitary operators"
- "bit ordering in qiskit" -> "Bit-ordering in the Qiskit SDK"
- "configure runtiem options" -> "configure runtime options"
- flatten execution modes sub-heading and move "manage jobs" above FAQs in list. rename FAQs  to "Execution modes FAQs". Also can drop "qiskit runtime" from sidebar titles (but keep in main page title)
- "retrieve and save results" page should go in the "manage jobs" section
- new visualization dropdown in tools that includes: visualize results, plot quantum states and visualize circuits.
- flatten qiskit serverless dropdown

scripts/patterns-reorg/entries.py
```diff
@@ -0,0 +1,538 @@
+# This code is a Qiskit project.
```
Eric-Arellano, line 1: This file can be mostly skimmed. We'll work closely with the content team to finalize the page titles and new URLs. This file is the Source Of Truth™️ for the migration.

scripts/patterns-reorg/entries.py
```diff
@@ -0,0 +1,538 @@
...
+    map_content,
+    postprocess_index_content,
+    optimize_content,
+    patterns_index_content,
+)
+
+
+# ------------------------------------------------------------------------------
+# Standalone pages
+# ------------------------------------------------------------------------------
```
Eric-Arellano, line 26: The comments in this section are probably worth reading. It took me a while to figure out how to handle the situation.

scripts/patterns-reorg/entries.py
```diff
@@ -0,0 +1,538 @@
...
+            VISUALIZE_RESULTS_PAGE,
+        ),
+    ),
+    SERVERLESS_PAGE,
+)
+
+
+TOP_LEVEL_ENTRIES = (
+    Entry("Get started"),
+    *GET_STARTED_CHILDREN,
```
Eric-Arellano, line 532: This will be revamped once we have proper label support

scripts/patterns-reorg/models.py
```diff
@@ -0,0 +1,115 @@
...
+import dataclasses
+from dataclasses import dataclass, field
+from pathlib import PurePath, Path
+
+
+@dataclass(frozen=True)
+class Entry:
+    title: str
+    slug: str | None = field(default=None, kw_only=True)
+    children: tuple[Entry, ...] = field(default_factory=lambda: (), kw_only=True)
```
Eric-Arellano, line 24: FYI `tuple[X, ...]` is the immutable version of `list[X]`. We need it to be immutable to set `frozen=True`. I originally was using `list` and didn't care about immutability, but it resulted in a bug with `filter_entries`, so I switched to this style.

scripts/patterns-reorg/models.py
```diff
@@ -0,0 +1,115 @@
...
+        else:
+            assert self.page_content is not None
+            content = self.page_content
+            extension = ".mdx"
+        (base_dir / f"{self.slug}{extension}").write_text(content)
+
+
+def entries_as_markdown_list(
+    entries: tuple[Entry, ...], *, indent: None | str = None
+) -> str:
```
Eric-Arellano, line 68: There might be a more readable way to write this than how I did it with `indent`. But it's working well, and this program is temporary, so it's probably fine

scripts/patterns-reorg/page_content.py
```diff
@@ -0,0 +1,205 @@
+# This code is a Qiskit project.
```
Eric-Arellano, line 1: No need to review closely this file. The content team will revamp it over time.

scripts/patterns-reorg/redirects.json
```diff
@@ -0,0 +1,69 @@
+{
```
Eric-Arellano, line 1: This page is auto-generated.

We may want the data in alternative formats for future uses like setting up redirects in closed source. We can easily re-generate the file however we want.

frankharkins (approved):
LGTM! Was very easy to understand

PR #1300, opened 2024-05-03
Test that Qiskit TocGrouping is in sync with index page

Eric-Arellano:
The Qiskit devs were worried about the index page from https://github.com/Qiskit/qiskit/pull/12333 falling out of sync with our custom grouping for `_toc.json`. 

This PR adds a test to ensure that the two files are in sync. In particular, it checks that the grouping and sequence of modules is identical. It does not check that the label for the groups are the same, though.

For now, we only check the dev docs because the latest docs (1.0) fail. https://github.com/Qiskit/documentation/issues/1341 will also start checking the latest docs once 1.1 is released.

### DIFF
--- a/scripts/lib/api/TocGrouping.test.ts
+++ b/scripts/lib/api/TocGrouping.test.ts
@@ -53,7 +53,7 @@ function validateTopLevelModuleAssumptions(): void {
       );
     } else if (entry.kind === "section" && entry.name.includes("qiskit.")) {
       throw new Error(
-        "Expected very `section` of QISKIT_TOC_GROUPING.entries to not have 'qiskit.' in the " +
+        "Expected every `section` of QISKIT_TOC_GROUPING.entries to not have 'qiskit.' in the " +
           "name. This will break the tests in this file. Either remove the module name or " +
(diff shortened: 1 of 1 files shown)


scripts/lib/api/TocGrouping.test.ts
```diff
@@ -0,0 +1,133 @@
...
+    ) {
+      throw new Error(
+        "Expected every top-level module of QISKIT_TOC_GROUPING to have the module name in " +
+          "its title, e.g. 'Quantum information (qiskit.quantum_info)'. This will break the " +
+          "tests in this file. Either add the module name to the title or rewrite these tests. " +
+          `Bad entry: ${entry.title}`,
+      );
+    } else if (entry.kind === "section" && entry.name.includes("qiskit.")) {
+      throw new Error(
+        "Expected very `section` of QISKIT_TOC_GROUPING.entries to not have 'qiskit.' in the " +
```
Eric-Arellano, line 56:
```suggestion
        "Expected every `section` of QISKIT_TOC_GROUPING.entries to not have 'qiskit.' in the " +
```

PR #1302, opened 2024-05-06
url change for serverless

abbycross:
The github.io URL has changed from quantum-serverless to qiskit-serverless, and there is no redirect.

https://qiskit-extensions.github.io/quantum-serverless/ -> https://qiskit-extensions.github.io/qiskit-serverless/

Update the URLs in our docs


abbycross:
In my ripgrep search I also turned up a result from the Qiskit Runtime API release notes. If we need to update the link there, it cannot be done in this repo.

Eric-Arellano (approved):

PR #1305, opened 2024-05-06
Refactor release notes handling of separate pages

Eric-Arellano:
This is a prefactor for the upcoming addition of the Cloud Transpiler. It does not yet have release notes, so we need to be able to disable the mechanism.


Eric-Arellano:
Each commit is distinct and has some description attached to it. I recommend reviewing by commit

frankharkins (approved):
Thanks for splitting it out into separate commits. These all seem like good improvements, thanks!

PR #1306, opened 2024-05-06
Allow packages to not have release notes

Eric-Arellano:
The upcoming cloud transpiler client docs will not have release notes at first.

### DIFF
--- a/scripts/lib/api/generateToc.test.ts
+++ b/scripts/lib/api/generateToc.test.ts
@@ -273,7 +273,7 @@ describe("generateToc", () => {
     });
   });
 
-  test("TOC with distinct release note files", () => {
+  test("TOC with separate release note files", () => {
     const toc = generateToc(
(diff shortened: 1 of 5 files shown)


scripts/lib/api/generateToc.test.ts
```diff
@@ -318,53 +318,54 @@ describe("generateToc", () => {
...
-      {
-        meta: {
-          apiType: "function",
-          apiName: "runSomething",
+  test("generate a toc without modules and releaes notes", () => {
+    const toc = generateToc(
+      Pkg.mock({
+        releaseNotesConfig: new ReleaseNotesConfig({ enabled: false }),
+      }),
+      [
```
Eric-Arellano, line 326: No changes made to this list - it's only formatting

frankharkins (approved):
Nice!

PR #1308, opened 2024-05-06
Allow not setting githubSlug

Eric-Arellano:
The new Qiskit transpiler service client will not have GitHub links when we first add it.

### DIFF
--- a/scripts/lib/api/Pkg.ts
+++ b/scripts/lib/api/Pkg.ts
@@ -33,7 +33,7 @@ type PackageType = "latest" | "historical" | "dev";
 export class Pkg {
   readonly name: string;
   readonly title: string;
-  readonly githubSlug: string;
+  readonly githubSlug?: string;
   readonly version: string;
   readonly versionWithoutPatch: string;
   readonly type: PackageType;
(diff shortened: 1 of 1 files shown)


scripts/lib/api/Pkg.ts
```diff
@@ -44,7 +44,9 @@ export class Pkg {
   constructor(kwargs: {
     name: string;
     title: string;
-    githubSlug: string;
+    // We use `null` rather than `undefined` to make sure the user really doesn't want to set this,
+    // rather than not realizing they should set it.
+    githubSlug: string | null;
```
frankharkins, line 49: What effect does this have on the code's behaviour? It doesn't look like L194 distinguishes between `null` and `undefined`.

Eric-Arellano: TypeScript expects you to set `githubSlug=null` in the constructor. Otherwise you could either set `githubSlug=undefined` or simply leave off `githubSlug` entirely. I'm nervous about people leaving off `githubSlug` without realizing they shouldn't.

But maybe I'm being overly nervous actually? We now have the error for the one place where we use `githubSlug`. So it probably is fine to use `undefined`?

frankharkins: Sounds good :)

frankharkins (approved):

PR #1311, opened 2024-05-07
Fix typo in TocGrouping

Eric-Arellano:
This temporarily broke docs regeneration. My bad!


scripts/lib/api/TocGrouping.ts
```diff
@@ -109,7 +109,6 @@ function qiskitModuleToSection(module: string): string | undefined {
       "qiskit.synthesis",
       "qiskit.synthesis",
       "qiskit.transpiler",
-      "qiskit.transpiler.passes",
```
Eric-Arellano, line 112: This wasn't wrong, but it's not necessary

arnaucasau (approved):
Thanks!

PR #1313, opened 2024-05-07
Add tutorial for error mitigation with Qiskit Runtime Estimator

### LINKED ISSUE #1168: Add tutorial on error mitigation with Qiskit Runtime Estimator
[ CURRENTLY UNDER PRODUCT REVIEW IN QTE REPO]

This tutorial is based on[ a summit demo](https://learning.quantum.ibm.com/tutorial/scale-quantum-computational-workflows) but uses Estimator instead of Sampler. It is a new notebook created by Saki and Pedro on Nate's team, please work with them on this issue.

- [ ] move [source file](https://github.com/IBM-Quantum-Technical-Enablement/quantum-enablement/tree/main/docs/notebooks/scaling-estimator) into this repo
- [ ] review and refactor if needed. currently only works well with estimator v1 but not v2. Address any feedback from product review

kevinsung:
Fixes #1168


javabster:
Phenomenal work Kevin! 

my comments:
- I think we might need to come up with a different title for this tutorial, atm it's too similar to the PEA utility notebook title 
- could you add a list of requirements to the top of this notebook (check out some of the other tutorial PRs to see how others have done it)
- in the intro when you select the efficientSU2 can you mention that this approach is generalisable to many types of circuits, I believe that was an important point from the original notebook.
- I like how clearly you set the goal of the tutorial in the intro, could you also add that first you will be showing a small example and then scaling it up in the second half
- Also in the intro, can you mention which of the EM techniques you will be comparing in the rest of the notebook
- more of a question - is there a reason we shouldn't just show the full 50q example from the start? The code example at the end looks very similar, maybe we could just use that from the get go?
- in general the explanation of the "scale up" section looks a bit sparse - are there any additional insights you could add there? atm it feels like that section was just lumped in at the end without much thought
- nice code comments!

Overall I think this is a great simplified version of the notebook, but i think we should definitely follow up with more of a deep dive into the different error mitigation techniques in additional documentation

abbycross:
@kevinsung does this [recent comment](https://github.ibm.com/IBM-Q-Software/ibm-quantum-learning-enablement/pull/114#issuecomment-79045702) about breaking changes to the Estimator result data for ZNE affect your tutorial?

kevinsung:
> @kevinsung does this [recent comment](https://github.ibm.com/IBM-Q-Software/ibm-quantum-learning-enablement/pull/114#issuecomment-79045702) about breaking changes to the Estimator result data for ZNE affect your tutorial?

I don't think so. I only access the ZNE results using the `evs` attribute which should be consistent with the non-ZNE results too.

kevinsung:
@javabster I've addressed your comments. I also added a "Conclusion" section and made a few other minor edits. Please take another look.

kevinsung:
>more of a question - is there a reason we shouldn't just show the full 50q example from the start? The code example at the end looks very similar, maybe we could just use that from the get go?

The main reason is simply ease of visualization. If you try to draw a 50-qubit circuit it's very slow and the output is too large to view comfortably. Also, in general it's preferable to build and test a workflow on small circuits first (I mention this now in the notebook).

javabster:
Looking great! Some last final small things:

- I really liked the bullet points at the end where you summarised the insights from the results 👏 
- it looks like you used numpy and mpl for the analyse results section, so those packages should be added to the requirements list
- for the mirror section, that's the same as the "compute-uncompute" technique right? If both terms are used by end users I think it would be good to mention that the mirror stuff is "otherwise known as compute-uncompute..."
- can you update the patterns subheadings to be in line with what the other tutorials are doing (e.g. analyse results should be post-process results)

kevinsung:
>it looks like you used numpy and mpl for the analyse results section, so those packages should be added to the requirements list

I don't think we should list Numpy and Matplotlib as explicit requirements because that would be redundant and noisy. Numpy is already a dependency of Qiskit (and just about every other Python library that works with numbers). And we already explicitly say Qiskit needs to be installed "with visualization support" which implies Matplotlib. Thoughts?

javabster:
I see your point about mpl but when you install qiskit does it also install numpy for you? I thought you had to install that separately

kevinsung:
> I see your point about mpl but when you install qiskit does it also install numpy for you? I thought you had to install that separately

Yes, installing Qiskit installs Numpy too.

kevinsung:
>- for the mirror section, that's the same as the "compute-uncompute" technique right? If both terms are used by end users I think it would be good to mention that the mirror stuff is "otherwise known as compute-uncompute..."
>- can you update the patterns subheadings to be in line with what the other tutorials are doing (e.g. analyse results should be post-process results)

Done

nonhermitian:
- Is it possible to get the error bars on the expectation values?  

- Can the operators be constructed using the sparse list method?

Looks very nice!

kevinsung:
> * Is it possible to get the error bars on the expectation values?
> 
>     * Can the operators be constructed using the sparse list method?
> 
> 
> Looks very nice!

Done!

javabster (approved):
LGTM 🚀

PR #1314, opened 2024-05-07
Add missing header - class without id

arnaucasau:
This PR adds the header of a class that was missing its id. The id was manually fixed in Box by modifying the HTML, and this is the regeneration as a follow-up of #1283


Eric-Arellano (approved):
Thanks <3

PR #1315, opened 2024-05-07
Deduplicate `generateApiComponents` dispatch table entries

arnaucasau:
This PR refactors the dispatch table used to prepare the props of each API component in order to simplify the `generateApiComponents.ts` script. The functions for "functions" and "methods" cannot be merged trivially because they check in different ways the `isDedicatedPage` prop. Using the same method for that gives us false positives on pages like [this](https://docs.quantum.ibm.com/api/qiskit/0.36/execute). Moreover, functions don't need any header in those cases.

Thanks @Eric-Arellano for the idea on #1283

### DIFF
--- a/scripts/lib/api/generateApiComponents.ts
+++ b/scripts/lib/api/generateApiComponents.ts
@@ -86,18 +86,29 @@ function prepareProps(
   apiType: ApiType,
   id: string,
 ): ComponentProps {
+  const prepClassOrException = () =>
+    prepareClassOrExceptionProps($, $child, $dl, githubSourceLink, id);
+  const prepFunction = () =>
+    prepareFunctionProps($, $child, $dl, githubSourceLink, id);
+  const prepMethod = () =>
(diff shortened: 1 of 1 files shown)


scripts/lib/api/generateApiComponents.ts
```diff
@@ -86,18 +86,29 @@ function prepareProps(
...
-      prepareAttributeProps($, $child, $dl, priorApiType, githubSourceLink, id),
-    function: () => prepareFunctionProps($, $child, $dl, id, githubSourceLink),
-    exception: () =>
-      prepareClassOrExceptionProps($, $child, $dl, githubSourceLink, id),
+    class: prepClassOrException,
+    exception: prepClassOrException,
+    property: prepAttributeOrProperty,
+    attribute: prepAttributeOrProperty,
+    method: prepMethod,
+    function: prepFunction,
```
Eric-Arellano, line 111: Nit that these aren't reused, so you could inline them. Either approach is fine to me

Eric-Arellano (approved):
Thanks!

PR #1316, opened 2024-05-07
Fix API docs script to tolerate empty docs folder

Eric-Arellano:
This was getting in the way of adding the new qiskit-transpiler-service docs, since its folder is empty. This also was a problem when rerunning the script if the folder had been previously deleted without having restored the deleted files before the rerun.

### DIFF
--- a/scripts/commands/updateApiDocs.ts
+++ b/scripts/commands/updateApiDocs.ts
@@ -128,10 +128,9 @@ async function prepareSphinxFolder(pkg: Pkg, args: Arguments): Promise<string> {
 
 async function deleteExistingMarkdown(pkg: Pkg): Promise<void> {
   const markdownDir = pkg.outputDir("docs");
-  if (pkg.isLatest() || (await pathExists(markdownDir))) {
-    console.log(
-      `Deleting existing markdown for ${pkg.name}:${pkg.versionWithoutPatch}`,
-    );
-    await rmFilesInFolder(markdownDir);
-  }
+  if (!(await pathExists(markdownDir))) return;
+  console.log(
+    `Deleting existing markdown for ${pkg.name}:${pkg.versionWithoutPatch}`,
+  );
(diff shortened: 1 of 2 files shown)


scripts/commands/updateApiDocs.ts
```diff
@@ -128,10 +128,9 @@ async function prepareSphinxFolder(pkg: Pkg, args: Arguments): Promise<string> {
 
 async function deleteExistingMarkdown(pkg: Pkg): Promise<void> {
   const markdownDir = pkg.outputDir("docs");
-  if (pkg.isLatest() || (await pathExists(markdownDir))) {
```
Eric-Arellano, line 131: I don't think it's necessary to check if `pkg.isLatest()` - it's already covered by the `pathExists` check. If the path does not exist, then there is nothing to delete, regardless of latest or not.

arnaucasau (approved):
Looks good, thanks! Tested locally and worked well. I left a suggestion to make the code simpler

scripts/lib/fs.ts
```diff
@@ -31,7 +29,18 @@ export async function pathExists(path: string) {
...
+  const dirMembers = await fs.readdir(dir);
+  await Promise.all(
+    dirMembers.map(async (member) => {
+      const fp = path.join(dir, member);
+      const stat = await fs.stat(fp);
+      if (stat.isFile()) {
+        await fs.unlink(fp);
+      }
+    }),
+  );
```
arnaucasau, line 45: That's a nice way of looking for the file. If we want to do it more succinct with the find command, I think we could add the -path parameter, something like:
```suggestion
 await $`find ${dir} -maxdepth 1 -type f -path "${dir}/*" | xargs rm -f {}`;
```

Eric-Arellano: TIL. Thanks Arnau!

PR #1317, opened 2024-05-07
Add Qiskit Transpiler Service Client docs

Eric-Arellano:
Docs for https://pypi.org/project/qiskit-transpiler-service/.

These docs are not ready to go live to production yet. But we can safely merge them in this repository to unblock infrastructure work in closed source. We won't have closed source sync over these files until we get the final okay from the transpiler service team.

PR #1318, opened 2024-05-07
Don't truncate API ToC by default

Eric-Arellano:
Turns out, this truncation does not work well with https://github.com/Qiskit/documentation/pull/1317. We agreed the best approach is to show the full module path for that package.

This PR also fixes https://github.com/Qiskit/documentation/pull/1308 to put the error in the correct location.

### DIFF
--- a/scripts/lib/api/Pkg.ts
+++ b/scripts/lib/api/Pkg.ts
@@ -186,12 +186,6 @@ export class Pkg {
    * `sphinx.ext.viewcode`, which means we need to deal with its quirks like handling `__init__.py`.
    */
   determineGithubUrlFn(): (fileName: string) => string {
-    if (!this.githubSlug) {
-      throw new Error(
-        `Encountered sphinx.ext.viewcode link, but Pkg.githubSlug is not set for ${this.name}`,
-      );
-    }
-
     // For files like `my_module/__init__.py`, `sphinx.ext.viewcode` will title the
     // file `my_module.py`. We need to add back the `/__init__.py` when linking to GitHub.
     const convertToInitPy = new Set([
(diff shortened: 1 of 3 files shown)


scripts/lib/api/Pkg.ts
```diff
@@ -27,6 +27,16 @@ export class ReleaseNotesConfig {
   }
 }
 
+export class TocConfig {
```
arnaucasau, line 30: What do you think about moving this to `TocGrouping.ts` and making it a `type` instead of a `class`? Something like:
```ts
export type TocConfig = {
  truncate?: boolean;
  tocGrouping?: TocGrouping;
};
```

arnaucasau (commented):
Nice! Thanks!

scripts/lib/api/Pkg.ts
Eric-Arellano: The class is nice because it allows us to use default args for the constructor. I used this same approach for `ReleaseNotesConfig`: https://github.com/Qiskit/documentation/pull/1305/commits/27b7a73fd21552ab17623d5bf23474a06926c05a

> What do you think about moving this to TocGrouping.ts

I could and we rename it to `TocConfig.ts`, but IMO the current spot is fine. `TocGrouping` is already pretty complex and focused entirely on the complex custom grouping mechanism. `Pkg.ts` is more generally about the config for a Pkg. Because `TocConfig` is such a simple class, I think it's fine to be near `Pkg`.

arnaucasau (approved):

scripts/lib/api/Pkg.ts
arnaucasau: Sounds good, thanks for the explanation!

PR #1320, opened 2024-05-07
Make generic scripts for new projects

Eric-Arellano:
Unblocks https://github.com/Qiskit/documentation/pull/1317

Context: we can't use `api/*` for globs because it would match against `api/migration-guides`.

### DIFF
--- a/scripts/commands/checkMetadata.ts
+++ b/scripts/commands/checkMetadata.ts
@@ -106,7 +106,7 @@ async function determineFiles(args: Arguments): Promise<[string[], string[]]> {
 }
 
 function isApi(filePath: string): boolean {
-  const apiFolders = Pkg.VALID_NAMES.map((pkg) => `/api/${pkg}`);
+  const apiFolders = Pkg.VALID_NAMES.map((pkg) => `/api/${pkg}/`);
   return apiFolders.some((urlPath) => filePath.startsWith(urlPath));
 }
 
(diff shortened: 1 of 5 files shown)


scripts/commands/checkMetadata.ts
```diff
@@ -104,11 +106,8 @@ async function determineFiles(args: Arguments): Promise<[string[], string[]]> {
 }
 
 function isApi(filePath: string): boolean {
-  return (
-    filePath.includes("/api/qiskit/") ||
-    filePath.includes("/api/qiskit-ibm-runtime/") ||
-    filePath.includes("/api/qiskit-ibm-provider/")
-  );
+  const apiFolders = Pkg.VALID_NAMES.map((pkg) => `/api/${pkg}`);
```
arnaucasau, line 109: The new urlPaths are missing the last `/`. I think the first one will match all the rest as all start with `/api/qiskit`.
```suggestion
  const apiFolders = Pkg.VALID_NAMES.map((pkg) => `/api/${pkg}/`);
```

arnaucasau (commented):
Nice work! Much better using the `Pkg.VALID_NAMES` for future extensions in the APIs

scripts/commands/checkMetadata.ts
Eric-Arellano: Ah, good catch!

arnaucasau (approved):

PR #1292, opened 2024-05-03
Changed naming from Quantum Serverless to Qiskit Serverless

pandasa123:
## Background

As Serverless is integrating into the Qiskit stack, we want to rebrand Quantum Serverless --> Qiskit Serverless to match. This specific PR changes naming only, but we have a later PR to adjust how we talk about Serverless and Patterns, common configuration elements like dependency management, etc. 

## Merge Date

**Do not merge yet**

We have to finish the [Github / PyPi side of things for Qiskit Serverless](https://github.com/Qiskit-Extensions/quantum-serverless/issues/1307) before this PR is merged. We expect everything to be merged latest by **May 20th**

## Stakeholders

@Eric-Arellano, we'd need a redirect since the page URL is changing

PR #1328, opened 2024-05-08
Update Qiskit Serverless URLs to final destination

Eric-Arellano:
This is the final planned URL.

I'll update the Runtime release notes link separately.


abbycross:
Should I go ahead with the naming changes included in #1292? @pandasa123 is that ready?

abbycross (approved):
👍 and I can do the naming changes in a different PR unless I should push them to this branch?

Eric-Arellano:
> Should I go ahead with the naming changes included in https://github.com/Qiskit/documentation/pull/1292? @pandasa123 is that ready?

Abby C and I discussed that the naming changes can be independent from changing the URL and setting up redirects. I recommend we wait to change the URL until we reorganize all the docs. But the name change is safe and can be done at any time.

Issue #1166, opened 2024-04-12
review/refactor Krylov tutorial (HIGH PRIORITY)


javabster:
This was originally [a summit demo](https://learning.quantum.ibm.com/tutorial/krylov-subspace-expansion) that has been converted into a 10q tutorial by Nate's team and is almost ready to deploy to learning as an open tutorial. Please work with Mirko (+ pedro?) on this issue

- [ ] move [source file](https://github.com/IBM-Quantum-Technical-Enablement/quantum-enablement/blob/main/docs/notebooks/krylov.ipynb) into this repo (retain author/commit history if possible)
- [ ] explain at a high level how to take this tutorial further and scale
- [ ] review and refactor 
- [ ] get a final product review and deploy to learning, update access controls to "open"

abbycross:
@frankharkins can you assist with moving this tutorial to this repo, preserving the author/commit history?

PR #1332, opened 2024-05-09
Refactor Krylov tutorial

frankharkins:
This tutorial was originally created in IBM-Quantum-Technical-Enablement/quantum-enablement#11. Part of #1166.


Eric-Arellano:
Maybe we need a blocklist mechanism for notebooks that completely fail? It'd be best to not try to fix the notebook in this PR, but we also don't want CI to be red.

frankharkins:
I think @abbycross will be working on this branch, so we can fix it in this PR

abbycross:
Hi @miamico! I've gone through to do voice/style edits on this tutorial, and am trying to run the notebook. The code throws an error in Step 3 - can you see what's going on?

Eric-Arellano:
> Hi @miamico! I've gone through to do voice/style edits on this tutorial, and am trying to run the notebook. The code throws an error in Step 3 - can you see what's going on?

Is it that the notebook is using 11 qubits? We're trying to run this notebook with our 5-qubit simulator in CI, but if it requires 11 qubits, that won't work. If this notebook can't be rewritten to work with the 5-qubit simulator, then we need to update `notebooks.toml` to move it from `notebooks_that_submit_jobs` to instead `notebooks_no_mock`. It's desirable to allow us to mock it if possible so that we get more CI validation.

miamico:
I'm working on a version of this notebook that will be using the hardware at utility scale. I'd hold off any further work on this version based on the simulator

abbycross:
> I'm working on a version of this notebook that will be using the hardware at utility scale. I'd hold off any further work on this version based on the simulator

@miamico ok thanks!!

frankharkins:
This is continued in inner-source.

PR #1335, opened 2024-05-09
Prettier ignore mypy_cache

Eric-Arellano:
If people run MyPy in the repo, they'll have the folder `.mypy_cache` which Prettier will format and Git will try to add.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ tsconfig.tsbuildinfo
 /.out/
 /.sphinx-artifacts/
 poetry.lock
+
+.mypy_cache/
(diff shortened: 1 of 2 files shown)


Eric-Arellano:
fyi @jakelishman

frankharkins (approved):

jakelishman:
It's also hitting `.github/**/*.yml` files for me, if you care about those.  Those may well be deliberate, though.

Eric-Arellano:
> It's also hitting .github/**/*.yml files for me, if you care about those. Those may well be deliberate, though.

Yep, deliberate. Same with README

PR #1336, opened 2024-05-09
Add option to use arbitrary Sphinx HTML artifact folder

jakelishman:
This adds a `--sphinx-artifact-folder` (alias `--sphinx-output`) option to manually specify the root of the Sphinx HTML output for a given package.  This implies `--skip-download`.  This is convenient for local API-docs development of the targetted packages, since I can use their regular docs-build processes and then just point `gen-api` at the right place without needing to copy anything into special folders.

For example, I've locally been using Sphinx as normal in Qiskit, then running
```bash
npm run gen-api -- -p qiskit -v 1.2.0 -a /path/to/docs/_build/html
```
to have the script directly target the Sphinx files without copying them.

I made the magic `/artifact` suffix a bit more local, rather than spread between two functions for two reasons:
1. it's easier to reason about magic that's only local.
2. if the `/artifact` is _always_ appended by `zxMain` outside of the control of `prepareSphinxFolder`, it defeated the main purpose of this option.

### DIFF
--- a/scripts/commands/updateApiDocs.ts
+++ b/scripts/commands/updateApiDocs.ts
@@ -67,7 +67,7 @@ const readArgs = (): Arguments => {
       implies: "skip-download",
       normalize: true,
       description:
-        "Skip downloading the the artifact from Box and instead use a given directory as the root of the Sphinx HTML output.",
+        "Skip downloading the artifact from Box and instead use the given directory as the root of the Sphinx HTML output.",
     })
     .parseSync();
 };
--- a/scripts/commands/updateApiDocs.ts
+++ b/scripts/commands/updateApiDocs.ts
@@ -26,6 +26,7 @@ interface Arguments {
   historical: boolean;
   dev: boolean;
   skipDownload: boolean;
+  sphinxArtifactFolder?: string;
 }
 
 const readArgs = (): Arguments => {
@@ -60,6 +61,14 @@ const readArgs = (): Arguments => {
       description:
         "Rather than downloading the artifact from Box, reuse what is already downloaded. This can save time, but it risks using an outdated version of the docs.",
     })
+    .option("sphinx-artifact-folder", {
+      alias: "sphinx-output",
+      type: "string",
+      implies: "skip-download",
+      normalize: true,
+      description:
+        "Skip downloading the the artifact from Box and instead use a given directory as the root of the Sphinx HTML output.",
+    })
     .parseSync();
 };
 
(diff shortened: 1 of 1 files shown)


scripts/commands/updateApiDocs.ts
```diff
@@ -60,6 +61,14 @@ const readArgs = (): Arguments => {
       description:
         "Rather than downloading the artifact from Box, reuse what is already downloaded. This can save time, but it risks using an outdated version of the docs.",
     })
+    .option("sphinx-artifact-folder", {
+      alias: "sphinx-output",
+      type: "string",
+      implies: "skip-download",
+      normalize: true,
+      description:
+        "Skip downloading the the artifact from Box and instead use a given directory as the root of the Sphinx HTML output.",
```
Eric-Arellano, line 70:
```suggestion
        "Skip downloading the artifact from Box and instead use the given directory as the root of the Sphinx HTML output.",
```

scripts/commands/updateApiDocs.ts
```diff
@@ -60,6 +61,14 @@ const readArgs = (): Arguments => {
       description:
         "Rather than downloading the artifact from Box, reuse what is already downloaded. This can save time, but it risks using an outdated version of the docs.",
     })
+    .option("sphinx-artifact-folder", {
+      alias: "sphinx-output",
```
Eric-Arellano, line 65: If you want an alias, I'd rather it be something short like `-a` for `artifact`. Right now, the two names feel quite similar and like we couldn't decide on the name

Eric-Arellano (commented):
Excellent! No need to add tests. @arnaucasau will document this in https://github.com/Qiskit/documentation/pull/1336

scripts/commands/updateApiDocs.ts
jakelishman: I don't really mind - this was mostly borne out of me not being to work out to have `yargs` have a command-line argument that differed from the name of the associated property in the object.  I wanted to _just_ have the command-line argument be `--sphinx-output`, but the internal details of the code want that property to be `sphinxArtifactFolder` for more consistency, which felt not-so-descriptive to me outside the context of the internal code.

Anyway, I don't mind much about the naming as long as the option exists - if you're rather I just remove the alias entirely, that's fine.

Eric-Arellano: Either remove the alias or use `-a` (or you can suggest something else). I like having the short alias to make it easier to use this script

Eric-Arellano (approved):
🚀

scripts/commands/updateApiDocs.ts
jakelishman: Done in 8c6434f5d1e14843f57cd8108934d510f959bb27.

Issue #1287, opened 2024-05-02
Add back `class` modifier in front of class signatures


Eric-Arellano:
This is how Sphinx renders https://docs.quantum.ibm.com/api/qiskit/dev/circuit#bits-and-registers

<img width="575" alt="Screenshot 2024-05-02 at 4 05 32 PM" src="https://github.com/Qiskit/documentation/assets/14852634/d8931a59-e7fb-45a5-b7ed-a3dec763b364">

Vs us:

<img width="475" alt="Screenshot 2024-05-02 at 4 05 41 PM" src="https://github.com/Qiskit/documentation/assets/14852634/d1ef8613-c98b-4e29-ae05-4e390f0f031f">

https://github.com/Qiskit/documentation/blob/eca9f437aa521d4419fe94e7537e60f405f30863/docs/api/qiskit/dev/circuit.mdx?plain=1#L260

Should we add back the `class` modifier? Should it only be for inline classes, or also dedicated pages like https://docs.quantum.ibm.com/api/qiskit/dev/qiskit.circuit.AnnotatedOperation?

<img width="695" alt="Screenshot 2024-05-02 at 4 09 32 PM" src="https://github.com/Qiskit/documentation/assets/14852634/a3ad8489-78fc-4029-bc44-24d753f7acce">

If so, should it be italicized? If it should be italicized, should we add a new attribute to the component like `signaturePrefix`?

--

What's interesting is some signatures are preserved, like `final class` and `static`. Why is this?

https://github.com/Qiskit/documentation/blob/eca9f437aa521d4419fe94e7537e60f405f30863/docs/api/qiskit/circuit_classical.mdx?plain=1#L63

https://github.com/Qiskit/documentation/blob/eca9f437aa521d4419fe94e7537e60f405f30863/docs/api/qiskit-ibm-runtime/qiskit_ibm_runtime.options.Options.mdx?plain=1#L90

Eric-Arellano:
We decided that we do want to do this. We still need to figure out the design. Devs should create a few examples for us to decide:

* `class` in the code section w/o italics
* `class` in the code section w/ italics (if possible)
* `class` before the code section w/o italics
* `class` before the code section w/ italics

We should audit if there are any other modifiers we've been dropping.

arnaucasau:
Here are the screenshots with an example of how each design would look like. I added the case where we have the modifier w/ bold in and before the code section.

### Inside the code section examples

#### Option 1. `class` in the code section

[...]

PR #1337, opened 2024-05-09
Add back modifiers in front of signatures

arnaucasau:
Part of  #1287

This PR adds modifiers to the signatures of our API components. We continue removing the modifiers for attributes and properties given that we don't show their signature on the website.

This change should be merged after changing the MDX components on IQP to support the new prop.


Eric-Arellano (approved):
This is great! Please wait to merge this until the internal change has landed & docs are synced.

PR #1146, opened 2024-04-05
Add output for code examples

beckykd:
closes #1143

PR #1342, opened 2024-05-10
Fix code examples in primitives-examples.mdx

Eric-Arellano:
Follow up to https://github.com/Qiskit/documentation/pull/1146. There were a few issues:

* all imports should be at the top of the code block
* we were defining Estimator a few times it was not used, or that it was redeclared later
* not using the n_qubits arg


abbycross (approved):
👍 but should also have a more code-literate person approve 😄

kevinsung (approved):

PR #1346, opened 2024-05-10
Refactor nb-tester execution process

frankharkins:
* Refactors to remove nbconvert, instead using nbclient directly.
* Rather than injecting cells into the notebook and cleaning up after, we patch runtime by running the patching code in the kernel before we execute the notebook. This should also make a future PR simpler too.


scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -197,26 +179,31 @@ async def execute_notebook(path: Path, args: argparse.Namespace, config: Config)
...
-        extra_arguments=["--InlineBackend.figure_format='svg'"]
     )
 
     # This runs the notebook, including possibly submitting jobs. We run it in a
     # new thread to avoid blocking other notebooks from submitting jobs.
-    with patch_runtime(nb, should_patch=not args.submit_jobs):
-        await asyncio.to_thread(processor.preprocess, nb)
-
-    if not args.write:
-        return nb
```
frankharkins, line 219: Slight logic change, but shouldn't affect the behaviour of the script. Before this PR, we exited early to skip cleaning the notebook. I think it's simpler to just always clean it.

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -197,26 +179,31 @@ async def execute_notebook(path: Path, args: argparse.Namespace, config: Config)
...
+    await notebook_client.async_execute()
```
Eric-Arellano, line 206: I think we need to reset the kernel between each notebook, otherwise we share state? This is `reset_kc` https://nbclient.readthedocs.io/en/latest/reference/nbclient.html#nbclient.client.NotebookClient.async_execute

Also the comment above about using a thread is no longer accurate. Imo you don't need a comment at all - it's already obvious enough with async/await that we're getting concurrency thanks to its usage. Before, we were explicitly wrapping synchronous code with an async context.

frankharkins: This function creates a new kernel and runs the notebook in it, so each notebook will get its own kernel. I'll update the docstring to make that clearer.

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -197,27 +179,39 @@ async def execute_notebook(path: Path, args: argparse.Namespace, config: Config)
...
-async def _execute_notebook(filepath: Path, args: argparse.Namespace) -> nbformat.NotebookNode:
+async def _execute_notebook(filepath: Path, args: argparse.Namespace, patch: bool) -> nbformat.NotebookNode:
     """
-    Use nbconvert to execute notebook.
+    Use nbclient to execute notebook. The steps are:
+    1. Read notebook from file
+    2. Create a new kernel
+    3. (Optional) Run some custom code to set up the kernel
+    4. Execute the notebook inside the kernel
+    5. Clean the notebook and return it
```
Eric-Arellano, line 189: This is great to list out the steps! Imo no need to duplicate it here and inline, though. It risks going out of sync. The inline comments are better

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
Eric-Arellano: Ahh, yeah I read that poorly. Agreed. Thanks!

Eric-Arellano (approved):

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
frankharkins: If you missed it then someone else will probably too, so best to make it clearer

Eric-Arellano (approved):

PR #1351, opened 2024-05-13
Refactor nb-tester config and args

frankharkins:
This script now has some non-trivial logic to do with skipping, mocking, writing, etc.

This refactor combines the `Config` and args objects so we have a single point of decision making.

### DIFF
--- a/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
+++ b/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
@@ -102,6 +102,21 @@ class Config:
 
             yield path
 
+    def should_patch(self, path: Path) -> bool:
+        if self.args.submit_jobs:
+            return False
+        if matches(path, self.notebooks_that_submit_jobs):
+            return True
+        return False
+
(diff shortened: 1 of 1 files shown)


scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -58,49 +59,67 @@ def all_job_submitting_notebooks(self) -> list[str]:
...
-            print(
-                f"ℹ️ Skipping {path} as it doesn't submit jobs and the --only-submit-jobs flag is set."
-            )
-            continue
+    def should_patch(self, path: Path) -> bool:
+        if self.args.submit_jobs:
+            return False
+        if matches(path, self.notebooks_that_submit_jobs):
+            return True
+        return False
```
Eric-Arellano, line 110:
```suggestion
        return matches(path, self.notebooks_that_submit_jobs)
```

Eric-Arellano (approved):
Nice improvement! Thanks for making this a dedicated PR

PR #1358, opened 2024-05-13
Revert truncation of Runtime left ToC

Eric-Arellano:
Reverts https://github.com/Qiskit/documentation/issues/1213. We've gotten mixed feedback about truncating the left table of contents. While it makes things less crowded, several people found it confusing. We'd rather aim for clarity with our docs, even if it's more verbose.


arnaucasau (approved):
Thanks!

PR #1359, opened 2024-05-13
Make sure all notebooks are classified in `notebooks.toml`

### LINKED ISSUE #1319: Require all notebooks to be classified
We currently classify notebooks in `notebooks.toml` into `notebooks_that_submit_jobs` and `notebooks_no_mock`. If the notebook is left off, it defaults to our third category of notebooks that do not submit jobs at all. 

It's easy to not realize when you should be changing `notebooks.toml`. After talking to some of the content team, we think it'd be better to explicitly categorize every notebook in `notebooks.toml`.

Tasks:
1. Create a new category like `notebooks_no_jobs`
2. Error if a notebook is not categorized with a helpful message saying what file to modify and explaining the three available categories

Feedback from Kevin: We should try to clarify what we mean by the notebook being "too big".

frankharkins:
Closes #1319.

* The script no longer uses `all-notebooks` as a glob pattern to search, instead it just iterates through the list of filepaths
* Added a check to detect unclassified notebooks

### DIFF
--- a/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
+++ b/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
@@ -124,12 +124,11 @@ class Config:
         ]
         if unclassified == []:
             return
-        print(
+        raise SystemExit(
             f"\nThe following notebooks are not classified in {self.args.config_path}:\n  "
             + "\n  ".join(map(str, unclassified))
             + "\nAdd them to the appropriate group so we know how to test them.\n"
         )
-        sys.exit(1)
 
     def should_patch(self, path: Path) -> bool:
         if self.args.submit_jobs:
--- a/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
+++ b/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
@@ -49,7 +49,7 @@ warnings.filterwarnings("ignore", message="Session is not supported in local tes
 @dataclass
 class Config:
     args: argparse.Namespace
-    all_notebooks: str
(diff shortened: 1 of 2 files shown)


scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -102,6 +117,20 @@ def notebooks_to_execute(self) -> Iterator[Path]:
...
+            if not matches(path, self.all_notebooks)
+        ]
+        if unclassified == []:
+            return
+        print(
+            f"\nThe following notebooks are not classified in {self.args.config_path}:\n  "
+            + "\n  ".join(map(str, unclassified))
+            + "\nAdd them to the appropriate group so we know how to test them.\n"
+        )
+        sys.exit(1)
```
Eric-Arellano, line 132: You could do this instead. https://adamj.eu/tech/2021/10/10/the-many-ways-to-exit-in-python/. It doesn't really matter, so feel free to disagree. Only practical difference here is writing to stderr rather than stdout.

```suggestion
        raise SystemExit(
            f"\nThe following notebooks are not classified in {self.args.config_path}:\n  "
            + "\n  ".join(map(str, unclassified))
            + "\nAdd them to the appropriate group so we know how to test them.\n"
        )
```

Eric-Arellano (approved):
This looks great! I like how you use `@property` to make the code extra readable.

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
frankharkins: Nice, thanks!

PR #1363, opened 2024-05-14
Exit notebook tests early on forks

### LINKED ISSUE #967: Work out how to handle notebook testing on forks
The "Test notebooks" workflow is failing on #956 because we can't share the account token with forks. Not sure what the best way to handle it is, but some options are:

- Skip the workflow on forks. I don't like this though as it makes it easier to merge untested code.
- Maybe allow merging to a feature branch first and then into main once the tests have passed. We should add a message somewhere in `nb-tester` explaining why its failing and what to do.

EDIT: Also need to fix the code that checks for trailing jobs as that crashes if no account is found

frankharkins:
Suggestion on how to handle notebook testing on forks. Resolves #967.

This PR proposes always failing the test on forks, with advice to merge to another branch in this repo to test it before merging to the main branch. While a little clunky, we very rarely get PRs from forks so this is probably simplest for now. We can review the process if it starts becoming a problem.


Eric-Arellano:
Thanks for looking into this! I don't think the feature branch approach will work super well. People without write access cannot create the initial feature branch to merge into. We could set up the feature branch for them and always use the same one like `fork-testing` - if there are issues, then they would not be able to push updates to the `fork-testing` branch so they would need to instead open another PR to modify `fork-testing`. PR review is often very iterative, so that might be clunky.

Instead, if we wanted to support testing from a fork, I think we would need to use something nifty like https://github.com/imjohnbo/ok-to-test. That is a bit complex, though, and I'm skeptical it's worth the complexity because we so rarely get contributions from people without write access.

Still, this PR is really valuable to error on forks because:

1. A lot of times people do have write access and don't realize they shouldn't be using a fork, since that goes against open source convention. It's good to course correct them.
2. For people without write access, it's a signal to the PR reviewers to be careful that it has not been tested.

I'd recommend we take this approach for now - still error on forks, but reword the error message:

- If you have write access, instead push to qiskit/documentation (upstream)
- If you don't have write access, please test it out locally by following the README instructions. Maybe include a note for code reviewers about how they should approach the situation.

Another concern with erroring on forks is we won't execute the code that validates all notebooks are classified. Ideally we could validate that, even if we skip running the notebooks. The best idea I can think of is to add a CLI arg to the notebook test that will skip all execution and only validate the classification.

Wdyt?

frankharkins:
I think we had similar ideas; I've edited the message to be clearer. I have an idea about the classification validation; will make a follow-up PR.

.github/workflows/notebook-test.yml
```diff
@@ -25,6 +25,24 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Exit early on forks
+        if: ${{ github.event.pull_request.head.repo.full_name != github.repository }}
+        shell: python
+        run: |
+          raise SystemExit(
+            "We can't run this test on pull requests form forks."
```
Eric-Arellano, line 33:
```suggestion
            "We can't run this test on pull requests from forks."
```

Eric-Arellano (approved):
Thanks!

PR #1364, opened 2024-05-14
Create dependabot.yml

### LINKED ISSUE #920: Automate updating requirements for notebook testing
Basically: set up dependabot for https://github.com/Qiskit/documentation/blob/main/scripts/nb-tester/requirements.txt. We want to always make sure we are testing the latest stable release of the docs.

Per https://github.com/Qiskit/documentation/issues/133, we should also test other versions of Qiskit when relevant, e.g. Qiskit 1.0 if the latest has been set to 2.0. But that can be a separate configuration.

--

If dependabot is not viable, then we could consider unpinning the versions in requirements.txt, but this is not desirable because it makes our CI undeterministic. It's far better to only change the requirement via a dedicated PR.

frankharkins:
Closes #920

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,4 +3,4 @@ updates:
   - package-ecosystem: "pip"
     directory: "scripts/nb-tester"
     schedule:
-      interval: "weekly"
+      interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "scripts/nb-tester"
+    schedule:
+      interval: "weekly"


Eric-Arellano (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "scripts/nb-tester"
+    schedule:
+      interval: "weekly"
```
Eric-Arellano, line 6: We might want daily to catch new releases of Qiskit sooner? It's important the docs don't have a big gap where they don't work with new releases.

frankharkins:
```suggestion
      interval: "daily"
```

PR #1373, opened 2024-05-14
Check for errors when executing initialization code

frankharkins:
We run some code before notebook execution to silence some warnings and set up mock backends. At the moment, errors are ignored and the kernel continues, often leading to unexpected behaviour.

This PR explicitly checks for errors when running the initialization code and exits early if needed.

### DIFF
--- a/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
+++ b/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
@@ -27,7 +27,7 @@ from typing import Iterator
 import nbclient
 import nbformat
 import tomli
-from jupyter_client.manager import start_new_async_kernel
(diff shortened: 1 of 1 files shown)


scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -225,6 +225,12 @@ async def execute_notebook(path: Path, config: Config) -> bool:
     print(f"✅ No problems in {path} (written)")
     return True
 
+async def _execute_in_kernel(kernel, code: str) -> None:
```
Eric-Arellano, line 228:
```suggestion
async def _execute_in_kernel(kernel: AsyncKernelClient, code: str) -> None:
```

This is imported `from jupyter_client.manager`

Eric-Arellano (approved):
Thanks!

PR #1375, opened 2024-05-14
Add check for files in `qiskit_bot.yaml`

### LINKED ISSUE #1330: Automate check that every file is in qiskit_bot.yaml
Abby C pointed out taht we risk qiskit_bot.yaml becoming stale, such as when we add new pages or rename pages. So, we should have a new check that ensures every non-API file has an owner in qiskit_bot.yaml. 

Of course, some pages don't need owners, so there should be an allowlist mechanism.  You have to explicitly opt-in to this allowlist. The default behavior is to error if the file has no owners.

The error message should explain well what the options are: adding an owner or allow-listing. It should account for the possibility the file has been renamed vs. new file. 

Implementation will be similar to the orphan page check.

arnaucasau:
Closes #1330

This PR adds a check for the qiskit bot config to avoid the `qiskit_bot.yaml` becoming stale.

The check shows two lists of files that can lead to an error:

The first one is for files that exist in the repo but don't have any entry in the qiskit bot entry. This case will be triggered when we add new files or we rename old ones without changing the configuration. If we don't want to track a new file, we can add its path to the `ALLOWED_OWNERLESS_FILES` set to avoid the error message.

The second list of files are the ones that don't exist in the repo but have an entry in the config file. This case will be triggered by files that change their name, removed files with the entry as a leftover, or files where we set the config beforehand. The last ones can be added to the `ALLOWED_INEXISTENT_FILES` set to avoid the error message.

The renamed files trigger both errors and will show the old name in the first list, and the new name in the second one. This could be useful to identify the old entries in the config file and to associate the files in case we have multiple errors.


arnaucasau:
An example of the new error messages can be found [here](https://github.com/Qiskit/documentation/actions/runs/9083668867/job/24962839660?pr=1375). Let me know if you think showing both lists is confusing or if it is helpful to check all the errors.

scripts/commands/checkQiskitBotFiles.ts
```diff
@@ -0,0 +1,131 @@
...
+  "docs/run/circuit-execution",
+  "docs/run/configure-runtime-compilation",
+  "docs/run/execution-modes-faq",
+  "docs/run/execution-modes",
+  "docs/run/native-gates",
+  "docs/run/save-jobs",
+  "docs/run/visualize-results",
+]);
+
+const ALLOWED_INEXISTENT_FILES = new Set([
```
Eric-Arellano, line 42: Nit: should be `NONEXISTENT`

scripts/commands/checkQiskitBotFiles.ts
```diff
@@ -0,0 +1,131 @@
...
+  "docs/migration-guides/qiskit-quantum-instance",
+  "docs/migration-guides/qiskit-algorithms-module",
+  "docs/migration-guides/qiskit-opflow-module",
+]);
+
+type qiskitBotConfig = {
+  always_notify: boolean;
+  notification_prelude: string;
+  notifications: { [id: string]: string[] };
+};
```
Eric-Arellano, line 58: Imo it's fine to not represent every key in the file because we don't use the first two. Instead, we only care about giving a type hint for `notifications`. Specifically, we want `getQiskitBotFiles` to still have the same type hint - but it's fine inside that function to use type casting with `as` whenever it's most convenient.

Removing the unused keys simplifies this script and makes it more future proof.

scripts/commands/checkQiskitBotFiles.ts
```diff
@@ -0,0 +1,131 @@
...
+
+const GLOBS = [
+  "docs/start/*",
```
Eric-Arellano, line 61: Does it fail if you have `docs/guides/*` as well?

scripts/commands/checkQiskitBotFiles.ts
```diff
@@ -0,0 +1,131 @@
...
+  const leftoverFiles = [...qiskitBotFiles].filter(
+    (file) => !filesToCheck.has(file) && !ALLOWED_INEXISTENT_FILES.has(file),
+  );
+
+  if (missingFiles.length > 0) showMissingFilesMessage(missingFiles);
+
+  if (leftoverFiles.length > 0) showLeftoverFilesMessage(leftoverFiles);
+
+  if (missingFiles.length == 0 && leftoverFiles.length == 0)
+    console.log("All files have an owner in the Qiskit bot config ✅\n");
```
Eric-Arellano, line 85: Style nit: if the single line can't fit on the same line as the `if`, it's better to wrap in `{ }`.

```suggestion
  if (missingFiles.length == 0 && leftoverFiles.length == 0) {
    console.log("All files have an owner in the Qiskit bot config ✅\n");
  }
```

scripts/commands/checkQiskitBotFiles.ts
```diff
@@ -0,0 +1,131 @@
...
+
+async function getFilesToCheck(): Promise<Set<string>> {
+  const files = (await globby(GLOBS)).map((file) => removeFileExtension(file));
+  return new Set(files);
+}
+
+function removeFileExtension(file: string): string {
+  const dir = path.parse(file).dir;
+  const name = path.parse(file).name;
+  return `${dir}/${name}`;
```
Eric-Arellano, line 105:
```suggestion
  const parsed = path.parse(file);
  return `${parsed.dir}/${parsed.name}`;
```

scripts/commands/checkQiskitBotFiles.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+function showMissingFilesMessage(missingFiles: string[]): void {
+  console.log(
```
Eric-Arellano, line 109: Better to use `console.error()` for all the error messages

scripts/commands/checkQiskitBotFiles.ts
```diff
@@ -0,0 +1,131 @@
...
+    "The following files don't have an owner in `qiskit_bot.yaml`:\n",
+  );
+  missingFiles.forEach((file) => console.log(`❌ ${file}`));
+
+  console.log(
+    "\nPlease, add an owner to the files if they are new or change their entries in case they have been renamed.",
```
Eric-Arellano, line 115:
```suggestion
    "\nAdd an owner to the files in qiskit_bot.yaml if they are new, or update the file names in qiskit_bot.yaml if the were renamed.",
```

scripts/commands/checkQiskitBotFiles.ts
```diff
@@ -0,0 +1,131 @@
...
+  );
+}
+
+function showLeftoverFilesMessage(leftoverFiles: string[]): void {
+  console.log(
+    "The following files don't exist, but they have an entry in `qiskit_bot.yaml`:\n",
+  );
+  leftoverFiles.forEach((file) => console.log(`❔ ${file}`));
+  console.log(
+    "\nPlease, remove the files' entries from the Qiskit bot yaml file or rename them if their name has changed.",
```
Eric-Arellano, line 126:
```suggestion
    "\nRemove the files from the qiskit_bot.yaml, or rename the file in qiskit_bot.yaml if the file was renamed.",
```

scripts/commands/checkQiskitBotFiles.ts
```diff
@@ -0,0 +1,131 @@
...
+  "docs/migration-guides/*",
+];
+
+async function main() {
+  const qiskitBotFiles = await getQiskitBotFiles();
+  const filesToCheck = await getFilesToCheck();
+
+  const missingFiles = [...filesToCheck].filter(
+    (file) => !qiskitBotFiles.has(file) && !ALLOWED_OWNERLESS_FILES.has(file),
+  );
```
Eric-Arellano, line 75: It'd be more readable if qiskitBotFiles returns an array so you don't have to do `[...]`. I'm not sure we need to be worried about duplicates with using a Set? 

It'd also be great to sort the file paths inside the helper functions so that the error message is easier to read.

Ditto `leftoverFiles`

Eric-Arellano (commented):
Great work!

scripts/commands/checkQiskitBotFiles.ts
arnaucasau: Probably not as we don't have a ton of files in the qiskit-bot config. We could have duplicated if we wrote the same file twice there but maybe it's not a big deal for this script. Having two messages of error in case of duplicated entries could even be helpful to detect them and remove them.

scripts/commands/checkQiskitBotFiles.ts
arnaucasau: No, it doesn't fail. Do you want me to add it to the `GLOBS`?

Eric-Arellano: Yes please, to future-proof this with the docs reorg.

Eric-Arellano (approved):
👏

PR #1382, opened 2024-05-14
Allow manually created class pages

Eric-Arellano:
This unblocks https://github.com/Qiskit/qiskit/pull/12403. Without this PR, we incorrectly treat the class as inline, which means it gets the blue left bar and an h3 incorrectly added. 

Without this PR, our test to check that the ToC and index page are in sync also fails, even though things behave how we expect (https://github.com/Qiskit/qiskit/pull/12403#pullrequestreview-2056366501)

### DIFF
--- a/scripts/lib/api/TocGrouping.test.ts
+++ b/scripts/lib/api/TocGrouping.test.ts
@@ -82,16 +82,27 @@ async function getIndexModuleGroups(fp: string): Promise<ModuleGroup[]> {
   const result: ModuleGroup[] = [];
   let currentGroup: ModuleGroup = [];
   for (const line of rawIndex.split("\n")) {
-    if (line.startsWith("* ")) {
-      if (line.includes("qiskit.")) {
-        const module = extractModuleName(line);
-        currentGroup.push(module);
+    // Each ModuleGroup represents an unordered list of entries starting with `*`.
+    // So, when we stop encountering `*`, we need to start a new ModuleGroup.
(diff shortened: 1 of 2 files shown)


Eric-Arellano:
fyi @jakelishman - this makes the preview better than what you had been seeing. https://github.com/Qiskit/documentation/pull/1380 is updated already to use this.

jakelishman:
oh cool, thanks!  Fwiw, I quite liked the blue bar on the standalone class pages anyway - it gave everything a nice consistency - but I also don't mind if it's not there.

Eric-Arellano:
> I quite liked the blue bar on the standalone class pages anyway - it gave everything a nice consistency - but I also don't mind if it's not there.

Design didn't want it for standalone pages because it adds too much noise to the page and increases indentation unnecessarily. We also don't use the purple bar for standalone function pages like https://docs.quantum.ibm.com/api/qiskit/qiskit.visualization.array_to_latex

jakelishman:
I'd kind of interpreted the purple bar as meaning "method" rather than "function", so it didn't feel inconsistent to me when that was missing.  The blue bar tying together a class didn't feel noisy or too indenting to me - it felt quite nice tying together an entire "object", regardless of where it was.

But anyway, that's not important at all.  If it wasn't _meant_ to have one, it's better that it doesn't accidentally get one because of mistaken parsing.

arnaucasau (approved):
Both the code and the preview look good, thanks!

Issue #1341, opened 2024-05-10
After Qiskit 1.1 release, run TocGrouping.test.ts on stable docs


Eric-Arellano:
This is a follow up to https://github.com/Qiskit/documentation/pull/1300. Once we switch Qiskit 1.1 to be the stable release, we should run that test on `latest`, `dev`, and generic code for the historical APIs starting with 1.1 and newer.

PR #1383, opened 2024-05-14
Generalize TocGrouping.test.ts for future Qiskit versions

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/1341.

### DIFF
--- a/scripts/lib/api/TocGrouping.test.ts
+++ b/scripts/lib/api/TocGrouping.test.ts
@@ -10,9 +10,9 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
-import { readFile } from "fs/promises";
+import { readFile, readdir } from "fs/promises";
 
-import { expect, test } from "@jest/globals";
+import { expect, describe, test } from "@jest/globals";
 
 import { QISKIT_TOC_GROUPING } from "./TocGrouping";
 import type { TocEntry } from "./generateToc";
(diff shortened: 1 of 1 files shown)


scripts/lib/api/TocGrouping.test.ts
```diff
@@ -121,13 +121,38 @@ async function getTocModuleGroups(fp: string): Promise<ModuleGroup[]> {
...
   const tocModuleGroups = await getTocModuleGroups(
-    "docs/api/qiskit/dev/_toc.json",
+    `docs/api/qiskit${dirName}/_toc.json`,
   );
   expect(indexModuleGroups).toEqual(tocModuleGroups);
+}
+
+describe("Qiskit ToC mirrors index page sections", () => {
+  test("validate assumptions", () => {
+    validateTopLevelModuleAssumptions();
```
Eric-Arellano, line 136: I could inline the function here, but I like that we get custom error messages by using the current function rather than `expect()`

scripts/lib/api/TocGrouping.test.ts
```diff
@@ -121,13 +121,38 @@ async function getTocModuleGroups(fp: string): Promise<ModuleGroup[]> {
...
+  });
+
+  test("dev", async () => {
+    await checkFolder("/dev");
+  });
+
+  test.failing("latest", async () => {
```
Eric-Arellano, line 143: Once 1.1 is released, we'll remove the `.failing()`

arnaucasau (approved):
Nice, thanks!

PR #1386, opened 2024-05-15
Fix notebooks.toml

frankharkins:
We didn't notice this until we tried executing notebooks that submit jobs.


Eric-Arellano (approved):
We didn't notice what?

frankharkins:
That the file doesn't exist

PR #1387, opened 2024-05-15
Fix the properties that were incorrectly marked as methods in Qiskit 0.24-0.28

### LINKED ISSUE #1352: Properties incorrectly being marked as functions
After merging https://github.com/Qiskit/documentation/pull/1340, search for `modifiers="property"`. I'm surprised those are being categorized as Functions rather than Attributes. I'd expect `@property` to show up in the Attributes section of classes.

arnaucasau:
Closes #1352

We have 18989 misclassified properties across different versions of Qiskit (0.24 - 0.28). To fix them, this PR added an early return to the `getApiType()` to return the correct `apyType` in case a member has the property modifier.

This is an example of the HTML of a property:

```html
<dl class="py method">
<dt id="qiskit.circuit.Measure.definition">
<em class="property">property </em><code class="sig-name descname">definition</code><a class="headerlink" href="#qiskit.circuit.Measure.definition" title="Permalink to this definition">¶</a></dt>
<dd><p>Return definition in terms of other basic gates.</p>
</dd></dl>
```

The script searches for the `property` modifier inside the `<dt>` tag inside the `<dl>` tag.


scripts/lib/api/generateApiComponents.ts
```diff
@@ -117,7 +117,7 @@ function prepareProps(
   // Remove the attributes and properties modifiers as we don't show their signatures,
   // but we still use them to create their headers
   if (apiType == "attribute" || apiType == "property") {
-    findByText($, $main, "em.property", apiType).remove();
+    findByText($, $child, "em.property", apiType).remove();
```
arnaucasau, line 120: This change is to restrict the search to the signature node. Before we were searching in the whole DOM which should be slower. Tested regenerating some historical versions (not all of them) as well as qiskit 1.0 and the dev version.

Eric-Arellano (approved):
Interesting, so the properties were already in the right section of the file (`## Attributes` vs `## Methods`). 

Neat fix! Thanks for doing this.

arnaucasau:
Actually, they are not in any section. For historical versions, both methods and attributes are all together alphabetically sorted (not sure if that happens always). I think the right ToC is a little confusing there because we assign an `<h3>` as the maximum header, and `Methods` and `Attributes` are `<h2>`s

Here is an example of that: https://docs.quantum.ibm.com/api/qiskit/0.24/qiskit.aqua.QuantumInstance#assemble

Eric-Arellano:
Good point! I opened https://github.com/Qiskit/documentation/issues/1388 and https://github.com/Qiskit/documentation/issues/1389.

PR #1392, opened 2024-05-15
External link checker: Add an HTTP header for the auth for GitHub

arnaucasau:
Part of #823

This PR adds a new HTTP header to the external link checker to set the auth for GitHub when we are checking GitHub links. The new header is only used for GitHub links following what we do in the closed-source repo.

In order to be able to get the environment variable from typescript with `process.env.GITHUB_TOKEN`, we need to define it in the workflows. In the case of running the script locally, the token could be undefined.

### DIFF
--- a/scripts/lib/links/ExternalLink.ts
+++ b/scripts/lib/links/ExternalLink.ts
@@ -65,7 +65,9 @@ function getHeaders(link: string) {
 
   if (link.startsWith("https://github.com")) {
     const ghToken = process.env.GITHUB_TOKEN;
-    headers["Authorization"] = ghToken ? `token ${ghToken}` : "";
+    if (ghToken) {
+      headers["Authorization"] = `token ${ghToken}`;
+    }
   }
 
   return headers;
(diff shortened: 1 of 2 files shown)


scripts/lib/links/ExternalLink.ts
```diff
@@ -57,3 +57,16 @@ export class ExternalLink {
...
 }
+
+function getHeaders(link: string) {
+  const headers: HeadersInit = {
+    "User-Agent": "qiskit-documentation-broken-links-finder",
+  };
+
+  if (link.startsWith("https://github.com")) {
+    const ghToken = process.env.GITHUB_TOKEN;
+    headers["Authorization"] = ghToken ? `token ${ghToken}` : "";
```
Eric-Arellano, line 68: Setting the header to the empty string is different semantically than not setting it at all. I haven't tested if it actually changes end-behavior, but I think this expresses intent better

```suggestion
    const ghToken = process.env.GITHUB_TOKEN;
    if (ghToken) {
      headers["Authorization"] = `token ${ghToken}`;
    }
```

Eric-Arellano (approved):
Thanks for doing this!

PR #1397, opened 2024-05-16
Pull notebook classification out into separate test

frankharkins:
We want to run the notebook classification check on every PR. This PR pulls that test out into a separate pytest suite and runs it as part of the standard CI job. I'd like to also add some more tests to this suite as the `nb-tester` has grown in complexity and there were some issues with it recently.


Eric-Arellano:
This is generally a good idea to have Python tests! In closed source, would this also be a Python test though?

frankharkins:
I think maybe we just duplicate this specific test in inner-source? It'd need a slightly different glob I think

Eric-Arellano (approved):
Thanks

PR #1398, opened 2024-05-16
Allow overriding default number of qubits for mock backend

frankharkins:
For notebooks that don't care, we'd like to keep the number of qubits in the mock backend low as it speeds up execution time. However, some notebooks do require more qubits to test properly (see #1332). This PR allows setting the number of qubits needed in the notebook metadata.


frankharkins:
Closing as it looks like this will not be needed for #1332. Can reopen if needed in the future.

PR #1419, opened 2024-05-22
Update Qiskit Transpiler Service docs

Eric-Arellano:


Eric-Arellano:
Merging to unblock the deploy and because it's low risk. FYI @frankharkins and @arnaucasau

PR #1421, opened 2024-05-23
Rewrite learning uploader

frankharkins:
The requirements for the learning uploader have changed since it was first created.

It was originally intended to be used locally by writers to quickly push content to the platform. I wrote it in Python so writers would not need to install node, and using string manipulations to interact with the REST API was good enough for just uploading the zip and linking the new file. It also included functionality for a writer to log into the database through a CLI.

Now, the uploader is only used in CI, so we don't care about logging in or sticking to Python, and we want to be able to do more complex manipulations such as adding new lessons and changing metadata. For this reason, this PR rewrites the uploader using the official [Directus SDK](https://docs.directus.io/guides/sdk/), which makes interacting with the API a lot easier.


scripts/tutorial-uploader/lib/api.test.ts
```diff
@@ -0,0 +1,171 @@
...
+import { API } from "./api";
+import { type LocalTutorialData } from "./local-tutorial-data";
+
+/* Skip tests if environment is not set */
+const maybeDescribe =
+  process.env.JEST_TEST_TUTORIAL_UPLOADER === "true" ? describe : describe.skip;
+
+/* Create test data */
+const createdSlugs: string[] = []; // To teardown afterwards
+function generateTutorialData(): LocalTutorialData {
```
Eric-Arellano, line 27: What if instead, you have a single variable inside the `describe` like `MOCK_TUTORIAL`? You can declare the object `as const` to ensure that one test doesn't overwrite the object and mess up other tests. Rather than `afterAll`, you'd use `afterEach` to remove `MOCK_TUTORIAL.slug`.

I think this might make things more clear. Right now, it's not very idiomatic to have `createdSlugs` and `generateTutorialData` outside the `desribe()`, and it took me a minute to convince myself that it's safe to use `afterAll` rather than `afterEach` because the tutorial slug will be unique.

--

If you take this recommendation, you don't need to use a random test ID. You can call it something like `test-tutorial`. 

Note that cleanup does not always happen properly, like if someone uses ctrl-c to exit out of the script. So, you might want to even delete this test entry in `beforeAll`.

scripts/tutorial-uploader/lib/api.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    category: "Workflow example",
+    topics: [],
+    reading_time: 50,
+    catalog_featured: false,
+  };
+}
+
+/* Just to be sure */
+if (/learning-api\.quantum\.ibm\.com/.test(process.env.LEARNING_API_URL!)) {
+  throw new Error("Tried to run tests against production!");
```
Eric-Arellano, line 46:
```suggestion
  throw new Error("Tried to run tests against production! Set the env var LEARNING_API_URL to either staging or local (see tutorial-uploader/README.md)");
```

scripts/tutorial-uploader/lib/api.ts
```diff
@@ -0,0 +1,260 @@
...
+import { type LocalTutorialData } from "./local-tutorial-data";
+
+/* To do:
+ *   [x] Fix types
+ *   [ ] Throw correctly on request failures
+ *   [ ] More helpful console logging
+ */
+
+export class API {
+  client: RestClient<LearningApiSchema>;
```
Eric-Arellano, line 42:
```suggestion
  readonly client: RestClient<LearningApiSchema>;
```
```suggestion
  client: RestClient<LearningApiSchema>;
```

scripts/tutorial-uploader/lib/api.ts
```diff
@@ -0,0 +1,260 @@
...
+      `Found ${ids.length} items for getId('${collection}', '${field}', '${value}'. ` +
+        `Expected one or none.`,
+    );
+  }
+
+  /**
+   * Tutorials can have many translations, but we only use English at the moment.
+   */
+  async getEnglishTranslationId(tutorialId: string): Promise<number> {
+    // TODO: This assumes the only translation is english (currently true)
```
Eric-Arellano, line 111: No need to handle multiple yet, but please do throw an exception so we know we need to deal with it

scripts/tutorial-uploader/lib/api.ts
```diff
@@ -0,0 +1,260 @@
...
+  /* Returns the file's ID */
+  async uploadLocalFolder(path: string): Promise<string> {
+    // Zip folder
+    const zippedFilePath = `${tmpdir()}/${randomBytes(8).toString(
+      "hex",
+    )}/tutorial.zip`;
+    await $`mkdir -p ${dirname(zippedFilePath)}`;
+    await $`(cd ${dirname(path)} && zip -qr ${zippedFilePath} ${basename(
+      path,
+    )})`;
```
Eric-Arellano, line 164: Probably more clear to use `cd` from zx: https://google.github.io/zx/api#cd

Should this be changing the cwd for all future processes, or only this specific one?

scripts/tutorial-uploader/lib/local-tutorial-data.ts
```diff
@@ -0,0 +1,82 @@
...
+  const raw = await readFile(path, "utf8");
+  const parsed = yaml.load(raw) as unknown[];
+  return parsed
+    .map((i) => verifyLocalTutorialData(i))
+    .map((i) => relativiseLocalPath(i, path));
+}
+
+const isString = (x: any) => {
+  return typeof x === "string";
+};
```
Eric-Arellano, line 50: Lodash has functions like https://lodash.com/docs/#isString

Eric-Arellano (commented):
This is exciting!

scripts/tutorial-uploader/lib/schema.ts
```diff
@@ -0,0 +1,60 @@
...
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+// Like keyof but can only be string (not number or Symbol)
+export type StringKeyOf<T extends object> = Extract<keyof T, string>;
+
+// To extract type of array
+export type ElementType<T extends any[]> = T extends (infer U)[] ? U : never;
```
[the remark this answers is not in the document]
frankharkins: It's [SO copy/paste](https://stackoverflow.com/a/52331580) (I should have attributed; I'll leave a comment). Given an array `U[]`, this returns the type of each element (i.e. `U`).

scripts/tutorial-uploader/lib/schema.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export interface LearningApiSchema {
+  tutorials: Tutorial[];
+  tutorials_translations: Translation[];
+  tutorials_topics: Topic[];
+  tutorials_categories: Category[];
+  tutorials_tutorials_topics: TutorialsTopicsRelation[];
```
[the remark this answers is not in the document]
frankharkins: Yep, this "collection" is a set of one-to-many mappings of "tutorials" to "tutorials_topics". Name is unfortunate but is out of our control. I'll add a comment.

scripts/tutorial-uploader/lib/schema.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+interface Tutorial {
+  id: string;
+  slug: string;
+  status: "draft" | "published" | "archived";
+  reading_time: number;
+  category: string; // ID
+  catalog_featured: boolean;
+  translations: number[] | Translation[];
```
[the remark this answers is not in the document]
frankharkins: It's an array of numbers. The API can either return a list of IDs or the object itself depending on the request you send. Here's the docs on setting up a schema: https://docs.directus.io/guides/sdk/types.html#adding-relational-fields.

scripts/tutorial-uploader/lib/schema.ts
```diff
@@ -0,0 +1,60 @@
...
+
+interface Category {
+  id: string;
+  name: string;
+}
+
+interface TutorialsTopicsRelation {
+  id: number;
+  tutorials_id: string | Tutorial;
+  tutorials_topics_id: string | Topic;
```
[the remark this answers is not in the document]
frankharkins: Yeah it can return a uuid or the object itself; see https://github.com/Qiskit/documentation/pull/1421/files#r1618651601.

scripts/tutorial-uploader/lib/schema.ts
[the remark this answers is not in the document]
frankharkins: Actually, re-reading that post, it looks like we can use

```ts
type Unpacked<T> = T extends (infer U)[]
```

Which is in the [TS documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html).

scripts/tutorial-uploader/lib/api.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+import { describe, expect, test } from "@jest/globals";
+import { readItem } from "@directus/sdk";
+
+import { API } from "./api";
+import { type LocalTutorialData } from "./local-tutorial-data";
+
+/* Skip tests if environment is not set */
+const maybeDescribe =
+  process.env.JEST_TEST_TUTORIAL_UPLOADER === "true" ? describe : describe.skip;
```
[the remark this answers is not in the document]
frankharkins: That's better, thanks!

scripts/tutorial-uploader/lib/api.ts
arnaucasau: I think that the intention is only to avoid having the full directory structure inside the zip file, but correct me if I'm wrong Frank. 

FWIW, for that, we could also modify the working directory with the `env -C` command in front of the `zip` command, something like `env -C ${path} zip -qr ${zippedFilePath} .` I like the use of `cd` as well, so that's just an alternative ([example](https://superuser.com/a/1842149))

scripts/tutorial-uploader/lib/schema.ts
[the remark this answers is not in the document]
frankharkins: FYI I actually came across a nice explanation of this the other day: https://youtu.be/eJ6R1knfsoc?feature=shared&t=1639

scripts/tutorial-uploader/lib/api.ts
frankharkins: Yep, I don't think zx cd is ideal as I don't want to change the working directory for other commands. @Eric-Arellano @arnaucasau do you have a preference between `env -C` or `(cd ... && ... )`. I'll add an explanation comment either way.

Eric-Arellano: `(cd ... && ... )` sgtm

arnaucasau: Yeah, I agree with Eric :+1:

scripts/tutorial-uploader/lib/api.ts
frankharkins: Fixed in 1696a8be75493290fafb945706a43689f6bfbc20; it now ignores non-English translations.

scripts/tutorial-uploader/lib/api.test.ts
frankharkins: Ok nice, I assumed the jest tests ran in parallel so I added the IDs to avoid conflicts. Fixed in ad843ecec.

scripts/tutorial-uploader/lib/api.test.ts
```diff
@@ -151,17 +134,21 @@ describe("Tutorial uploader API", () => {
...
         languages_code: "en-US",
       },
     ]);
   });
 
-  afterAll(async () => {
-    for (const slug of createdSlugs) {
-      await api.deleteTutorial(slug);
+  beforeAll(async () => {
+    if (await api.getId("tutorials", "slug", MOCK_TUTORIAL.slug)) {
```
Eric-Arellano, line 148: Nice rewrite! This line is now repeated several times and could be a helper function

scripts/tutorial-uploader/README.md
```diff
@@ -0,0 +1,54 @@
...
+     Consider using [direnv](https://direnv.net/) to handle this.
+
+- Create two categories with names `Workflow example` and `How-to`; our script
+  fails if these categories don't exist. To create a category:
+
+  1. Go to <http://0.0.0.0:8055/admin/content/tutorials_categories/+>
+  2. Add the name and click the tick in the top-right corner to save
+
+- Create the following tutorial topics at <http://0.0.0.0:8055/admin/content/tutorials_topics>.
+  > TODO: Can we automate this?
```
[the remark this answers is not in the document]
frankharkins: Added in a1095a15e0563fff4758bace546be59f77b868c4

scripts/tutorial-uploader/README.md
```diff
@@ -0,0 +1,40 @@
...
+  - password: `password`
+
+- Create a token for local testing.
+
+  1. In the local CMS, go to "User directory" (in the leftmost navbar)
+  2. Click "Create item"
+  3. Create a new user with the "Content creator admin" role and generate a new
+     static token. Copy the token to your clipboard. Then click the tick on the
+     top-right of the page to save the user.
+  4. To test the script in this repo, export the following envrionment
```
arnaucasau, line 27: Nit:
```suggestion
  4. To test the script in this repo, export the following environment
```

scripts/tutorial-uploader/README.md
```diff
@@ -0,0 +1,40 @@
...
+This script uploads tutorials to IBM Quantum Learning.
+
+## Developing
+
+To work on this script, you'll need to set up the `saiba-api` project locally.
+
+There are some extra steps you'll need to take to set up `saiba-api` for
+developing this script:
+
+- Follow instruction in the README in the saiba-api repo. Don't forget the
```
arnaucasau, line 12: Nit: Maybe I'm wrong here and this is correct, but just in case :)

```suggestion
- Follow the instructions in the README in the saiba-api repo. Don't forget the
```

scripts/tutorial-uploader/lib/local-tutorial-data.ts
```diff
@@ -0,0 +1,73 @@
...
+  return tutorial;
+}
+
+export async function readTutorialData(
+  path: string,
+): Promise<LocalTutorialData[]> {
+  const raw = await readFile(path, "utf8");
+  const parsed = yaml.load(raw) as unknown[];
+  return parsed
+    .map((i) => verifyLocalTutorialData(i))
```
arnaucasau, line 45: I like how you are verifying if the object matches the interface with `verifyLocalTutorialData` giving a detailed error for the attributes that fail. 

Optional suggestion: I think we could make it more explicit that the function is also type-casting the objects into `LocalTutorialData` to make the code more readable and less prone to errors if in the future we need to change the verification somehow.

One option could be to change the function's name. Alternatively, we could also change the `verifyLocalTutorialData` function to return `void` and move the check before this return using a `forEach`. With this strategy the second map could to be `.map((i) => relativiseLocalPath(i as LocalTutorialData, path));`

tutorials/README.md
```diff
@@ -3,60 +3,47 @@
...
+>
+> One potential gotcha is that the learning platform ignores the top-level
+> heading of the notebook. These headings are only included in the notebook for
+> writers' convenience. If you want to change the title of a notebook, edit the
+> title in `tutorials/learning-api.conf.yaml`.
+
+## Adding new tutorials
+
+To create a new tutorial, add a new notebook to the `tutorials/` folder and
+make new entry to `learning-api.conf.yaml`. The next time the deploy action is
```
arnaucasau, line 25: Nit:
```suggestion
make a new entry to `learning-api.conf.yaml`. The next time the deploy action is
```

arnaucasau (commented):
Awesome work Frank! It looks great! :rocket:

scripts/tutorial-uploader/README.md
frankharkins: No you're completely right, thanks!

scripts/tutorial-uploader/lib/local-tutorial-data.ts
frankharkins: Good idea, thanks! Done in ae35611a332d77b925b2d0d65ff43615a168afd5.

frankharkins:
This is ready for final review, but it might be best to avoid merging until I get back in case there are teething issues.

scripts/tutorial-uploader/lib/api.test.ts
```diff
@@ -0,0 +1,151 @@
...
+  category: "Workflow example",
+  topics: [],
+  reading_time: 50,
+  catalog_featured: false,
+};
+
+/* Just to be sure */
+if (/learning-api\.quantum\.ibm\.com/.test(process.env.LEARNING_API_URL!)) {
+  throw new Error(
+    "Tried to run tests against production! Set the env var LEARNING_API_URL to either staging or local (see tutorial-uploader/README.md)",
```
Eric-Arellano, line 35: Given Abdón's request, should we no longer be mentioning staging? Maybe we should error if it's anything other than localhost? At that point, should we hardcode the value?

scripts/tutorial-uploader/lib/api.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    // Build form
+    const file = new Blob([await readFile(zippedFilePath)], {
+      type: "application/zip",
+    });
+    const formData = new FormData();
+    formData.append("title", zippedFilePath);
```
Eric-Arellano, line 203: What does this do? I'm surprised we'd want the title to have the random hex code and `tutorial.zip`

scripts/tutorial-uploader/lib/api.ts
```diff
@@ -0,0 +1,299 @@
...
+    // Clean up
+    await $`rm -r ${dirname(zippedFilePath)}`;
+
+    return response.id;
+  }
+
+  async updateExistingTutorial(
+    tutorialId: string,
+    localData: LocalTutorialData,
+  ) {
```
Eric-Arellano, line 218:
```suggestion
  ): Promise<void> {
```

scripts/tutorial-uploader/lib/api.ts
```diff
@@ -0,0 +1,299 @@
...
+      ],
+    };
+
+    await this.client.request(updateItem("tutorials", tutorialId, newTutorial));
+    await this.updateTutorialTopics(tutorialId, localData.topics);
+  }
+
+  /*
+   * Only sets minimum data required for API to accept the creation request
+   * updateExistingTutorial should be called immediately after
```
Eric-Arellano, line 245: Idea prompted by this usage note: make most the functions private by putting `#` in front of the name so that it's clear what you expect other files to use or not. That reduces the risk of someone encountering this "gotcha" that they need to also use updateExistingTutorial

scripts/tutorial-uploader/lib/api.ts
```diff
@@ -0,0 +1,299 @@
...
+    const response = await this.client.request(
+      createItem("tutorials", tutorial),
+    );
+    return response.id;
+  }
+
+  /**
+   * Update tutorial if it exists, otherwise create new
+   */
+  async upsertTutorial(localData: LocalTutorialData) {
```
Eric-Arellano, line 278:
```suggestion
  async upsertTutorial(localData: LocalTutorialData): Promise<void> {
```

scripts/tutorial-uploader/lib/api.ts
```diff
@@ -0,0 +1,299 @@
...
+    if (id === null) {
+      id = await this.createTutorial(localData);
+    }
+    await this.updateExistingTutorial(id, localData);
+  }
+
+  /**
+   * For testing
+   */
+  async deleteTutorial(tutorialSlug: string) {
```
Eric-Arellano, line 289:
```suggestion
  async deleteTutorial(tutorialSlug: string): Promise<void> {
```

scripts/tutorial-uploader/lib/local-tutorial-data.ts
```diff
@@ -0,0 +1,73 @@
...
+  tutorial.local_path = join(dirname(path), tutorial.local_path);
+  return tutorial;
+}
+
+export async function readTutorialData(
+  path: string,
+): Promise<LocalTutorialData[]> {
+  const raw = await readFile(path, "utf8");
+  const parsed = yaml.load(raw) as unknown[];
+  parsed.forEach((i) => assertIsLocalTutorialData(i));
```
Eric-Arellano, line 44: Nice! This is definitely an improvement.

scripts/tutorial-uploader/lib/local-tutorial-data.ts
```diff
@@ -0,0 +1,73 @@
...
+  return parsed.map((i) => relativiseLocalPath(i as LocalTutorialData, path));
```
Eric-Arellano, line 45: Is the `as` still required, thanks to the type guard from assertIsLocalTutorialData?

```suggestion
  return parsed.map((i) => relativiseLocalPath(i, path));
```

scripts/tutorial-uploader/setup-for-testing.ts
```diff
@@ -0,0 +1,56 @@
...
+  "Chemistry",
+  "Dynamic circuits",
+  "Error mitigation",
+  "Optimization",
+  "Qiskit Patterns",
+  "Scheduling",
+  "Transpilation",
+];
+
+const CATEGORIES = ["Workflow example", "How-to"];
```
Eric-Arellano, line 32: What's the impact of these becoming stale? If we use a new value in the config file that is not yet defined here, we'd fail in the test, which would be a signal to update this file?

tutorials/README.md
```diff
@@ -3,60 +3,47 @@
...
-workflow. This will push the notebooks on the main branch to the environment
-you select.
+workflow. This will push the notebooks and their metadata from the main branch
+to the environment you select.
 
-You should always start with deploying to "Learning platform (staging)". This
-will deploy to https://learning.www-dev.quantum.ibm.com/catalog/tutorials.
-Check that your tutorial renders properly.
+You should always start with deploying to "Learning platform (staging)". And
+check that your tutorial renders properly.
```
Eric-Arellano, line 37:
```suggestion
You should always start with deploying to "Learning platform (staging)", and
check that your tutorial renders properly. Ask a teammate for the staging link
if you need it.
```

tutorials/learning-api.conf.yaml
```diff
@@ -1,31 +1,135 @@
...
+# Description of values:
+#
+#   title: The title of the page (top-level headings in the notebook are ignored)
+#   short_description: The description users will see in page previews (max 160 characters)
+#   slug: The last part of the URL; do not change this (see note above)
+#   local_path: The path to the tutorial folder, relative to `tutorials/`
+#   status: Can be "draft", "published", or "archived". Do not set to
+#           "archived" until appropriate redirects are set up.
+#   category: Can be "Workflow example" or "How-to"
+#   topics: List of topic names. Choose from
```
Eric-Arellano, line 16: Maybe add a note in this file that when you make changes to these two, update the setup test script.

Eric-Arellano (commented):
This is an epic improvement!

scripts/tutorial-uploader/lib/api.ts
frankharkins: From what I can tell, it's the human-readable title used in the CMS. Since the file gets destroyed after it's linked to the tutorial, it shouldn't matter much. If a human did see the file, I think the hex code helps convey it's for machines.

scripts/tutorial-uploader/lib/api.test.ts
frankharkins: I think mentioning staging is ok, it's just the URLs that are the problem? I'll seek more guidance when we're back.

scripts/tutorial-uploader/lib/local-tutorial-data.ts
frankharkins: I tried that but unfortunately the type checker isn't that clever. It did work when I made a function that asserted `parsed` was of type `LocalTutorialData[]`, but I thought this was just as clear and a little simpler.

scripts/tutorial-uploader/setup-for-testing.ts
frankharkins: Yes, adding a category that doesn't exist would break the uploader. I'll make a check that verifies the list against production.

frankharkins:
This is ready for final review, but we can't merge until the token has the right permissions.

scripts/tutorial-uploader/lib/api.test.ts
```diff
@@ -0,0 +1,151 @@
...
+  test("upload new tutorial", async () => {
+    expect(await api.getTutorialIdBySlug(MOCK_TUTORIAL.slug)).toBeNull();
+
+    await api.upsertTutorial(MOCK_TUTORIAL);
+    const tutorialId = await api.getTutorialIdBySlug(MOCK_TUTORIAL.slug);
+    expect(tutorialId).toBeTruthy();
+
+    const retrievedTutorial = await api.client.request(
+      // @ts-ignore
+      readItem("tutorials", tutorialId, { fields: ["*", "translations.*"] }),
```
arnaucasau, line 51: With this `readItem`, are you trying to read nested fields? If that's the case, FYI (probably you already know), I found this alternative [syntaxis](https://docs.directus.io/guides/sdk/getting-started.html#read-nested-fields). Maybe this could help with the `@ts-ignore` :thinking:. If that doesn't help or the `readItem` had another purpose, I still like how you have made it work. I didn't know about `@ts-ignore` before :rocket:

arnaucasau (approved):
Awesome work Frank :rocket:. I've been looking at the improvements you did from the comments Eric left last week and the code itself and all looks great! I left some feedback to try to understand better some parts and a possible nit

scripts/tutorial-uploader/lib/api.test.ts
frankharkins: Awesome! Yes always best to fix the type rather than ignore when possible. That fixed all but one.

frankharkins:
This comment is a reminder to update `learning-api.conf.yaml` before merging as tutorial data has changed since this PR was created.

frankharkins:
This is ready to go! I'll use it deploy the QAOA tutorial then I'll move it to inner-source to hand off to the new team. Thanks for your patience.

scripts/tutorial-uploader/lib/api.test.ts
```diff
@@ -0,0 +1,154 @@
...
+      {
+        title: modifiedTutorial.title,
+        short_description: modifiedTutorial.short_description,
+        content: "Here's some basic content.\n",
+        languages_code: "en-US",
+      },
+    ]);
+  });
+
+  beforeAll(async () => {
```
Eric-Arellano, line 145: Nit that the convention is for this and `afterEach` to be near the top of the describe block

tutorials/learning-api.conf.yaml
```diff
@@ -1,31 +1,38 @@
...
+#             * Chemistry
+#             * Dynamic circuits
+#             * Error mitigation
+#             * Optimization
+#             * Qiskit Patterns
+#             * Scheduling
+#             * Transpilation
+#           If these change, make sure to update scripts/tutorial-uploader/setup-for-testing.ts
```
Eric-Arellano, line 24: This comment will need to change when migrating

tutorials/learning-api.conf.yaml
```diff
@@ -1,31 +1,38 @@
...
-  - path: explore-composer
-    urlStaging: tutorials/28f53e87-52ba-4fa7-a1f9-9ad7f548a7e0
-    urlProduction: tutorials/967371a7-159d-459a-a7a6-210c55b14766
-  - path: combine-error-mitigation-options-with-the-estimator-primitive
-    urlStaging: tutorials/58d982db-9510-4815-a13c-c81abae1507f
-    urlProduction: tutorials/5f0507b3-3eae-4ccd-9a7a-74edd5a157a3
+# This file contains all the metadata for tutorials controlled by this repo.
+#
+# WARNING: Do not change any "slugs" until appropriate redirects are set up.
+# Speak to Abby Mitchell, Eric Arellano, or Frank Harkins to set up redirects.
```
Eric-Arellano, line 4: This is no longer accurate. I think this should be the learning team

Eric-Arellano (approved):
Great job!

PR #1426, opened 2024-05-23
use hellinger_fidelity function from qiskit

kevinsung:


Eric-Arellano (commented):
Is it okay the outputs didn't change at all?

PR #1438, opened 2024-05-28
Add missing metadata for Qiskit Transpiler Service

Eric-Arellano:


scripts/lib/api/specialCaseResults.ts
```diff
@@ -46,6 +49,16 @@ export function specialCaseResults(results: HtmlToMdResultWithUrl[]): void {
         hardcodedFrontmatter: RUNTIME_INDEX_META,
       };
     }
+
+    if (
+      result.url.endsWith("/index") &&
+      result.url.includes("qiskit-transpiler-service/")
```
Eric-Arellano, line 55: this check is so that we will still match on `qiskit-transpiler-service/dev/index`

arnaucasau (approved):
Looks good! Thanks Eric!

PR #1456, opened 2024-05-29
Update final output for Hello world

### LINKED ISSUE #1434: Fix Hello World plot

On this page, the results in the last figure do not agree with the text explaining it: https://docs.quantum.ibm.com/start/hello-world

### Select all that apply

- [ ] typo
- [X] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

### Describe the fix.

The plot should look like the one in the Hello World video.  We should verify that the code is correct and fix the output.

beckykd:
closes #1434


beckykd:
@frankharkins @Eric-Arellano 

I was super confused when re-running the notebook because I came across this cell in the notebook that I downloaded: 
"# Check there are six observables.
# If not, edit the comments in the previous cell and update this test."

After getting everything updated, I was going to ask about that cell and realized that it's not in the published version.  So that cell is clearly one that just exists to verify that everything gets updated when we make changes.  

The problem is that it's in the downloaded version of the notebook and was super confusing to me as a user.  Is there any way we can fix that?  Clarify that that cell is for notebook verification and isn't actually part of the tutorial?

abbycross:
> The problem is that it's in the downloaded version of the notebook and was super confusing to me as a user. Is there any way we can fix that? Clarify that that cell is for notebook verification and isn't actually part of the tutorial?

FWIW: I noticed in the Challenge this year that the intro lab does a nice job explaining the grader cells that appear in all notebooks -- they look like they aren't doing anything but are important and can't be altered. We should find the right language that goes along with each of the verification cells

Eric-Arellano:
The current plan is to remove the cells from the download, tracked by https://github.ibm.com/IBM-Q-Software/iqp-channel-docs/issues/1351.

beckykd:
> The current plan is to remove the cells from the download, tracked by https://github.ibm.com/IBM-Q-Software/iqp-channel-docs/issues/1351.

OK - when that happens it'll be important that we don't use those to update anything.  It's probably obvious and maybe I'm the only one doing it, but it seemed so handy :)

Eric-Arellano:
> when that happens it'll be important that we don't use those to update anything. It's probably obvious and maybe I'm the only one doing it, but it seemed so handy :)

Yeah, these testing blocks should solely be to verify, like "read-only".

Eric-Arellano (approved):

PR #1457, opened 2024-05-29
Clarify title for Qiskit SDK simulator page

Eric-Arellano:
We should clarify this is the Qiskit SDK. 

We don't change the URL because it will be redirected anyways soon to its new home in the `guides/` section, so it's not worth renaming yet.


abbycross (approved):
👍

PR #1472, opened 2024-05-31
Update conversion pipeline integration test

Eric-Arellano:
qiskit-sphinx-theme was updated to better show inline classes.


arnaucasau (approved):
Thanks!

PR #1479, opened 2024-06-02
Spellcheck done API docs

ljcamargo:
Hi @Eric-Arellano 

Related to https://github.com/Qiskit/qiskit/issues/12067

I had to change the package.json check:spelling script because the ignorePaths didn't manage well the partial exclusion of the past versions, perhaps changing dir structure for the long term solution will be ideal.

I also updated the custom dictionaries of people.txt and qiskit.txt in order to add some lastnames and terms, I sorted them alphabetically as well.

I avoided changing references to papers where "randomised" appeared in the title and I added this word to the dictionary in order to avoid the lint failure in the future. I took care not to change any other references, but this was the only case for now.


Eric-Arellano:
Thanks, @ljcamargo! We appreciate the patch. All the Qiskit changes need to be done in the Qiskit/qiskit repository, as explained at https://github.com/Qiskit/qiskit/issues/12067. That is because the docs here are generated from Qiskit/qiskit and it's not safe for us to edit the MDX files directly here because they will be overwritten. 

Those changes will usually happen in Python files in their docstring. A tool like ripgrep can be useful to quickly find where the bad string is located in Qiskit/qiskit.

1ucian0:
I'm removing `[Unitary Hack 2024]` from the title and "This closes, https://github.com/Qiskit/qiskit/issues/12067" is replaced by "related to https://github.com/Qiskit/qiskit/issues/12067"

frankharkins:
Closing as per https://github.com/Qiskit/documentation/pull/1479#issuecomment-2145616634

PR #1484, opened 2024-06-03
Change index entry from patterns `_toc.json`

arnaucasau:
This PR changes the patterns script to use '`"url": "/guides"`' instead of `"url": "/guides/index"'` for the index entry of the guides `_toc.json`. The PR doesn't modify the `entries.py` script to avoid having an empty slug.

### DIFF
--- a/scripts/patterns-reorg/main.py
+++ b/scripts/patterns-reorg/main.py
@@ -37,7 +37,6 @@ def write_guides_dir() -> None:
         "children": [entry.to_json("guides") for entry in TOP_LEVEL_ENTRIES],
     }
     text = json.dumps(result, indent=2)
-    text = text.replace('"url": "/guides/index"', '"url": "/guides"')
     (folder_path / "_toc.json").write_text(text)
 
 
--- a/scripts/patterns-reorg/main.py
+++ b/scripts/patterns-reorg/main.py
@@ -37,6 +37,7 @@ def write_guides_dir() -> None:
         "children": [entry.to_json("guides") for entry in TOP_LEVEL_ENTRIES],
     }
     text = json.dumps(result, indent=2)
+    text = text.replace('"url": "/guides/index"', '"url": "/guides"')
     (folder_path / "_toc.json").write_text(text)
 
 
--- a/scripts/patterns-reorg/models.py
+++ b/scripts/patterns-reorg/models.py
@@ -63,7 +63,8 @@ class Entry:
             assert self.page_content is not None
             content = self.page_content
             extension = ".mdx"
-        Path(base_dir / f"{self.get_relative_path_from_slug()}{extension}").write_text(content)
+        dest = base_dir / f"{self.relative_path_from_slug()}{extension}"
+        dest.write_text(content)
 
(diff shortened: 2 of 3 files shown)


scripts/patterns-reorg/main.py
```diff
@@ -37,6 +37,7 @@ def write_guides_dir() -> None:
         "children": [entry.to_json("guides") for entry in TOP_LEVEL_ENTRIES],
     }
     text = json.dumps(result, indent=2)
+    text = text.replace('"url": "/guides/index"', '"url": "/guides"')
```
Eric-Arellano, line 40: I think a better fix would be in `models.py`, probably with how we store the slug in the first place. For example, allow `slug=""` for the index. (That would mean you change other slugs to be `slug="/my-page"`, rather than `slug="my-page"`

We want both this Toc.json thing _and_ redirects to be using `/guides` rather than `/guides/index`. So it's better to fix this at the source.

scripts/patterns-reorg/models.py
```diff
@@ -63,7 +63,12 @@ def ensure_slugs_exist(self, base_dir: Path) -> None:
             assert self.page_content is not None
             content = self.page_content
             extension = ".mdx"
-        (base_dir / f"{self.slug}{extension}").write_text(content)
+        Path(base_dir / f"{self.get_relative_path_from_slug()}{extension}").write_text(content)
+
+    def get_relative_path_from_slug(self):
```
Eric-Arellano, line 68:
```suggestion
    def relative_path_from_slug(self) -> str | None:
```

scripts/patterns-reorg/models.py
```diff
@@ -63,7 +63,12 @@ def ensure_slugs_exist(self, base_dir: Path) -> None:
             assert self.page_content is not None
             content = self.page_content
             extension = ".mdx"
-        (base_dir / f"{self.slug}{extension}").write_text(content)
+        Path(base_dir / f"{self.get_relative_path_from_slug()}{extension}").write_text(content)
```
Eric-Arellano, line 66:
```suggestion
        dest = base_dir / f"{self.relative_path_from_slug()}{extension}"
        dest.write_text(content)
```

scripts/patterns-reorg/models.py
```diff
@@ -63,7 +63,12 @@ def ensure_slugs_exist(self, base_dir: Path) -> None:
...
+
+    def get_relative_path_from_slug(self):
+        if self.slug is None:
+            return
+        return (self.slug if self.slug != "" else "/index").removeprefix("/")
```
Eric-Arellano, line 71:
```suggestion
        if self.slug == "":
            return "index"
        return self.slug.removeprefix("/")
```

Eric-Arellano (commented):
Thanks! Can you please run the script to confirm that the redirects file doesn't change at all? You'll need to `rm -rf docs/guides` after

Eric-Arellano (approved):

arnaucasau:
Checked, the script only created the `docs/guides` folder and it doesn't change the redirects file :+1:

Issue #1466, opened 2024-05-30
Docs reorg: audit new URLs


Eric-Arellano:
URLs are hard to change, so we want the new URLs to be well named. We want to be confident that they will last us in the future, even if the slug is longer. 

To do this, you'd look at https://github.com/Qiskit/documentation/blob/main/scripts/patterns-reorg/redirects.json. Each line has the original URL on the left, and the new URL on the right, but without the `guides/` prefix showing. For example, line 2 `"start/install": "install-qiskit",` means `start/install` -> `guides/install-qiskit`

--

If you want to look at the page title for context, it may be helpful to instead look at `entries.py`, like this:

https://github.com/Qiskit/documentation/blob/640bd7ae0764aca8b3c05fc1f0020331e3f50d9c/scripts/patterns-reorg/entries.py#L35-L39

The first value is the page title. `from_file` is the old URL (minus the extension). `slug` is the new URL without the `guides/` prefix.

--

You can leave suggestions as a comment on this issue.

abbycross:
I do have a vague, maybe-unfounded concern that page titles do change more often than URLs, so if we try to be *too* precise in our filenames, those could go stale. Thoughts @beckykd ?

Eric-Arellano:
> I do have a vague, maybe-unfounded concern that page titles do change more often than URLs, so if we try to be too precise in our filenames, those could go stale. Thoughts @beckykd ?

I totally agree. For example, `qiskit-serverless` is titled "Qiskit Serverless Workloads", and right now it's the only page about Qiskit Serverless. But we could imagine adding more pages about Qiskit Serverless in the future. The `qiskit-serverless` page would still probably act like the introduction page to the tool, so maybe we'd do something like this:

```
Qiskit Serverless
|-  Introduction to Qiskit Serverless (qiskit-serverless)
|_ Advanced workflows (qiskit-serverless-advanced-workflows)
```

In other words, it is good to keep the page name generic as `qiskit-serverless` because it is flexible to us renaming `Qiskit Serverless Workflows` to `Introduction to Qiskit Serverless`. Whereas the URL `qiskit-serverless-workflows` would have been overly precise.

--

On the other end of the spectrum, a page like `configuration` is way too generic.

Eric-Arellano:
> Are actual file names changing then? For example "start/install": "install-qiskit" ?

Yes, every single file is going to need a redirect from the old folders like `start/` to the new `guides/` folder. That means that now is our opportunity to rename the files if we want.

[...]

PR #1487, opened 2024-06-03
Update URLs for docs reorganization

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/1466.


Eric-Arellano:
@javabster, Abby C and Becky have both now reviewed the new URLs as part of https://github.com/Qiskit/documentation/issues/1466. I don't know if we want to close https://github.com/Qiskit/documentation/issues/1466, or if you'll want to do one final review first? 

It's one of our most important decisions with the reorg so that we can avoid needing redirects in the future.

abbycross (approved):
looks good!

Issue #1297, opened 2024-05-03
Patterns: migrate docs to new "guides" folder


Eric-Arellano:
We're flattening the docs to be under a new "guides" section, like proposal 3 of https://github.com/Qiskit/documentation/pull/1294. This issue tracks renaming the files and setting up the new ToC.

This change needs to be well coordinated with internal changes like changing the top nav bar and adding redirects.

Eric-Arellano:
This script is now added, and should be ready to go after https://github.com/Qiskit/documentation/pull/1493 and upcoming PRs to close https://github.com/Qiskit/documentation/issues/1460 and https://github.com/Qiskit/documentation/issues/1462

The final migration PR will look like this:

1. scripts/patterns-reorg/update_internal_links.py
2. git commit
3. scripts/patterns-reorg/main.py --delete-existing
4. git commit

That PR should have no manual changes, to reduce risk of code review.

## What about if we need to edit existing content?

The migration script cannot handle making edits to existing content. Instead, we copy over exactly whatever the original content was. So far, we've been able to safely land all the changes we need _before_ the docs reorg, and hopefully we can keep doing that.

If we cannot make certain changes to _existing_ content beforehand, then we have two approaches we can take:

1. Make the manual changes in the final PR 
2. Add a new mechanism to the migration script. We need to still preserve Git history, so this would probably look like still copying over the original file, then somehow updating the content after, like applying a Git diff. The content update should probably be done in a dedicated program so that it can be a dedicated Git commit.

PR #1493, opened 2024-06-04
Allow deleting all existing docs guides

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/1297. Now when you run `scripts/patterns-reorg/main.py --delete-existing`, Git will recognize which files were renamed and preserve their Git history.


arnaucasau (approved):
Looks good to me, tested locally and works well!

PR #1495, opened 2024-06-04
Use internal links for new index page

Eric-Arellano:
We shouldn't be using external links in general.

This PR also updates to use the new `guides/` URLs. Note that we have some risk that the links we hardcode here diverge from our source of truth in `entries.py`. We can't have `page_content.py` load the final URLs from `entries.py` because it would result in a circular import. However, our link checker will complain if there are any divergences, so it should be fine.


arnaucasau (approved):
Thank you Eric!

PR #1497, opened 2024-06-05
Set redirects for future deleted pages in the patterns reorg

arnaucasau:
This PR adds a new class called `DeletedPage` that we can use to set redirects for pages that will be removed after the reorg.

The PR also adds some redirects for the index pages of the old folders duplicating the entries to match the two ways they can be used. An index page can be accessed by using the name of the parent folder or by specifying the index file. For example, the `run` index can be accessed by going to `/run/` or `/run/index`.


scripts/patterns-reorg/models.py
```diff
@@ -96,6 +96,12 @@ def entries_as_markdown_list(
     return "\n".join(result)
 
 
+@dataclass(frozen=True)
+class DeletedPage:
+    slug: str
+    from_file: str
```
Eric-Arellano, line 102: I recommend calling it `old_slug` and `redirect_to`

scripts/patterns-reorg/models.py
```diff
@@ -111,12 +117,24 @@ def filter_entries(
 
 
 def determine_redirects(
-    entries: tuple[Entry, ...], *, prefix: str = ""
+    entries: tuple[Entry, ...] | tuple[DeletedPage, ...], *, prefix: str = ""
 ) -> dict[str, str]:
     result = {}
     for entry in entries:
-        if entry.slug and entry.from_file:
+        if entry.slug is not None and entry.from_file:
```
Eric-Arellano, line 124: This is getting pretty big. You can invert the conditional and early return with `continue`

scripts/patterns-reorg/models.py
```diff
@@ -111,12 +117,24 @@ def filter_entries(
...
+            # have two links possible. For example, `/run/index` and `/run/`
+            # point to the same page.
+            old_url_split = old_url.split('/')
+            old_file_name = old_url_split[-1]
+            old_folder = old_url_split[-2]
+            if old_file_name == "index":
+                result[f"{old_folder}/"] = f"{prefix}{entry.slug.removeprefix('/')}"
+
+        if type(entry) is Entry:
+            result.update(determine_redirects(entry.children))
```
Eric-Arellano, line 138: The more idiomatic approach is `isinstance`, which handles inheritance correctly

```suggestion
        if isinstance(entry, Entry):
            result.update(determine_redirects(entry.children))
```

scripts/patterns-reorg/models.py
```diff
@@ -111,12 +117,24 @@ def filter_entries(
...
             old_url = str(PurePath(entry.from_file).with_suffix(""))
             result[old_url] = f"{prefix}{entry.slug.removeprefix('/')}"
-        result.update(determine_redirects(entry.children))
+
+            # We need to add two links for each index entry because we can
+            # have two links possible. For example, `/run/index` and `/run/`
+            # point to the same page.
+            old_url_split = old_url.split('/')
+            old_file_name = old_url_split[-1]
+            old_folder = old_url_split[-2]
```
Eric-Arellano, line 133:
```suggestion
            old_folder, old_file_name = old_url.split('/')
```

scripts/patterns-reorg/models.py
```diff
@@ -111,12 +117,24 @@ def filter_entries(
...
     result = {}
     for entry in entries:
-        if entry.slug and entry.from_file:
+        if entry.slug is not None and entry.from_file:
             old_url = str(PurePath(entry.from_file).with_suffix(""))
             result[old_url] = f"{prefix}{entry.slug.removeprefix('/')}"
-        result.update(determine_redirects(entry.children))
+
+            # We need to add two links for each index entry because we can
+            # have two links possible. For example, `/run/index` and `/run/`
```
Eric-Arellano, line 129:
```suggestion
            # have two links possible. For example, `/run/index` and `/run`
```

scripts/patterns-reorg/redirects.json
```diff
@@ -65,5 +67,11 @@
   "build/circuit-visualization": "visualize-circuits",
   "verify/plot-quantum-states": "plot-quantum-states",
   "run/visualize-results": "visualize-results",
-  "run/quantum-serverless": "qiskit-serverless"
+  "run/quantum-serverless": "qiskit-serverless",
+  "/build/index": "map-problem-to-circuits",
```
Eric-Arellano, line 71: Should be `build/index` without the leading `/`

scripts/patterns-reorg/models.py
```diff
@@ -111,12 +117,22 @@ def filter_entries(
...
             old_url = str(PurePath(entry.from_file).with_suffix(""))
             result[old_url] = f"{prefix}{entry.slug.removeprefix('/')}"
-        result.update(determine_redirects(entry.children))
+
+            # We need to add two links for each index entry because we can
+            # have two links possible. For example, `/run/index` and `/run`
+            # point to the same page.
+            old_folder, old_file_name = old_url.split('/')
+            if old_file_name == "index":
+                result[f"{old_folder}/"] = f"{prefix}{entry.slug.removeprefix('/')}"
```
Eric-Arellano, line 133: You can have a new variable called `redirect_to` to DRY f"{prefix}{entry.slug.removeprefix('/')}"

scripts/patterns-reorg/models.py
arnaucasau: I think the names you suggest are better, but in this case, it is convenient to have the same names as the class `Entry` to avoid having to check more things in the `determine_redirects` function. With the same names, we can treat them as `Entry`s indistinctly, and all is more general. If you have a strong opinion about it, I can change it though

Eric-Arellano: Yeah, I would still change it. `from_file` is a misnomer because you would need the file extension set. It would be better for the code paths to be distinct in the for loop

scripts/patterns-reorg/redirects.json
```diff
@@ -18,6 +18,7 @@
   "build/interoperate-qiskit-qasm3": "interoperate-qiskit-qasm3",
   "build/qasm-feature-table": "qasm-feature-table",
   "transpile/index": "transpile",
+  "transpile/": "transpile",
```
Eric-Arellano, line 21: This should probably be `transpile` without the `/`.

scripts/patterns-reorg/models.py
```diff
@@ -124,20 +133,15 @@ def determine_redirects(
...
-
-
+            if entry.slug is None or not entry.from_file:
+                continue
+            
+            old_url = str(PurePath(entry.from_file).with_suffix(""))
+            redirect_to = f"{prefix}{entry.slug.removeprefix('/')}"
+            add_redirect_to_dict(result, old_url, redirect_to)
+
+        elif isinstance(entry, DeletedPage):
```
arnaucasau, line 143: I used an `elif` and not a `continue` in the first case because I think it's more readable.

scripts/patterns-reorg/deleted_entries.py
```diff
@@ -0,0 +1,7 @@
+from models import DeletedPage
+
+DELETED_PAGES = (
+    DeletedPage(redirect_to="/map-problem-to-circuits", old_slug="build/index"),
+    DeletedPage(redirect_to="/execute-on-hardware", old_slug="run/index"),
+    DeletedPage(redirect_to="", old_slug="start/index"),
```
Eric-Arellano, line 6: Nit

```suggestion
    DeletedPage(old_slug="build/index", redirect_to="/map-problem-to-circuits"),
    DeletedPage(old_slug="run/index", redirect_to="/execute-on-hardware"),
    DeletedPage(old_slug="start/index", redirect_to=""),
```

scripts/patterns-reorg/models.py
```diff
@@ -109,14 +115,33 @@ def filter_entries(
         result.append(new_entry)
     return tuple(result)
 
+def add_redirect_to_dict(dict: dict[str, str], old_url: str, redirect_to: str)-> None:
```
Eric-Arellano, line 118: `dict` is a dangerous name because you're shadowing the builtin value `dict`

Eric-Arellano: Also it'd be good to put a `_` in front of the function name since this is a private helper

Eric-Arellano (commented):
Thanks!

Eric-Arellano (approved):

PR #1498, opened 2024-06-05
Add script to update all non-API links to `guides/`

### LINKED ISSUE #1460: Patterns: update all non-API links to guides/
Part of https://github.com/Qiskit/documentation/pull/1294.

We need to update internal links to the new `guides/` URLs. We should use that PR as the source of truth for what the redirects are. 

The updated links should be done in a single automated Git commit, made by the migration program. Although, we aren't ready at the moment with the rest of the changes for this Git commit idea. Instead, I recommend having a dedicated program with its own `main()`, like `update_links.py`. It won't do anything with Git - only update the URLs. That program can of course import the rest of the files in `scripts/patterns-reorg`, like computing the redirects.

arnaucasau:
Closes #1460

This PR adds a script to update all the non-API links to point to `guides/` taking into account that all the files will now be in the same folder, and there's no need to preserve the relative paths


scripts/patterns-reorg/update_internal_links.py
```diff
@@ -0,0 +1,71 @@
...
+    else:
+        search_key = f"{folder}/{link_split[-1]}"
+
+    if search_key in redirects:
+        new_link = redirects[search_key]
+        if new_link == "":
+            new_link = "/guides/"
+        return markdown.replace(markdown, markdown.replace(link, f"{new_link}{anchor}"))
+    
+    return markdown
```
Eric-Arellano, line 54:
```suggestion
    if search_key not in redirects:
        return

    new_link = redirects[search_key]
    if new_link == "":
        new_link = "/guides/"
    return markdown.replace(markdown, markdown.replace(link, f"{new_link}{anchor}"))
```

Eric-Arellano: Also what's up with the double markdown.replace()?

scripts/patterns-reorg/update_internal_links.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def main() -> None:
+    inline_link_re = re.compile(r'\[([^\]]+)\]\(([^)]+)\)')
+    folders = ["build", "run", "start", "transpile", "verify"]
```
Eric-Arellano, line 59: This can be DRYed with main.py - you can make it a module-level constant

scripts/patterns-reorg/update_internal_links.py
```diff
@@ -0,0 +1,71 @@
...
+    if anchor_index == -1:
+        link_without_anchor = link
+        anchor = ""
+    else:
+        link_without_anchor = link[:anchor_index]
+        anchor = link[anchor_index:]
+
+    link_split = link_without_anchor.split('/')
+    if link.startswith("/") or link.startswith("../"):
+        search_key = f"{link_split[-2]}/{link_split[-1]}"
```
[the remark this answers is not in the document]
arnaucasau: Not exactly because we have links like `../start/install` where I'm skipping the two dots at the beginning and searching one by `start/install`.

scripts/patterns-reorg/update_internal_links.py
arnaucasau: Oh, that's a leftover, thanks! Before, I was replacing the markdown of the whole file, and without the double step, links like `build/pulse` used as a relative link in the same folder as `pulse` were indistinguishable from the word "pulse" found in the docs. With the new approach, using `re.sub`, we only replace the content in the link, so it should be okay to remove it as long as we don't have it in the link text (I'll look closely at that case).

scripts/patterns-reorg/main.py
```diff
@@ -52,15 +54,19 @@ def write_guides_dir() -> None:
     (folder_path / "_toc.json").write_text(text)
 
 
+def generate_redirects() -> dict[str, str]:
```
Eric-Arellano, line 57: This naming is pretty similar to `determine_redirects`. Instead, you can do a module level constant `REDIRECTS = (*TOP_LEVEL_ENTRIES, *DELETED_PAGES)`

Eric-Arellano (commented):
Nice! Could you please have this generate the links to be like `[quantum circuits](./construct-circuits)` rather than `[quantum circuits](construct-circuits)`, i.e. with the dot? While it's not necessary, imo it's more obvious to the reader what's going on. Better for grep also

scripts/patterns-reorg/main.py
```diff
@@ -23,6 +23,8 @@
 from entries import TOP_LEVEL_ENTRIES
 from deleted_entries import DELETED_PAGES
 
+OLD_FOLDERS = ["start", "run", "verify", "transpile", "build"]
+REDIRECTS = (*TOP_LEVEL_ENTRIES, *DELETED_PAGES)
```
Eric-Arellano, line 27: I meant this:

```suggestion
REDIRECTS = determine_redirects((*TOP_LEVEL_ENTRIES, *DELETED_PAGES))
```

You'll need to update call sites

Eric-Arellano (approved):
Thanks! Looks good after this last change

Eric-Arellano (approved):

PR #1499, opened 2024-06-05
Revamp options page

### LINKED ISSUE #1247: Document primitive option precedence rules
### URL, if applicable

_No response_

When a user sets options like this,

```python
from qiskit_ibm_runtime import EstimatorV2, QiskitRuntimeService
from qiskit import QuantumCircuit

backend = service.backend("ibm_auckland")

estimator = EstimatorV2(backend)

estimator.options.default_shots = 100
estimator.options.resilience_level = 0
estimator.options.resilience.zne_mitigation = True
```

it might not be clear how the resilience level might interact with the other options, because the resilience level itself can modify the mitigation strategy. In this example, does setting the resilience level to 0 have the effect of turning off the attempt to add zne_mitigation? (spoiler: no)

What needs to be documented is that that the resilience level is FIRST used to set a base configuration on any options it thinks are relevant, and THEN any other user-specified options are updated on top of that. So it would be possible, for example, though pointless, to completely undo the effect of setting the resilience level by reversing all of its intentions manually.

### LINKED ISSUE #1416: Qiskit Runtime primitive options difficult to find

_No response_

### Select all that apply

- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [X] other

### Describe the fix.

At the Partner Forum I heard again that Qiskit Runtime primitive options are difficult to find in the docs. We attempted to fix this by adding a table in #1124, but perhaps that wasn't enough. This issue is meant to brainstorm on how we can improve on the experience.

### LINKED ISSUE #1453: Add information on how to turn off error mitigation / suppression
### URL, if applicable

https://docs.quantum.ibm.com/run/configure-error-mitigation

Many researchers want total control over their experiments and don't want the primitives to transform their circuits. For `EstimatorV2`, one can just set `resilience_level=0` and `optimization_level=0`. It is however not obvious how to do this for `SamplerV2`, since there is no `resilience_level` in SamplerV2. 

We should have a few words in the docs on how to accomplish this. For SamplerV2, measurement twirling is enabled by default. So one needs to set `options.twirling.enable_measure = False`. 

These defaults can change in the future, so we should decide whether to show code to turn them all off now, or update the docs if/when defaults change. 

My first thought was to add this to the error mitigation page... but this applies to both error mitigation _and_ suppression. Perhaps we can repurpose the compilation page to error suppression, since our compilation is just error suppression now (i.e. no layout and routing). 

### LINKED ISSUE #1529: [ NEW PAGE ] runtime feature compatibility table
@jyu00 is working on a "feature compatibility table" that shows which runtime features can be used together or not. We should add this table to the docs, either as a separate page or as part of an existing runtime options page

beckykd:
Closes #1416 
Closes #1453 
Closes #1247
Closes #1529 

Comment from Blake: the discoverability of the detailed Sampler and Estimator options is very poor. i.e. a user can only find these by navigating the Qiskit Runtime IBM Client API docs. Can we add links to [SamplerOptions](https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.options.SamplerOptions) and [EstimatorOptions](https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.options.EstimatorOptions) to the [Get started with primitives](https://docs.quantum.ibm.com/guides/get-started-with-primitives) page?


beckykd:
@arnaucasau or @frankharkins , I need to change the order of these files in the TOC in the entries.py page.  **Overview should be above error suppression.**  I assume you just move them around, but Eric said it's complicated so we should ask you.  Could one of you please make whatever change is necessary? 

This is what it looks like now: 

        children=(
            Entry(
                "Configure error suppression",
                slug="/configure-error-suppression",
                from_file="run/configure-error-suppression.mdx",
            ),
            Entry(
                "Configure runtime error mitigation",
                slug="/configure-error-mitigation",
                from_file="run/configure-error-mitigation.mdx",
            ),
            Entry(
                "Qiskit Runtime options overview",
                slug="/runtime-options-overview",
                from_file="run/runtime-options-overview.mdx",

arnaucasau:
Yes, in this case, we only need to change the order of the entries. I changed it in https://github.com/Qiskit/documentation/pull/1499/commits/af0b8677875df2c207d5fc1f09f5a51f0e4095eb and as you can see in the following screenshots, we get the correct order in the left ToC and in the index of the `execute-on-hardware` page

Captura desde 2024-06-06 19-57-04

Captura desde 2024-06-06 19-57-17

javabster:
I got some feedback from the enablement team that they have been using this table in some of their workshops with users and it has been well received, do you think there is a way to incorporate this table (or something similar) into the options page?
Screenshot 2024-06-07 at 2 30 03 AM

beckykd:
> do you think there is a way to incorporate this table (or something similar) into the options page?

Since we're planning on putting in options for Estimator V1, V2 and Sampler V1, V2, it seems like the best thing to do would be to move that table into this topic (and add 3 more like it), rather than duplicating it.  Thoughts?

beckykd:
I made two trial versions of the tables / info based on ideas that @abbycross  and I brainstormed, so you can compare side-by-side: 

https://qiskit-docs-preview-pr-1499.1799mxdls7qz.us-south.codeengine.appdomain.cloud/run/runtime-options-overview#options-classes

https://qiskit-docs-preview-pr-1499.1799mxdls7qz.us-south.codeengine.appdomain.cloud/run/runtime-options-overview-2#options-classes

I'd love your suggestions @javabster @jyu00

jyu00:
> I got some feedback from the enablement team that they have been using this table in some of their workshops with users and it has been well received

Ha, I saw the same table which prompted [this comment](https://github.com/Qiskit/documentation/issues/1416#issuecomment-2152394712), as I too found the table very helpful!

jyu00:
> I made two trial versions of the tables / info based on ideas that @abbycross and I brainstormed, so you can compare side-by-side:
> 
> https://qiskit-docs-preview-pr-1499.1799mxdls7qz.us-south.codeengine.appdomain.cloud/run/runtime-options-overview#options-classes
> 
> https://qiskit-docs-preview-pr-1499.1799mxdls7qz.us-south.codeengine.appdomain.cloud/run/runtime-options-overview-2#options-classes


I like the first table better, since it's a one-stop-shop. However, at least on my screen, the right most column is not immediately visible, and it was not obvious at all that you can scroll to the right (because Mac doesn't show scroll bars...). I assume there is a standard way to tell people there's more on the right? 

image

abbycross:
I wonder if we need tabs for the option descriptions before the tabbed table, since the descriptions don't change among the versions - just which primitives use which options, right? I'd lean toward simply mentioning which apply to which primitive/version when that is applicable, and only have one set of descriptions.

beckykd:
> I wonder if we need tabs for the option descriptions before the tabbed table, since the descriptions don't change among the versions - just which primitives use which options, right? I'd lean toward simply mentioning which apply to which primitive/version when that is applicable, and only have one set of descriptions.

Right now they're all the same, since I just copied them all in there; I think there will be different versions for V1 and V2 though, because there are (I believe) different options.  I can work on that and see what we end up with.

beckykd:
> I like the first table better, since it's a one-stop-shop.

It's probably easier to write, too, but I don't like that you have to scroll vertically AND horizontally to find all of the information.  It wish we could at least lock that top row.

javabster:
I don't have a strong preference for option 1 or option 2. Option 1 is nice to have it all together, but I also like how in option 2 you can use the right hand sidebar to jump to the specific option you care about. I'm happy to defer to Jessie's preference.

To help with the scroll issue in the interim we could just add a note in the text reminding people to scroll to the right if they cant see it in their browser

for the heading "Options classes" could we call it "Options classes summary" instead?

docs/run/runtime-options-overview.mdx
```diff
@@ -0,0 +1,559 @@
...
+
+<Tabs>
+  <TabItem value="EstimatorV2" label="Estimator V2">
+| Options              | Sub-options              | Sub-sub-options         | Choices                                                                                   | Default                   |
+|----------------------|--------------------------|-------------------------|-------------------------------------------------------------------------------------------|---------------------------|
+| default_shots        |                          |                         |  integer in the range[0, `backend.max_shots`]                                             | `4096`                      |
+| default_precision    |                          |                         |  float > 0                                                                                        | `0.015625 (1 / sqrt(4096))`     |
+| dynamical_decoupling |                          |                         |                                                                                           |   `FieldInfo(annotation=Union[DynamicalDecouplingOptions, Dict], required=False, default_factory=DynamicalDecouplingOptions)`                         |
```
jyu00, line 138: I don't think this `FieldInfo(...)` is intentional...

docs/run/runtime-options-overview.mdx
```diff
@@ -0,0 +1,559 @@
...
+
+<Admonition type="caution" title="Important">
+Scroll to see the full table.
+</Admonition>
+
+<Tabs>
+  <TabItem value="EstimatorV2" label="Estimator V2">
+| Options              | Sub-options              | Sub-sub-options         | Choices                                                                                   | Default                   |
+|----------------------|--------------------------|-------------------------|-------------------------------------------------------------------------------------------|---------------------------|
+| default_shots        |                          |                         |  integer in the range[0, `backend.max_shots`]                                             | `4096`                      |
```
jyu00, line 136: `4096` is the default value for SamplerV2. The default for EstimatorV2 is `None`

docs/run/runtime-options-overview.mdx
```diff
@@ -0,0 +1,559 @@
...
+| default_precision    |                          |                         |  float > 0                                                                                        | `0.015625 (1 / sqrt(4096))`     |
```
jyu00, line 137:
```suggestion
| default_precision    |                          |                         |  float > 0                                                                                        | `0.015625, or 1 / sqrt(4096)`     |
```

docs/run/runtime-options-overview.mdx
```diff
@@ -0,0 +1,559 @@
...
+|                      |                          | noise_gain              |  `auto`/ float in the range `[0, 1)`                                                     | `auto`                    |
+|                      | pec_mitigation           |                         | `True`/`False`                                                                           | `False`                      |
+|                      | zne_mitigation           |                         | `True`/`False`                                                                                | `False`                     |
+|                      | zne                      | noise_factors           |   list of floats; each float >= 1                                                                   | `(1, 3, 5)`                 |
+|                      |                          | extrapolator            | `'exponential'`<br />`'linear'`<br />`'double_exponential'`<br />`'polynomial_degree_(1 <= k <= 7)'` | (`'exponential'`, `'linear'`) |
```
jyu00, line 165: ZNE amplifiers are missing here

jyu00: Also this choices column typically means you can pick from one of the values. But this option is actually 1 or more of these extrapolator values (hence the default is `("exponential", "linear")`)

docs/run/runtime-options-overview.mdx
```diff
@@ -0,0 +1,559 @@
...
+|                      | measure_mitigation       |                         | `True`/`False`                                                                                | `True`                      |
+|                      | measure_noise_learning   | num_randomizations      |  integer >= 1 (Note that this value strongly influences execution time.)                                                                             | `32`                      |
+|                      |                          | shots_per_randomization |  integer                                                                                 | 'auto'                    |
+|                      | pec                      | max_overhead            |  `None`/ integer >1                                                                      | `100`                      |
+|                      |                          | noise_gain              |  `auto`/ float in the range `[0, 1)`                                                     | `auto`                    |
```
jyu00, line 161:
```suggestion
|                      |                          | noise_gain              |  `auto`/ float in the range `[0, 1]`                                                     | `auto`                    |
```

docs/run/runtime-options-overview.mdx
```diff
@@ -0,0 +1,559 @@
...
+|                      | coupling_map             |                         |  List where each entry specifies a directed two-qubit interaction, for example: [[0, 1], [0, 3], [1, 2], [1, 5], [2, 5], [4, 1], [5, 3]]     | full connectivity                   |
+|                      | basis_gates              |                         |  Set of single- and two-qubit gates that provide a universal gates set, meaning that any quantum operation can be decomposed into those gates. Example: `['ecr', 'x', 'sx', 'rz']` | The set of all basis gates supported by [Qiskit Aer simulator](https://qiskit.github.io/qiskit-aer/stubs/qiskit_aer.AerSimulator.html) |
+| transpilation        | skip_transpilation       |                         | `True`/`False`                                                                                      | `False`                     |
+|                      | initial_layout           |                         |                                                                                         |                         |
```
jyu00, line 174: EstimatorV2 doesn't have transpilation options

docs/run/runtime-options-overview.mdx
```diff
@@ -0,0 +1,559 @@
...
+| twirling             | enable_gates             |                         | `True`/`False`                                                                                      | `False`                     |
+|                      | enable_measure           |                         | `True`/`False`                                                                                      | `True`                      |
+|                      | num_randomizations       |                         |  `auto`/ integer >= 1 (Note that this value strongly influences execution time.)            | `'auto'`                    |
+|                      | shots_per_randomization  |                         |   `auto`/ integer >= 1.  The maximum depends on the maximum shots allowed by the system. To find this value, run `backend.max_shots`.                                                                      | `'auto'`                    |
```
jyu00, line 178: The description seems wordier than the rest. Maybe just "`auto` / integer >= 1 and <= `backend.max_shots`"

docs/run/runtime-options-overview.mdx
```diff
@@ -0,0 +1,559 @@
...
+|                      | job_tags                 |                         |List of tags                                                                                             |`None`                           |
+| execution            |                          |                         |                                                                                                         |                           |
+|                      | init_qubits              |                         | `True`/`False`                                                                         | `True`                    |
+|                      | rep_delay                |                         | Value in the range supplied by `backend.rep_delay_range`.                                | Given by `backend.default_rep_delay`.|
+| max_execution_time   |                          |                         | integer number of seconds in the range (0, 5000]                                                                   | `10800` (3 hours)                         |
```
jyu00, line 150: Just out of curiosity, where did you get the `5000`? The valid range is actually [1, 10800]. Side note, is there a standard way to specify a range? Do we use `()` or `[]` or something else?

docs/run/runtime-options-overview.mdx
```diff
@@ -0,0 +1,559 @@
...
+|                      | noise_model              |                         |   [Qiskit Aer NoiseModel](../verify/building_noise_models), or its representation        | `None`                      |
+|                      | seed_simulator           |                         |   integer                                                                               | `None`                    |
+|                      | coupling_map             |                         |  List where each entry specifies a directed two-qubit interaction, for example: [[0, 1], [0, 3], [1, 2], [1, 5], [2, 5], [4, 1], [5, 3]]     | full connectivity                   |
+|                      | basis_gates              |                         |  Set of single- and two-qubit gates that provide a universal gates set, meaning that any quantum operation can be decomposed into those gates. Example: `['ecr', 'x', 'sx', 'rz']` | The set of all basis gates supported by [Qiskit Aer simulator](https://qiskit.github.io/qiskit-aer/stubs/qiskit_aer.AerSimulator.html) |
```
jyu00, line 172: Rather than "choices", these two have basically the full description. Maybe for `coupling_map` we can say "List of directed two-qubit interaction". And for `basis_gates` "List of basis gate names to unroll to".

jyu00 (commented):
I think it'd also be useful to link the API reference for each options category, so people can read the full description if they need to

ihincks (commented):
Looking great @beckykd !

docs/run/runtime-options-overview.mdx
```diff
@@ -0,0 +1,559 @@
...
+  - [Execution](../api/qiskit-ibm-runtime/qiskit_ibm_runtime.options.ExecutionOptions): Primitive execution options, including whether to initialize qubits and the repetition delay.
+  - [Twirling](../api/qiskit-ibm-runtime/qiskit_ibm_runtime.options.TwirlingOptions): Twirling options, such as whether to apply two-qubit gate twirling and the number of shots to run for each random sample.
+  - [Environment](../api/qiskit-ibm-runtime/qiskit_ibm_runtime.options.EnvironmentOptions): Execution environment options such as the logging level to set and job tags to add.
+  - [Simulator](../api/qiskit-ibm-runtime/qiskit_ibm_runtime.options.SimulatorOptions): Simulator options, such as the basis gates, simulator seed, and coupling map. Applies to **local testing mode only**.
+  </TabItem>
+
+</Tabs>
+
+<span id="options-table"></span>
+### Available options
```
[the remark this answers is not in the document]
beckykd: > We'll have to keep this list up-to-date with the option classes. Maybe we could have a label in qiskit-ibm-runtime about option doc changes or something?

Are you thinking of a tag in that repo?  LMK if you decide on something like that so I can follow it.

abbycross (approved):
👍

arnaucasau:
@beckykd 

[Yesterday](https://github.com/Qiskit/documentation/pull/1600) we moved all the images for `start`, `build`, `transpile`, `verify`, and `run` to a folder called `guides/` in preparation for the patterns reorg. To avoid creating the `run` folder again in the main branch when this PR gets merged, could you move the new images to `public/images/guides/`?

The change should be `public/images/run/runtime-options-overview/* -> public/images/guides/runtime-options-overview/*`.

In case you have to change a lot of links here, I'm happy to help updating them, just let me know :+1:

beckykd:
@jyu00 , @ihincks - are the options for SamplerV1 the same as those for EstimatorV1?

beckykd:
> @jyu00 , @ihincks - are the options for SamplerV1 the same as those for EstimatorV1?

Got confirmation that the options are the same

beckykd:
follow-up here: https://github.com/Qiskit/documentation/pull/1723

PR #1502, opened 2024-06-06
Change docker image

arnaucasau:
Use preview app for PR previews


Eric-Arellano (approved):

Dockerfile
```diff
@@ -31,10 +31,10 @@
 #   2. ❯ docker run --rm -p 3000:3000 -t qiskit-docs-preview
```
Eric-Arellano, line 31: the instructions are now stale

PR #1468, opened 2024-05-30
[do not merge] Demo collapsible get started

Eric-Arellano:

PR #1289, opened 2024-05-03
[do not merge] Preview new headers hierarchy

arnaucasau:
`Do not merge`

This PR is to trigger the preview of the new header hierarchy and make the review of #1283 easy

PR #1512, opened 2024-06-10
Use `pull_request_target` for PR preview teardown

frankharkins:
Should fix 1034.

## Investigation

It seems PR previews _are_ being cleaned up correctly for _merged_ PRs; the teardown action is running, and the previews are no longer accessible after merging (see following screenshot as example).

Screenshot 2024-06-07 at 13 35 48

It seems the problem is if a PR is closed but not merged (e.g. #1468, #1380, and #1289 still have active previews). I believe this is the reason: https://github.com/orgs/community/discussions/26657#discussioncomment-3252753.

## Solution

This PR tries to fix this by using the `pull_request_target` event to trigger teardowns. This should trigger correctly when PRs are closed.

We need to be careful with this event as it allows untrusted PRs to trigger events using secrets. Since this action runs on the target branch (rather than a merge commit of the target and PR branches), an untrusted user shouldn't be able to do anything malicious, but we need to make sure we **never** checkout the PR branch with this event type, or read any user-defined inputs (such as the PR title) as it could lead to an injection. I think this is unlikely, but I've left a warning in the action just in case.

### DIFF
--- a/.github/workflows/preview-docs-teardown.yml
+++ b/.github/workflows/preview-docs-teardown.yml
@@ -1,0 +1,51 @@
+# This code is a Qiskit project.
+#
+# (C) Copyright IBM 2023.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# WARNING: This event type uses secrets even when triggered by an untrusted PR.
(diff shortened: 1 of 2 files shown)


arnaucasau (approved):
Looks good to me. Thanks for adding the warning as well!

Very interesting the use of `pull_request_target`. Have you considered using `workflow_run` as the GitHub Support team suggested in the [comment](https://github.com/orgs/community/discussions/26657#discussioncomment-3252753) you linked in the PR description? I'm thinking that we would have the same problem as with the original workflow in that case :thinking:

frankharkins:
Thanks!

> Have you considered using `workflow_run`...

From what I understand, I don't think that would work as that triggers a workflow based on the completion of another workflow, whereas we want the trigger to be PR closure.

PR #1513, opened 2024-06-10
Explain why decompose is used

### LINKED ISSUE #1439: Explain why decompose should be used

It is not clear why an user should use decompose for re-synthesis as done here: https://docs.quantum.ibm.com/build/unitary-synthesis#re-synthesis-for-circuit-optimization  (comment from Paul n)

### Select all that apply

- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [X] other

### Describe the fix.

Explain this process more fully

kaelynj:
Closes #1439 
 
Decompose is basically only used here to visualize the circuit correctly. If it wasn't, it  just be an opaque block when calling `draw()`


javabster (approved):
LGTM 🚀

Issue #1462, opened 2024-05-30
Patterns: get CI green for guides migration


Eric-Arellano:
1. Run `scripts/patterns-reorg/main.py`
2. Get `npm run check` to be green, and you also want to run the commands we use in `api-checks.yml`.

Fixes will probably require https://github.com/Qiskit/documentation/issues/1460, then using ignores to work around https://github.com/Qiskit/documentation/issues/1461. We also need to add metadata for the new pages, which can be done in `page_content.py`.

Issue #1461, opened 2024-05-29
Patterns: Update API links to new `guides/` destinations


Eric-Arellano:
Follow up to https://github.com/Qiskit/documentation/issues/1460. 

Even though we will have redirects in place, it's a best practice to update our API docs to use the correct URL directly. This should be done by changing the docs in the relevant repositories for versions of the docs that are still being updated. Otherwise, we should update the HTML in the uploaded Box artifacts.

Update July 18: they should also link to `/migration-guides` rather than `/api/migration-guides`, once that goes out.

Eric-Arellano:
PRs are up in both repositories.

PR #1520, opened 2024-06-10
Get CI green for guides migration

arnaucasau:
Part of #1462

This PR prepares our checks for the guides migration to get green CI when we make the change. Changes:

- Added API docs links to `ignore.ts`. Those links need to be updated from the other repos and from the Box artifacts for historical versions. Once the migration is done we'll have redirects set up for those links as explained [here](https://github.com/Qiskit/documentation/issues/1461).
- Added some words to the cspell dictionary and fixed other complaints from the checker
- Added a script to fix the Qiskit bot files by changing the paths to `guides/`. The PR also changes the ignores we had to `guides/`, leaving the old ones that should be removed once the migration is done.
- Fixed the `update_internal_links.py` script to switch over more links and consider the migration guides.
- Added entries for the notebooks with their names on the `guides/` folder. Once the migration is done, we need to remove the old ones.


scripts/patterns-reorg/update_qiskit_bot.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3.11
```
arnaucasau, line 1: This script, together with update_internal_links, is meant to be used only one time. After the migration, they will be removed

scripts/patterns-reorg/update_qiskit_bot.py
```diff
@@ -0,0 +1,86 @@
...
+            )
+        )
+        # Use a placeholder to have the correct indentation for the notifications
+        yaml.dump(
+            {"INDENTATION": notifications},
+            file,
+            default_flow_style=False,
+            default_style='"',
+        )
+    replace_indentation_placeholder(qiskit_bot_file_path)
```
arnaucasau, line 76: This trick is done to avoid changing the style of the current qiskit bot config file and to have a cleaner git diff

arnaucasau:
In the commit [d1f8726](https://github.com/Qiskit/documentation/pull/1520/commits/d1f8726a514b1d3151fba5e6853df101ac344e2a), I executed the following commands to generate the new pages and remove the old ones, updating all the links.

```bash
./scripts/patterns-reorg/update_internal_links.py
./scripts/patterns-reorg/update_qiskit_bot.py
./scripts/patterns-reorg/main.py --delete-existing
npm run fmt
````

The commit passed all the tests except [Execute notebooks](https://github.com/Qiskit/documentation/actions/runs/9455788573/job/26046328417). The notebooks are the same as in the main branch just copied over to the `guides/` folder. I also tested it in this [PR](https://github.com/Qiskit/documentation/pull/1537) and it also failed, so probably out of the scope of this PR, and it should be fixed in the main branch

**Review by frankharkins — APPROVED** [review]
Thanks!

File: scripts/nb-tester/notebooks.toml [hunk 3]
```diff
@@ -51,8 +80,10 @@ notebooks_that_submit_jobs = [
 # A job is "too big" if a cell can't run in under 5 mins, or we run out of
 # memory on a reasonable device.
 notebooks_no_mock = [
-    "docs/start/hello-world.ipynb",
+    "docs/guides/hello-world.ipynb",
```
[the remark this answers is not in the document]
arnaucasau: Thanks for the review! Yes, both should exist for the moment until we make the migration. I moved the `docs/start/hello-world.ipynb` entry to the end of the list (line 88) to clarify that the comment only affects that line, just in case.

.github/workflows/notebook-test.yml
```diff
@@ -56,6 +56,11 @@ jobs:
         with:
           # Add your notebook to this list if it needs latex or graphviz to run
           files: |
+            docs/guides/visualize-circuits.ipynb
+            docs/guides/custom-backend.ipynb
+            docs/guides/transpiler-stages.ipynb
+            docs/guides/represent-quantum-computers.ipynb
+            docs/guides/common-parameters.ipynb
```
Eric-Arellano, line 63: It'd be good to comment below that these should be removed.

scripts/commands/checkQiskitBotFiles.ts
```diff
@@ -16,6 +16,28 @@ import { load } from "js-yaml";
...
+  "docs/guides/execution-modes-faq",
+  "docs/guides/execution-modes",
+  "docs/guides/native-gates",
+  "docs/guides/save-jobs",
+  "docs/guides/visualize-results",
+  "docs/guides/create-transpiler-plugin",
+  "docs/guides/intro-to-patterns",
+  "docs/guides/optimize-for-hardware",
+  "docs/guides/postprocess-results",
+  // Todo: remove the following old paths
```
Eric-Arellano, line 40: It'd be helfpul to use an easily searchable format like `TODO(patterns-reorg):`. Ditto the others.

Eric-Arellano (commented):
Nice work! Thank you!

PR #1537, opened 2024-06-12
[do not merge] Test execute notebooks CI

arnaucasau:
**Do not merge**, this PR is to try if CI breaks executing the modified notebook

### DIFF
--- a/docs/run/save-jobs.ipynb
+++ b/docs/run/save-jobs.ipynb
@@ -17,6 +17,7 @@
    "source": [
     "## Retrieve jobs from IBM Quantum\n",
     "\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


frankharkins:
This is failing due to #1552, not because of a problem with your script. I'm not sure how this notebook made it into the repo... maybe before we started testing with Python 3.9?

arnaucasau:
Oh, I see. Thanks for looking into this Frank! I can make the changes if you want, so feel free to assign the issue to me.

I found that `/run/save-jobs.ipynb` also fails with the message `Unable to retrieve result for job csnfd2mp1vzg008ab670. Job was cancelled.` as you can see in [3606f1b](https://github.com/Qiskit/documentation/pull/1537/commits/3606f1b0c788d5c5bcb861cd75d8d23e436e6168).

PR #1536, opened 2024-06-12
Add "Latest updates" link to patterns reorg

frankharkins:
After a meeting about #1301, we decided the most appropriate action was to link out to the existing "Product updates" page.

PR #1541, opened 2024-06-13
Revert "Add "Latest updates" link to patterns reorg (#1536)"

frankharkins:
Turns out we don't want to do this any more.


arnaucasau (approved):
Thanks!👍🏻

Issue #1299, opened 2024-05-03
Patterns: validate Patterns index pages include all guides


Eric-Arellano:
In https://github.com/Qiskit/documentation/pull/1294, each of the 4 patterns pages links out to the rest of the docs app. It provides an alternative organization than the tools-based approach.

We are going to manually maintain both `guides/_toc.json` and those 4 index pages. There are a few edge cases that would be too complex for us to generate the index files in a clear way. It's best to let humans decide what goes where.

However, we want an automated check to ensure that every page shows up both in the Tools folders & in the Workflow index pages. Likewise, a page should never show up >1 time in the `_toc.json` or the index pages. The check should not care what TOC section/workflow index the specific page shows up in; we only need to validate that it shows up exactly once in both places.

This code is probably useful for how to parse the index pages:

https://github.com/Qiskit/documentation/blob/2974d289268980131d967b3cd2224d80e99d23f0/scripts/lib/api/TocGrouping.test.ts#L80

Note that the index pages may have sublists, like `*` with indentation.

We probably need an allowlist mechanism. Right now, Qiskit Serverless only shows up in the Tools and not Workflow index pages.

PR #1550, opened 2024-06-14
Add patterns index validation

arnaucasau:
Closes https://github.com/Qiskit/documentation/issues/1299

This PR adds a script that validates that every page shows up both in the Tools folders and in the Workflow index pages.

### DIFF
--- a/scripts/commands/checkPatternsIndex.ts
+++ b/scripts/commands/checkPatternsIndex.ts
@@ -192,10 +192,11 @@ async function main() {
       ...getExtraIndexPagesErrors(indexPage, indexEntries, toolsEntries),
     );
 
-    // Remove index entries from the tools entries list
-    toolsEntries = new Set(
-      [...toolsEntries].filter((page) => !indexEntries.has(page)),
-    );
+    toolsEntries.forEach((page) => {
+      if (indexEntries.has(page)) {
+        toolsEntries.delete(page);
+      }
+    });
   }
 
   const extraToolsEntriesErrors = getExtraToolsEntriesErrors(toolsEntries);
--- a/scripts/commands/checkPatternsIndex.ts
+++ b/scripts/commands/checkPatternsIndex.ts
@@ -23,7 +23,7 @@ const INDEX_PAGES = [
 
 const TOC_PATH = `docs/guides/_toc.json`;
 
-async function getIndexPages(indexPage: string): Promise<string[]> {
+async function getIndexEntries(indexPage: string): Promise<string[]> {
   const rawIndex = await readFile(indexPage, "utf-8");
   const result: string[] = [];
   for (const line of rawIndex.split("\n")) {
@@ -77,7 +77,7 @@ async function getToolsTocEntriesToCheck(): Promise<string[]> {
   return toolsPages.filter((page) => !IGNORE_URL.includes(page));
 }
 
-async function getPagesAndDuplicatesErrors(
+async function deduplicateEntriesAndGetErrors(
   src: string,
   entries: string[],
 ): Promise<[string[], string[]]> {
@@ -133,23 +133,23 @@ async function main() {
   const extraIndexEntriesErrors: string[] = [];
   const extraToolsEntriesErrors: string[] = [];
 
-  const toolsEntries = await getToolsTocEntriesToCheck();
-  let [toolsPages, toolsErrors] = await getPagesAndDuplicatesErrors(
+  const allToolsEntries = await getToolsTocEntriesToCheck();
+  let [toolsEntries, toolsErrors] = await deduplicateEntriesAndGetErrors(
     TOC_PATH,
-    toolsEntries,
+    allToolsEntries,
   );
   duplicatesErrors.push(...toolsErrors);
 
   for (const indexPage of INDEX_PAGES) {
-    const indexEntries = await getIndexPages(indexPage);
-    let [indexPages, indexErrors] = await getPagesAndDuplicatesErrors(
+    const allIndexEntries = await getIndexEntries(indexPage);
+    let [indexEntries, indexErrors] = await deduplicateEntriesAndGetErrors(
       indexPage,
-      indexEntries,
+      allIndexEntries,
     );
     duplicatesErrors.push(...indexErrors);
 
-    const ExtraIndexPages = indexPages.filter(
-      (page) => !toolsPages.includes(page),
+    const ExtraIndexPages = indexEntries.filter(
+      (page) => !toolsEntries.includes(page),
     );
     if (ExtraIndexPages.length > 0) {
       ExtraIndexPages.forEach((page) =>
@@ -159,11 +159,12 @@ async function main() {
       );
     }
 
-    toolsPages = toolsPages.filter((page) => !indexPages.includes(page));
+    // Remove index entries from the tools entries list
+    toolsEntries = toolsEntries.filter((page) => !indexEntries.includes(page));
   }
 
-  if (toolsPages.length > 0) {
-    toolsPages.forEach((page) =>
+  if (toolsEntries.length > 0) {
+    toolsEntries.forEach((page) =>
       extraToolsEntriesErrors.push(
         `❌ The entry ${page} is not present on any index page`,
       ),
--- a/scripts/commands/checkPatternsIndex.ts
+++ b/scripts/commands/checkPatternsIndex.ts
@@ -12,8 +12,9 @@
 
 import { readFile } from "fs/promises";
 import { pathExists } from "../lib/fs";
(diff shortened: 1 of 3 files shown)


scripts/commands/checkPatternsIndex.ts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+async function getToolsTocEntriesToCheck(): Promise<string[]> {
+  const toc = JSON.parse(await readFile(TOC_PATH, "utf-8"));
+  const toolsNode = toc.children.find((child: any) => child.title == "Tools");
+  const toolsPages = getTocSectionPageNames(toolsNode);
+  return toolsPages.filter((page) => !IGNORE_URL.includes(page));
+}
+
+async function getDeduplicateEntriesAndAddErrors(
```
Eric-Arellano, line 82: A couple suggestions on this function:

1. It'd be better to move the `const entries =` back into `main`, imo. This function is doing a lot right now, which you can tell by the function name. Instead, this function would take `entries: string[]` as an arg
2. The function should be returning a `Set<string>` for the deduplicated entries since it has O(1) lookup
3. You can call the function something like `deduplicateEntries`
4. Rather than mutating the `errors` arg, more idiomatic is to return the errors as a new value. I know that you call this function twice; you'd concatenate the two error arrays in the final place that you use them. You'd end up returning `[Set<string>, string[]]`, or you could define an object to be more explicit.

scripts/commands/checkPatternsIndex.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  return deduplicatedPages;
+}
+
+function addExtraIndexPagesErrors(
+  indexPage: string,
+  indexEntries: string[],
+  toolsEntries: string[],
+  errors: string[],
```
Eric-Arellano, line 107: Ditto that you should return a new array rather than mutating

scripts/commands/checkPatternsIndex.ts
```diff
@@ -0,0 +1,212 @@
...
+): void {
+  const ExtraIndexPages = indexEntries.filter(
```
Eric-Arellano, line 109: variable should be lowercase

scripts/commands/checkPatternsIndex.ts
```diff
@@ -0,0 +1,212 @@
...
+    (page) => !toolsEntries.includes(page),
+  );
+  if (ExtraIndexPages.length > 0) {
```
Eric-Arellano, line 112: Why the if statement? You can also do `return indexEntries.filter().map()` directly

Eric-Arellano (commented):
Thanks for working on this!

scripts/commands/checkPatternsIndex.ts
```diff
@@ -0,0 +1,212 @@
...
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+import { readFile } from "fs/promises";
+import { pathExists } from "../lib/fs";
+
+const IGNORE_URL = ["/guides/qiskit-serverless"];
```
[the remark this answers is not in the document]
arnaucasau: I tried it, but the prettier removes the comma and inlines the entry :thinking:

scripts/commands/checkPatternsIndex.ts
arnaucasau: Changed it to follow your comment :+1:

scripts/commands/checkPatternsIndex.ts
```diff
@@ -0,0 +1,212 @@
...
+
+function getTocSectionPageNames(sectionNode: TocEntry): string[] {
+  let results = [];
+  if (sectionNode.url) {
+    results.push(sectionNode.url);
+  }
+
+  if (sectionNode.children) {
+    for (const child of sectionNode.children) {
+      results = [...results, ...getTocSectionPageNames(child)];
```
Eric-Arellano, line 72:
```suggestion
      results.push(...getTocSectionPageNames(child));
```

scripts/commands/checkPatternsIndex.ts
```diff
@@ -0,0 +1,212 @@
...
+  const toc = JSON.parse(await readFile(TOC_PATH, "utf-8"));
+  const toolsNode = toc.children.find(
+    (child: TocEntry) => child.title == "Tools",
+  );
+  const toolsPages = getTocSectionPageNames(toolsNode);
+  return toolsPages.filter((page) => !IGNORED_URLS.includes(page));
+}
+
+async function deduplicateEntries(
+  src: string,
```
Eric-Arellano, line 89: `src` isn't a great name. Maybe `filePath`?

scripts/commands/checkPatternsIndex.ts
```diff
@@ -0,0 +1,212 @@
...
+  toolsEntries: Set<string>,
+): string[] {
+  const extraIndexPages = [...indexEntries].filter(
+    (page) => !toolsEntries.has(page),
+  );
+
+  return extraIndexPages.map(
+    (page) =>
+      `❌ ${indexPage}: The entry ${page} doesn't appear in the \`Tools\` menu.`,
+  );
```
Eric-Arellano, line 118:
```suggestion
  return [...indexEntries].filter(
    (page) => !toolsEntries.has(page),
  ).map(
    (page) =>
      `❌ ${indexPage}: The entry ${page} doesn't appear in the \`Tools\` menu.`,
  );
```

scripts/commands/checkPatternsIndex.ts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+function getExtraToolsEntriesErrors(toolsEntries: Set<string>): string[] {
+  return [...toolsEntries].map(
+    (page) => `❌ The entry ${page} is not present on any index page`,
+  );
```
Eric-Arellano, line 124: This is a little surprising to not have logic in here to do the check, and that it only creates the error messages. I see why you do it that way, but the naming should reflect it. Such as calling the arg `remainingToolsEntries`. Or, move the filtering logic inside this function.

scripts/commands/checkPatternsIndex.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  let extraIndexEntriesErrors: string[] = [];
+  for (const indexPage of INDEX_PAGES) {
+    const indexAllEntries = await getIndexEntries(indexPage);
+    let [indexEntries, indexDuplicatedErrors] = await deduplicateEntries(
+      indexPage,
+      indexAllEntries,
+    );
+    duplicatesErrors = [...duplicatesErrors, ...indexDuplicatedErrors];
```
Eric-Arellano, line 189:
```suggestion
    duplicatesErrors.push(...indexDuplicatedErrors);
```

scripts/commands/checkPatternsIndex.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    extraIndexEntriesErrors = [
+      ...extraIndexEntriesErrors,
+      ...getExtraIndexPagesErrors(indexPage, indexEntries, toolsEntries),
+    ];
```
Eric-Arellano, line 194:
```suggestion
    extraIndexEntriesErrors.push(
      ...getExtraIndexPagesErrors(indexPage, indexEntries, toolsEntries),
    );
```

scripts/commands/checkPatternsIndex.ts
```diff
@@ -0,0 +1,211 @@
...
+    duplicatesErrors.push(...indexDuplicatedErrors);
+
+    extraIndexEntriesErrors.push(
+      ...getExtraIndexPagesErrors(indexPage, indexEntries, toolsEntries),
+    );
+
+    // Remove index entries from the tools entries list
+    toolsEntries = new Set(
+      [...toolsEntries].filter((page) => !indexEntries.has(page)),
+    );
```
Eric-Arellano, line 198: Taken from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/delete - this should be more clear and you could probably get rid of the comment

```suggestion
    toolsEntries.forEach((page) => {
      if (indexEntries.has(page)) {
        toolsEntries.delete(page)
       }
       }
    );
```

Eric-Arellano (approved):
Outside of the last nit, this looks good. Thanks!

scripts/commands/checkPatternsIndex.ts
arnaucasau: Awesome, thank you Eric!

PR #1562, opened 2024-06-16
Added clarification regarding job mode in V2

### LINKED ISSUE #1486: Clarify `backend=backend` runs things in job mode for v2 primitives
If you set `backend=backend` in a v2 primitive it will run the program in job mode, even if it's inside a batch/session context
```
with Batch(backend=backend):
    sampler = SamplerV2(backend=backend)
```

so in general we should emphasise to users that they should not do this and instead use this formats:
```
with Batch(backend=backend):
    sampler = SamplerV2()
```
(note `session=batch` is confusing and will likely be changed in future)

arujjval:
Closes #1486

added a note at https://docs.quantum.ibm.com/run/run-jobs-batch

### DIFF
--- a/docs/run/run-jobs-batch.mdx
+++ b/docs/run/run-jobs-batch.mdx
@@ -46,3 +46,8 @@ The following example shows how you can partition a long list of circuits into m
          job = sampler.run(partitioned_circuits)
          jobs.append(job)
  ```
+
+
+<Admonition type="note">
+    If you set `backend=backend` in a v2 primitive it will run the program in job mode, even if it's inside a batch/session context. So, preferred way to use it is `sampler = SamplerV2()` when running in batch mode.
+</Admonition>


docs/run/run-jobs-batch.mdx
```diff
@@ -46,3 +46,8 @@ The following example shows how you can partition a long list of circuits into m
          job = sampler.run(partitioned_circuits)
          jobs.append(job)
  ```
+
+
+<Admonition type="note">
+    If you set `backend=backend` in a v2 primitive it will run the program in job mode, even if it's inside a batch/session context. So, preferred way to use it is `sampler = SamplerV2()` when running in batch mode.
```
frankharkins, line 52: I wonder if we should phrase this more strongly; what do you think of this?

```suggestion
<Admonition type="caution">
    If you set `backend=backend` in a v2 primitive it will run the program in job mode, even if it's inside a batch/session context. Avoid setting the backend parameter if you want to run in batch/session mode.
```

arujjval: I think its a good idea

frankharkins: No; if you're happy with the suggestion we can commit it and merge the PR

arujjval: Sure, go ahead.

frankharkins (approved):
Confirmed this was accurate with @jyu00

Issue #605, opened 2024-01-10
[ EPIC] Add open source related information to "Additional resources" tab


javabster:
Because of the time crunch of the last milestone we never got round to finding a good home in docs for some of the open-source related information including:

- version strategy
- operating system support
- contributing guidelines + maintainer guidelines
- deprecation policy

some of these pages were added to https://docs.quantum.ibm.com/start/install, others were just added as markdown files in the qiskit repo. But the current placement doesn't feel like it exactly "fits" so we should come up with a better place.

Terris also started looking into designs for a "contributing" section of the docs, I think it would be good to re-open this discussion now that things have calmed down a bit

```[tasklist]
### Tasks
- [ ] #796
- [ ] #716
- [ ] https://github.com/Qiskit/documentation/issues/1698
```

abbycross:
Just so we don't lose it, mentioning here that another place where the open-source info got placed is on the Support page: https://docs.quantum.ibm.com/support#open-source-governance

javabster:
related #716

Eric-Arellano:
We plan to have a new "Additional resources" section in the top nav bar. Open source will live here. We can use this epic to plan creation of the content for that folder.

abbycross:
Right now the Error Code Registry is under the API reference menu - is it a candidate for "Additional resources"?

Eric-Arellano:
> Right now the Error Code Registry is under the API reference menu - is it a candidate for "Additional resources"?

We were thinking that stays under API references since it's very API-like, but it could be moved if we want. Migration guides would move, though, since they never really belonged in API reference.

Eric-Arellano:
This is now live. The ecosystem documentation is tracked by https://github.com/Qiskit/documentation/issues/1711, and citations / FAQs by https://github.com/Qiskit/documentation/issues/716 and https://github.com/Qiskit/documentation/issues/1699

PR #1563, opened 2024-06-17
Explain QPY support in install#release-schedule

1ucian0:
Qiskit contract with respect of QPY is kind of strong and deserves an explanation. Adding it to https://docs.quantum.ibm.com/start/install#release-schedule

### DIFF
--- a/docs/guides/install-qiskit.mdx
+++ b/docs/guides/install-qiskit.mdx
@@ -327,6 +327,12 @@ ensures that you see any deprecation warnings before a
 major version release. 
 Without the cap, `pip` installs 
 the newest version available by default.
+
+The QPY serialization format is backwards-compatible so that a new Qiskit release can always load a QPY
+file generated with an earlier release of Qiskit. However, the format isn't forward-compatible so, in principle, it's not possible
+to load QPY files generated with a newer version of Qiskit using an older release.  To facilitate user migration across major version releases, the (`qiskit.qpy.dump()`](/api/qiskit/qpy#dump) function will always support at least one overlapping version between the `X.0.0` and the `X-1.Y.0` release (where `Y` is the last minor version of 
+that series). The parameter `qiskit.qpy.dump(..., version=...)` will enable saving QPY format files that can be loaded by both major versions from the newer
+release. See more details in [RFC 0020](https://github.com/Qiskit/RFCs/blob/master/0020-release_cycle.md#qpy-support).
 </details>
 
 


frankharkins (approved):

docs/start/install.mdx
```diff
@@ -360,6 +360,16 @@ release had some issue with the sdist package publishing, and a post-release
...
 </details>
 
+<details>
+<summary id="qpy-support">
+QPY support</summary>
+The QPY serialization format is backwards compatible so that a new Qiskit can release can always load a QPY
+file generated with an earlier release of Qiskit. However, the format isn't forward compatible so, in principle, it's not possible
+to load QPY files generated with newer version of Qiskit using an older release.  To facilitate user migration across major version releases the `qiskit.qpy.dump()` function will always support at least one overlapping version between the `X.0.0` and the `X-1.Y.0` release (where `Y` is the last minor version of 
+that series). This will enable saving QPY format files that can be loaded by both major versions from the newer
+release. See more details in [RFC 0020](https://github.com/Qiskit/RFCs/blob/master/0020-release_cycle.md#qpy-support).
```
abbycross, line 370:
```suggestion
The QPY serialization format is backwards-compatible so that a new Qiskit release can always load a QPY
file generated with an earlier release of Qiskit. However, the format isn't forward-compatible so, in principle, it's not possible
to load QPY files generated with a newer version of Qiskit using an older release.  To facilitate user migration across major version releases, the `qiskit.qpy.dump()` function will always support at least one overlapping version between the `X.0.0` and the `X-1.Y.0` release (where `Y` is the last minor version of 
that series). This will enable saving QPY format files that can be loaded by both major versions from the newer
release. See more details in [RFC 0020](https://github.com/Qiskit/RFCs/blob/master/0020-release_cycle.md#qpy-support).
```

abbycross (approved):
👍

docs/start/install.mdx
```diff
@@ -360,6 +360,17 @@ release had some issue with the sdist package publishing, and a post-release
 0.25.2.1 only fixed the packaging issue for the release.
 </details>
 
+<details>
+<summary id="qpy-support">
+QPY support
+</summary>
```
mtreinish, line 366: I would move this up to be adjacent to the upgrade strategy section before the pre and post-release sections. Realistically my old repeated complaint is that we need more hierarchy to the headings in this section of the docs is really showing here, because this will look pretty awkward on the rendered docs.

docs/start/install.mdx
```diff
@@ -360,6 +360,17 @@ release had some issue with the sdist package publishing, and a post-release
...
+The QPY serialization format is backwards-compatible so that a new Qiskit release can always load a QPY
+file generated with an earlier release of Qiskit. However, the format isn't forward-compatible so, in principle, it's not possible
+to load QPY files generated with a newer version of Qiskit using an older release.  To facilitate user migration across major version releases, the `qiskit.qpy.dump()` function will always support at least one overlapping version between the `X.0.0` and the `X-1.Y.0` release (where `Y` is the last minor version of 
+that series). This will enable saving QPY format files that can be loaded by both major versions from the newer
+release. See more details in [RFC 0020](https://github.com/Qiskit/RFCs/blob/master/0020-release_cycle.md#qpy-support).
```
mtreinish, line 371: The only thing missing here is I think a link to the `qpy` `dump()` docs and also maybe calling out the `version` kwarg because that's the mechanism to support doing this. I might prefer linking to the qpy docs instead of the rfc.

docs/start/install.mdx
1ucian0: @javabster @Eric-Arellano what do you think?

docs/start/install.mdx
```diff
@@ -360,6 +360,17 @@ release had some issue with the sdist package publishing, and a post-release
 0.25.2.1 only fixed the packaging issue for the release.
 </details>
 
+<details>
+<summary id="qpy-support">
+QPY support
+</summary>
+The QPY serialization format is backwards-compatible so that a new Qiskit release can always load a QPY
+file generated with an earlier release of Qiskit. However, the format isn't forward-compatible so, in principle, it's not possible
+to load QPY files generated with a newer version of Qiskit using an older release.  To facilitate user migration across major version releases, the `qiskit.qpy.dump()` function will always support at least one overlapping version between the `X.0.0` and the `X-1.Y.0` release (where `Y` is the last minor version of
```
1ucian0, line 369: Addressing https://github.com/Qiskit/documentation/pull/1563/files#r1644918215
```suggestion
to load QPY files generated with a newer version of Qiskit using an older release.  To facilitate user migration across major version releases, the (`qiskit.qpy.dump()`](/api/qiskit/qpy#dump) function will always support at least one overlapping version between the `X.0.0` and the `X-1.Y.0` release (where `Y` is the last minor version of 
```

docs/start/install.mdx
1ucian0: Link included in 8291ba9d03bbdb5f695a51a32724389c4e6fba08

`version` is mentioned in fb822da64387ab533272f2dbabd8fc92fb3225af

docs/start/install.mdx
javabster: moving it next to upgrade strategy makes sense to me. Also we will soon begin working on [this issue](https://github.com/Qiskit/documentation/issues/605) to move some of the more open-source focused information into a dedicated section, instead of trying to squeeze it all into the install page, so hopefully this should help ease the header hierarchy issues

1ucian0: This means we should merge this as suggested in this PR and move it with #605 ?

javabster: it means we can move it next to the upgrade strategy for this PR. I just mentioned #605 in response to matthews comment about not liking the header hierarchy in general

1ucian0: done and squashed. Thanks!

PR #1578, opened 2024-06-19
Make code example in "save-jobs" more robust

frankharkins:
This code example fails in our tests if the most recent job is cancelled. This PR filters by job status to avoid that.


Eric-Arellano (approved):
Thanks!

PR #1579, opened 2024-06-20
Fix circuit in sampler in migration guide

Gopal-Dahale:
## Purpose

Fix circuit in sampler in migration guide

## Details

In the migration guide **Migrate from cloud simulators to local simulators**, the section on Clifford simulation creates a `circuit` but does not use it in the pass manager and sampler. This is fixed here.


kaelynj (approved):
Apologies for taking a while to review this. Thanks for catching this!

PR #1587, opened 2024-06-21
Extracting Qubit Indices

### LINKED ISSUE #1469: Better documentation on Registers and Qubits
### URL, if applicable

Probably https://docs.quantum.ibm.com/build/circuit-construction

This is user feedback from Qiskit Slack. A user asked this SO question: https://quantumcomputing.stackexchange.com/questions/38565/unable-to-extract-the-quantum-registers-information-from-qiskit-quantum-circuit

> slack user 2: Once you have your Qubit isolated you can retrieve the index by doing:
qc.qubits.index(your_qubit)
where qc is your QuantumCircuit object

>> slack user 1: Thank you very much! I have been searching all over the internet for the answer for the past 2 days

> Yes I find the documentation to be not super clear on how Registers and Qubits are handled

no

AbdullahKazi500:
the issue brought up by the users on Slack regarding how to retrieve the index of a qubit within a QuantumCircuit object. The original documentation lacked clarity on this point, which led to confusion among users. The updated version explains how to extract the index of a qubit from a Qubit object using both QuantumCircuit.find_bit() and QuantumCircuit.qubits.index(), ensuring that users can easily understand how to manage qubits and registers in their circuits.
fixes #1469

### DIFF
--- a/docs/build/decompose.ipynb
+++ b/docs/build/decompose.ipynb
@@ -1,0 +1,89 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "bc54d7bc-b2d8-4b30-9b0b-05689e07a463",
+   "metadata": {},
+   "source": [
+    "# Extract Quantum Register Information from Qiskit Quantum Circuit Data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c50d8e43-ae82-4e41-8d17-a37332d1bf6d",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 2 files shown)


docs/build/decompose.ipynb
```diff
@@ -0,0 +1,89 @@
...
+    "    print(\"Name of gate:\", name_of_gate)\n",
+    "    \n",
+    "    # Method 1: Using find_bit to get the index\n",
+    "    qarg_indices_find_bit = [qc.find_bit(qarg)[0] for qarg in qargs]\n",
+    "    print(\"qargs indices (find_bit):\", qarg_indices_find_bit)\n",
+    "    \n",
+    "    # Method 2: Using qubits.index to get the index\n",
+    "    qarg_indices_index = [qc.qubits.index(qarg) for qarg in qargs]\n",
+    "    print(\"qargs indices (index):\", qarg_indices_index)\n",
+    "    print(\"\\n\")"
```
mtreinish, line 64: We should **not** recommend this, it will have significant performance overhead which scales quadratically with the number of qubit it's almost never the right thing to do. The `find_bit()` pattern was added specifically to have a constant time lookup for this and should be the only thing we recommend.

mtreinish (changes requested):

frankharkins:
I'm closing this PR as this issue is already being worked on by another contributor.

PR #1590, opened 2024-06-25
Use new Docker image for previews

Eric-Arellano:
We removed the remote refresh mechanism from the previews, so port 5001 is no longer relevant. 

We also can now use the Dockerhub image for both `./start` and IBM Code Engine PR previews, so we can be consistent and stop using the IBM Container Registry version. That simplifies our infra.


arnaucasau (approved):
Nice, looks good! Thank you Eric

PR #1594, opened 2024-06-26
Update patterns redirects file

arnaucasau:
The PR updates the `scripts/patterns-reorg/redirects.json` file to a more convenient format that aligns better with our middleware. Having the `/` at the beginning of the old paths and the `/guides/` prefix in the new paths will allow us to have a more generic redirect file that could be used in the future for different URLs than patterns, and it will help us to set the redirects with a simpler logic.

e.g., `url.pathname = REDIRECTS[url.pathname]`

The PR also adds a missing redirect for `/start/latest-updates`.


scripts/patterns-reorg/redirects.json
```diff
@@ -1,80 +1,81 @@
...
-  "build/index": "map-problem-to-circuits",
-  "build": "map-problem-to-circuits",
-  "run/index": "execute-on-hardware",
-  "run": "execute-on-hardware",
-  "start/index": "",
-  "start": ""
+  "/start/install": "/guides/install-qiskit",
+  "/start/setup-channel": "/guides/setup-channel",
+  "/start/hello-world": "/guides/hello-world",
+  "/start/latest-updates": "/guides/latest-updates",
```
arnaucasau, line 5: This is the only new redirect. The rest are the same, but adding the slash in the old path and the `/guides/` prefix in the redirect.

arnaucasau, line 5: This is the only new redirect. The rest are the same, but adding the slash in the old path and the `/guides/` prefix in the redirect.

Eric-Arellano (approved):

PR #1605, opened 2024-06-27
Fix `update_internal_links` script

arnaucasau:
This PR fixes the `update_internal_links.py` script that was broken because we changed the `redirects.json` file to have a different style.

### DIFF
--- a/scripts/patterns-reorg/update_internal_links.py
+++ b/scripts/patterns-reorg/update_internal_links.py
@@ -52,10 +52,11 @@ def update_link(markdown: str, folder: str, link: str, prefix: str) -> str:
         else:
             search_key = f"{folder}/{link_split[-1]}"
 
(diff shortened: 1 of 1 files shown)


scripts/patterns-reorg/update_internal_links.py
```diff
@@ -52,10 +52,11 @@ def update_link(markdown: str, folder: str, link: str, prefix: str) -> str:
         else:
             search_key = f"{folder}/{link_split[-1]}"
 
+    search_key = f"/{search_key.removeprefix('/')}"
```
arnaucasau, line 55: I used the `removeprefix('/')` because we have links like `/verify` or `verify/index`. Removing the slash to all links allows us to add it back to all the links safely without having to use a conditional for the former case which would end up being `//verify` otherwise.

Eric-Arellano (approved):

PR #1626, opened 2024-07-01
Tweak redirects for runtime options pages

Eric-Arellano:
Prep for https://github.com/Qiskit/documentation/pull/1499. 

* `/configure-runtime-compilation` will become `/configure-error-suppression`. 
* `/advanced-runtime-options` is being deleted and replaced with `/runtime-options-overview`, which isn't necessarily the same content. Still, it costs us nothing to do a redirect; it avoids a 404 for users and keeps our SEO we've built up for the page.

I didn't update the page titles yet, which will be done in https://github.com/Qiskit/documentation/pull/1499. We only need to get the URLs right here because those are hard to change.


beckykd (approved):
Thanks!

PR #1647, opened 2024-07-02
Don't use "system" with docs reorg

Eric-Arellano:
We'll update the redirect rules in closed source.

The page content will be updated in a follow up. For now, we need to get the URLs and left ToC right to unblock the migration.


docs/guides/_toc.json
```diff
@@ -368,27 +368,27 @@
           ]
         },
         {
-          "title": "IBM Quantum systems",
+          "title": "IBM Quantum computers",
```
javabster, line 371:
```suggestion
          "title": "IBM quantum computers",
```

javabster (approved):
ah I guess we'll need to set up redirects for the urls that are changing too?

Eric-Arellano:
> ah I guess we'll need to set up redirects for the urls that are changing too?

Yep, but we will simply update the redirects we already were setting up for the old `run/` folder

PR #1653, opened 2024-07-04
Complete guides reorg To-Do

arnaucasau:
This PR finishes with all the To-Do we had from the guides reorg


Eric-Arellano (approved):
Thanks!

PR #1655, opened 2024-07-04
Remove README entry about Composer component

Eric-Arellano:
This component is unused and being removed.


arnaucasau (approved):
Nice, thank you!

PR #1668, opened 2024-07-09
Fix formatting in new QAOA tutorial

frankharkins:
Fixes a number of formatting problems in the new tutorial that became apparent after previewing on staging. I've uploaded the this fixed notebook to staging. To view, go to `<staging-url>/tutorial/quantum-approximate-optimization-algorithm`.

## Changes

* **Wrap math in `$$` and remove `align`:** The learning platform requires `$$` to recognize math. I believe Jupyter notebooks are more forgiving which is why this probably didn't cause any problems in review. I also removed the `align` environment because it's for aligning lines of equations and we only have one line at a time.
* **Removed HTML:** The learning platform doesn't support HTML, it either gets stripped or escaped. I've replaced any HTML with the closest standard markdown.
* **Replaced top-level headings:** All top-level headings are stripped by the learning platform and replaced with the title from the CMS. In general, it's best practice to only have one top-level heading anyway.
* Ran linter on the notebook.


Eric-Arellano (approved):
Thank you!

PR #1676, opened 2024-07-10
Extract more images from QAOA notebook

frankharkins:
Bit embarassed to be making a third PR. Not sure how I missed this, but there were still two attached images in the QAOA notebook, which are not supported by the learning platform.


Eric-Arellano (approved):
Thanks!

PR #1683, opened 2024-07-10
Add PR preview GitHub infrastructure

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/1043.

This deploys a new folder at https://qiskit.github.io/documentation/pr-{x}. For now, it's a contrived `index.html` only to get the technique working. We will switch to the full built preview app once the Docker image is accessible.

The GitHub workflow follows this logic:

1. Build the static site
2. Deploy by pushing the files to the `gh-pages` branch under the folder `pr-{x}`
3. Check if the deploy has worked by polling https://qiskit.github.io/documentation/pr-{x}, with a one minute timeout.

A follow up will add the cleanup mechanism.

### DIFF
--- a/.github/workflows/pr-preview.yml
+++ b/.github/workflows/pr-preview.yml
@@ -19,7 +19,7 @@ on:
 
 jobs:
   pr-preview:
-    name: PR preview cleanup
+    name: PR preview
     runs-on: ubuntu-latest
     if: github.event.pull_request.head.repo.full_name == github.repository
     env:
@@ -42,6 +42,8 @@ jobs:
           ref: gh-pages
       - name: Deploy to GitHub Pages
         run: |
+          git config user.name github-actions[bot]
+          git config user.email github-actions[bot]@users.noreply.github.com
           git add .
           git commit -m "Deploy PR preview for PR ${{ github.event.pull_request.number }}"
           git push
--- a/.github/workflows/pr-preview.yml
+++ b/.github/workflows/pr-preview.yml
@@ -1,0 +1,47 @@
+# This code is a Qiskit project.
+#
+# (C) Copyright IBM 2024.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+name: PR preview
+on: 
+  pull_request:
+    # paths:
+    #   - "docs/**/*"
+    #   - "public/images/**/*"
+
+jobs:
+  pr-preview:
+    name: PR preview cleanup
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.head.repo.full_name == github.repository
+    env:
+      PR_PREVIEW_PATH: pr-${{ github.event.pull_request.number }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 18
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+  
+      - name: Build static site
+        run: scripts/pr_previews/builder.py ${{ env.PR_PREVIEW_PATH }} --proof-of-concept
+
+      - name: Switch to gh-pages branch
+        uses: actions/checkout@v4
+        with:
+          ref: gh-pages
+      - name: Deploy to GitHub Pages
+        run: |
+          git add .
+          git commit -m "Deploy PR preview for PR ${{ github.event.pull_request.number }}"
+          git push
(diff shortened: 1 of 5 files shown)


.github/workflows/pr-preview.yml
```diff
@@ -0,0 +1,66 @@
...
+# obtain a copy of this license in the LICENSE file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+name: PR preview
+on:
+  pull_request:
```
Eric-Arellano, line 15: In https://github.com/Qiskit/documentation/pull/1595, we specified which PR events to run this on. That's because we cared about doing clean up very eagerly, as soon as the PR was closed. With this new GH Pages approach, we only do cleanup on pushes to `main`, that it will see if there are any stale folders in `gh-pages` branch. It's not a very big deal to have an extra folder hanging around for a bit.

.github/workflows/pr-preview.yml
```diff
@@ -0,0 +1,66 @@
...
+      PR_PREVIEW_URL: https://qiskit.github.io/documentation/pr-${{ github.event.pull_request.number }}/
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 18
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
```
Eric-Arellano, line 36: We'll need to set up IBM Container Registry once that is set up by infra.

.github/workflows/pr-preview.yml
```diff
@@ -0,0 +1,66 @@
...
+        uses: bobheadxi/deployments@v1
+        id: deployment
+        with:
+          step: start
+          token: ${{ secrets.GITHUB_TOKEN }}
+          env: pr_preview
+          ref: ${{ github.event.pull_request.head.ref }}
+
+      - name: Build static site
+        run: scripts/pr_previews/builder.py ${{ env.PR_PREVIEW_PATH }} --proof-of-concept
```
Eric-Arellano, line 47: We'll remove `--proof-of-concept` once safe to do so.

Eric-Arellano (commented):
You can see an example deployment button in the PR overview, which links to https://qiskit.github.io/documentation/pr-1683/. Example commit triggered by this workflow: https://github.com/Qiskit/documentation/commit/28cb071cebc6c8637f452522982d588962478b91

I tested these edge cases:
* first time the PR builds a preview
* preview already exists for that PR
     * no changes made
     * changes made
* deployment issues (we time out after 1 minute of trying)

.github/workflows/pr-preview.yml
```diff
@@ -0,0 +1,66 @@
...
+
+      - name: Deploy to GitHub Pages
+        run: scripts/pr_previews/deploy.py ${{ env.PR_PREVIEW_PATH }}
+      - name: Switch back to original branch
+        uses: actions/checkout@v4
```
frankharkins, line 52: When do we change branches? And is it necessary to change back?

scripts/pr_previews/deploy.py
```diff
@@ -0,0 +1,70 @@
...
+        run_subprocess(["git", "add", "."])
+        run_subprocess(["git", "commit", "-m", f"Deploy PR preview for {folder}"])
+        run_subprocess(["git", "push"])
+        logger.info("Pushed updates to gh-pages branch")
+    else:
+        logger.info("No changed files detected, so no push made to gh-pages")
+
+    logger.warning(
+        "The branch is set to gh-pages. You probably want to `git switch` back to your original branch"
+    )
```
frankharkins, line 65: Can we just do `git switch -` here?

scripts/pr_previews/poll_deployment.py
```diff
@@ -0,0 +1,72 @@
...
+    parsed_url = urlparse(url)
+    conn = http.client.HTTPSConnection(parsed_url.netloc, timeout=5)
+    try:
+        conn.request("HEAD", parsed_url.path)
+        return conn.getresponse().status
+    except Exception as e:
+        logger.error(f"Error checking server: {e}")
+        return -1
+    finally:
+        conn.close()
```
frankharkins, line 67: Could this be replaced with `requests.get(url).status_code`?

frankharkins (approved):
This seems like a good approach and the code is very clean, thanks!

scripts/pr_previews/deploy.py
Eric-Arellano: I tried, but we need to determine what the original branch name is. Using `git rev-parse --abbrev-ref HEAD` was returning `HEAD`, whereas locally it returns the branch name for me. I think this is because of how GH Actions is set up.

I considered instead using an env var set up by GH Actions to get the original branch name. But it seemed simpler to use the checkout action.

scripts/pr_previews/poll_deployment.py
Eric-Arellano: It could, but we would need to install requests. This uses the standard library

PR #1697, opened 2024-07-12
Clean up PR previews

frankharkins:
Adds a script and action to delete PR previews for closed PRs. See 3a29f15ec9aedc97574bf0d6cf282162efa4b966 for an example.


scripts/pr_previews/utils.py
```diff
@@ -28,6 +29,21 @@ def configure_logging() -> None:
...
+        ["git", "config", "user.email", "github-actions[bot]@users.noreply.github.com"]
+    )
+
+
+@contextmanager
+def switch_branch(branchname: str) -> None:
+    run_subprocess(["git", "fetch", "origin", branchname])
+    run_subprocess(["git", "switch", branchname])
+    yield
+    run_subprocess(["git", "checkout", "-"])
```
Eric-Arellano, line 44: This fails in GitHub actions. You could still keep this function but have the logging after the yield

https://github.com/Qiskit/documentation/actions/runs/9900538685/job/27351493566#step:7:17

Eric-Arellano (commented):
This looks great!

scripts/pr_previews/utils.py
frankharkins: Are you sure? It looks like the job succeeds: https://github.com/Qiskit/documentation/actions/runs/9909483488/job/27377633039

It did fail when I used `switch` though.

scripts/pr_previews/cleanup.py
```diff
@@ -0,0 +1,60 @@
...
+from utils import configure_logging, run_subprocess, setup_git_account, switch_branch
+
+
+def main() -> None:
+    setup_git_account()
+
+    with switch_branch("gh-pages"):
+        delete_closed_pr_folders()
+
+        changed_files = run_subprocess(["git", "status", "--porcelain"]).stdout.strip()
```
Eric-Arellano, line 31: This could be a helper function

scripts/pr_previews/cleanup.py
```diff
@@ -0,0 +1,60 @@
...
+        if not changed_files:
+            logger.info("No changed files detected, so no push made to gh-pages")
+            return
+
+        run_subprocess(["git", "add", "."])
+        run_subprocess(["git", "commit", "-m", f"Clean up PR previews"])
+        run_subprocess(["git", "push"])
```
Eric-Arellano, line 38: This could be a helper function, with `-m` as the arg

Eric-Arellano (approved):
Good find with `git checkout` vs `git switch`! Could you please fix the `raise SystemExit()` in `run_subprocess to be `raise SystemExit(1)`? I realized we haven't been setting the exit code correctly, which messes up GH Actions.

frankharkins:
Good call with the exit code, I wondered why the action continued

scripts/pr_previews/utils.py
```diff
@@ -28,6 +29,31 @@ def configure_logging() -> None:
...
 
 
+def setup_git_account() -> None:
+    run_subprocess(["git", "config", "user.name", "github-actions[bot]"])
+    run_subprocess(
+        ["git", "config", "user.email", "github-actions[bot]@users.noreply.github.com"]
+    )
+
+
+def changed_files() -> None:
```
Eric-Arellano, line 39:
```suggestion
def changed_files() -> str:
```

scripts/pr_previews/utils.py
```diff
@@ -28,6 +29,31 @@ def configure_logging() -> None:
...
+
+
+def commit_all_and_push(commit_message: str) -> None:
+    run_subprocess(["git", "add", "."])
+    run_subprocess(["git", "commit", "-m", commit_message])
+    run_subprocess(["git", "push"])
+
+
+@contextmanager
+def switch_branch(branchname: str) -> None:
```
Eric-Arellano, line 50: I think you need to import from `typing`
```suggestion
def switch_branch(branchname: str) -> Iterator[None]:
```

Eric-Arellano (approved):

scripts/pr_previews/cleanup.py
```diff
@@ -0,0 +1,64 @@
...
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+import json
+import logging
+import shutil
+from pathlib import Path
```
frankharkins, line 18:
```suggestion
from pathlib import Path
from typing import Iterator
```

PR #1726, opened 2024-07-17
Report page render times

frankharkins:
I'm using this for testing but figured it might be helpful to report in our weekly checks too.

### DIFF
--- a/scripts/commands/checkPagesRender.ts
+++ b/scripts/commands/checkPagesRender.ts
@@ -15,6 +15,7 @@ import fs from "fs/promises";
 import { globby } from "globby";
 import yargs from "yargs/yargs";
 import { hideBin } from "yargs/helpers";
+import { mean } from "lodash";
 
 import { zxMain } from "../lib/zx";
 import { Pkg } from "../lib/api/Pkg";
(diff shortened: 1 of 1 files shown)


scripts/commands/checkPagesRender.ts
```diff
@@ -83,17 +83,27 @@ zxMain(async () => {
...
+    } else {
       console.error(`❌ Failed to render: ${fp}`);
       failures.push(fp);
     }
 
     // This script can be slow, so log progress every 10 files.
     if (numFilesChecked % 10 == 0) {
-      console.log(`Checked ${numFilesChecked} / ${files.length} pages`);
+      const averageTime = (
+        renderTimes.reduce((a, b) => a + b, 0) / renderTimes.length
```
Eric-Arellano, line 101: Lodash has a `mean()` function https://www.geeksforgeeks.org/lodash-_-mean-method/

scripts/commands/checkPagesRender.ts
```diff
@@ -83,17 +83,27 @@ zxMain(async () => {
...
+      ).toFixed(0);
+      console.log(
+        `Checked ${numFilesChecked} / ${files.length} pages (~${averageTime}ms per page)`,
```
Eric-Arellano, line 104: Thoughts on instead computing the average for the batch of 10 files? That could help us to understand if certain files are acting up, like if we have outliers. This was indeed the problem with the static site building.

Eric-Arellano (commented):
Great idea!

scripts/commands/checkPagesRender.ts
frankharkins: That's what I wanted it to do, I reset `renderTimes` in the L106

Eric-Arellano (approved):

PR #1730, opened 2024-07-18
Improve types in page check

frankharkins:
This really doesn't matter much but it bugged me that I had to use `!` on line 93. I think this is better.


Eric-Arellano (approved):
Nice, good improvement

PR #1740, opened 2024-07-18
Fix categorization of dev dependencies

Eric-Arellano:
We're going to start having some projects install the scripts from this project as a GitHub VCS requirement. They won't install any of our dev deps, so we need to fix our categorization.

This only recategorizes and sorts deps. It does not change any version numbers.


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "docs",
+  "name": "qiskit-documentation",
```
Eric-Arellano, line 2: This is the name people will use in `dependencies` in their `package.json`, so it should not be so vague.

frankharkins (approved):

PR #1745, opened 2024-07-22
Refactor tests to prepare for Playwright

Eric-Arellano:
To unblock packaging our scripts for the Qiskit Functions repos, I plan to migrate us to Playwright. I was having issues with getting Jest to work with TypeScript and ES Modules (vs CommonJS modules), whereas Playwright works out-of-the-box.

Playwright does not have `toMatchInlineSnapshot`, and `toMatch` has different semantics; so, this switches to `toEqual`. It also removes unnecessary `describe()` blocks.


frankharkins (approved):
Nice, thanks!

PR #1746, opened 2024-07-22
Remove `getRoot()` helper function

Eric-Arellano:
`getRoot()` isn't going to work well when we package `scripts/js` for the Qiskit Functions repositories. One particular challenge is that in this repo, we run scripts from `scripts/js/commands/`—but the Functions repos will run from `node_modules/qiskit-documentation/dist/commands`—so the relative path we were using cannot work for both situations.

Another issue is that Playwright does not define `__dirname` because it runs with ES Modules rather than CommonJS.

Turns out, we don't even need to use `getRoot()`. The relative paths will be relative to the cwd, which should be the repo root anyways.


frankharkins (approved):
Nice! Thanks

PR #1747, opened 2024-07-22
Allow running `scripts/js` from other Node projects

Eric-Arellano:
This PR allows other repositories to depend on this repo as a VCS requirement, then to run our scripts, like this:

```json
  "scripts": {
    "check": "node node_modules/qiskit-documentation/dist/commands/checkMetadata.js"
  },
  "dependencies": {
    "qiskit-documentation": "qiskit/documentation#3422358984f9ede3a6ac8a2bbe7df517fbd306cb"
  },
```

```
❯ npm run check                       

> my-function-docs@0.1.0 check:metadata
> node node_modules/qiskit-documentation/dist/commands/checkMetadata.js
```

`node_modules/qiskit-documentation` will include only the contents of `src/js` but compiled to JavaScript, along with `package.json`. That results in the other project bringing in all of the transitive dependencies of this repo.

Running scripts locally still works.

## Migration to ES Modules

Before, we were using CommonJS for how imports and module loading works. However, this was causing issues when trying to run in other repos. The easiest fix was to migrate to ES Modules by setting `"type": "module"` in `package.json`.

This required that we set a `.js` file extension in all of our import statements, and that we use `lodash-es` rather than `lodash`.

## Migration to Playwright

I could not get Jest to play nicely with TypeScript & ES Modules. The simplest fix was to migrate to Playwright. 

We already use Playwright in qiskit-sphinx-theme, so this brings nice consistency to the open source projects.

## Followup work: fix individual scripts

Some of the scripts may need changes to work in other repositories, like our internal link checker. For now, `checkMetadata.ts` and `checkExternalLinks.ts` definitely work.


package.json
```diff
@@ -4,24 +4,29 @@
   "description": "The documentation content home for https://docs.quantum.ibm.com.",
   "author": "Qiskit Development Team",
   "license": "Apache-2.0",
+  "type": "module",
```
Eric-Arellano, line 7: This switches us to ES Modules. The default is CommonJS

package.json
```diff
@@ -4,24 +4,29 @@
...
+  "files": [
+    "dist"
+  ],
```
Eric-Arellano, line 10: This tells NPM to only include `dist/` in our package and not other files like `docs/`

package.json
```diff
@@ -4,24 +4,29 @@
...
   "scripts": {
+    "prepare": "tsc",
```
Eric-Arellano, line 12: `prepare` is a special script. VCS requirements run `prepare` after Git cloning the repo.

tsconfig.json
```diff
@@ -4,12 +4,13 @@
     "module": "ESNext",
     "allowJs": false,
     "strict": true,
-    "noEmit": true,
     "esModuleInterop": true,
     "moduleResolution": "node",
+    "declaration": true,
```
Eric-Arellano, line 9: This generates TypeScript files so that other repos have our types.

package.json
frankharkins: Where is `dist/`?

frankharkins: I see, it's generated by `tsconfig.json`?

frankharkins (approved):
LGTM! Thanks for the explanation.

package.json
Eric-Arellano: Ah yes, exactly. We don't check it in to version control intentionally because we would have to manually regenerate the folder every time we make a PR.

PR #1748, opened 2024-07-22
Move config files to scripts/config

Eric-Arellano:
This is meant to better split config files vs source code, to make it more obvious what should be changed by everyday script users.

Not all of the scripts use `config/` yet, but could be refactored in the future if we want.


Eric-Arellano:
@frankharkins this ended up not being necessary to DRY our spelling script. So it's only an aesthetic preference. Feel free to disagree with the change :)

package.json
```diff
@@ -10,7 +10,7 @@
     "typecheck": "tsc",
     "check": "npm run check:qiskit-bot && npm run check:patterns-index && npm run check:metadata && npm run check:spelling && npm run check:internal-links && npm run check:orphan-pages && npm run check:fmt",
     "check:metadata": "node -r esbuild-register scripts/js/commands/checkMetadata.ts",
-    "check:spelling": "cspell --relative --no-progress docs/**/*.md* docs/api/**/*.md* --config cspell/cSpell.json",
```
Eric-Arellano, line 13: `--relative` seems to be the default: https://cspell.org/docs/getting-started/#command-lint--spell-checking. Note that search only shows `--no-relative`, rather than `--relative`.

We now always use `.mdx`, so it's more obvious to configure that.

frankharkins (approved):
Thanks, this make it clearer what the different kinds of files are at a glance.

Issue #1720, opened 2024-07-16
Pages rendering check started timing out - have local previews gotten slower?


Eric-Arellano:
Checking that every page renders has been timing out: https://github.com/Qiskit/documentation/issues/1662 https://github.com/Qiskit/documentation/issues/1617 https://github.com/Qiskit/documentation/issues/1706

Note Frank's comment:

> The "pages render" job timed out, taking over 6hrs. It seems to be working correctly, just extremely slow. The last few times this ran, the "pages render" jobs only took ~30 mins, so this seems to be a regression.

Anecdotally, `./start` seems unusually slow.

## Testing idea: bisect if there was a regression

Locally, try running `./start` and timing how long it takes to load `/api/migration-guides` (i.e. a small folder). Change the image in `./start` to a local copy of the dynamic Docker image from closed source, where you are trying different builds of that Dockerfile to see if we introduced a regression in closed source.

Maybe the migration to a universal base image is relevant? That's the first commit I would bisect. If that is a problem, see https://sharp.pixelplumbing.com/install#linux-memory-allocator as a possible culprit.

## Profiling

You could also try profiling Next.js to see what is taking so long. I was profiling slow static site builds with `NEXT_TURBOPACK_TRACING=1 NEXT_CPU_PROF=1 yarn build`. Here, you could try setting the env vars in `./start` and see if it generates a `trace` file. If it doesn't, you may need to modify the Dockerfile in closed source. It may also be easier to use `NEXT_TURBOPACK_TRACING=1 NEXT_CPU_PROF=1 yarn dev` with the preview app in closed source and skip the Dockerfile. 

You then need to analyze the trace. You can set up a new npm project locally with this code from Next.js's scripts:

<details><summary>package.json</summary>

```json
{
  "name": "next-trace",
  "version": "1.0.0",
  "main": "index.js",
  "scripts": {
    "start": "node trace-to-tree.mjs"
  },
  "author": "",
  "license": "ISC",
  "description": "",
  "dependencies": {
    "event-stream": "^4.0.1"
  }
}
```
</details>

<details><summary>picocolors.mjs</summary>

```javascript
// ISC License

// Copyright (c) 2021 Alexey Raspopov, Kostiantyn Denysov, Anton Verinov

// Permission to use, copy, modify, and/or distribute this software for any
// purpose with or without fee is hereby granted, provided that the above
// copyright notice and this permission notice appear in all copies.

[...]

PR #1750, opened 2024-07-23
Allow skipping API docs in local previews

frankharkins:
The preview app is slow due because the it searches through the docs folder each time we load a page. Eric had the idea to improve the speed by removing API docs from the image. Since the API docs are so large, this roughly halves the time taken per page load. Writers can ask to include API docs using the `--apis` option.

Helps with https://github.com/Qiskit/documentation/issues/1720.

### DIFF
--- a/start
+++ b/start
@@ -35,7 +35,7 @@ def skip_apis() -> tuple[str]:
     """Mounts an empty directory to /docs/api/ to effectively exclude it"""
     if "--apis" in sys.argv:
         return ()
-    print("Skipping API docs for speed, use --apis to include them")
+    print("Skipping API docs for speed; use --apis to include them")
     return ("-v", "/home/node/app/docs/api")
 
 def main() -> None:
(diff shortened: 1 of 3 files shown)


start
```diff
@@ -31,6 +31,12 @@ def translation_volume_mounts() -> Iterator[str]:
             continue
         yield from ["-v", f"{path}:/home/node/app/docs/{path.name}"]
 
+def skip_apis() -> tuple[str]:
+    """Mounts an empty directory to /docs/api/ to effectively exclude it"""
+    if "--apis" in sys.argv:
+        return ()
+    print("Skipping API docs for speed, use --apis to include them")
```
Eric-Arellano, line 38:
```suggestion
    print("Skipping API docs for speed; use --apis to include them")
```

start
```diff
@@ -44,6 +50,7 @@ def main() -> None:
         "run",
         "-v",
         f"{PWD}/docs:/home/node/app/docs",
+        *skip_apis(),
         *translation_volume_mounts(),
```
Eric-Arellano, line 54: Can you please remove this while at it?

Eric-Arellano (approved):
Nice work!

Eric-Arellano:
Oh, the GitHub actions will need to specify `--api` - I think only the extended checks

Issue #1741, opened 2024-07-18
Spellcheck Jupyter notebooks


Eric-Arellano:
Not great that we skip those files...

https://github.com/Qiskit/documentation/blob/150ff528df126b75a4c420e6e1f09fd4ab173a1c/package.json#L10

To fix this, we could use our file `markdownReader.ts` to extract out markdown from Jupyter notebooks, and spellcheck from there.

This means that we would run the spellchecker from our custom scripts/commands file rather than directly invoking cspell. That will probably be helpful anyways as part of the effort to allow reusing our scripts in other repositories.

PR #1752, opened 2024-07-23
Spellcheck Jupyter notebooks & allow running checker from other repos

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/1741. We use the cspell library to check Jupyter notebooks:

```
❯ npm run check:spelling

> qiskit-documentation@0.1.0 check:spelling
> tsx scripts/js/commands/checkSpelling.ts

$ npx cspell --no-progress --config scripts/config/cspell/cSpell.json docs/**/*.mdx !docs/api/**/*.mdx docs/api/migration-guides/*.mdx
CSpell: Files checked: 53, Issues found: 0 in 0 files
❌ Unrecognized word in docs/guides/hello-world.ipynb: typoooooo
```

There is now an `--apis` flag, although there are currently failures.

--

Other repositories can also now run cspell with this `script`:

```
"check:spelling": "node node_modules/qiskit-documentation/dist/commands/checkSpelling.js --config node_modules/qiskit-documentation/scripts/config/cspell/cspell.json"
```

### DIFF
--- a/scripts/js/commands/checkSpelling.ts
+++ b/scripts/js/commands/checkSpelling.ts
@@ -10,12 +10,16 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
-import { zxMain } from "../lib/zx";
 import { $ } from "zx";
 import yargs from "yargs/yargs";
 import { hideBin } from "yargs/helpers";
+import { spellCheckDocument } from "cspell-lib";
+import { globby } from "globby";
+import pMap from "p-map";
 
-import { Pkg } from "../lib/api/Pkg";
+import { zxMain } from "../lib/zx.js";
+import { Pkg } from "../lib/api/Pkg.js";
+import { readMarkdown } from "../lib/markdownReader.js";
 
 interface Arguments {
   [x: string]: unknown;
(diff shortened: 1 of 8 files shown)


scripts/js/commands/checkSpelling.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  if (args.apis) {
+    const apiFolders = Pkg.VALID_NAMES.map((api) => `docs/api/${api}/**/*.mdx`);
+    await $`${cspellCmd} ${apiFolders}`;
+  }
+
+  await checkAllNotebooks(args.config);
+});
+
+async function checkAllNotebooks(configPath: string): Promise<void> {
```
Eric-Arellano, line 60: Thoughts appreciated on how to make the logging more useful. Please take a look at the PR description for example output

I don't know if we want to unconditionally log that we're checking n Jupyter notebooks? Might be noisy.

Two ideas I just thought of:

* Add an emoji warning for each failure
* If it's the first failure encountered, use `\n\n` so it stands out from the logs above it.

frankharkins: I think a "❌" emoji usually works well

scripts/js/commands/checkSpelling.ts
```diff
@@ -0,0 +1,101 @@
...
+  const paths = await globby("docs/**/*.ipynb");
+  let allGood = true;
+  await pMap(paths, async (path) => {
+    const errors = await checkForNotebookMistakes(path, configPath);
+    if (errors.length) {
+      allGood = false;
+      const deduplicated = new Set(errors);
+      deduplicated.forEach((mistake) =>
+        console.error(`Spelling mistake in ${path}: ${mistake}`),
```
frankharkins, line 69:
```suggestion
        console.error(`Unrecognized word in ${path}: ${mistake}`),
```
Nit to make it clearer to users that we're matching against a dictionary and they can add new words if necessary.

scripts/js/commands/checkSpelling.ts
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Returns any spelling mistakes in the notebook's markdown blocks, if any.
+ */
+async function checkForNotebookMistakes(
+  fp: string,
+  configPath: string,
+): Promise<string[]> {
+  // We should only check the markdown blocks, not code blocks.
+  const text = await readMarkdown(fp);
```
frankharkins, line 87: Don't block on this, but isn't cspell able to check code too?

frankharkins (approved):
Nice! Very easy to read

Issue #1756, opened 2024-07-24
Spellcheck: check code blocks?


Eric-Arellano:
Should we remove these ignores?

https://github.com/Qiskit/documentation/blob/3ac8a1d103619ba965fbc92825c56e4614df8b0e/scripts/config/cspell/cSpell.json#L33-L40

frankharkins:
I'd leave the `$$` which is for latex, the rest I think can be removed.

PR #1758, opened 2024-07-24
Spellcheck code blocks

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/1756. While this sometimes results in annoying false positives, we decided it was worth it to us.

For notebooks, we only check the source code we write, not the output generated.


docs/guides/run-jobs-session.mdx
```diff
@@ -65,15 +65,19 @@ The context manager automatically opens and closes the session.
 <Tabs>
   <TabItem value="PrimV2" label="V2 primitives">
 ``` python
+from qiskit_ibm_runtime import Session, Samplerv2 as Sampler, Estimatorv2 as Estimator
```
Eric-Arellano, line 68: The original config I was using didn't like `estimatorv2`. I added a generic rule so that any word with a number is fine, since it's probably code. So, we can go back to `estimatorv2` if we want. 

But I figured it's better to switch to the same style we use elsewhere.

scripts/config/cspell/dictionaries/qiskit.txt
```diff
@@ -158,11 +158,21 @@ mathrm
...
+misassignment
+toric
+gphase
+cobyla
+nelder
+stds
+accum
+optim
+synthes
+depolar
```
Eric-Arellano, line 178: These ones are word. We never use the shorthand, it's always the full word like `optimizing`. But cspell is cutting off the suffix and complaining about `optim`. Adding the full word `optimizing` to the dictionary didn't fix it.

scripts/js/commands/checkSpelling.ts
```diff
@@ -83,15 +83,14 @@ async function checkForNotebookMistakes(
   fp: string,
```
Eric-Arellano, line 83: These changes and below should be reviewed by @frankharkins

abbycross (approved):
Thanks!

scripts/config/cspell/dictionaries/qiskit.txt
frankharkins: Weird design choice by cspell, thanks for the explanation though.

frankharkins (approved):
Nice! Looks like it's caught a few good ones already

PR #1759, opened 2024-07-24
Add internal link checker for other repositories

Eric-Arellano:
The Functions repositories need to check their internal links are valid. 

To keep things clean, I added a new command file specifically for other repos, rather than modifying `checkInternalLinks.ts`.

The main challenge is that they won't have access to which other docs pages exist, like `/guides/hello-world`, since those docs won't live in the repository. To solve this, we save and distribute a snapshot of the current state of links in `scripts/config/internal-links.json`. This snapshot does risk falling out of date, which means we will occasionally want to regenerate it. I went with bundling the file—rather than something like fetching it from GitHub—to keep things simple (e.g. no network flakes) and to keep the build deterministic.

Example:

```markdown
# Page in another repo

[Another page](/guides/my-function)     (this is a file in the other repo)
[Another page](/guides/hello-world)
[Another page](/guides/hello-world2)
```

```
❯ npm run check:internal-links

> my-function-docs@0.1.0 check:internal-links
> node node_modules/qiskit-documentation/dist/commands/checkInternalLinksFromOtherRepo.js



Checking internal links for docs
❌ Could not find link '/guides/hello-world2'. Appears in:
    docs/guides/index.mdx    ❓ Did you mean '/guides/hello-world'?

Some internal links appear broken 💔
```


frankharkins (approved):
Looks good!

PR #1762, opened 2024-07-25
Migrate migration guides to not have `/api` in URL

Eric-Arellano:
We'll set up redirects in closed source.

Motivations:

* Shorter URL for users
* Simpler config. We had to do weird special-casing to treat `/api/migration-guides` like guides but everything else in `/api` like API docs
* More clear analytics


Eric-Arellano:
Please do not hit the merge button! We need to coordinate merging this with changes in closed source.

frankharkins (approved):
Awesome!

PR #1763, opened 2024-07-25
Fix spelling in migration guides

Eric-Arellano:


scripts/config/cspell/cSpell.json
```diff
@@ -45,9 +45,7 @@
     "</?thead>",
     "</?iframe",
     // Method use, e.g. Class.foo()
-    "([a-zA-Z_]\\w*)\\.[a-zA-Z_]\\w*",
```
Eric-Arellano, line 48: The new regex now matches multiple periods

scripts/config/cspell/cSpell.json
```diff
@@ -45,9 +45,7 @@
...
-    // Observable labels like IIZII
-    "[XIZ]{3,6}",
```
Eric-Arellano, line 50: Turns out this was explaining the `accum` vs `accumulating` thing, somehow 🤷 Removing this fixes it

frankharkins: Turns out cspell is [case insensitive by default](https://cspell.org/docs/case-sensitive/) which is why it was matching "izing" words. No idea why it matches "accumulating" though 🤔 

Either way, I think better to explicitly ignore the observable labels per page.

docs/api/migration-guides/qiskit-opflow-module.mdx
```diff
@@ -3,6 +3,8 @@ title: qiskit.opflow migration guide
 description: How to update your code to stop using the deprecated `qiskit.opflow` module.
 ---
 
+{/* cspell:ignore XYZY, SLSQP */}
```
frankharkins, line 6:
```suggestion
{/* cspell:ignore XYZY */}
```

Nit: You ignore this in `cSpell.json`.

frankharkins (approved):
Nice! Impressed there was only one actual spelling correction

PR #1773, opened 2024-07-29
Revert CI workaround for setuptools

Eric-Arellano:
https://github.com/Qiskit/documentation/pull/1772 should no longer be necessary because setuptools 72 was yanked.


frankharkins (approved):
That was quick!

PR #1777, opened 2024-07-29
Add PEA to error techniques topic

beckykd:
Will get more information from @samantha Barron, but I'm adding some information here as a placeholder.


docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -120,6 +120,74 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "e5dcab7b",
+   "metadata": {},
+   "source": [
+    "## Probabilistic Error Amplification (PEA)\n",
+    "\n",
+    "PEA is an aplification technique used with [Zero-noise extrapolation(ZNE)](#zne).  The main challenge in successfully implmenting ZNE is to have an accurate model for noise in the expectation value and to amplify the noise in a known way.  For utility-scale experiments, PEA is often the best choice because it can be applied to any circuit that can be run with native noise factor $(\\lambda - 1)$. \n",
```
SamFerracin, line 130:
```suggestion
    "PEA is an amplification technique used with [Zero-noise extrapolation(ZNE)](#zne).  The main challenge in successfully implementing ZNE is to have an accurate model for noise in the expectation value and to amplify the noise in a known way.  For utility-scale experiments, PEA is often the best choice because it can be applied to any circuit that can be run with native noise factor $(\\lambda - 1)$. \n",
```

SamFerracin: Perhaps you could say something about ZNE, such as:

PEA is an amplification technique used with [Zero-noise extrapolation(ZNE)](#zne). Assuming that the "target" circuit of interest suffers noise of strength $\alpha$, ZNE requires implementing many copies of the target circuit, each one suffering a noise whose strength is artificially amplified to values higher than $\alpha$. Finally, the mitigated results are estimated by fitting the noisy data and extrapolating the value at noise strength $\alpha=0$.

One of the main challenges in ZNE is to accurately amplify the noise affecting the target circuit. Many techniques (such as "gate folding") provide easy ways of performing this amplification, but are potentially inaccurate and may lead to incorrect results. PEA is a more sophisticated technique and for utility-scale experiments, it is often the best choice.

docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -120,6 +120,74 @@
...
+    "\n",
+    "Because PEA is a ZNE amplification technique, you also need to enable ZNE by setting `resilience.zne_mitigation = True`. Other `resilience.zne` options can be used in addition to set extrapolators, amplification levels, and so on. PEA requries a noise model, which is automatically generated when using primitives. \n",
+    "\n",
+    "When the Estimator primitive is used to call `noise-learner`, PEA can reuse the returned noise model. **LINK TO NOISE LEARNER DOCS** "
```
SamFerracin, line 134: Here, you are not using the noise learner. I would simply phrase things as follows:

The following snippet provides an example where PEA is used to mitigate the result of an estimator job:
```
estimator = Estimator(mode=backend)
estimator.options.resilience.zne_mitigation = True
estimator.options.resilience.zne.amplifier = "pea"

estimator.run(pubs, precision=0.01)
```

docs/guides/error-mitigation-and-suppression-techniques.ipynb
SamFerracin: A few references:
- You can use [this](https://ieeexplore.ieee.org/document/9259940) one for ZNE and for gate folding.
- For "but are potentially inaccurate and may lead to incorrect results" you can use [this](https://www.nature.com/articles/s41567-022-01914-3#Sec7) (see [page 4 of supplementary material](https://arxiv.org/pdf/2108.09197) on the arxiv, where it says that "However, we caution that the method can provide noise amplification results that arbitrarily deviate from the desired result when only slightly more complicated noise models are considered")

docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -120,6 +120,74 @@
...
+   ]
```
SamFerracin, line 135: We are still not 100% sure if we'll enable reusing the noise learning data for follow-up experiments. If we do (and I will give you a head-up), you may want to add this after [this snippet](https://github.com/Qiskit/documentation/pull/1777/files#r1715951505):

When choosing to use PEA, the estimator automatically performs the noise reconstruction in preparation for the actual mitigation experiment. The results of this "noise learning" stage are reported in the metadata of the estimator's results. They can be re-used in subsequent experiments, potentially leading to significantly improved runtimes:

```
estimator = Estimator(mode=backend)
estimator.options.resilience.zne_mitigation = True 
estimator.options.resilience.zne.amplifier = "pea" 

# the first job performs the noise reconstruction in preparation to the actual estimation task
job1 = estimator.run(pubs, precision=0.01)

# the noise data is available in the results
noise_data = job1.result().metadata["resilience"]["layer_noise_model"]

# it can be re-used in subsequent estimation jobs to skip the noise-learning stage
estimator.options.resilience.layer_noise_model = noise_data
job2 = estimator.run(pubs, precision=0.01)
```
The noise learning stage can also be performed independently of the estimation task via the `NoiseLearner` class:
```
learner = NoiseLearner(backend)
learner_job = learner.run(pubs)  # `run` accepts pubs or a list of circuits
noise_data = job.result()

estimator = Estimator(mode=backend) 
estimator.options.resilience.zne_mitigation = True 
estimator.options.resilience.zne.amplifier = "pea"
estimator.options.resilience.layer_noise_model = noise_data
estimator_job = estimator.run(pubs, precision=0.01)

File: docs/guides/error-mitigation-and-suppression-techniques.ipynb (hunk 1 above)
Comment (beckykd) [reply docs/guides/error-mitigation-and-suppression-techniques.ipynb:130]: > * You can use [this](https://ieeexplore.ieee.org/document/9259940) one for ZNE and for gate folding

This requires a subscription.  Is there a free source?

Comment (SamFerracin) [reply docs/guides/error-mitigation-and-suppression-techniques.ipynb:130]: You can use the [arXiv url](https://arxiv.org/abs/2005.10921) if you prefer, but I'm not sure what the policy is regarding citations

File: docs/guides/error-mitigation-and-suppression-techniques.ipynb [hunk 4]
```diff
@@ -207,11 +208,142 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "e5dcab7b",
+   "metadata": {},
+   "source": [
+    "## Probabilistic Error Amplification (PEA)\n",
+    "\n",
+    "[PEA](https://www.nature.com/articles/s41586-023-06096-3) is a noise amplification technique used with [Zero-noise extrapolation (ZNE)](#zne) where the noise is amplified using knowledge of a [sparse Pauli-Lindblad noise model](https://www.nature.com/articles/s41567-023-02042-2), which needs to be learned by running additional circuits. Assuming that the \"target\" circuit of interest suffers noise of strength $\\alpha$, ZNE requires implementing many copies of the target circuit, each one suffering from noise whose strength is artificially amplified to values higher than $\\alpha$. Finally, the mitigated results are estimated by fitting the noisy data and extrapolating the value at noise strength $(\\alpha = 0)$. \n",
```
SamFerracin, line 218: I know that I asked you to add this paragraph, but I had not realized that the previous section is called "ZNE" and it explains all of this in more words :sweat_smile: I think that we can just get rid of this first paragraph, as it overlaps.

The second paragraph reads like a very natural continuation of the discussion. I would keep it as is, perhaps removing the reference and quotation marks for gate folding (it's already introduced in the "ZNE" section)

docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -207,11 +208,142 @@
...
+    "\n",
+    "One of the main challenges in ZNE is to accurately amplify the noise affecting the target circuit. Many techniques, such as \"gate folding\" (see [Digital zero noise extrapolation for quantum error mitigation](https://arxiv.org/abs/2005.10921)) provide easy ways to perform this amplification, but are potentially inaccurate and might lead to incorrect results.  See the article [\"Scalable error mitigation for noisy quantum circuits produces competitive expectation values\"](https://www.nature.com/articles/s41567-022-01914-3), and specifically page 4 of the [supplementary material](https://arxiv.org/pdf/2108.09197) for details. \n",
```
SamFerracin, line 220:
```suggestion
    "One of the main challenges in ZNE is to accurately amplify the noise affecting the target circuit. Gate folding provides an easy way to perform this amplification, but is potentially inaccurate and might lead to incorrect results.  See the article [\"Scalable error mitigation for noisy quantum circuits produces competitive expectation values\"](https://www.nature.com/articles/s41567-022-01914-3), and specifically page 4 of the [supplementary material](https://arxiv.org/pdf/2108.09197) for details. \n",
```

docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -207,11 +208,142 @@
...
+    "estimator_job = estimator.run(pubs, precision=0.01)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "68268837",
+   "metadata": {},
+   "source": [
+    "<Admonition type=\"note\"> \n",
+    "`EstimatorV2.run` invokes noise learning each time it is called.  Therefore, to ensure that you are using a consistent noise model, it is important to only call `run()` once. \n",
```
SamFerracin, line 305: This feels like out of place here. I would remove it, together with the examples of where the noise learning is run twice

docs/guides/error-mitigation-and-suppression-techniques.ipynb
beckykd: Well, @kevinsung suggested that I move it there, which is why you didn't notice the ZNE section before :)

docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -207,11 +208,98 @@
...
+   "source": [
+    "## Probabilistic Error Amplification (PEA)\n",
+    "\n",
+    "One of the main challenges in ZNE is to accurately amplify the noise affecting the target circuit. Gate folding provides an easy way to perform this amplification, but is potentially inaccurate and might lead to incorrect results.  See the article [\"Scalable error mitigation for noisy quantum circuits produces competitive expectation values\"](https://www.nature.com/articles/s41567-022-01914-3), and specifically page 4 of the [supplementary material](https://arxiv.org/pdf/2108.09197) for details. \n",
+    "\n",
+    "PEA is a more sophisticated technique that performs preliminary experiments to reconstruct the noise and then uses this information to perform an accurate amplification. For utility-scale experiments, it is often the best choice.\n",
+    "\n",
+    "Because PEA is a ZNE amplification technique, you also need to enable ZNE by setting `resilience.zne_mitigation = True`. Other `resilience.zne` options can be used in addition to set extrapolators, amplification levels, and so on. PEA requires a noise model, which is automatically generated when using primitives. \n",
```
kevinsung, line 222:
```suggestion
    "Because PEA is a ZNE noise amplification technique, you also need to enable ZNE by setting `resilience.zne_mitigation = True`. Other `resilience.zne` options can be used in addition to set extrapolators, amplification levels, and so on. PEA requires a noise model, which is automatically generated when using primitives. \n",
```

kevinsung (approved):

SamFerracin (approved):

docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -207,11 +208,98 @@
...
+   "execution_count": null,
+   "id": "4947f442",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "estimator = Estimator(mode=backend)\n",
+    "estimator.options.resilience.zne_mitigation = True\n",
+    "estimator.options.resilience.zne.amplifier = \"pea\"\n",
+    "\n",
+    "estimator.run(pubs, precision=0.01)"
```
SamFerracin, line 238:
```suggestion
```

docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -207,11 +208,98 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "aae27947",
+   "metadata": {},
+   "source": [
+    "When choosing to use PEA, Estimator automatically performs the noise reconstruction to prepare for the actual mitigation experiment. The results of this \"noise learning\" stage are reported in Estimator's result metadata. The results can be reused in subsequent experiments, potentially leading to significantly improved runtimes:"
```
SamFerracin, line 246: Remove the rest of this section (a stand-alone noise learner guide is coming)

jyu00:
I think it's worth calling out PEA on [this page](https://docs.quantum.ibm.com/guides/configure-error-mitigation#configure-error-mitigation) as well, even though it's technically ZNE (especially since the description about ZNE is focused on gate folding), i.e. this sentence should say + link to PEA:

```
Primitives support several error mitigation techniques, including TREX, ZNE, and PEC.
```

docs/guides/error-mitigation-and-suppression-techniques.ipynb
SamFerracin: I think it's good

abbycross:
> I think it's worth calling out PEA on [this page](https://docs.quantum.ibm.com/guides/configure-error-mitigation#configure-error-mitigation) as well, even though it's technically ZNE (especially since the description about ZNE is focused on gate folding), i.e. this sentence should say + link to PEA:
> 
> ```
> Primitives support several error mitigation techniques, including TREX, ZNE, and PEC.
> ```

Looks like I should use this link for PEA - @jyu00 confirm?: https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.options.ZneOptions

jyu00:
> OR - do you want to link to the new PEA section in the error techniques docs page instead?

@abbycross The new PEA section in the [techniques](https://docs.quantum.ibm.com/guides/error-mitigation-and-suppression-techniques) page, to match the others.

Issue #846, opened 2024-02-19
Guide for the backend primitives


beckykd:
### Describe the new content you are requesting.

@ElePT suggested creating a guide for the backend primitives.  This can probably wait until the V2 backend primitives are out. For now, the content has been pulled from https://docs.quantum.ibm.com/verify/simulate-with-qiskit-primitives into https://docs.quantum.ibm.com/run/primitives-get-started

### If this new content request is accepted, do you want to write the content?

no

javabster:
do we have enough content to fill a whole page for this? Or is the small section in the primitives page basically it?

beckykd:
Done

PR #1794, opened 2024-08-02
V2 additions to backend primitives section

ElePT:
This PR is a proposal to partially address https://github.com/Qiskit/documentation/issues/846. The issue was written pre-backend primitives V2, when there were more differences in the interface (*) between runtime and backend primitives, so proposing a separate guide seemed reasonable at the time. Now (V2), I think the differences are small enough to keep the backend primitives section as an appendix to the get-started page, where I just added the usual V1-V2 tabs and a section at the end explaining differences. 

This is just a rough proposal and it has been a while since I've written a guide, so feel free to edit it to fit the style/conventions that are used now.

---

(*) Backend primitives handled different kinds of pass managers to run before and after parameter binding, but V2 primitives no longer do any transpilation, so the way you run and configure backend and runtime primitives is almost the same. I didn't add the V1 explanation for the interface differences because V1 primitives are deprecated since qiskit 1.2 (released yesterday) and I think it would clutter the page and confuse users to add an explanation of how they worked now.

### DIFF
--- a/docs/guides/get-started-with-primitives.mdx
+++ b/docs/guides/get-started-with-primitives.mdx
@@ -337,7 +337,7 @@ backend primitives are designed to run locally in the user's machine.
 
 - The [`qiskit.primitives.BackendSamplerV2`](/api/qiskit/qiskit.primitives.BackendSamplerV2) class requires a backend that supports the `memory` option.
 
-- The backend primitives V2 interfaces expose custom `Sampler` and `Estimator` `Options` for that are different to the Runtime implementations.
+- The backend primitives V2 interfaces expose custom [`Sampler`](/api/qiskit/qiskit.primitives.BackendSamplerV2) and [`Estimator`](/api/qiskit/qiskit.primitives.BackendEstimatorV2) `Options` that are different from the Runtime implementations.
 
 
 ## Next steps
--- a/docs/guides/get-started-with-primitives.mdx
+++ b/docs/guides/get-started-with-primitives.mdx
@@ -264,13 +264,32 @@ print(f"  > Metadata: {result.metadata[0]}")
 <span id="backend"></span>
 ## Get started with the backend primitives
 
-The `Sampler` primitive can be run with any provider by using [`qiskit.primitives.BackendSampler`](/api/qiskit/qiskit.primitives.BackendSampler). Likewise, the `Estimator` primitive can be run with any provider using [`qiskit.primitives.BackendEstimator`](../api/qiskit/qiskit.primitives.BackendEstimator). Equivalent implementations in V2 are coming soon.
+Unlike provider-specific primitives, backend primitives are generic implementations that can be used with an arbitrary 
+`backend` object, as long as it implements the [`Backend`](/api/qiskit/qiskit.providers.Backend) interface.
+
+The `Sampler` primitive can be run with any provider by using [`qiskit.primitives.BackendSamplerV2`](/api/qiskit/qiskit.primitives.BackendSamplerV2)
+or the legacy V1 implementation, [`qiskit.primitives.BackendSampler`](/api/qiskit/qiskit.primitives.BackendSampler). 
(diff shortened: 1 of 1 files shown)


docs/guides/get-started-with-primitives.mdx
```diff
@@ -264,31 +264,79 @@ print(f"  > Metadata: {result.metadata[0]}")
...
 backend = provider.get_backend('backend_name')
 sampler = BackendSampler(backend)
 ```
+  </TabItem>
+
+### Similarities and differences between backend and Runtime primitives (V2)
+
+- The inputs to [`qiskit.primitives.BackendSamplerV2`](/api/qiskit/qiskit.primitives.BackendSamplerV2) and [`qiskit.primitives.BackendEstimatorV2`](../api/qiskit/qiskit.primitives.BackendEstimatorV2)
+follow the same PUB format as the primitives in Qiskit Runtime, so do the outputs, see (link to https://docs.quantum.ibm.com/guides/primitive-input-output). 
+However, there can be differences in the fields of the returned metadata.
```
ElePT, line 331: I decided not to go into further detail because the metadata isn't a fixed part of the interface and the content has been changing quite a lot, so listing specific fields I think increases the maintenance load of the page.

docs/guides/get-started-with-primitives.mdx
```diff
@@ -264,31 +264,79 @@ print(f"  > Metadata: {result.metadata[0]}")
...
+
+- The [`qiskit.primitives.BackendEstimatorV2`](/api/qiskit/qiskit.primitives.BackendEstimatorV2) class offers no measurement or gate error mitigation implementations out-of-the-box, as
+backend primitives are designed to run locally in the user's machine. 
+
+- The [`qiskit.primitives.BackendSamplerV2`](/api/qiskit/qiskit.primitives.BackendSamplerV2) class requires a backend that supports the `memory` option.
+
+- The backend primitives V2 interfaces expose custom `Sampler` and `Estimator` `Options` for that are different to the Runtime implementations.
```
ElePT, line 338: I wanted to link to the specific option classes but I see that they are not listed in the docs, which might be by design, not sure. I am referring to:

- BackendEstimatorV2 options class: https://github.com/Qiskit/qiskit/blob/main/qiskit/primitives/backend_estimator_v2.py#L41-L57
- BackendSamplerV2 options class: https://github.com/Qiskit/qiskit/blob/main/qiskit/primitives/backend_sampler_v2.py#L43-L54

Getting into details about the options in this page seems to be a bit too much (as it is a get-started-with page) that's why I thought linking to the API docs would be a good solution. I guess we could also omit this item.

beckykd: That seems like an error.  Investigating.

docs/guides/get-started-with-primitives.mdx
beckykd: Seems reasonable.

docs/guides/get-started-with-primitives.mdx
beckykd: I tracked this down.  Those classes are not public, so all we can do is point them to the BackendEstimatorV2 class docs and the BackendSamplerV2 class docs

Eric-Arellano: It might be unintentional that they are not documented, if someone forgot to list them in `qiskit/primitives/__init__.py` docstring. Consider asking in the Qiskit dev Slack what should be done

beckykd: > It might be unintentional that they are not documented, if someone forgot to list them in `qiskit/primitives/__init__.py` docstring. Consider asking in the Qiskit dev Slack what should be done

It's intentional.  They're not documented because they are not public.  I was told that it would take a little more work to make them public because the name would be overloaded.  It's under consideration for the next release.  For now, we just have to link to the docs I mentioned in my previous comment.

beckykd (approved):

javabster (approved):
LGTM 🚀

PR #1800, opened 2024-08-05
[wip] Check images for alt text

### LINKED ISSUE #1651: Accessibility: new check for image alt text

shraddha-aangiras:
Implemented to ensure all images have some alt text
Fixes #1651


shraddha-aangiras:
@Eric-Arellano please review this when you're free, thanks!

README.md
```diff
@@ -284,6 +284,20 @@ npm run check:metadata -- --apis
 npm run check
 ```
 
+## Check image alt text
+
+Every image needs to have `alt text` for accessibility. The `lint` job in CI will fail if images do not have alt text.
```
Eric-Arellano, line 289: Nit, "alt text" isn't code so no need to escape it.

```suggestion
Every image needs to have alt text for accessibility. The `lint` job in CI will fail if images do not have alt text.
```

README.md
```diff
@@ -284,6 +284,20 @@ npm run check:metadata -- --apis
...
+
+You can also check for valid metadata locally:
+
+```bash
+# Only check file metadata
```
Eric-Arellano, line 294:
```suggestion
# Only check alt text
```

README.md
```diff
@@ -284,6 +284,20 @@ npm run check:metadata -- --apis
...
+npm run check:alt-text
+
+# Or, run all the checks. Although this only checks non-API docs.
```
Eric-Arellano, line 297:
```suggestion
# Or, run all the checks
```

scripts/js/commands/checkAltText.ts
```diff
@@ -0,0 +1,97 @@
...
+
+interface Arguments {
+    [x: string]: unknown;
+  }
+
+const readArgs = (): Arguments => {
+    return yargs(hideBin(process.argv))
+      .version(false)
+      .parseSync();
+  };
```
Eric-Arellano, line 28: No need for this. This program doesn't need to take arguments

scripts/js/commands/checkAltText.ts
```diff
@@ -0,0 +1,97 @@
...
+
+async function determineFiles(args: Arguments): Promise<[string[], string[]]> {
+  const mdGlobs = ["docs/guides/*.mdx", "docs/migration-guides/*.mdx"];
+  const notebookGlobs = ["docs/guides/*.ipynb", "docs/migration-guides/*.ipynb"];
+  return [await globby(mdGlobs), await globby(notebookGlobs)];
```
Eric-Arellano, line 33: No need to differentiate between MDX and ipynb. We do that for checkMetadata because you set up the metadata very differently there, but we don't do that for alt text.

Also, we want to check all folders other than `docs/api`. You can do something like:

```
return await globby(["docs/**/*.{ipynb,mdx}", "!docs/api/**/*.mdx"])
```

At that point, this function isn't really "pulling its weight". It would be cleaner to inline the code, meaning you get rid of the helper function and directly have `const files = await globby(["docs/**/*.{ipynb,mdx}", "!docs/api/**/*.mdx"])` in your main function

scripts/js/lib/links/extractAltText.test.ts
```diff
@@ -0,0 +1,36 @@
...
+//
+// This code is licensed under the Apache License, Version 2.0. You may
+// obtain a copy of this license in the LICENSE file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+import { expect, test } from "@jest/globals";
```
Eric-Arellano, line 13: We don't use Jest anymore. We've now migrated to Playwright. You can look at some other test files for inspiration.

scripts/js/lib/links/extractAltText.ts
```diff
@@ -0,0 +1,49 @@
...
+ 
+export interface AltText {
+  imageName: Set<string>; 
+  altText: Set<string | null | undefined>;
+};
+
+
+export async function extractAlt(
+  markdown: string,
+): Promise<AltText> {
```
Eric-Arellano, line 29: You can simply this functionality by making it more focused. All we care about is finding which images are missing alt text. We don't need to keep track of which images do have alt text. Note that if you are missing alt text, then `altText` would be empty. So, all we need is to return a `string[]` for the image names we're missing. You could rename the function something like `findImagesWithoutAltText`

scripts/js/commands/checkAltText.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+
+function getNullAlt(
```
Eric-Arellano, line 37: You won't need this function if you take my recommendation in `extractAltText.ts`

Eric-Arellano (commented):
Thanks for working on this!

Eric-Arellano:
Superseded by https://github.com/Qiskit/documentation/pull/2349. Thanks for kickstarting this!

PR #1809, opened 2024-08-08
V1 is no longer supported

### LINKED ISSUE #1802: Upcoming removal of V1 primitive support
### URL, if applicable

_No response_

We plan to drop V1 primitives support on Aug 15. Various pages will need to be cleaned up to remove V1 reference. 

beckykd:
closes #1802


beckykd:
With V1 going away, we could change `from qiskit_ibm_runtime import EstimatorV2 as Estimator`  to `from qiskit_ibm_runtime import Estimator as Estimator` ? But @jyu00 and I discussed this and decided that it would be too confusing for anyone who isn't actually using the most current version of Runtime (because they would get the v1 Estimator instead of v2).  We can revisit changing the code in a while.

Eric-Arellano:
> from qiskit_ibm_runtime import Estimator as Estimator

FYI this would be simply `from qiskit_ibm_runtime import Estimator` since you're not changing the name. I agree it's a good idea to wait 2-3 months at least until making this simplification.

docs/guides/primitives.mdx
```diff
@@ -109,9 +77,9 @@ estimator.run([(circuit1, observable1, param_values1),(circuit2, observable2, pa
 ```
 
 <span id="broadcast-rules"></span>
-##### Broadcasting rules
+### Broadcasting rules
```
jyu00, line 80: Do we need broadcasting rules here given they are repeated in the `Primitive inputs and outputs` page?

docs/guides/specify-runtime-options.mdx
```diff
@@ -8,16 +8,12 @@ description: Specify options when building with Qiskit Runtime primitives
...
 You can use options to customize the Estimator and Sampler primitives. This section focuses on how to specify Qiskit Runtime primitive options. While the interface of the primitives' `run()`  method is common across all implementations, their options are not. Consult the corresponding API references for information about the [`qiskit.primitives`](/api/qiskit/primitives#primitives) and [`qiskit_aer.primitives`](https://qiskit.github.io/qiskit-aer/apidocs/aer_primitives.html) options.
 
-Options are specified differently depending on whether you are using the V1 or V2 interface.
-
-## V2 primitives
-
-Options are specified differently in the V2 primitives in these ways:
+Notes about specifying options in the primitives:
 
 - `SamplerV2` and `EstimatorV2` have separate options classes. You can see the available options and update option values during or after primitive initialization.
```
jyu00, line 13: Do these need to be `Sampler` and `Estimator`?

docs/guides/specify-runtime-options.mdx
```diff
@@ -333,26 +218,14 @@ estimator = Estimator(mode=backend)
...
 
 You can turn off all error mitigation and suppression if you are, for example, doing research on your own mitigation techniques. To accomplish this, for EstimatorV2, set `resilience_level = 0`. For SamplerV2, no changes are necessary because no error mitigation or suppression options are enabled by default.
 
 Example:
 
-Turn off all error mitigation and suppression in EstimatorV2.
+Turn off all error mitigation and suppression in Estimator.
 
 ```python
 from qiskit_ibm_runtime import EstimatorV2 as Estimator, Options, QiskitRuntimeService
```
jyu00, line 231: This should not import `Options` since it's only used for V1 and will be removed.

docs/guides/run-jobs-session.mdx
```diff
@@ -72,19 +55,6 @@ with Session(backend=backend):
...
-
-backend = service.least_busy(operational=True, simulator=False)
-with Session(backend=backend):
-  estimator = Estimator()
-  sampler = Sampler()
-```
-  </TabItem>
-</Tabs>
 
 When you start a session,  you must specify a QPU. This can be done by specifying its name or by passing a `backend` object.
```
jyu00, line 59: Specifying a QPU name is deprecated (the code snippet also doesn't do that). I thought this was fixed in the other PR 🤔

docs/guides/simulate-with-qiskit-sdk-primitives.mdx
```diff
@@ -15,13 +15,8 @@ The following sections show how to use the reference primitives to run your work
 
 ## Use the reference `Estimator`
```
jyu00, line 16: This does not render well, at least not on my browser: 

image

jyu00 (commented):
The pages look much less cluttered now! 

Side question, when can we remove the ISA warning? Given we imposed that almost 6 months ago.

docs/guides/run-jobs-session.mdx
beckykd: I thought so too - I'll try merging main into this PR again

docs/guides/specify-runtime-options.mdx
beckykd: > Do these need to be `Sampler` and `Estimator` (without `V2`)?

I thought these were the class names and needed to keep the V2.

docs/guides/simulate-with-qiskit-sdk-primitives.mdx
beckykd: This is fixed in prod, but the issue has not yet been fixed in previews.  It's safe to ignore.

docs/guides/specify-runtime-options.mdx
Eric-Arellano: I think we were keeping the V2 for a bit longer to avoid confusion with people using old versions of qiskit-ibm-runtime

beckykd:
> Side question, when can we remove the ISA warning? Given we imposed that almost 6 months ago.

I'm asking about this.  I'm ready to delete it whenever.

docs/guides/specify-runtime-options.mdx
jyu00: I'm asking because I see both `Sampler`/`SamplerV2` and `Estimator`/`EstimatorV2` being used. For example

image

docs/guides/primitives.mdx
beckykd: No!  good catch!

docs/guides/specify-runtime-options.mdx
beckykd: I hopefully addressed this in my latest commit:
- Only use code when referring to actual class names
- The class names are xxxV2.
- There are still some OK uses of the old class names in the migration guides.  Hopefully I didn't get overzealous and remove any.

docs/migration-guides/qiskit-algorithms-module.mdx
```diff
@@ -352,7 +352,7 @@ extends [`qiskit.algorithms.minimum_eigensolvers.SamplingVQE`](/api/qiskit/0.46/
 <Admonition type="note">
   If you want to run QAOA on a non-diagonal operator, use the [`qiskit.circuit.library.QAOAAnsatz`](/api/qiskit/qiskit.circuit.library.QAOAAnsatz) with
   [`qiskit.algorithms.minimum_eigensolvers.VQE`](/api/qiskit/0.46/qiskit.algorithms.minimum_eigensolvers.VQE), but there will be no state result.
-  If your application requires the final probability distribution, instantiate a `Sampler`
+  If your application requires the final probability distribution, instantiate a Sampler
```
beckykd, line 355:
```suggestion
  If your application requires the final probability distribution, instantiate `SamplerV2`
```

docs/migration-guides/qiskit-runtime-examples.mdx
```diff
@@ -262,9 +262,9 @@ expectation:  -1.065734058826613
 The Sampler primitive is used to design an algorithm that samples
 circuits and extracts probability distributions.
 
-Both `Sampler` and `backend.run()` take in circuits as inputs. The main
+Both `SamplerV2` and `backend.run()` take in circuits as inputs. The main
```
beckykd, line 265:
```suggestion
Both `Sampler` and `backend.run()` take in circuits as inputs. The main
```

docs/migration-guides/qiskit-runtime-examples.mdx
```diff
@@ -262,9 +262,9 @@ expectation:  -1.065734058826613
...
 difference is the format of the output: `backend.run()` outputs
-**counts**, while `Sampler` returns **per-shot measurements** (but has convenience methods to also return counts).
+**counts**, while `SamplerV2` returns **per-shot measurements** (but has convenience methods to also return counts).
```
beckykd, line 267:
```suggestion
**counts**, while `Sampler` returns **per-shot measurements** (but has convenience methods to also return counts).
```

docs/guides/primitives.mdx
```diff
@@ -60,163 +46,35 @@ For hardware providers, implementing primitives natively means you can provide y
...
-```
-
-#### Sampler V2
-
-* Takes one parameter: PUBs in the format  (`<single circuit>`, `<optional one or more parameter values>`, `<optional shots>`), where there can be multiple `parameter values` items, and each item can be either an array or a single parameter, depending on the chosen circuit.
-* Elements from each are aggregated. For example, each array of parameter values in the PUB is applied to the PUB's circuit.
+* Elements from each PUB are aggregated. For example, each array of parameter values in the PUB is applied to the PUB's circuit.
 * Obeys program outputs. Typically this is a bit array but can also be an array of complex numbers (measurement level 1).
 * Returns raw data type. Data from each shot is returned (analogous to `memory=True` in the `backend.run` interface), and post-processing is done by using convenience methods.
 * Output data is grouped by output registers. You need the classical register name to get the results. By default, it is named `meas` if you use `measure_all()`. You can find the classical register name by running `<circuit_name>.cregs`. For example, `qc.cregs`.
```
jyu00, line 80: This list was originally written as a comparison between V1 and V2. It's a bit odd to have it here. In fact I don't think we need them anymore since the inputs/outputs [page](https://docs.quantum.ibm.com/guides/primitive-input-output) already covers them.

jyu00 (approved):
One more comment but otherwise LGTM!

Eric-Arellano (approved):
I did not closely review. I'm deferring to Jessie's review, in case @abbycross you wanted to double check

PR #1824, opened 2024-08-14
Generate qiskit-transpiler-service 0.4.5

Eric-Arellano:
This means that we now have our first historical version of qiskit-transpiler-service: 0.3. So, we need to update `checkInternalLinks.ts` to start checking historical versions. 

We don't yet add a dev version because it's annoying to maintain and the users are too few for now.


docs/api/qiskit-transpiler-service/0.3/_package.json
```diff
@@ -0,0 +1,4 @@
+{
```
Eric-Arellano, line 1: This 0.3 folder is simply moving the files we had before. You can skip over these files

scripts/config/api-html-artifacts.json
```diff
@@ -52,6 +52,7 @@
     "0.14": "https://ibm.box.com/shared/static/t37e7jjsi0hii4j3xoorwpwso5m03jqn.zip"
   },
   "qiskit-transpiler-service": {
+    "0.4": "https://ibm.box.com/shared/static/9y66qk3nim6oejzq8phpctxkt0j6o19z.zip",
```
Eric-Arellano, line 55: This was human-written

scripts/config/historical-pages-to-latest.json
```diff
@@ -18556,5 +18556,7 @@
     "0.8": {},
     "0.9": {}
   },
-  "qiskit-transpiler-service": {}
+  "qiskit-transpiler-service": {
+    "0.3": {}
+  }
```
Eric-Arellano, line 18561: Auto-generated

scripts/js/commands/checkInternalLinks.ts
```diff
@@ -106,6 +106,7 @@ const RUNTIME_GLOBS_TO_LOAD = [
   "docs/guides/*.{mdx,ipynb}",
```
Eric-Arellano, line 106: This file should be closely reviewed. It's necessary because this is our first time having historical versions for qiskit-transpiler-service

kaelynj (approved):
Looks good to me! :rocket:

Issue #1826, opened 2024-08-14
Address Serverless 0.15 requiring Python 3.11


Eric-Arellano:
Options:

1. Ask the team to restore Python 3.9 and 3.10 coverage
2. Give up on us testing Python 3.9, which means we risk using invalid syntax that will break some users.

While addressing this, consider having changes to requirements.txt prompt runs of the notebook execution job. That would have caught this problem.

PR #1831, opened 2024-08-14
Restore Serverless 0.15.2 and test notebooks when changing requirements

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/1826.

## Switches to Python 3.11

Qiskit Serverless requires Python 3.11+ now, even though Qiskit SDK still supports 3.9 and 3.10. While I'm talking to the Serverless team, it seems very unlikely they are going to change their mind on this due to their own constraints.

Unfortunately, this means that we need to always use Python 3.11 with our repository so that we can install serverless for the files that use it. Our other alternative is to stop testing the files with Serverless, but that seems even worse.

The risk with using 3.11 in CI rather than 3.9 is that we may use syntax not in Python 3.9 and 3.10. However, given the actual history of our docs, this seems unlikely: we don't use new Python syntax like type hints. If we do use Python 3.11-syntax, it's not the end of the world.

## Tests all notebooks on changes to requirements.txt

This was an oversight that we would not test the notebooks when changing requirements.txt. So, we merged a change that worked locally but broke CI. Now, we test all notebooks when the nb-tester program and config changes.

## Also upgrades Runtime

This is necessary to make Serverless happy. According to https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/release-notes#0270-2024-08-08, there are no breaking changes. CI also shows everything passes.

### DIFF
--- a/.github/workflows/notebook-test.yml
+++ b/.github/workflows/notebook-test.yml
@@ -67,7 +67,7 @@ jobs:
         with:
           # Install Linux deps if the specific guides were changed, or
           # if all files are being tested.
-          install-linux-deps: ${{ steps.linux-changed-files.outputs.any_changed || !steps.all-changed-files.outputs.any_changed }}
+          install-linux-deps: ${{ steps.linux-changed-files.outputs.any_changed == 'true' || steps.all-changed-files.outputs.any_changed == 'false' }}
           ibm-quantum-token: ${{ secrets.IBM_QUANTUM_TEST_TOKEN }}
 
       - name: Check lint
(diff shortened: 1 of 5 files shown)


.github/workflows/notebook-test.yml
```diff
@@ -91,12 +94,22 @@ jobs:
...
+              args.extend(files.split("\n"))
           subprocess.run(args, check=True)
 
+      - name: Detect changed notebooks
+        id: changed-notebooks
+        if: "!cancelled()"
+        run: |
+          echo "CHANGED_NOTEBOOKS<<EOF" >> $GITHUB_OUTPUT
+          git diff --name-only >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
```
Eric-Arellano, line 108: This step comes from our cron job

kaelynj (approved):
Everything here makes sense to me (and feels like the best option we have in terms of being able to keep CI working when including serverless docs).

PR #1833, opened 2024-08-15
Fix artifact upload when no changes made

Eric-Arellano:
The artifact upload was failing because `path` was set to the empty string. It's fine to give an invalid file path like `no-changes`, so long as we give something. The action will warn but not fail in this case.


abbycross (approved):
👍

PR #1837, opened 2024-08-15
Aug 15 updates to Qiskit Runtime

jyu00:


abbycross:
@jyu00 We had tentatively been thinking to leave the last three months of updates on the page at a time. Maybe that feels like it makes the page too long? Happy to stick with only the latest Qiskit Runtime version update on the page, if you prefer. It will be time to create the archive page for the older Runtime updates if so, which I will put together.

Eric-Arellano (approved):

PR #1840, opened 2024-08-16
Upgrade testing deps to Qiskit 1.2 and Runtime 0.28

Eric-Arellano:


kaelynj:
Fixed the warning in the `circuit-library` notebook, but it looks like the warning from the `local-testing-mode` notebook is caused by a bug within `qiskit-aer`. I've added an `ignore-warnings` tag to the affected cell so it should pass CI for now and I'll open a separate issue to track it.

kaelynj (approved):
Changes look good to me now and CI is green :100: :green_circle:

PR #1848, opened 2024-08-16
Fix images in Qiskit release notes for historical versions

Eric-Arellano:
Before, we never copied images for the release notes file to the historical API folder. For example, if we had the file `public/images/qiskit/release_notes_0_0.png`, we would not copy it over to `public/images/qiskit/1.1/release_notes_0_0.png` when we convert 1.1 from the latest release to instead be a historical release stored at `qiskit/1.1`. 

That is the right behavior for Runtime and Provider, because they only have a single `release_notes.mdx` file. However, it does not work correctly with Qiskit, which has one release note file per release.

This PR differentiates between the two use cases. It keeps things generic by using our `Pkg.hasSeparateReleaseNotes` mechanism.

### DIFF
--- a/scripts/js/commands/api/convertApiDocsToHistorical.ts
+++ b/scripts/js/commands/api/convertApiDocsToHistorical.ts
@@ -44,22 +44,19 @@ const readArgs = (): Arguments => {
 zxMain(async () => {
   const args = readArgs();
 
-  const pkgName = Pkg.VALID_NAMES.find((pkgName) => pkgName === args.package);
-  if (pkgName === undefined) {
-    throw new Error(`Unrecognized package: ${args.package}`);
-  }
+  const pkg = await Pkg.fromArgs(args.package, "ignored", "ignored", "latest");
 
-  const packageFile = await readFile(`docs/api/${pkgName}/_package.json`, {
+  const packageFile = await readFile(`docs/api/${pkg.name}/_package.json`, {
     encoding: "utf8",
   });
   const packageInfo = JSON.parse(packageFile);
   const versionMatch = packageInfo.version.match(/^(\d+\.\d+)/);
   const versionWithoutPatch = versionMatch[0];
 
-  const projectNewHistoricalFolder = `docs/api/${pkgName}/${versionWithoutPatch}`;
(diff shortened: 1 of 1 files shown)


scripts/js/commands/api/convertApiDocsToHistorical.ts
```diff
@@ -143,12 +144,22 @@ async function generateJsonFiles(
...
   const imageDirDest = `public/images/api/${pkgName}/${versionWithoutPatch}`;
   await mkdirp(imageDirDest);
-  await $`find ${imageDirSource}/* -maxdepth 0 -type f | grep -v "release_notes" | xargs -I {} cp -a {} ${imageDirDest}`;
+
+  // If the project only has a single release notes file, we should not copy the release notes images.
+  if (hasSeparateReleaseNotes) {
+    await $`find ${imageDirSource}/* -maxdepth 0 -type f | xargs -I {} cp -a {} ${imageDirDest}`;
+  } else {
+    await $`find ${imageDirSource}/* -maxdepth 0 -type f | grep -v "release_notes" | xargs -I {} cp -a {} ${imageDirDest}`;
+  }
```
Eric-Arellano, line 162: I didn't try DRYing this more because I think it can make the escaping confusing with `zx`. Imo it's not that bad to have it duplicated, and we're at low risk of them falling out of sync since the two lines are so close together.

kaelynj: I think the input defs for `copyObjectsInv` should include the `hasSeparateReleaseNotes` var correct?

Eric-Arellano: Shouldn't be relevant for `objectsInv`. No matter what, we move the `objectsInv` file to be in the new historical folder at `public/api/<pkg>/<version>`. Doesn't matter whether release notes are a single page or one-per-release

kaelynj: Ope nvm I misread what these lines are replacing. This looks good to me!

kaelynj (approved):

PR #1881, opened 2024-08-27
Use consistent name for Qiskit Runtime Client

Eric-Arellano:
We were being inconsistent with calling Runtime "Qiskit Runtime IBM Client" but "Qiskit IBM Runtime REST API". We also don't use IBM with "Qiskit Transpiler Service Client" and "Qiskit SDK". So, `@jyu00` and `@javabster` proposed that we simplify to remove the redundant IBM.

This PR updates the left page titles we use. A closed source PR will update the top dropdown.

We don't update the index pages for Runtime API docs because those are auto-generated.


jyu00 (approved):
LGTM!

frankharkins (approved):

PR #1885, opened 2024-08-27
Fix misleading file paths in script error messages

Eric-Arellano:


beckykd (approved):
Super fast fix!  Thanks!

PR #1888, opened 2024-08-28
Run all notebooks when passed requirements file as argument

### LINKED ISSUE #1844: Dependency upgrades should always run every notebook, not only changed notebooks
https://github.com/Qiskit/documentation/pull/1831 now executes all notebooks when requirements are changed. That's a big improvement so that we know we aren't breaking notebooks unintentionally when upgrading dependencies.

However, there's one edge case that can be deceptive: as soon as you change one of the notebooks, we stop testing every notebook and instead test only the changed notebook. Imagine this scenario: 5 notebooks have issues with the upgrade, and the author fixes one of them. CI will stop running on those 4 notebooks, so they will look okay even though they are not.

If `scripts/nb-tester` has changed, we should _always_ run on every notebook.

frankharkins:
Fixes #1844

### DIFF
--- a/.github/workflows/notebook-test.yml
+++ b/.github/workflows/notebook-test.yml
@@ -50,8 +50,8 @@ jobs:
           files: "docs/**/*.ipynb"
           separator: "\n"
 
-      - name: Get changed dependency files
-        id: changed-dependencies
+      - name: Get changed config files
+        id: changed-config
         uses: tj-actions/changed-files@af2816c65436325c50621100d67f6e853cd1b0f1
         with:
           files: "scripts/nb-tester/**/*"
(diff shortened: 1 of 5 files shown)


frankharkins:
That's neater, thanks! Should be working now, see https://github.com/Qiskit/documentation/actions/runs/10600858185/job/29379326069?pr=1888

.github/workflows/notebook-test.yml
```diff
@@ -50,6 +50,12 @@ jobs:
           files: "docs/**/*.ipynb"
           separator: "\n"
 
+      - name: Get changed dependency files
```
Eric-Arellano, line 53: Nit: maybe call this "changed config"? We rerun the tests when changing any file in nb-tester, which I think is a good idea to ensure we don't break things.

Eric-Arellano (approved):
Huzzah, thanks!

Eric-Arellano (approved):

Issue #1896, opened 2024-08-30
Dark mode for Patterns diagram


Eric-Arellano:
This looks bad at https://docs.quantum.ibm.com/guides/intro-to-patterns

<img width="836" alt="Screenshot 2024-08-30 at 1 49 09 PM" src="https://github.com/user-attachments/assets/bd45671c-c9e8-427d-8c4e-0cc1e2530cde">

Ditto the four specific index pages we have.

gracelindsell:
@Eric-Arellano The illustrations have been adjusted so that they arent as start on dark mode. They can be found [here](https://www.figma.com/design/sNXucK1uMaZ8XXyLTZrdrs/IBM-Quantum-Documentation?node-id=7461-37577&t=G6FylF6fKlyoJgR6-1)

PR #1904, opened 2024-09-03
Use dark-mode friendly Patterns SVGs

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/1896. Light mode stays the same, but dark mode now looks like:
Screenshot 2024-09-03 at 3 45 52 PM

Screenshot 2024-09-03 at 3 45 41 PM


frankharkins (commented):
Do we support the `<picture>` element? If so, we might be able to keep the light version and switch to the dark version only in dark mode: https://github.blog/changelog/2022-05-19-specify-theme-context-for-images-in-markdown-beta/

Eric-Arellano:
>  If so, we might be able to keep the light version and switch to the dark version only in dark mode:

Light mode doesn't change with this PR :) 

Screenshot 2024-09-04 at 9 17 38 AM

frankharkins (approved):

abbycross (approved):

Eric-Arellano:
> How does that work? Transparency?

Grace set a bunch of values like this in the SVG in Figma:

```
fill-opacity="0.6"
```

PR #1906, opened 2024-09-04
Add contributors for each file

### LINKED ISSUE #1876: Implement contributor names on each guides page
Implement the design from https://www.figma.com/design/sNXucK1uMaZ8XXyLTZrdrs/IBM-Quantum-Documentation?node-id=7174-39884&t=GIeSutsuCJkUNVjg-1

Screenshot 2024-08-26 at 10 58 16 AM

Because the component is not in the footer, it should probably be a component like `<ContributorsAttribution>`. The other design I was considering is putting metadata at the top of the file.

## How to get usernames

We need to figure out how to store the contributor GitHub usernames/handles. It will be much easier if we save the value to the file in qiskit/documentation, rather than somehow magically pulling the value at build-time in closed source. It would be hard to do the latter because it would require dynamically pulling Git information, even though the closed source and open source repos are not meant to have tight coupling.

I suspect the best approach is to have a script that dynamically adds the usernames to the file. We probably need users to manually run it, unfortunately. CI would lint to check everything is up-to-date.

frankharkins:
This adds a list of contributors for each file so we can attribute them. This PR includes the list, plus a script to generate the list. The script is a bit slow (~20s) so I'm not sure when is best to run it, maybe in a cron job?

***

Closes #1876


scripts/js/commands/generateContributors.ts
```diff
@@ -0,0 +1,83 @@
...
+  return (await $(command, rest)).stdout;
+}
+
+async function getCommits(filepath: string): Promise<CommitHash[]> {
+  return (await run`git log --format='%h' --follow -- ${filepath}`)
+    .split("\n")
+    .filter((s) => s !== "");
+}
+
+async function getFileAuthors(filepath: string): Promise<string[]> {
```
frankharkins, line 43:
```suggestion
async function getFileAuthors(filepath: string): Promise<AuthorName[]> {
```

frankharkins:
This is gonna live in closed source instead

PR #1908, opened 2024-09-04
Change internal link checker's Qiskit SDK strategy

Eric-Arellano:
This is to unblock https://github.com/Qiskit/documentation/issues/495. Our goal is to fix all broken links for Qiskit 0.46+, whereas we don't plan to fix <0.46. So, we want our script to make it ergonomic to check 0.46+.

This PR results in us now checking 0.46, 1.0, and 1.1 docs. It does that with two changes:

1. Rename `--skip-broken-historical` to `--include-broken-historical` so that, by default, `--historical-apis` "just works".
1. `--qiskit-release-notes` is renamed to `--qiskit-legacy-release-notes`. 
    - It now only checks <0.46. 
    - `--current-apis` will check the release notes for the latest version, whereas before you had to remember to  specify `--qiskit-release-notes`
    - For Qiskit 0.46+, `--historical-apis` will check the release note for each version, whereas before you had to remember to specify `--qiskit-release-notes`

| Option                          | Checks                                                  |
|--------------------------------:|:--------------------------------------------------------|
| `--current-apis`                | Current API docs, including latest Qiskit release note |
| `--qiskit-legacy-release-notes` | Only Qiskit release notes < 0.46                        |
| `--historical-apis`             | All Runtime and Transpiler historical; Qiskit 0.46+, including those release notes        |
| `--historical-apis --include-broken-historical`             | All historical, including Qiskit historical release notes        |

### DIFF
--- a/scripts/js/commands/checkInternalLinks.ts
+++ b/scripts/js/commands/checkInternalLinks.ts
@@ -62,7 +62,7 @@ const readArgs = (): Arguments => {
       type: "boolean",
       default: false,
       description:
-        "Also check historical releases that are known to still fail (currently Qiskit <0.46). " +
+        "Also check historical releases that are known to still fail (currently all of Qiskit). " +
         "Intended to be used alongside `--historical-apis`.",
     })
     .option("qiskit-legacy-release-notes", {
@@ -146,9 +146,9 @@ async function determineFileBatches(args: Arguments): Promise<FileBatch[]> {
     {
       check: args.historicalApis,
       hasSeparateReleaseNotes: true,
-      // Qiskit docs are broken on <0.46.
-      skipVersions: (version) =>
-        !args.includeBrokenHistorical && +version < 0.46,
+      // Qiskit docs are broken for all historical versions. We should
+      // fix 0.46+ and only skip <0.46 after that.
+      skipVersions: () => !args.includeBrokenHistorical,
     },
   );
 
--- a/scripts/js/commands/checkInternalLinks.ts
+++ b/scripts/js/commands/checkInternalLinks.ts
@@ -34,8 +34,8 @@ interface Arguments {
   currentApis: boolean;
   devApis: boolean;
   historicalApis: boolean;
-  qiskitReleaseNotes: boolean;
-  skipBrokenHistorical: boolean;
+  qiskitLegacyReleaseNotes: boolean;
+  includeBrokenHistorical: boolean;
 }
 
 const readArgs = (): Arguments => {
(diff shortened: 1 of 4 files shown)


scripts/js/commands/checkInternalLinks.ts
```diff
@@ -178,42 +187,27 @@ async function determineCurrentDocsFileBatch(
...
 
-  if (!args.currentApis) {
-    toCheck.push(`!{public,docs}/api/*/*`);
-    toLoad.push(`docs/api/*/*`);
-  }
-
-  if (args.qiskitReleaseNotes) {
-    const currentVersion = JSON.parse(
+  if (args.currentApis) {
+    const currentQiskitVersion = JSON.parse(
```
Eric-Arellano, line 196: We now check the release notes file for the current version if `--current-apis` is set. Before, you had to also set `--qiskit-release-notes`

scripts/js/commands/checkInternalLinks.ts
```diff
@@ -238,41 +232,43 @@ async function determineDevFileBatches(): Promise<FileBatch[]> {
...
-      );
-    }
+    const toCheck: string[] = [
+      `docs/api/${projectName}/${folder.name}/*`,
+      `public/api/${projectName}/${folder.name}/objects.inv`,
+    ];
+    const toLoad = [...extraGlobsToLoad];
 
-    if (checkSeparateReleaseNotes && !isBeforeQiskit0_45) {
+    // Also check the relesae note file for this version, if the package has
```
Eric-Arellano, line 264:
```suggestion
    // Also check the release note file for this version, if the package has
```

scripts/js/commands/checkInternalLinks.ts
```diff
@@ -294,17 +290,15 @@ async function determineQiskitLegacyReleaseNotes(): Promise<FileBatch> {
     .map((releaseNotesPath) =>
       releaseNotesPath.split("/").pop()!.split(".").slice(0, -1).join("."),
     )
-    .filter(
-      (version) => +version < 1 && version != "0.45" && version != "0.46",
-    );
+    .filter((version) => +version < 1 && version != "0.46");
```
Eric-Arellano, line 293: This is inside determineQiskitLegacyReleaseNotes(). 0.46 is now checked as part of `--historical-apis`. I'm not sure why 0.45 wasn't being checked in this batch previously.

frankharkins:
Thanks! This seems to work well, but I'm a bit confused about the options. Is this correct?

| Option                          | Checks                                                  |
|--------------------------------:|:--------------------------------------------------------|
| `--current-apis`                | Current API docs, including latest Qiskit release notes |
| `--qiskit-legacy-release-notes` | Only Qiskit release notes < 0.46                        |
| `--historical-apis`             | All API docs, including all Qiskit release notes        |

If so, does it still make sense to check release notes separately? I think it would be easier to have just two options: `--current-apis` and `--historical-apis`, and have these include the current and historical release notes respectively.

Eric-Arellano:
> If so, does it still make sense to check release notes separately? I think it would be easier to have just two options: --current-apis and --historical-apis, and have these include the current and historical release notes respectively.

Your table is correct. The reason we still have `--qiskit-legacy-release-notes` is that Qiskit <0.46 is broken and we don't have plans to fix it. It is skipped unless you use `--include-broken-historical`, which we don't really plan to use.

If we fixed Qiskit <0.46, we could remove both `--include-broken-historical` and `--qiskit-legacy-release-notes`.

frankharkins (approved):

PR #1910, opened 2024-09-04
Eagerly log output of `regen-apis`

Eric-Arellano:
The script is quite slow, so it's better to log the output eagerly so it doesn't look like things hung.

### DIFF
--- a/scripts/js/commands/api/regenerateApiDocs.ts
+++ b/scripts/js/commands/api/regenerateApiDocs.ts
@@ -55,7 +55,6 @@ zxMain(async () => {
   const args = readArgs();
   await validateGitStatus();
 
-  const results = new Map<string, string[]>();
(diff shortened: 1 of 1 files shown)


scripts/js/commands/api/regenerateApiDocs.ts
```diff
@@ -139,13 +121,13 @@ async function regenerateVersion(
     await $`${command}`;
     if ((await gitStatus()) !== "") {
       await gitCommit(`Regenerate ${pkgName} ${version}`);
-      return `✅ ${pkgName} ${version} regenerated correctly`;
+      console.log(`🚀 ${pkgName} ${version} regenerated correctly`);
```
frankharkins, line 124: Nice use of emoji

frankharkins (approved):

PR #1911, opened 2024-09-04
Regen Runtime and Qiskit

Eric-Arellano:
There were some changes that we didn't properly set up by regenerating the API docs. This runs `npm run regen-apis`. 

The file docs/api/qiskit-ibm-runtime/0.23/options.mdx was manually changed in the uploaded Box artifact to avoid a regression where the link changed to point to current API docs for a page that was deleted.


docs/api/qiskit-ibm-runtime/0.23/options.mdx
```diff
@@ -45,7 +45,7 @@ estimator.options.update(resilience_level=1)
 Refer to [`SamplerOptions`](qiskit_ibm_runtime.options.SamplerOptions "qiskit_ibm_runtime.options.SamplerOptions") and [`EstimatorOptions`](qiskit_ibm_runtime.options.EstimatorOptions "qiskit_ibm_runtime.options.EstimatorOptions") for V2 Sampler and V2 Estimator options, respectively.
 
 <Admonition title="Note" type="note">
-  If an option is not specified, the server default value is used. The default values are subject to change. See the [API Reference](/api/qiskit-ibm-runtime/0.23/options) for the most up-to-date default values.
+  If an option is not specified, the server default value is used. The default values are subject to change.
```
Eric-Arellano, line 48: This link is meant to go to the current page itself, so I removed the link because it doesn't really make sense.

public/api/qiskit-ibm-runtime/0.14/objects.inv
Eric-Arellano, line 1: I think this is a whitespace change or something. I couldn't really tell through the local diff what changed. Note that only 18 bytes changed, and the file still is 56KB.

frankharkins: If you're using https://github.com/Qiskit/documentation?tab=readme-ov-file#view-diff-for-objectsinv and it doesn't show any change, it means there was no difference after decompression. I'm not sure why this occasionally happens; might be something to do with zlib versions.

frankharkins (approved):

public/api/qiskit-ibm-runtime/0.14/objects.inv
Eric-Arellano: Yeah, I am using that extension. Thanks!

PR #1914, opened 2024-09-05
Fix release note image path for Qiskit historical versions

Eric-Arellano:
https://github.com/Qiskit/documentation/pull/1848 only fixed `npm run make-historical`, but not `npm run gen-api` and `npm run regen-apis`. With this PR, 1.1 stays the same when regenerating, whereas before the release note changed.

I also manually changed the Sphinx artifact for 1.0 to fix a link to a page we migrated.


frankharkins (approved):
Nice, thanks!

PR #1915, opened 2024-09-05
Fix API aliases being orphaned

Eric-Arellano:
We have some attribute pages that are aliases, like `Sampler` being an alias for `SamplerV1`. These pages weren't showing up in the ToC (orphans) because of an oversight.

This PR also makes these improvements:

* Fixes our code for copying release note images to the correct location for Qiskit, as a continuation of https://github.com/Qiskit/documentation/pull/1914
* Uses `ApiType` rather than `string` in more places so that we get stronger typing
* Removes an unused arg, `$main`


docs/api/qiskit/release-notes/1.0.mdx
```diff
@@ -587,7 +587,7 @@ This guide is meant to complement the release notes and have a targeted advice f
     qc.draw("mpl")
     ```
 
-    ![\_images/release\_notes-1.png](/images/api/qiskit/release_notes-1.png)
+    ![\_images/release\_notes-1.png](/images/api/qiskit/1.0/release_notes-1.png)
```
Eric-Arellano, line 590: This was changed in https://github.com/Qiskit/documentation/pull/1914. I didn't think to regenerate 1.0

scripts/js/lib/api/generateToc.ts
```diff
@@ -74,9 +74,15 @@ function getModulesAndItems(
...
-      result.meta.apiType === "exception",
+      result.meta.apiType &&
+      [
+        "class",
+        "function",
+        "exception",
+        "method",
+        "property",
+        "attribute",
+      ].includes(result.meta.apiType),
```
frankharkins, line 85: Is this what makes the Sampler and Estimator appear in the TOC?

frankharkins (approved):
Thanks!

scripts/js/lib/api/generateToc.ts
Eric-Arellano: Yes, exactly. In the upcoming PR to add `..py:data::`, we'll add `data` here as well which fixes the remaining orphan pages we have in API docs 🚀

Issue #1401, opened 2024-05-16
Properly support global variables in API docs (`.. py:data::`) 


Eric-Arellano:
This builds on top of https://github.com/Qiskit/documentation/issues/935, which is about broken cross-references. This issue is to also have a more intentional visual design for `.. py:data::`, such as an MDX component like `<GlobalVariable>` (name tbd).

<details><summary>unannotated screenshot</summary>

<img width="869" alt="Screenshot 2024-05-17 at 1 31 41 PM" src="https://github.com/Qiskit/documentation/assets/14852634/b80417e4-018d-4e4c-a1cb-696f5e287074">
</details>

<details><summary>annotated screenshot</summary>

<img width="869" alt="Screenshot 2024-05-17 at 1 31 41 PM" src="https://github.com/Qiskit/documentation/assets/14852634/56657b16-a913-4709-9ca7-88440dfd46ee">
</details>

<details><summary>offending code</summary>

https://github.com/Qiskit/documentation/blob/394ca77e604b7c58dd916c7ad8f104b6c1e339ac/docs/api/qiskit/dev/circuit.mdx?plain=1#L883-L893
</details>

I propose the design should look the same as our other components: a purple or blue left bar, header inside the component (https://github.com/Qiskit/documentation/issues/1395) with whatever the correct surrounding level would be, and ensure there is a `<span id>` or that the header has the ID.

jakelishman:
Oh boy - "A `EquivalenceLibrary" haha. No points for English for me.

Issue #935, opened 2024-02-29
Fix missing anchors for global variables (`.. py:data::`)


Eric-Arellano:
That is, remove these link checker ignores:

https://github.com/Qiskit/documentation/blob/c7bbfb1b4e86b318c3512c555112b84d8bccf3cc/scripts/lib/links/ignores.ts#L100-L132

They correspond to this RST:

https://github.com/Qiskit/qiskit/blob/d26d3069254a669a43991da73e161abd57f897d3/qiskit/utils/optionals.py#L28-L45

Which results in this markdown:

https://github.com/Qiskit/documentation/blob/c7bbfb1b4e86b318c3512c555112b84d8bccf3cc/docs/api/qiskit/utils.md?plain=1#L329-L336

--

To avoid dealing with the table nuance, it may be easier to look at `qasm2.md`:

https://github.com/Qiskit/documentation/blob/c7bbfb1b4e86b318c3512c555112b84d8bccf3cc/scripts/lib/links/ignores.ts#L95-L97

https://github.com/Qiskit/qiskit/blob/d26d3069254a669a43991da73e161abd57f897d3/qiskit/qasm2/__init__.py#L454-L470

https://github.com/Qiskit/documentation/blob/c7bbfb1b4e86b318c3512c555112b84d8bccf3cc/docs/api/qiskit/qasm2.md?plain=1#L497-L509

--

We should be adding a `<span id="qiskit.utils.optionals.HAS_AER"></span>` for example.

Eric-Arellano:
Fixed by https://github.com/Qiskit/documentation/pull/1917. The utils table is broken on historical versions, but we're not going to fix that because it's too finicky with modifying the HTML.

PR #1917, opened 2024-09-06
Handle Sphinx data type

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/1401 and https://github.com/Qiskit/documentation/issues/935.

We reuse `<Attribute>`, which actually works out quite well! It understands how to parse when the data object has a value assigned to it like `= "my_value"`. It can also properly detect `isDedicatedPage`.

image

There is one weird edge case: Qiskit was improperly using `.. py:data::` inside a table in `utils.mdx`. Naively, it would cause issues with trying to render the component inside the table, which does not work. I fixed it in Qiskit `main`, but it would be too time consuming to fix on 0.35+. So, we now have a line to skip over converting to an `Attribute` Component if the element is inside a table. That means the table will continue to render the same as before on those historical docs - the biggest downside is that anchor links still will be broken for the page.

Screenshot 2024-09-06 at 9 40 51 AM

A follow up will simplify our linters like checking for orphan pages, since this PR fixes a lot of those problems.


frankharkins (approved):
Nice work!

Issue #1841, opened 2024-08-16
Figure out what to do with Dependabot


Eric-Arellano:
Due to now testing all notebooks when requirements.txt change (https://github.com/Qiskit/documentation/pull/1831), dependabot now fails. That's because it cannot access the GitHub Actions Secret due to GitHub's security defaults. See https://github.com/Qiskit/documentation/pull/1839 for an example failure.

We have two options:

1. Stop using Dependabot and manually update the versions when new releases happen.
2. Ask the security team if we can give Dependabot access to the secret.

Update: we agree to get rid of Dependabot for Python requirements. TODOs:
- Set up Dependabot for JavaScript dependencies
- Document the process for new API releases (gen docs, update search index, update testing requirement version)

frankharkins:
I think keeping the Python dependabot is still useful. All we need is for a maintainer to push a commit to that branch to get CI to run (see my merge commit in https://github.com/Qiskit/documentation/pull/1879: the token loads correctly).

Eric-Arellano:
Apparently dependabot already runs for JavaScript. Frank has a good point that it's still valuable to keep for Python.

So I think the only possible action is if we want to update the README? I'm not sure if that's necessary.

PR #1927, opened 2024-09-10
Document how to upgrade noteboook requirements.txt

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/1841.

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,7 +423,7 @@ The add the following to your `.gitconfig` (usually found at `~/.gitconfig`).
 
 When a new version of an API is released, we should also update `nb-tester/requirements.txt` to ensure that our notebooks still work with the latest version of the API. You can do this upgrade either manually or wait for Dependabot's automated PR.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -419,6 +419,14 @@ The add the following to your `.gitconfig` (usually found at `~/.gitconfig`).
   textconv = sh -c 'sphobjinv convert plain "$0" -'
 ```
 
+### Dependabot - upgrade notebook testing version
+
+When a new version of an API is released, we should also update `nb-tester/requirements.txt` to ensure that our notebooks still work with the latest version of the API. You can do this upgrade either manually or wait for Dependabot's automated PR.
+
+Dependabot will fail to run at first due to not having access to the token. To fix this, have someone with write access push a commit to the PR, such as by merging main or using `git commit --allow-empty`.
```
frankharkins, line 426:
```suggestion
Dependabot will fail to run at first due to not having access to the token. To fix this, have someone with write access trigger CI for the PR, such as by merging main or closing then reopening the issue.
```
Closing/reopening is a bit easier since you can do it from GitHub.

frankharkins (approved):
Thanks!

Issue #798, opened 2024-02-13
Some page URLs end in .rst


frankharkins:
The following files in `docs/api/qiskit` end in `.rst.md`:

```
qiskit.pulse.library.Constant_class.rst.md
qiskit.pulse.library.Cos_class.rst.md
qiskit.pulse.library.Drag_class.rst.md
qiskit.pulse.library.Gaussian_class.rst.md
qiskit.pulse.library.Sawtooth_class.rst.md
qiskit.pulse.library.Sech_fun.rst.md
qiskit.pulse.library.Sin_class.rst.md
qiskit.pulse.library.Square_fun.rst.md
qiskit.pulse.library.Triangle_class.rst.md
```

The `.rst` ends up in the URL. For example, see https://docs.quantum.ibm.com/api/qiskit/qiskit.pulse.library.Square_fun.rst.

Eric-Arellano:
Yeah, this quirk has caused a lot of problems, including with our redirect rules. We should fix this, but make sure we set up redirects correctly.

Eric-Arellano:
I've opened up https://github.com/Qiskit/qiskit/pull/13107. Assuming that gets merged, we'll need to set up redirects.

PR #1935, opened 2024-09-12
Generate Qiskit 1.2.1, 1.3.0b1 and Runtime improvements

Eric-Arellano:
This brings in a fix for https://github.com/Qiskit/documentation/issues/798, which means that we need to set up our historical redirects for those RST files in Qiskit to go to the now renamed files. I will set up redirects in closed source.


frankharkins (approved):
Thanks! Good call with the historical redirects too

PR #1953, opened 2024-09-16
Adjust function name to Tensor-network error mitigation

abbycross:
By request - updating the Algorithmiq function name to Tensor-network error mitigation


Eric-Arellano (approved):
My only concern is capitalization: it's only our sentence case function

Eric-Arellano:
> Also, the IBM function is also sentence case.)

Sort of. We capitalize the Circuit weirdly

PR #1960, opened 2024-09-16
Add tests for code in `serverless-manage-resources.ipynb`

frankharkins:
This PR fixes the code in `serverless-manage-resources.ipynb` and adds tests to check it runs correctly.

Link to preview: https://qiskit.github.io/documentation/pr-1960/guides/serverless-manage-resources


docs/guides/serverless-manage-resources.ipynb
```diff
@@ -22,22 +22,53 @@
   },
```
Eric-Arellano, line 22: I think line 20 should be `transpile_remote.py`. /guides/serverless-first-program#get-program-arguments also uses the wrong file name

docs/guides/serverless-manage-resources.ipynb
```diff
@@ -73,18 +153,115 @@
...
+     "output_type": "stream",
+     "text": [
+      "Appending to ./source_files/transpile_remote.py\n"
+     ]
+    }
+   ],
+   "source": [
+    "%%writefile --append ./source_files/transpile_remote.py\n",
+    "# This cell is hidden from users. It checks the transpilation ran correctly.\n",
+    "assert isinstance(results[0], QuantumCircuit)"
```
Eric-Arellano, line 206:
```suggestion
    "assert all(isinstance(result, QuantumCircuit) for result in results)"
```

docs/guides/serverless-manage-resources.ipynb
```diff
@@ -73,18 +153,115 @@
...
+    "import time\n",
+    "import uuid\n",
+    "\n",
+    "TITLE = \"transpile_remote_serverless_test_\" + uuid.uuid4().hex[:8]\n",
+    "serverless = IBMServerlessClient()\n",
+    "transpile_remote_demo = QiskitFunction(\n",
+    "    title=TITLE,\n",
+    "    entrypoint=\"transpile_remote.py\",\n",
+    "    working_dir=\"./source_files/\",\n",
+    "\tdependencies=[\"qiskit-aer==0.14.1\"],\n",
```
Eric-Arellano, line 247: Why is this hardcoded? I'm not very familiar with serverless

docs/guides/serverless-manage-resources.ipynb
```diff
@@ -133,18 +340,27 @@
...
+     "output_type": "stream",
+     "text": [
+      "Writing ./source_files/transpile_parallel.py\n"
+     ]
+    }
+   ],
    "source": [
-    "```python\n",
-    "# source_files/transpile_parallel.py\n",
+    "%%writefile ./source_files/transpile_parallel.py\n",
```
Eric-Arellano, line 357: Do you know why we change file names here?

docs/guides/serverless-manage-resources.ipynb
```diff
@@ -307,6 +516,66 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "# This cell is hidden from users.\n",
+    "# It uploads the serverless program and checks it runs.\n",
+    "\n",
+    "from qiskit.circuit.random import random_circuit\n",
+    "from qiskit_serverless import IBMServerlessClient, QiskitFunction\n",
+    "import time\n",
```
Eric-Arellano, line 547: It'd probably be good to DRY this code. Can you define a hidden helper function at the top that gets called by both?

Eric-Arellano (commented):
Thanks for working on this!

docs/guides/serverless-manage-resources.ipynb
frankharkins: I think it's just to demonstrate that you can set dependencies for your serverless code

Eric-Arellano: Hm, I wonder if we can remove it. I'm worried about the dep falling out of sync. Think about how many people copy and paste code, or even Qiskit Code Assistant learning from our docs. We don't want them installing an old version.

frankharkins: lol I'm sorry; I didn't realise this was part of the testing code I added. I copied it from https://docs.quantum.ibm.com/guides/serverless-first-program. I'll remove the dpe.

> Think about how many people copy and paste code

👀

docs/guides/serverless-manage-resources.ipynb
frankharkins: Not really, I guess they could both be `transpile_remote`

frankharkins:
Converting to draft as I'm struggling to work around a limitation. The problem is that `IBMServerlessClient()` (without API token) fails on the server. I thought it'd work since `QiskitRuntimeService()` works ok, but it turns out that's because it can get user details from an environment variable (which is set on the server), whereas serverless can only use a config file (set locally but not on the server). To get this to work, we'd need to inject our API token into the code at somepoint, and I don't think we can do that without adding more logic to `nb-tester`. I think the best course of action is to wait until this is possible before picking this PR back up.

frankharkins:
Superceded by #2398

PR #1963, opened 2024-09-16
Add back page toc on intro-to-qiskit

abbycross:
Page is getting long and could use the navigational guides. Also restores the option for a reader to vote on the page's helpfulness, and to report a bug on GitHub.


Eric-Arellano (approved):
I'm happy with this, but please double check with product like Abby M if you haven't already

abbycross:
Already did check w Abby - thanks!

PR #1972, opened 2024-09-17
Regen Qiskit 0.46.3 to fix broken link

Eric-Arellano:
Also makes the Runtime link ignores more precise. I spent twenty minutes trying to figure out why `objects.inv` still has a broken link `index#qiskit-runtime-version-api-docs-preview` in Runtime but not Transpiler Service & Qiskit. I couldn't figure it out, so moved on.


scripts/js/lib/links/ignores.ts
```diff
@@ -83,37 +83,42 @@ const _QISKIT_QPY_IGNORES = Object.fromEntries(
...
+        [
+          `/api/qiskit-ibm-runtime/${vers}index#next-steps`,
+          `/api/qiskit-ibm-runtime/${vers}index#qiskit-runtime-version-api-docs-preview`,
+        ],
+      ],
+    ),
+  );
+  const newVersions = Object.fromEntries(
     [
-      `/api/qiskit-ibm-runtime/${vers}qiskit_ibm_runtime.RuntimeEncoder#key_separator`,
```
Eric-Arellano, line 111: This is no longer broken

frankharkins (approved):

PR #1988, opened 2024-09-18
add measurement information

### LINKED ISSUE #1925: Add information somewhere about measuring qubits
### URL, if applicable

maybe in the construct-circuits topic

We talk about measuring qubits and where measurements can go in a couple of error messages but we actually never mention how to add a measurement in the guides.

beckykd:
Closes #1925


docs/guides/measure-qubits.mdx
```diff
@@ -0,0 +1,96 @@
...
+### `measure_all` method
+
+To measure all qubits into the corresponding classical bits, use the [`measure_all`](/api/qiskit/qiskit.circuit.QuantumCircuit#measure_all) method. By default, this method adds new classical bits in a `ClassicalRegister` to store these measurements.
+
+```python
+from qiskit import QuantumCircuit
+
+qc = QuantumCircuit(3, 1)
+qc.x([0, 2])
+qc.measure_all # Measure all qubits.
```
Cryoris, line 47: `measure_all` and `measure_active` are methods, so they need the brackets 🙂 
```suggestion
qc.measure_all() # Measure all qubits.
```

docs/guides/measure-qubits.mdx
```diff
@@ -0,0 +1,96 @@
...
+### `measure_active` method
+
+To measure all qubits that are not idle, use the [`measure_active`](/api/qiskit/qiskit.circuit.QuantumCircuit#measure_active) method. This method creates a new `ClassicalRegister` with a size equal to the number of non-idle qubits being measured.
+
+```python
+from qiskit import QuantumCircuit
+
+qc = QuantumCircuit(3, 1)
+qc.x([0, 2])
+qc.measure_active # Measure qubits that are not idle.
```
Cryoris, line 64:
```suggestion
qc.measure_active() # Measure qubits that are not idle.
```

docs/guides/measure-qubits.mdx
```diff
@@ -0,0 +1,96 @@
+---
+title: Measure qubits
+description: Learn how to measure qubits, including constraints on where measurements can be used.
+---
+
+# Measure qubits
+
+To get information about a qubit's state, you have to _measure_ it onto a [classical bit](/api/qiskit/circuit#qiskit.circuit.Clbit).  This is an irreversible operation, and usually destroys entanglement and phase coherence between the target qubit and the rest of the system. Thus, after a qubit is measured, you can't perform any more instructions on that qubit, and each qubit can only be measured once. Additionally, there must be at least one classical register to store the measurement information.
```
Cryoris, line 8: Is the goal of this docs page to explain measurements themselves or how they can be used in Qiskit? The former is definitely more involved than the latter. Maybe it would be good to restrict to just explaining how they work in Qiskit? 🙂 

Some notes as to why explaining measurements is more involved:
* There are other ways of obtaining information about qubit states aside from directly measuring it (see e.g. "weak measurement" if you're interested), to be technically correct maybe "can" is a better word 🙂 
* Measurements don't need to be destructive (i.e. we don't know the qubit state after) -- in fact many quantum computing stacks (incl. IBMs) assume that it is projective, meaning that the qubit is projected onto the outcome we observed. 

How about reformulating as the following?
```suggestion
To get information about a qubit's state, you can _measure_ it onto a [classical bit](/api/qiskit/circuit#qiskit.circuit.Clbit). In Qiskit, measurements are performed in the computational basis, i.e., the single-qubit Pauli-$Z$ basis. A measurement will thus yield 0 or 1 depending on the overlap with the Pauli-$Z$ eigenstates $|0\rangle$ and $|1\rangle$.
```

I don't know how to apply math correctly, but we could add this equation to be even more explicit
image

docs/guides/measure-qubits.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+## Apply a measurement to a circuit
+
+There are several ways to apply measurement to a circuit:
+
+### `Measure` class
+
+The Qiskit [Measure](/api/qiskit/circuit#qiskit.circuit.Measure) class measures the specified qubits.
```
Cryoris, line 16: How about renaming this to use the `QuantumCircuit.measure` method instead of the `Measure` class? That's what the code snippets use and usually the main entry point 🙂 You could also add a `Measure` class part below which explicitly uses it, i.e.
```
from qiskit.circuit import Measure

....
qc.append(Measure(), [0], [0])  # measure qubit 0 into clbit 0
```

docs/guides/measure-qubits.mdx
```diff
@@ -0,0 +1,96 @@
...
+from qiskit import QuantumCircuit
+
+qc = QuantumCircuit(3, 1)
+qc.x([0, 2])
+qc.measure_all # Measure all qubits.
+```
+
+******************************************************
+QUESTION:
+**WHEN I RUN THE ABOVE CODE AND ADD qc.draw("mpl"), IT DOESN'T LOOK LIKE ANY MEASUREMENTS ARE ACTUALLY ADDED, BUT IF I USE qc.measure(1, 0) INSTEAD, I SEE A MEASUREMENT. WHAT IS HAPPENING?**
```
Cryoris, line 52: The brackets are missing, see the comments above 🙂

docs/guides/measure-qubits.mdx
```diff
@@ -0,0 +1,96 @@
...
+```python
+from qiskit.circuit.random import random_circuit
+
+circ = random_circuit(2, 2, measure=True)
+circ.draw(output='mpl')
+```
+
+<Admonition type="attention" title="Important notes">
+
+* Qubit measurement must be the last instruction on a qubit.  That is, no instructions can come after a measurement on that qubit (unless it's a dynamic circuit).
```
Cryoris, line 82: I would suggest to turn this around, it's perfectly valid to have operations after measurements but it would be good to point out that there could be issues in running those circuits.
```suggestion
* Circuits that contain operations _after_ a measurement are called dynamic circuits. Note, that not all backends or simulators might support these.
```

docs/guides/measure-qubits.mdx
```diff
@@ -0,0 +1,96 @@
...
+* Each qubit can be measured only once.
```
Cryoris, line 83: Dynamic circuits can measure qubits multiple times 🙂 
```suggestion
```

docs/guides/measure-qubits.mdx
```diff
@@ -0,0 +1,96 @@
...
+You can add measurements when generating a random circuit. Each qubit is measured, as shown in the following example.  For more information, see the [`random_circuit`](/api/qiskit/circuit#random_circuit) method documentation.
+
+Example:
+
+```python
+from qiskit.circuit.random import random_circuit
+
+circ = random_circuit(2, 2, measure=True)
+circ.draw(output='mpl')
+```
```
Cryoris, line 78: This is a special case that might be best just documented on the `random_circuit` function, it's likely not necessary to point this out in general docs
```suggestion
```

docs/guides/measure-qubits.mdx
beckykd: Interesting - I've been editing error messages, and one of them says you can't have operations after measurements.  Or maybe I'm missing a nuance there:  
image

docs/guides/measure-qubits.mdx
beckykd: I got this one from an error message too: 
image
 
I can clarify that it only applies to non-dynamic circuits though.

docs/guides/measure-qubits.mdx
beckykd: Oh dear - I generated this from our new AI tool :(

beckykd:
I probably need to fix the equation.  I'll do that and then it should be ready for final review.

docs/guides/measure-qubits.mdx
```diff
@@ -0,0 +1,97 @@
...
+---
+
+# Measure qubits
+
+To get information about a qubit's state, you can _measure_ it onto a [classical bit](/api/qiskit/circuit#qiskit.circuit.Clbit). In Qiskit, measurements are performed in the computational basis, that is, the single-qubit Pauli-$Z$ basis. A measurement will thus yield 0 or 1, depending on the overlap with the Pauli-$Z$ eigenstates $|0\rangle$ and $|1\rangle$:
+
+$$
+|q\rangle \xrightarrow{measure}\begin{cases}
+      |0\rangle (\text{outcome}+1), \text{with probability } p_0=|\langle q|0\rangle|^{2}\text{,} \\
+      |1\rangle (\text{outcome}-1), \text{with probability } p_1=|\langle q|1\rangle|^{2}\text{.}
```
Cryoris, line 13: Sorry I missed this before, but this is a minor nitpick: we don't strictly state whether the measurements are projective so it would be more correct to just state the measurement output (0 or 1) instead of the post-measurement state ($|0\rangle$ or $|1\rangle$)
```suggestion
      0 (\text{outcome}+1), \text{with probability } p_0=|\langle q|0\rangle|^{2}\text{,} \\
      1 (\text{outcome}-1), \text{with probability } p_1=|\langle q|1\rangle|^{2}\text{.}
```

docs/guides/measure-qubits.mdx
```diff
@@ -0,0 +1,97 @@
...
+    \end{cases}
+$$
+
+## Apply a measurement to a circuit
+
+There are several ways to apply measurement to a circuit:
+
+### `QuantumCircuit.measure` method
```
Cryoris, line 21: Below you're not pre-pending `QuantumCircuit` to the methods `measure_active` and `measure_all`, should it also be removed in this title?
```suggestion
### `measure` method
```

docs/guides/measure-qubits.mdx
```diff
@@ -0,0 +1,97 @@
...
+### `measure_active` method
+
+To measure all qubits that are not idle, use the [`measure_active`](/api/qiskit/qiskit.circuit.QuantumCircuit#measure_active) method. This method creates a new `ClassicalRegister` with a size equal to the number of non-idle qubits being measured.
+
+```python
+from qiskit import QuantumCircuit
+
+qc = QuantumCircuit(3, 1)
+qc.x([0, 2])
+qc.measure_active() # Measure qubits that are not idle.
```
Cryoris, line 77: Just a suggestion, maybe users don't know what "idle" means 🙂 
```suggestion
qc.measure_active() # Measure qubits that are not idle, i.e., qubits 0 and 2.
```

docs/guides/measure-qubits.mdx
```diff
@@ -0,0 +1,97 @@
...
+```
+
+
+<Admonition type="attention" title="Important notes">
+
+* Circuits that contain operations _after_ a measurement are called dynamic circuits. Note, that not all backends or simulators might support these.
+* Each qubit can be measured only once, except when using dynamic circuits.
```
Cryoris, line 84: IMO it would be better to remove this, it can be a bit confusing. Above you already mention that in non-dynamic circuits the measurement must be the last operation, which I would argue means it can only be measured once 🙂

Cryoris (commented):
Some minor comments otherwise LGTM!

docs/guides/measure-qubits.mdx
```diff
@@ -9,16 +9,16 @@ To get information about a qubit's state, you can _measure_ it onto a [classical
...
+      1 (\text{outcome}-1), \text{with probability } p_1=|\langle q|1\rangle|^{2}\text{.}
     \end{cases}
 $$
 
 ## Apply a measurement to a circuit
 
 There are several ways to apply measurement to a circuit:
 
-### `QuantumCircuit.measure` method
+### `measure` method
```
Eric-Arellano, line 21: Why this change? It's useful to include the class name

Cryoris: For consistency (see below) -- but it would of course also be an option to include the class in the other titles 🙂

Eric-Arellano: Got it. @beckykd I recommend putting the prefix in front of every method

abbycross (approved):

PR #1992, opened 2024-09-19
benchmarking info for intro pg

### LINKED ISSUE #1674: [ NEW PAGE ] Benchmarking performance
With performance becoming a big feature of qiskit for the rest of the year we should include a page somewhere in docs (not sure where yet) that explains the following:

- what do we mean by "performance" (e.g. speed of transpilation, memory usage, number of cnot gates). Discuss challenges of quantum performance metrics and why they change often (Explain Speed, Quality and Completion)
- why are these metrics important? why should I, a quantum developer, care about these? what value does it add to my quantum program?
- where can I find benchmarking information, or how could I run the benchmarks myself?
- lists of papers relevant to each metric
- small section in latest update page about latest performance improvements?

~Options for where this content could live:~
~- the "additional resources" tab~
~- the transpilation specific information could live in the transpile section, not sure about the other information~
~- if there are multiple pages that could be created maybe it's worth creating a "performance" or "benchmarking" folder in the guides section? Might be confusing though if we also want to put performance discussion interweaved with the other sections on error mitigation or transpilation thought~

update: put "performance" folder in guides section, with at least 2 pages (1x SDK performance, 1x hardware/other performance)

questions to discuss:
~- do we just want to talk about the software performance (e.g. transpile time) or also hardware (quantum volume, CLOPS etc.)~
- how much is SDK specific vs runtime vs transpiler service

abbycross:
Closes #1674

### DIFF
--- a/docs/guides/index.mdx
+++ b/docs/guides/index.mdx
@@ -69,6 +69,12 @@ The Qiskit Transpiler Service ([package name `qiskit-ibm-transpiler`](https://py
 
 Qiskit addons are a collection of research capabilities for utility-scale algorithm discovery. These capabilities build upon Qiskit’s performant foundation of tools for creating and running quantum algorithms. Addons are modular software components that plug into a workflow to scale or design new quantum algorithms. To learn more about the set of available Qiskit addons and how to get started using them, visit the [documentation](/guides/addons).
 
+## Benchmarking and the Benchpress package
+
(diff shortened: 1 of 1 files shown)


abbycross:
Very broad language here, pointing to the blog and the github repo for the package itself, in the intro to qiskit page. To discuss with @javabster about expanding this content.

javabster:
actually now that im looking at the full page preview, I think this would be better if included as a subsection of the Qiskit SDK section. bc the benchmarking only applies to the SDK (afaik) so it feels weird to have to scroll past all the other qiskit things to get to this section. otherwise content LGTM

javabster (approved):
LGTM 🚀

PR #1999, opened 2024-09-20
Add functions owner

beckykd:


qiskit_bot.yaml
```diff
@@ -313,18 +313,21 @@ notifications:
     - "@fpietra"
     - "@panadestein"
     - "@matteoacrossi"
+    - "@@johannesgreiner"
```
Eric-Arellano, line 316: Every entry has a double `@` which will break

Eric-Arellano (approved):

PR #2003, opened 2024-09-20
toc nesting audit

### LINKED ISSUE #1895: Audit sidebar nesting
On the left sidebar there are some pages that users might want to get to quickly (like intro to primitives and intro to transpilation) but they are buried in many layers of nesting. I think it would be good to go through our sidebar and make some adjustments overall to reduce nesting where we can. We should keep the top level of nesting but i think the amount of second and third level of nesting could be reduced

abbycross:
Closes #1895

### DIFF
--- a/docs/guides/_toc.json
+++ b/docs/guides/_toc.json
@@ -125,83 +125,73 @@
         {
           "title": "Circuits and operators",
           "children": [
-            {
-              "title": "Build circuits with the Qiskit SDK",
-              "children": [
-                {
-                  "title": "Circuit library",
-                  "url": "/guides/circuit-library"
-                },
-                {
-                  "title": "Construct circuits",
-                  "url": "/guides/construct-circuits"
-                },
-                {
-                  "title": "Classical feedforward and control flow",
-                  "url": "/guides/classical-feedforward-and-control-flow"
-                },
-                {
-                  "title": "Synthesize unitary operators",
-                  "url": "/guides/synthesize-unitary-operators"
-                },
-                {
-                  "title": "Bit-ordering in the Qiskit SDK",
-                  "url": "/guides/bit-ordering"
-                },
-                {
-                  "title": "Save circuits to disk",
-                  "url": "/guides/save-circuits"
-                }
-              ]
-            },
-            {
-              "title": "Build operators with the Qiskit SDK",
-              "children": [
-                {
-                  "title": "Operators module overview",
-                  "url": "/guides/operators-overview"
-                },
-                {
-                  "title": "Specify observables in the Pauli basis",
-                  "url": "/guides/specify-observables-pauli"
-                },
-                {
-                  "title": "The Operator class",
-                  "url": "/guides/operator-class"
-                }
-              ]
-            },
-            {
-              "title": "Other circuit building tools",
-              "children": [
-                {
-                  "title": "Pulse schedules",
-                  "url": "/guides/pulse"
-                },
-                {
-                  "title": "OpenQASM",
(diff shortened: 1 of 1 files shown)


docs/guides/_toc.json
```diff
@@ -211,67 +201,52 @@
...
-                {
-                  "title": "Represent quantum computers",
-                  "url": "/guides/represent-quantum-computers"
-                }
-              ]
+              "title": "Transpiler stages",
+              "url": "/guides/transpiler-stages"
             },
             {
-              "title": "Advanced transpilation resources",
```
Eric-Arellano, line 252: I think this is still a useful subgroup. I'm not sure if there is a better name like "Custom transpilation and plugins"

docs/guides/_toc.json
```diff
@@ -125,83 +125,73 @@
...
+                "url": "/guides/specify-observables-pauli"
+              },
+              {
+                "title": "The Operator class",
+                "url": "/guides/operator-class"
+              }
+            ]
+          },
+          {
+            "title": "Other circuit-building tools",
```
javabster, line 167: the "other circuit building tools section" should still be inside the circuits and operators top level section

docs/guides/_toc.json
javabster: yeah im happy to keep this subfolder, I would just move it so it is the very last item (i.e. after the transpiler servcie)

docs/guides/_toc.json
abbycross: oops! Yes indeed, will fix

docs/guides/_toc.json
abbycross: I'm restoring the subfolder - do we want to rename or keep the current name? @javabster

javabster: i think keep the current name, "custom" could also imply custom passmanager stuff which isnt in this advanced section

abbycross: confirmed - keeping the current name. Subfolder is restored.

Eric-Arellano (approved):
This lgtm, although please wait for Abby M. 

My only suggested tweak is to restore Transpiler how it was before outside of flattening "Get started with the Qiskit transpiler" -- I thought it made sense for "Configure preset pass managers" to be a folder name. I'm also not sure why Abby M wants Qiskit Transpiler Service to come before "Advanced transpilation resources"?

Thanks for working on this! It's a great improvement.

javabster:
ok yeah Eric raises a good point, now that we've flattened it its not clear which pages are for custom passmanagers vs using generate preset passmanager. 

Ok this is what i think we should do here:

- Introduction to transpilation
- Transpiler stages
- Transpile with pass managers
- Create a passmanager for dynamical decoupling (move page out of advanced section)
- Configure preset pass managers (dropdown)
- Qiskit transpiler service (dropdown)
- Advanced transpilation resources (dropdown)

with this version the intro pages and the custom passmanager content (i.e. the content we most want to encourage people to read) isnt hidden in a dropdown and are logically grouped together. and its clear which pages are about the preset passmanagers bc theyre in their dedicated section. I want to put the transpiler service above the advanced dropdown because it feels logical to have the most advanced stuff be at the end of the section, and transpiler service is actually for people that are less advanced not more advanced

beckykd (commented):
LGTM, but I'll let @javabster approve!

javabster (approved):
LGTM 🚀 it's so nice to have fewer clicks to find what I need!

Eric-Arellano:
Agreed! Excellent job @abbycross!!!

PR #2010, opened 2024-09-23
Fix bird naming

abbycross:
Any instances of IBM + [Processor family name] should change to IBM Quantum [Processor family name]


abbycross:
Note: got confirmation that it's acceptable to say IBM QPU (rather than (IBM Quantum QPU)

Eric-Arellano (approved):
This was the only file with issues?

abbycross:
> This was the only file with issues?

Weirdly, yes. I searched 
`rg 'IBM&reg; Heron'` (also replaced with Eagle, Falcon, etc.) and
`rg 'IBM Heron'` (same process) and that was the only one. I did see plenty of `IBM QPUs` but clarified with the blog team that that was acceptable.

PR #2015, opened 2024-09-23
Fix code examples in "Manage Qiskit Serverless compute and data resources"

frankharkins:
This makes the code examples runnable. Unfortunately we can't add tests just yet (will happen in #1960), but this does mean users will be able to run all the code on the page.

I've removed the section on automatic QPU selection after discussion with the serverless team. This feature is being developed but isn't ready just yet. We'll add the section back in when it's ready.

### DIFF
--- a/docs/guides/serverless-manage-resources.ipynb
+++ b/docs/guides/serverless-manage-resources.ipynb
@@ -40,7 +40,7 @@
     "    \"\"\"Transpiles an abstract circuit (or list of circuits) into an ISA circuit for a given backend.\"\"\"\n",
     "    pass_manager = generate_preset_pass_manager(\n",
     "        optimization_level=optimization_level,\n",
-    "\t\tbackend=service.backend(backend)\n",
+    "        backend=service.backend(backend)\n",
     "    )\n",
     "    isa_circuit = pass_manager.run(circuit)\n",
     "    return isa_circuit"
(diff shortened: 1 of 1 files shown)


frankharkins:
PR did deploy, check out the page here: https://docs.quantum.ibm.com/guides/serverless-manage-resources

docs/guides/serverless-manage-resources.ipynb
```diff
@@ -108,81 +126,16 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "%%writefile --append ./source_files/transpile_remote.py\n",
+    "\n",
     "@distribute_task(target={\n",
     "    \"cpu\": 16,\n",
-    "    \"mem\": 32 * 1024 * 1024 * 1024\n",
+    "    \"mem\": 2 * 1024 * 1024 * 1024\n",
```
Eric-Arellano, line 133: Why this change?

Eric-Arellano (approved):
Thanks!

docs/guides/serverless-manage-resources.ipynb
frankharkins: The prose says:

>  To allocate 2 GB of memory for your worker, you need to allocate `"mem": 2 * 1024 * 1024 * 1024`

As a reader, I was mildly surprised when the code example allocated 32GB rather than 2.

Issue #745, opened 2024-02-01
Improve `dev` version of API docs to use continuous deployment


Eric-Arellano:
In https://github.com/Qiskit/documentation/issues/316, you have to manually trigger a regeneration of the dev docs. That was done for expediency. Better would be to automate this so that we regenerate the dev docs ~every day with whatever new changes there are, if any. 

Doing that relates to https://github.com/Qiskit/documentation/issues/501, that we need a way to update docs via CI/CD.

Eric-Arellano:
I think this is high priority. We've been very bad at remembering to update this, so we often go 2+ weeks without any regeneration.

While I don't think we should do CD for general API docs (https://github.com/Qiskit/documentation/issues/501#issuecomment-2334766670), I do think we should still do it for dev versions. The major difference is that these dev versions should be safe to use GitHub URLs in scripts/config/api-html-artifacts.json, rather than Box URLs. That should be fine because the docs will be regenerated so frequently that we're not at much risk of the GitHub artifact expiring (we generally want the URLs to still work in 1+ years)

We should not do automatic updates for `rc` releases, though. Those should correspond to the actual released artifact. So, we need to check `_package.json` to see if it's `-dev` or `rc*`.

To implement this:

1. Figure out how to get the latest GitHub Actions artifact corresponding to the `main` branch
2. Download the artifact
3. Update `api-html-artifacts.json` with the URL
4. `npm run gen-api`
5. Open the PR

Create a GH action that runs this every 1-3 nights (TBD frequency).

Note that a human still has to do the merge & open-source sync.

PR #2017, opened 2024-09-23
Update dev docs in nightly job

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/745. This follows the spec laid out there:

* Do nothing if the dev version is an rc
* Otherwise, use GitHub API to get latest artifact from `main`
* Update `api-html-artifacts.json` to the GitHub URL
* Generate the docs
* Create a PR if there were changes

To be able to call `updateApiDocs.ts` from `syncDevDocs.ts`, we now use the special value ``if (import.meta.url === `file://${process.argv[1]}`) ``, which is like Python's `if __name__ == "__main__":` so that you can import the file without executing the script. This works because we use ES6 modules now.


.github/workflows/dev-docs-sync.yml
```diff
@@ -0,0 +1,62 @@
...
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+name: Dev docs sync
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 1 * * *" # Every day at 1 am UTC
```
Eric-Arellano, line 17: Our other jobs run at midnight, so I figured this reduces contention such as if we get rate limited by GitHub by our external link checker

.github/workflows/dev-docs-sync.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Make pull request
+        if: ${{ !cancelled() && steps.changed-docs-files.outputs.any_changed == 'true' }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git config user.name "github-actions[bot]"
+          git switch -c actions/cron-${{ github.run_id }}
+          git add .
+          git commit -m "Re-run notebooks"
```
Eric-Arellano, line 51:
```suggestion
          git commit -m "Sync latest dev docs"
```

scripts/js/commands/api/updateApiDocs.ts
```diff
@@ -73,33 +73,19 @@ const readArgs = (): Arguments => {
     .parseSync();
 };
 
-zxMain(async () => {
-  const args = readArgs();
```
Eric-Arellano, line 77: No changes made to the logic or sequencing of this code block. I only repositioned it so that the new function `generateVersion()` abstracts some of it, and the main is at the bottom of the file.

scripts/js/lib/links/ExternalLink.ts
```diff
@@ -58,16 +58,22 @@ export class ExternalLink {
...
+    link.startsWith("https://api.github.com")
+  ) {
     const ghToken = process.env.GITHUB_TOKEN;
-    if (ghToken) {
-      headers["Authorization"] = `token ${ghToken}`;
+    if (!ghToken) {
+      throw new Error(
+        "The env variable GITHUB_TOKEN is not set. You can get your token by running `gh auth token`.",
+      );
     }
```
Eric-Arellano, line 75: If the token isn't set, you'll get a 403 when generating the dev docs from a GitHub URL. This is relevant for when you do e.g. `regen-apis` locally. So, we should error eagerly.

Eric-Arellano (commented):
See https://github.com/Qiskit/documentation/pull/2014 for an example.

frankharkins (approved):
Nice work! Very easy to read and the assumptions seem robust.

PR #2023, opened 2024-09-24
Wait for docker pull in CI

frankharkins:
CI can fail if `docker pull ...` takes longer than expected. This PR adds a `--pull-only` option to `./start`, then waits for it to complete in CI.


Eric-Arellano (approved):
Nice!

PR #2027, opened 2024-09-24
Update ibm cf parameter names

jyu00:
As an effort to unify the interfaces of all circuit functions, IBM circuit function has renamed the `backend` parameter to `backend_name`, and removed `precision` (since there is already `default_precision` in the options). This PR updates the docs accordingly.


Eric-Arellano (approved):
LGTM. This is ready to go live now, right? Like the old code was broken?

jyu00:
> LGTM. This is ready to go live now, right? Like the old code was broken?

Yes. The new code is already live 😅

PR #2035, opened 2024-09-25
Add cell output for IBM Circuit function

jyu00:
Add the cell output per [this comment](https://github.com/Qiskit/documentation/pull/2019#discussion_r1773740955).

Also it appears some kind of git merge brought back `precision` as an option. So this PR also got rid of that.


Eric-Arellano (approved):
Thank you!

PR #2045, opened 2024-09-27
Remove bad `figcaption` from Qiskit 0.40

Eric-Arellano:
For whatever reason, Qiskit 0.40 docs did not include the normal HTML/CSS classes that `figcaption` sets for us to be able to remove the values. So, we have broken links like this:

Screenshot 2024-09-27 at 2 40 48 PM

There was no programmatic way to fix this easily, with HTML like this:

```html
<p>(<a class="reference external" href="../apidoc/transpiler-17.py">Source code</a>, <a class="reference external" href="../apidoc/transpiler-17.png">png</a>, <a class="reference external" href="../apidoc/transpiler-17.hires.png">hires.png</a>, <a class="reference external" href="../apidoc/transpiler-17.pdf">pdf</a>)</p>
```

So, I (semi-automatically) deleted all the bad HTML directly from the Sphinx HTML and used Git to verify I didn't accidentally remove other things.

--

Thanks to this PR, Qiskit 0.40 now has no broken links.


frankharkins (approved):
Nice, thanks!

PR #2046, opened 2024-09-27
Fix broken anchor link in NFT Qiskit class

Eric-Arellano:
The citation was improperly configured in Qiskit 0.24-0.28 and it caused a broken anchor link. This required manually fixing the HTML. I simplified the code to inline the citation.


frankharkins (approved):
Looks a lot neater, thanks

PR #2050, opened 2024-09-30
Fix typo in example code of v1.1.0 changelog

gschurck:


Eric-Arellano (approved):
Thanks for the contribution, @gschurck! I appreciate you taking the time to make Qiskit better.

FYI for posterity: I updated the HTML we store for Qiskit 1.1 in Box to have this change, so that the next time we regenerate the docs, we won't accidentally revert the fix.

PR #2056, opened 2024-10-01
Removed unnecessary imports from circuit drawer example code

emilkovacev:
Simple fix, just removed imports that aren't used to make the code block work.


emilkovacev:
Hi Eric, it seemed like I was supposed to open this issue in Qiskit repository instead, since the code is in a code block. Is that correct?

Eric-Arellano:
Ah, yes, that's correct. I didn't notice that. Thanks!

Issue #1649, opened 2024-07-03
Improve fork failure for notebook testing


Eric-Arellano:
When testing notebooks, we fail on forks:

https://github.com/Qiskit/documentation/blob/77f07b691a2f01e9c87d69e09fad933bbfd2782f/.github/workflows/notebook-test.yml#L28-L44

This is a good idea. However, there is no need to fail on the fork if the notebook is skipped anyways because it cannot be tested in CI.

--

We should also make sure all subsequent steps are canceled properly.

--

We should run the notebook lint, which is safe on forks.

PR #897, opened 2024-02-27
Replace Squeaky CI advice with custom advice

frankharkins:
Closes https://github.com/Qiskit/documentation/issues/838.

Replaces the message in CI with correct instructions for our repository. I've added a switch to remove the advice from the printed message, I think this is the neatest net logic change: https://github.com/frankharkins/squeaky/commit/33e24f8eb633d25758c787c4efd5d645458c352f.

Screenshot 2024-02-27 at 17 20 42

See https://github.com/frankharkins/documentation/pull/2 for the test PR.

This doesn't display the message when running `tox -e lint`, but I think it's unlikely a contributor would run that command locally without first encountering it in CI.

PR #2062, opened 2024-10-01
Improve fork behavior for notebook job

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/1649.

1. Don't run the last three steps if earlier steps failed.
    - We were missing `if: success()`
2. Move the lint job to normal CI. 
    - This is useful because it means we won't accidentally merge something with bad linting. That was possible before because the Execute notebooks job is expected to fail on forks, although it is still important the lint job passes.
    - We already were setting up the Python environment in main CI, so there is not much downside to this
3. Don't fail CI if none of the notebooks could have been tested anyways because they're not mockable, e.g. `hello-world`.

Now the failure looks like this:

Screenshot 2024-10-01 at 2 46 44 PM

### DIFF
--- a/.github/workflows/notebook-test.yml
+++ b/.github/workflows/notebook-test.yml
@@ -57,7 +57,6 @@ jobs:
           ibm-quantum-token: ${{ secrets.IBM_QUANTUM_TEST_TOKEN }}
 
       - name: Execute notebooks
-        if: ${{ success() && !cancelled() }}
         shell: python
         env:
           PR_REPOSITORY: ${{ github.event.pull_request.head.repo.full_name }}
@@ -75,14 +74,12 @@ jobs:
 
       - name: Detect changed notebooks
         id: changed-notebooks
-        if: ${{ success() && !cancelled() }}
         run: |
           echo "CHANGED_NOTEBOOKS<<EOF" >> $GITHUB_OUTPUT
           git diff --name-only >> $GITHUB_OUTPUT
           echo "EOF" >> $GITHUB_OUTPUT
 
       - name: Upload executed notebooks
-        if: ${{ success() && !cancelled() }}
         uses: actions/upload-artifact@v4
         with:
           name: Executed notebooks
--- a/.github/workflows/notebook-test.yml
+++ b/.github/workflows/notebook-test.yml
@@ -25,37 +25,17 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Exit early on forks
-        if: ${{ github.event.pull_request.head.repo.full_name != github.repository }}
-        shell: python
-        run: |
-          raise SystemExit(
-            "We can't run this test on pull requests from forks."
-            "\n\n"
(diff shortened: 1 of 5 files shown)


Eric-Arellano:
I used https://github.com/Qiskit/documentation/pull/2063 to experiment with this.

.github/workflows/notebook-test.yml
```diff
@@ -76,46 +56,33 @@ jobs:
...
-              "notebook from this run by clicking \"Summary\" on the upper-left "
-              "of this page, and downloading the \"Executed notebooks\" artifact."
-              "\n\n"
-              "For more information, see https://github.com/Qiskit/documentation/blob/main/README.md#lint-notebooks"
-            )
-            sys.exit(1)
-
       - name: Execute notebooks
-        if: "!cancelled()"
+        if: ${{ success() && !cancelled() }}
```
frankharkins, line 60: I think this is now just the default behaviour if an `if` condition isn't included. For context, we added `!cancelled()` for https://github.com/Qiskit/documentation/pull/897#pullrequestreview-1904336782

```suggestion
```

.github/workflows/notebook-test.yml
```diff
@@ -76,46 +56,33 @@ jobs:
...
           if changed_notebooks and not changed_config:
               args.extend(changed_notebooks.split("\n"))
+          if os.environ["PR_REPOSITORY"] != os.environ["GITHUB_REPOSITORY"]:
+              args.append("--is-fork")
           subprocess.run(args, check=True)
 
       - name: Detect changed notebooks
         id: changed-notebooks
-        if: "!cancelled()"
+        if: ${{ success() && !cancelled() }}
```
frankharkins, line 78:
```suggestion
```

.github/workflows/notebook-test.yml
```diff
@@ -76,46 +56,33 @@ jobs:
...
         run: |
           echo "CHANGED_NOTEBOOKS<<EOF" >> $GITHUB_OUTPUT
           git diff --name-only >> $GITHUB_OUTPUT
           echo "EOF" >> $GITHUB_OUTPUT
 
       - name: Upload executed notebooks
-        if: "!cancelled()"
+        if: ${{ success() && !cancelled() }}
```
frankharkins, line 85:
```suggestion
```

frankharkins (approved):
Thanks! I see quite a few formatting changes, is this from a linter? Maybe we should add it to avoid noise in future PRs

Eric-Arellano:
> I see quite a few formatting changes, is this from a linter?

Yeah, I ran Black to fix some new code I added, then it changed other parts. Pardon the noise.

> Maybe we should add it to avoid noise in future PRs

Eh, I've thought about it in the past and am unsure because it adds even more things to worry about in our repo. We have so little Python code I'm not convinced it's worth it.

(If we were using a tool like Pantsbuild or Bazel, I would 100% have set it up a long time ago because it would allow us to `pants fmt ::` to format all code across all languages without having to memorize each different workflow.)

PR #2066, opened 2024-10-02
Fix rendering of times symbol on job limits page

blakejohnson:
The unicode times symbol was rendering as an $x$. Trying to replace it with $\times$.


abbycross:
My local preview shows that your change fixed the problem
Screenshot 2024-10-02 at 1 43 49 PM

abbycross (approved):

PR #2068, opened 2024-10-02
Generate Qiskit 1.2.4

Eric-Arellano:
The release notes rendered well locally with `./start --apis`.


mtreinish:
We ended up having to yank 1.2.3 because there was a bug in the fix that slipped through a hole in our CI coverage. We only discovered it after the release was live, so it might make sense to wait until 1.2.4. There won't be any changes except for adding additional ci coverage to close the hole and changing a `!= "0"` to "!= 0`.

kaelynj (approved):
LGTM :rocket:

frankharkins (approved):

PR #2069, opened 2024-10-02
Handle Runtime dev docs' CI having integration test failures

Eric-Arellano:
The cron job was failing because Runtime's recent `main` branch CI runs have failing integration tests. Those are irrelevant to the docs build, so we should tolerate it.

### DIFF
--- a/scripts/js/commands/api/syncDevDocs.ts
+++ b/scripts/js/commands/api/syncDevDocs.ts
@@ -47,10 +47,10 @@ zxMain(async () => {
 
   await updateHtmlArtifacts({ qiskitUrl, runtimeUrl });
 
-  if (!qiskitIsRc) {
+  if (qiskitUrl) {
     await regenDocs("qiskit", qiskitVersion);
   }
-  if (!runtimeIsRc) {
+  if (runtimeUrl) {
     await regenDocs("qiskit-ibm-runtime", runtimeVersion);
   }
 });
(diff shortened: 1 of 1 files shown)


scripts/js/commands/api/syncDevDocs.ts
```diff
@@ -47,10 +47,10 @@ zxMain(async () => {
 
   await updateHtmlArtifacts({ qiskitUrl, runtimeUrl });
 
-  if (!qiskitIsRc) {
+  if (qiskitUrl) {
```
Eric-Arellano, line 50: This change has no impact because `qiskitUrl` is only set if `!qiskitIsRc`. I did it because 
a) I think it's clearer code, and 
b) I was manually commenting out the code that sets `qiskitUrl` and was annoyed that this conditional still was executing.

scripts/js/commands/api/syncDevDocs.ts
```diff
@@ -67,12 +67,25 @@ async function getArtifactUrl(
...
+  }
+
+  const urlProc =
     await $`gh api repos/qiskit/${repoName}/actions/runs/${runId}/artifacts --jq '.artifacts[0].archive_download_url'`;
-  return url.stdout.trim();
+  const url = urlProc.stdout.trim();
+  if (!url) {
+    throw new Error(
+      `Failed to get artifact URL for https://github.com/Qiskit/${repoName}/actions/runs/${runId}`,
+    );
```
frankharkins, line 86: Should we throw or just `console.error`? I assume throwing will kill regeneration for all packages, even if only one fails

Eric-Arellano: I think throw. If something is wrong, I want to know about it so that we can investigate. No one checks logs for cron jobs, so we'd risk an API not being updated for a long time.

We can always manually generate locally or loosen this if we find a good reason. But I'd rather start defensive.

frankharkins (approved):
Sounds good!

PR #2080, opened 2024-10-03
Clarify transpiler service limitations

### LINKED ISSUE #2034: Clarify transpiler service limitation

https://docs.quantum.ibm.com/guides/qiskit-transpiler-service

### Select all that apply

- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [X] other

### Describe the fix.

I was late to the party for https://github.com/Qiskit/documentation/pull/1941, so I'll leave my comments here:

> The Qiskit Transpiler Service can transpile up to 1 million two-qubit gates in either ai mode ai="true", ai="false", or ai="auto".

We should clarify that it's 1 million two-qubit gates _per job_, not per circuit. Also is it really necessary to say it works for `ai=nn` give this applies to all 3 `ai` options ?

> The maximum time allowed to run a transpilation process is 30 minutes.

Also should clarify it's per job. Users may not understand what a transpilation process is (e.g. the service may break it into multiple processes for all they know). 

> You must retrieve the transpilation result from the service within 20 minutes after the transpilation process ends. After 20 minutes, your job is removed.

I'd suggest changing this to say "After 20 minutes, your job result is discarded". Saying the job is removed gives the impression the job itself is cancelled. 

@cbjuan please keep me honest here! And we should discuss the 120-minute queueing time limit 🙂

beckykd:
Closes #2034

### DIFF
--- a/docs/guides/qiskit-transpiler-service.mdx
+++ b/docs/guides/qiskit-transpiler-service.mdx
@@ -98,7 +98,7 @@ The following examples demonstrate how to transpile circuits using the Qiskit Tr
 
 Following are the most relevant limitations of the service:
 
-- The Qiskit Transpiler Service can transpile up to 1 million two-qubit gates per job in any `ai` mode.
+- The maximum number of two-qubit gates per circuit in a transpilation job in any `ai` mode is 1 million.
 - The maximum time allowed to run a transpilation process is 30 minutes per job.
 - You must retrieve the transpilation result from the service within 20 minutes after the transpilation process ends. After 20 minutes, the job result is discarded.
 - The maximum time a set of circuits can live in the internal queue while waiting to be transpiled is 120 minutes. After that time, if the job has not been transpiled, it is discarded.
--- a/docs/guides/qiskit-transpiler-service.mdx
+++ b/docs/guides/qiskit-transpiler-service.mdx
@@ -98,9 +98,9 @@ The following examples demonstrate how to transpile circuits using the Qiskit Tr
 
 Following are the most relevant limitations of the service:
 
-- The Qiskit Transpiler Service can transpile up to 1 million two-qubit gates in either `ai` mode `ai="true"`, `ai="false"`, or `ai="auto"`.
-- The maximum time allowed to run a transpilation process is 30 minutes.
-- You must retrieve the transpilation result from the service within 20 minutes after the transpilation process ends. After 20 minutes, your job is removed.
+- The Qiskit Transpiler Service can transpile up to 1 million two-qubit gates per job in any `ai` mode.
+- The maximum time allowed to run a transpilation process is 30 minutes per job.
+- You must retrieve the transpilation result from the service within 20 minutes after the transpilation process ends. After 20 minutes, the job result is discarded.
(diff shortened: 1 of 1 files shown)


jyu00 (approved):
Thanks @beckykd!

docs/guides/qiskit-transpiler-service.mdx
```diff
@@ -98,9 +98,9 @@ The following examples demonstrate how to transpile circuits using the Qiskit Tr
 
 Following are the most relevant limitations of the service:
 
-- The Qiskit Transpiler Service can transpile up to 1 million two-qubit gates in either `ai` mode `ai="true"`, `ai="false"`, or `ai="auto"`.
-- The maximum time allowed to run a transpilation process is 30 minutes.
-- You must retrieve the transpilation result from the service within 20 minutes after the transpilation process ends. After 20 minutes, your job is removed.
+- The Qiskit Transpiler Service can transpile up to 1 million two-qubit gates per job in any `ai` mode.
```
cbjuan, line 101:
```suggestion
- The Qiskit Transpiler Service can transpile up to 1 million two-qubit gates per any circuit part of a transpilation job in any `ai` mode.
```

I know the suggestion is not clear wording, yet the point is that it will fail if any of the circuits submitted as part of the same transpilation job to the transpiler service contains > 1 M two-qubit gates

cbjuan (changes requested):

docs/guides/qiskit-transpiler-service.mdx
cbjuan: Basically a transpilation job could contain N circuits. If any of those N circuits has > 1M two-qubit gates, it will fail

With the current text (IMO), someone could think that the job will fail if it contains >1M two-qubit gates in total (sum of two-qubit gates per circuit or similarly)

abbycross: How about "The maximum number of two-qubit gates _per circuit_ in a transpilation job is 1 million."

abbycross (approved):

docs/guides/qiskit-transpiler-service.mdx
cbjuan: > > Basically a transpilation job could contain N circuits. If any of those N circuits has > 1M two-qubit gates, it will fail
> > With the current text (IMO), someone could think that the job will fail if it contains >1M two-qubit gates in total (sum of two-qubit gates per circuit or similarly)
> 
> Oh interesting, because I thought it's per job (since that's how Qiskit transpiler works). My apologies.

no worries! Basically my comment was to be more accurate. Thanks for raising the issue @jyu00 :)

cbjuan: could we use @abbycross suggestion instead of my own one, @beckykd ?

https://github.com/Qiskit/documentation/pull/2080#discussion_r1787838987

beckykd: @cbjuan - do we want to keep the part about any ai mode?

cbjuan: I think so. Thank you!

cbjuan (approved):

Issue #2087, opened 2024-10-06
Fix incorrect `meas` attribute in example code for sampler


srdtrk:
### URL to the relevant documentation

https://docs.quantum.ibm.com/guides/get-started-with-primitives#4-invoke-the-sampler-and-get-results

### Select all that apply

- [ ] typo
- [x] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

## Description

The code snippet in the [Qiskit Runtime Sampler documentation](https://docs.quantum.ibm.com/guides/get-started-with-primitives#4-invoke-the-sampler-and-get-results) appears to contain an error when retrieving measurement results.

The current example provided:

```py
print(f"Counts for the meas output register: {pub_result.data.meas.get_counts()}")
```

produces an error because DataBin does not have an attribute named `meas`.

To successfully retrieve the counts, the following code should be used instead:

```py
print(f"Counts for the meas output register: {pub_result.data.c.get_counts()}")
```

This change allows the code to run correctly by accessing the appropriate slot for the measurement output.

I’d love to open a PR to contribute this fix if agreed.

abbycross:
Hello and thanks for opening the issue! Yes, please do go ahead and work on this - I've assigned you.

jyu00:
The field after `pub_result.data` is the name of the classical register used. In the [code example](https://docs.quantum.ibm.com/guides/get-started-with-primitives#2-create-a-circuit) above, it uses `circuit.measure_all()`, which by default name the classical register `meas`. 

Now it's possible that this is not true for all Qiskit versions. @srdtrk, do you know which Qiskit version you were using when you encountered this?

srdtrk:
Thanks for info @jyu00. I'm actually using qiskit `v1.2.4` and qiskit-ibm-runtime `v0.30.0`. I'm not experienced enough in qiskit or python to actually find the code-snippet in qiskit. (Although the sampler should be in `qiskit-ibm-runtime`).  This is the code snippet that works for me in my notebook:
```py
# make a new simulator object
backend = AerSimulator()
sampler = Sampler(backend)

# run the experiment
job = sampler.run([qcEx1])

# get the results
result = job.result()[0]

# interpret the results as a "counts" dictionary
result.data.c.get_counts()
```
and the imports are as follows:
```py
### Import the required packages

from qiskit import QuantumCircuit
from qiskit_aer import AerSimulator
from qiskit_ibm_runtime import QiskitRuntimeService
[...]

PR #2101, opened 2024-10-10
fix: incorrect 'meas' attribute in example code for Sampler

srdtrk:
fixes: #2087


jyu00:
Please see my [comment](https://github.com/Qiskit/documentation/issues/2087#issuecomment-2406097717) in the original issue. `meas` is the default name used in the latest Qiskit (1.2.4).

PR #2137, opened 2024-10-15
Switch `CodeAssistantPrompt` format

frankharkins:
We're changing how the prompt works. We'll have to merge with the broken preview as we need to merge the MDX changes and the component changes to inner-source at the same time. I've copied the files over to inner-source to test it out with the new component and make sure the syntax is correct.


Eric-Arellano (dismissed):
Looks good! Please wait to merge until closed source is ready.

Eric-Arellano (approved):

PR #2138, opened 2024-10-15
Fix forks running job cancellation code in PRs

Eric-Arellano:
Fixes https://github.com/Qiskit/documentation/pull/2062. 

While we now correctly detect in forks when no notebooks need to be run, we still were trying to call cleanup code, even though there are no jobs to clean up. That cleanup code requires logging into Runtime, so it fails on forks.

Instead, we should early exit if there are no notebooks to run.


kaelynj (approved):
LGTM :fire:

PR #2143, opened 2024-10-16
Fix dev docs sync detection of changed files

Eric-Arellano:
The dev docs sync PR is correctly running and regenerating the dev docs, but it's skipping the step to open PRs. This is because `@frankharkins` pointed out tj-actions/changed-files compares files across branches, rather than based on the current working directory.

### DIFF
--- a/.github/workflows/dev-docs-sync.yml
+++ b/.github/workflows/dev-docs-sync.yml
@@ -40,9 +40,9 @@ jobs:
           # images in public/images or changes to scripts/config/api-html-artifacts.json
           # to avoid noisy PRs. It's common for images to change because some of the
           # images are generated non-deterministically.
-          files: docs/api/**/*.mdx
+          files: "docs/api/**/*.mdx"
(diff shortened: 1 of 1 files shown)


frankharkins:
I wonder if this is because the "changed files" action compares files changed across checked-out branches, rather than in the current working directory? Could you try `git diff` instead? It works correctly with the notebook CI.

https://github.com/Qiskit/documentation/blob/3d70b115f7d07d8f2e5ea3bf13a4f8d1ff112ba0/.github/workflows/notebook-test-cron.yml#L35-L41

Eric-Arellano:
> I wonder if this is because the "changed files" action compares files changed across checked-out branches, rather than in the current working directory? Could you try git diff instead? It works correctly with the notebook CI.

Oh...duh. Thanks for pointing that out!

frankharkins (approved):
> Oh...duh

The action _is_ called "changed files", I only worked it out by looking at the notebook test action.

PR #2147, opened 2024-10-17
Fix external link checker exception handling

Eric-Arellano:
We weren't properly catching errors due to reusing the symbol `error`. That resulted in a test recently saying that a link we expected to 404 was instead a valid link.

This PR also improves the tests to be deterministic by mocking `fetch()`. This removes the risk of flakiness and makes our test faster.

### DIFF
--- a/scripts/js/lib/links/ExternalLink.test.ts
+++ b/scripts/js/lib/links/ExternalLink.test.ts
@@ -20,21 +20,38 @@ test("ExternalLink constructor ignores anchors", () => {
 });
 
 test.describe("ExternalLink.check()", () => {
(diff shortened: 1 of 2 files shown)


scripts/js/lib/links/ExternalLink.test.ts
```diff
@@ -20,21 +20,38 @@ test("ExternalLink constructor ignores anchors", () => {
 });
 
 test.describe("ExternalLink.check()", () => {
+  let originalFetch: typeof global.fetch;
```
Eric-Arellano, line 23: Note that this approach would not be safe if Playwright ran the tests inside a file in parallel. However, it runs them sequentially.

PR #2154, opened 2024-10-21
Revert api-redirect-data.json

Eric-Arellano:
Reverts https://github.com/Qiskit/documentation/pull/2125. Turns out, we didn't need to take that approach so we can instead use the simpler generateHistoricalRedirects.ts.


Eric-Arellano:
> We still need the custom redirects in the new file, right?

Those already exist via `generateHistoricalRedirects.ts`. I was recreating them.

frankharkins (approved):
Ok nice, thanks

PR #2156, opened 2024-10-21
Add table of when to use each method

### LINKED ISSUE #2041: Recommendation on error mitigation techniques
### URL, if applicable

_No response_

Even though we have a [page](https://docs.quantum.ibm.com/guides/error-mitigation-and-suppression-techniques) describing the different error suppression / mitigation methods, not all users can translate that into the best ones to use for their workloads. It'd be nice to give users some pointers on when to use what, and the cost associated. For example, dynamical decoupling is mostly "free" (does not incur additional QPU time), but should only be used when there is idle time in the circuit. In addition, a longer DD sequence like XY4 should be used if the idle time is long, vs XX for shorter time. 

abbycross:
Closes #2041 . Thank you @miamico for providing this content!


docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -278,6 +278,23 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "a6c87cd7",
+   "metadata": {},
+   "source": [
+    "## When to use each method\n",
+    "\n",
+    "| Method | When to use | Notes|\n",
+    "| --- | --- | --- |\n",
+    "| Dynamical decoupling | Use with circuits in which one or more qubits have long idle times | Circuits with dense structure of two-qubit gates (for example, see the paper [Evidence for the utility of quantum computing before fault tolerance](https://doi.org/10.1038/s41586-023-06096-3)). Also, Heron devices are less susceptible to the dominating noise that is suppressed by simple DD sequences (cross-talk) |\n",
```
beckykd, line 290:
```suggestion
    "| Dynamical decoupling | Use with circuits in which one or more qubits have long idle times | Circuits with dense two-qubit gates structures (for example, see the paper [Evidence for the utility of quantum computing before fault tolerance](https://doi.org/10.1038/s41586-023-06096-3)). Also, Heron devices are less susceptible to the dominating noise that is suppressed by simple DD sequences (cross-talk) |\n",
```

docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -278,6 +278,23 @@
...
+    "| Pauli twirling | Always use | There is a trade-off in the execution time (devices execute more shots faster than more circuits); therefore, using Pauli twirling with enough randomization (even when lowering the shots to keep the total number of shots constant) will result in a slowdown |\n",
```
beckykd, line 291:
```suggestion
    "| Pauli twirling | Always use | There is a trade-off with the execution time (devices execute more shots faster than more circuits); therefore, using Pauli twirling with enough randomization (even when lowering the shots to keep the total number of shots constant) causes a slowdown |\n",
```

docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -278,6 +278,23 @@
...
+    "| TREX | Always use, if already Pauli twirling the circuits | There is some indication that TREX can lead to unphysical expectation values in some cases, but the risk is likely worth it, given the strength of measurement noise and low overhead |\n",
+    "| ZNE | Use if the noisy expectation value measured without ZNE is at least 30% of the ideal value (otherwise, extrapolation is likely to break down, leading to biased results)| Increases runtime by X times (where X is the number of noise factors) |\n",
```
beckykd, line 293:
```suggestion
    "| ZNE | Use if the noisy expectation value measured without ZNE is at least 30% of the ideal value. Otherwise, extrapolation is likely to break down, leading to biased results.| Increases runtime by X times, where X is the number of noise factors. |\n",
```

docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -278,6 +278,23 @@
...
+    "| PEA | Use if the noisy expectation value measured without ZNE is at least 30% of the ideal value (otherwise extrapolation is likely to break down, leading to biased results); and if the number of unique layers in the circuit is relatively small (fewer than 20)| Increases runtime by X times (where X is the number of noise factors), and needs time for learning the noise model |\n",
```
beckykd, line 294: So many phrases - I feel like this would result in poor translation!  I'm trying to rewrite to make the structure a little more simple.

beckykd, line 294:
```suggestion
    "| PEA | Use if the noisy expectation value measured without ZNE is at least 30% of the ideal value (otherwise extrapolation is likely to break down, leading to biased results); and if there are fewer than 20 unique layers in the circuit.| Increases runtime by X times, where X is the number of noise factors, and needs time to learn the noise model |\n",
```

docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -278,6 +278,23 @@
...
+    "| PEC | Use when you need an unbiased estimate | In most cases PEA has a better overhead in terms of number of total shots (across randomizations) |"
```
beckykd, line 295: Should these all have periods at the ends?

beckykd (approved):
Our users will appreciate this new content!  Thanks!

docs/guides/error-mitigation-and-suppression-techniques.ipynb
abbycross: I think the rule is just to be consistent - and since most of the table's content did not have periods, I made the rest of it the same, and only added a period between sentences when there was more than one sentence in a cell. But if you think it looks better to have all periods, I don't mind that either.

docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -278,6 +278,23 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "a6c87cd7",
+   "metadata": {},
+   "source": [
+    "## When to use each method\n",
+    "\n",
+    "| Method | When to use | Notes|\n",
+    "| --- | --- | --- |\n",
+    "| Dynamical decoupling | Use with circuits in which one or more qubits have long idle times | Circuits with dense two-qubit gates structures (for example, see the paper [Evidence for the utility of quantum computing before fault tolerance](https://doi.org/10.1038/s41586-023-06096-3)). Also, Heron devices are less susceptible to the dominating noise that is suppressed by simple DD sequences (cross-talk) |\n",
```
abbycross, line 290: @miamico checking that I'm citing the right paper here?

jyu00, line 290: > Use with circuits in which one or more qubits have long idle times

How do I know if my qubits have long idle times? 

> Circuits with dense two-qubit gates structures (for example, see the paper [Evidence for the utility of quantum computing before fault tolerance](https://doi.org/10.1038/s41586-023-06096-3)).

What is this sentence trying to say? That I can find an example of a dense circuit in the paper? How do I know if my circuit is considered "dense" or not? 

In addition, what kind of DD sequences should I use?

docs/guides/error-mitigation-and-suppression-techniques.ipynb
```diff
@@ -278,6 +278,23 @@
...
+    "| Pauli twirling | Always use | There is a trade-off with the execution time (devices execute more shots faster than more circuits); therefore, using Pauli twirling with enough randomization (even when lowering the shots to keep the total number of shots constant) causes a slowdown |\n",
```
jyu00, line 291: > Always use

That kind of contradicts with the note that says there is a trade off. E.g. it's really "always use if you can afford it" lol. How do users know whether to make that trade off?

jyu00: And is gate twirling always necessary if I'm not doing mitigation?

jyu00 (commented):
I think these recommendations are a bit too general. Perhaps we need an actual tutorial that can show users, for example, how to identify idling time in your circuits, instead of a simple summary table. Users that need this kind of recommendation typically require more hand holding.

abbycross:
@jyu00 is looking for more context around these recommendations, and has proposed a tutorial on Learning to serve the purpose. Closing this PR for now.

Issue #2159, opened 2024-10-21
Misleading error message suggestion for photos in internal link checker


Eric-Arellano:
```
❌ Could not find link '../../images/guides/img.png'. Appears in:
    docs/guides/my-guide.ipynb    ❓ Did you mean 'public/images/guides/img'?
```

The URL should be `/images/guides/img.png`.

PR #2163, opened 2024-10-21
Fix suggestion for image link checker

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/2159.

This change also results in more suggestions happening than we had before, as shown in the tests.

### DIFF
--- a/scripts/js/lib/links/InternalLink.ts
+++ b/scripts/js/lib/links/InternalLink.ts
@@ -114,14 +114,14 @@ export class InternalLink {
     let suggestionPath: String = "";
     let suggestionPathAnchors: string[] = [];
 
-    const possiblePaths = this.possibleFilePaths(originFile);
-    const pathNoExtension = possiblePaths[0].replace(/\.[^\/.]+$/, "");
-
     existingFiles.forEach((file) => {
-      let score = levenshtein.get(pathNoExtension, file.path);
+      const candidatePath = file.path.startsWith("public/")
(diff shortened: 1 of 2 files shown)


scripts/js/lib/links/InternalLink.ts
```diff
@@ -114,14 +114,14 @@ export class InternalLink {
...
     let suggestionPathAnchors: string[] = [];
 
-    const possiblePaths = this.possibleFilePaths(originFile);
-    const pathNoExtension = possiblePaths[0].replace(/\.[^\/.]+$/, "");
-
     existingFiles.forEach((file) => {
-      let score = levenshtein.get(pathNoExtension, file.path);
+      const candidatePath = file.path.startsWith("public/")
+        ? file.path.replace(/^public/, "")
+        : file.path.replace(/\.[^\/.]+$/, "").replace(/^docs/, "");
```
Eric-Arellano, line 120: The first `replace` removes the file extension, which would be wrong to do for `public/`.

scripts/js/lib/links/InternalLink.ts
```diff
@@ -114,14 +114,14 @@ export class InternalLink {
...
+      let score = levenshtein.get(this.value, candidatePath);
```
Eric-Arellano, line 121: `this.value` is the problematic link. The anchor is stripped and stored separately in `this.anchor`.

frankharkins (approved):
Nice, deduplicated some code too

PR #2167, opened 2024-10-22
Add Qiskit addon API docs as a baseline

Eric-Arellano:
We're in the process of adding six Qiskit addon API docs. This PR sets up the initial projects so that it's easier for us to iterate on them, such as running `npm run gen-api -- -a <path-to-html>` to see locally how changes to Sphinx config impact the MDX pipeline. 

The docs are _not_ ready to go live and they will not yet be synced by closed source or deployed.

The quality issues will fixed in the project repositories, such as https://github.com/Qiskit/qiskit-addon-obp/pull/24.

## API naming scheme

We call the API by their full name, without their abbrevation. For example, "Operator Backpropagation" rather than "OBP" or "Operator Backpropagation (OBP)".

We do not explicitly mention "Qiskit addon" in the API name. It's already evident these are addons by looking at the left ToC which shows the full module path like `qiskit_addon_obp`.

image_720

## URL scheme

Each package's URL matches their PyPI name, like `qiskit-addon-obp`.

Note that we may want to normalize the URLs for individual pages to follow SEO best practices, e.g. `/api/qiskit-addon-obp/qiskit_addon_obp.utils.simplify` would be `/api/qiskit-addon-obp/qiskit-addon-obp-utils-simplify`. We would make this change in a follow up PR.


Eric-Arellano:
The commits are distinct.

scripts/js/lib/api/Pkg.ts
```diff
@@ -86,7 +92,6 @@ export class Pkg {
       return new Pkg({
         ...args,
         title: "Qiskit SDK",
-        name: "qiskit",
```
Eric-Arellano, line 89: No need to set this. `name` was already set in `..args`. It was validated that it equals `qiskit` via the `if` statement right above.

scripts/js/commands/checkMetadata.ts
```diff
@@ -72,6 +72,7 @@ const main = async (): Promise<void> => {
   const mdErrors = [];
   for (const file of mdFiles) {
     if (ALLOWED_VIOLATIONS.has(file)) continue;
+    if (file.includes("api/qiskit-addon-")) continue;
```
frankharkins, line 75: Why are they allowed to violate MDX?

frankharkins (approved):

PR #2169, opened 2024-10-23
Algorithmiq TEM documentation: change parameter in example

ludmilaasb:
Updating the Algorithmiq TEM documentation in the Example section. The parameter `max_execution_time` was removed.


Eric-Arellano (commented):
Would it be helpful to demonstrate another option? It looks like there will be no examples of setting `options` anymore, which might be fine but I'm curious what you think.

matteoacrossi:
We can show setting default_precision maybe. The issue with the execution time is that 60s is too little. So the alternative is to increase that.

Eric-Arellano (approved):

PR #2171, opened 2024-10-23
make paths safer  (tweak to latest updates pg)

abbycross:


Eric-Arellano (approved):
In general, I recommend always using absolute paths.

Should this bring in the other copy editing changes Becky suggested?

Issue #2174, opened 2024-10-23
Normalize URLs for Addon API docs for better SEO


Eric-Arellano:
Before: /api/qiskit-addon-obp/qiskit_addon_obp.utils.MyClass
After: /api/qiskit-addon-obp/utils-my-class

This was a recommendation by the SEO team because the search engines expect URLs to comply with the above format. With Addons, we can get this format right before the docs go live to avoid having to set up redirects in the future.

This mechanism should only be enabled for the new Addon projects. It should be enabled in `Pkg.ts`.

Ensure objects.inv is updated correctly. Warning that we ignore objects.inv for the Addons for now because there is so much noise from the non-API docs being included in them.

PR #2177, opened 2024-10-24
Shorten and use kebab-case for Addon URLs

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/2174. We now convert URLs like `qiskit_addon_obp.MyClass` to `my-class`.

This is safe to change because the Addons are not yet public. We may eventually want to do this for Qiskit, Runtime, and Transpiler Service - for now, they stay the same.


scripts/js/lib/api/updateLinks.ts
```diff
@@ -67,36 +69,46 @@ export function normalizeUrl(
...
-  const [path, hash] = last(urlParts)!.split("#") as [
+  const [page, hash] = last(urlParts)!.split("#") as [
     string,
     string | undefined,
   ];
 
+  const normalizedPage = kwargs.kebabCaseAndShorten
+    ? kebabCaseAndShortenPage(page, kwargs.pkgName)
+    : page;
+  const normalizedBase = [...initialUrlParts, normalizedPage].join("/");
```
Eric-Arellano, line 80: Lmk if you can think of a better name for this. It's the entire normalized URL minus the anchor.

Eric-Arellano (commented):
The three commits are distinct

frankharkins (approved):
Looks good! Thanks for all the tests

PR #2192, opened 2024-10-28
Mention mthree on configure-error-mitigation

### LINKED ISSUE #2104: Mention M3 with Sampler

_No response_

### Select all that apply

- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [x] other

### Describe the fix.

One complaint we recently received from the users was that V2 Sampler no longer supports error mitigation. While that's unlikely to change anytime soon, we could recommend people to use [mthree](https://github.com/Qiskit-Extensions/mthree), which was used in V1 Sampler and is still suitable for many use cases.

abbycross:
Closes #2104

### DIFF
--- a/docs/guides/configure-error-mitigation.mdx
+++ b/docs/guides/configure-error-mitigation.mdx
@@ -58,7 +58,9 @@ applied at each resilience level.
 
 ## Configure Estimator with resilience levels
 
-You can use resilience levels to specify error mitigation techniques, or you can set custom techniques individually as described in [Custom error settings.](#advanced-error) You cannot specify resilience levels in Sampler. However, you can set custom techniques individually.
+You can use resilience levels to specify error mitigation techniques, or you can set custom techniques individually as described in [Custom error settings.](#advanced-error) 
+
+Sampler does not support error mitigation, but you can use the [mthree](https://qiskit.github.io/qiskit-addon-mthree/) (matrix-free measurement mitigation) package to do error mitigation locally.
 
 <details>
 <summary>Resilience Level 0</summary>
--- a/docs/guides/configure-error-mitigation.mdx
+++ b/docs/guides/configure-error-mitigation.mdx
@@ -11,7 +11,13 @@ results in quantum pre-processing overhead related to model training and
 classical post-processing overhead to mitigate errors in the raw results
 by using the generated model.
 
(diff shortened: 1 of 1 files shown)


docs/guides/configure-error-mitigation.mdx
```diff
@@ -58,7 +58,9 @@ applied at each resilience level.
 
 ## Configure Estimator with resilience levels
 
-You can use resilience levels to specify error mitigation techniques, or you can set custom techniques individually as described in [Custom error settings.](#advanced-error) You cannot specify resilience levels in Sampler. However, you can set custom techniques individually.
+You can use resilience levels to specify error mitigation techniques, or you can set custom techniques individually as described in [Custom error settings.](#advanced-error) 
+
+Sampler does not support error mitigation, but you can use the [mthree](https://qiskit.github.io/qiskit-addon-mthree/) (matrix-free measurement mitigation) package to do error mitigation locally.
```
abbycross, line 63: linking to the ecosystem docs here - could link to the github repo instead or in addition - lmk if there is a preference for one of those options.

Eric-Arellano, line 63: Should it be "The Sampler primitive"?

abbycross: We use them interchangeably

jyu00: I find it a bit odd to see this paragraph under `Configure Estimator with resilience levels`, since this is not related to resilience levels. However, I cannot decide if it's better under `Custom error settings, since the code example for Sampler there shows error _suppression_, not mitigation.

beckykd: I can't figure out how to add a comment to a part of the file that wasn't edited, but maybe this would work: 
1 - At the top where we say "Primitives support several error mitigation techniques...", change that to "Estimator supports..." (since sampler doesn't support any)
2 - Move this m3 content after that updated paragraph as a note.

beckykd, line 63:
```suggestion
Sampler does not support error mitigation, but you can use the [mthree](https://qiskit.github.io/qiskit-addon-mthree/) (matrix-free measurement mitigation) package to perform error mitigation locally.
```

beckykd: I'm not a giant fan of "perform" either, but "do" sounds weird to me.

beckykd (approved):
Just a couple of suggestions.  Otherwise, LGTM

docs/guides/configure-error-mitigation.mdx
abbycross: I definitely like `perform` better than `do`. 👍

abbycross: I was going to suggest it being a note too. @jyu00 do you have a preference for one of Becky's suggestions?

jyu00: I think Becky was suggesting we do both? Change the top to say just Estimator, and follow that with a note saying you can use M3 for Sampler. I do like this approach.

abbycross: ah yes, I don't know why I assumed it was one or the other 🤣 both it is!

abbycross:
@jyu00 ready for final review... I removed the text from the bottom and put it in a note near the top -- also changed primitives to estimator:
Screenshot 2024-10-30 at 2 58 39 PM

PR #2197, opened 2024-10-29
Add linting with ruff

frankharkins:
Adds linting and formatting with [Ruff](https://docs.astral.sh/ruff).

This caught a couple of undefined variables, but it also enforces consistent code styling across our code examples. You may find it easier to review this PR commit-wise.


docs/guides/qunasys-quri-chemistry.ipynb
```diff
@@ -357,14 +360,10 @@
...
     "\n",
     "while True:\n",
-    "    try:\n",
-    "        print(datetime.datetime.now(), job.status())\n",
-    "        time.sleep(10)\n",
-    "        if job.status() in [\"ERROR\", \"DONE\"]:\n",
-    "            break\n",
-    "    except:\n",
-    "        print(\"Retrieve job status failed.\")\n",
-    "        pass\n",
```
frankharkins, line 367: I removed the exception as we actually do want the notebook to halt on error here, otherwise `job.result()` will fail immediately after. Same for L559.

docs/ruff.toml
```diff
@@ -0,0 +1 @@
+line-length=78
```
frankharkins, line 1: From experimentation, 78 chars is when the horizontal scroll appears on the website.

tox.ini
```diff
@@ -11,10 +11,17 @@ setenv = PYDEVD_DISABLE_FILE_VALIDATION=1
...
+  squeaky==0.7.0
+  ruff==0.7.1
+
+commands =
+    lint: ruff check {posargs:docs}
+    lint: ruff format --check {posargs:docs}
+    lint: squeaky --check --no-advice {posargs:docs}
+    fix: squeaky {posargs:docs}
+    fix: ruff format {posargs:docs}
+    fix: ruff check --fix {posargs:docs}
```
frankharkins, line 24: Unlike `ruff format` and `squeaky`, `ruff check` can't always be autofixed. This affects the order we want to call these commands.

For `lint` (i.e. check but don't fix), we want to show anything requiring manual attention first. Formatting can be done automatically. This is the other way round for `fix`: We want to fix everything we can, then hand over to the user for anything requiring manual attention.

docs/guides/qedma-qesem.ipynb
```diff
@@ -73,6 +73,22 @@
...
+    "tags": [
+     "remove-cell"
+    ]
+   },
+   "outputs": [],
+   "source": [
+    "# This cell is hidden from users\n",
+    "# It stops the linter complaining about undefined variables\n",
+    "# ruff: noqa: F821"
+   ]
```
frankharkins, line 90: Fixing this notebook is out of scope of this PR. Tracked in #2198.

Eric-Arellano:
Thanks for doing this! Maybe split up the Ruff formatting into a dedicated PR? It's useful in code review & future Git blame to differentiate between purely style changes vs semantic changes.

Eric-Arellano (approved):
Thanks for doing this!

PR #2204, opened 2024-10-29
Fix hash capitalization in API links

Eric-Arellano:
Fixes https://github.com/Qiskit/documentation/pull/2177. We weren't normalizing the hash in all cases.

### DIFF
--- a/scripts/js/lib/api/updateLinks.test.ts
+++ b/scripts/js/lib/api/updateLinks.test.ts
@@ -26,10 +26,11 @@ test.describe("updateLinks", () => {
 [link3](qiskit_ibm_runtime.RuntimeJob.job#wut)
 [link4](../stubs/qiskit_ibm_runtime.RuntimeJob)
 [link5](../apidocs/qiskit_ibm_runtime.RuntimeJob#qiskit_ibm_runtime.RuntimeJob)
-[link6](qiskit_ibm_runtime.RuntimeJob)
-[link7](#qiskit_ibm_runtime.RuntimeJob.job)
-[link8](https://qiskit.org/documentation/apidoc/algorithms.html)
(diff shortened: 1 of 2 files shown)


scripts/js/lib/api/updateLinks.test.ts
```diff
@@ -26,10 +26,11 @@ test.describe("updateLinks", () => {
 [link3](qiskit_ibm_runtime.RuntimeJob.job#wut)
 [link4](../stubs/qiskit_ibm_runtime.RuntimeJob)
 [link5](../apidocs/qiskit_ibm_runtime.RuntimeJob#qiskit_ibm_runtime.RuntimeJob)
-[link6](qiskit_ibm_runtime.RuntimeJob)
```
Eric-Arellano, line 29: I added a new link 6, so updated the link numbers for all these

scripts/js/lib/api/updateLinks.ts
```diff
@@ -104,20 +108,10 @@ export function normalizeUrl(
...
       : parentName;
-    return [...initialUrlParts, normalizedParentName].join("/") + "#" + member;
+    url = [...initialUrlParts, normalizedParentName].join("/") + "#" + member;
   }
 
-  if (!hash) return normalizedUrlWithoutHash;
-
-  // Anchors generated from markdown headings are always lower case but, if these
-  // headings are API references, Sphinx sometimes expects them to include
-  // uppercase characters.
```
Eric-Arellano, line 114: This came from the helper function `lowerCaseIfMarkdownAnchor`. I accidentally didn't delete it last PR. Its implementation was more robust so I use that rather than inlining

frankharkins (approved):
Nice, thank you!

PR #2208, opened 2024-10-30
List package versions in executed notebooks

### LINKED ISSUE #922: Add a way to dynamically render the version of Qiskit, Runtime, and Aer to notebooks
Split out of https://github.com/Qiskit/documentation/issues/133.

We want to render something like:

```
qiskit: 1.0.1
qiskit-ibm-runtime: 0.20.0
qiskit-aer: 13.3.0
```

It should use the values from the virtual environment used to run the notebook, so the values should be auto-generated and come from e.g. `qiskit.__version__` or whatever the import is. That means this will need to be a code block rather than markdown block. Open question if we can make the formatting look nicer, though.

We discussed it being nice to not show Aer if Aer is not used in the file. We should see if we can automate that by seeing if Aer has already been imported earlier. Otherwise, we discussed if code authors manually mark it, but that is error-prone.

I imagine we'll need to copy-and-paste the same code snippet, rather than DRYing it with a helper function. That's because the notebooks need to run on their own and not depend on other notebooks. In the past, Qiskit itself had a version table to solve this problem; I think that's been removed?

frankharkins:
Adds the ability for `nb-tester` to list package versions in executed notebooks. Closes #922.

To add version information to a notebook, add a markdown cell with tag `version-info`. The script will replace the markdown in that cell with the version information. I've added this cell to one notebook and executed it to demonstrate (preview: https://qiskit.github.io/documentation/pr-2208/guides/plot-quantum-states). If we go with this approach, I can add empty cells to all notebooks and they'll be populated as they're run.

This is just plain markdown for now, but we could also consider putting it the information in a component or adding it to the notebook metadata so we can display it somewhere else (e.g. the right sidebar).


frankharkins:
I've added the tagged cell to most notebooks, the exception being the function notebooks. I also re-ran notebooks that don't need mocking to populate those with version information. The cells are still in the other notebooks (e.g. hello world), but will be invisible until those notebooks are next executed.

abbycross (approved):
Did we decide explicitly not to bold the title of the twistie? If so, it looks good to go! (I know we considered it both ways and I don't recall where we landed with it)

beckykd:
> Did we decide explicitly not to bold the title of the twistie?

I thought we decided to bold the title

frankharkins:
You're right, knew I'd missed something 😂

PR #2209, opened 2024-10-30
Regenerate addon API docs with reorganization

Eric-Arellano:
All five addons now use the structure we expect of having a flat list of module pages, along with an index page.

Thanks to this PR, we can now run `checkOrphanPages.ts` on the addon projects.


Eric-Arellano:
The first commit is generated. Second commit is manual.

frankharkins (approved):

Issue #2210, opened 2024-10-31
Broken link for inlined classes in addon docs


Eric-Arellano:
https://github.com/Qiskit/documentation/blob/8b17877094259e36dff221c919781ce38fa9d636/scripts/js/lib/links/ignores.ts#L335-L342

The issue is that the header for methods doesn't repeat the class name. The anchor is `p_norm`, not `TransactionBudget.p_norm`. Note the original Sphinx is set to `qiskit_addon_obp.utils.truncating.TruncationErrorBudget.p_norm`, which we do preserve via the MDX component's ID.

This happens because:

https://github.com/Qiskit/documentation/blob/8b17877094259e36dff221c919781ce38fa9d636/scripts/js/lib/api/updateLinks.ts#L95-L99

Three ways to fix this:

1. Add a span  to the heading with the `id` set to what's expected
2. Have updateLinks.ts realize it's an inlined class and remove the class
3. Don't have updateLinks.ts remove the prefix `qiskit_addon_obp.utils.truncating.`. Ensure the link checker correctly loads `id="qiskit_addon_obp.utils.truncating.TruncationErrorBudget.p_norm"` from the `<Function>` component.

PR #2230, opened 2024-11-01
Link checker parses headings inside component

Eric-Arellano:
The link checker was failing to detect inline class methods (related to https://github.com/Qiskit/documentation/issues/2210) because it does not peek inside `<Class>` components.

To fix this, we now use our own parsing code for headings. 

It's possible there will be future edge cases this PR does not account for in its heading parsing. Our link checker will fail in that case and we can forward fix.

### DIFF
--- a/scripts/js/lib/links/extractLinks.test.ts
+++ b/scripts/js/lib/links/extractLinks.test.ts
@@ -29,6 +29,16 @@ test("parseAnchors()", () => {
 
   ## \`code-header\`
 
+  ## Header.with periods-and wild! punctuation?? and numbers 1234 8 (parentheses)
+
+  ## header_using\_underscores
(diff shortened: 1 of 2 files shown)


frankharkins (approved):
Thanks! This solution looks good. Though if we hit more problems like this we might want to consider spinning up the server and making requests. It would be slower, but it'd be guaranteed correct.

Eric-Arellano:
> Though if we hit more problems like this we might want to consider spinning up the server and making requests. It would be slower, but it'd be guaranteed correct.

I don't think so - think about how slow checkPagesRender.ts is. It's important our link checker handles all historical API docs as well; the closed source link checker doesn't do that because its live render approach is too slow.

PR #2241, opened 2024-11-04
Add versions to API metadata & autogenerate index page metadata

Eric-Arellano:
An issue with our API docs is confusing which version specific pages correspond to because we have so many duplicates. That is especially an issue with SEO, but also rich previews & tab titles. So, this PR adds the version to both the `title` and `description`.

We use minor versions rather than patch versions because search engines can be slow to update and we don't want their metadata to be out-of-date. Likewise, we call it `(latest version)` and `(dev version)` for this same reason.

Finally, this PR automates setting metadata for the index page whereas before we hardcoded it.


Eric-Arellano:
CI fails on this because pagesRender has too many changed files

frankharkins (approved):
Sorry I missed this, looks great!

PR #2252, opened 2024-11-06
Expand cspell ignore instructions

beckykd:


README.md
```diff
@@ -309,7 +309,7 @@ npm run check
 
 There are two ways to deal with cSpell incorrectly complaining about a word, such as abbreviations.
 
-1. Ignore the word in the local markdown file by adding a comment to the file, like below. The word is not case-sensitive, and the comment can be placed anywhere.
+1. Ignore the word in the local markdown file by adding a comment to the file, like below. The word is not case-sensitive, and the comment can be placed anywhere (inside a markdown cell, in the source part). Note that in the text editor, each line is surrounded by quotes and ends with /n.  So it looks like this:  "{/* cspell:ignore hellllooooo, ayyyyy */}/n"
```
Eric-Arellano, line 312: These new instructions only apply if it's a Jupyter notebook. I recommend something like this:

If the file is a Jupyter notebook, add the comment inside a markdown cell in the `source` part. Each line should be surrounded by quotes and end with `\n`. 

Then you could add a second example code block with the language set to `json`:

```json
  {
   "cell_type": "markdown",
   "id": "552b1077",
   "metadata": {},
   "source": [
    "# Hello world",
    "{/* cspell:ignore helloooooo */}/n"
   ]
  },
```

frankharkins:
```suggestion
1. Ignore the word in the local markdown file by adding a comment to the file, like below. The word is not case-sensitive, and the comment can be placed anywhere (inside a markdown cell, in the source part). Note that in the text editor, each line is surrounded by quotes and ends with /n.  So it looks like this:  "{/* cspell:ignore hellllooooo, ayyyyy */}\n"
```

Eric-Arellano: @beckykd this is still unsolved. Frank's suggestion only fixed a typo and didn't yet apply my suggestion

README.md
```diff
@@ -444,7 +444,7 @@ Refer to the [Common Markdown syntax](https://commonmark.org/) for a primer on M
 
 ## How to add a new page
 
-Choose which existing folder from `docs/` your new page belongs to (probably `guides`).
+Choose which existing folder from `docs/` your new page belongs to (probably `guides`) (probably `guides`).
```
Eric-Arellano, line 447:
```suggestion
Choose which existing folder from `docs/` your new page belongs to (probably `guides`).
```

Eric-Arellano (approved):

Issue #2269, opened 2024-11-08
`make-historical` has incorrect page metadata


Eric-Arellano:
We need to update the metadata for historical API docs to have the version, rather than `latest`

https://github.com/Qiskit/documentation/blob/bc16d038abe279da9332ad2b6ccd754419167bcf/docs/api/qiskit-ibm-runtime/0.32/debug_tools.mdx?plain=1#L2

I wonder if we should remove `make-historical` as a script and instead have developers simply run `npm run gen-api -- --historical`?

PR #2312, opened 2024-11-13
Stop using make-historical

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/2269.

That issue made me realize I'm not sure why we have been using `make-historical` all this time rather than `npm run gen-api` with `--historical`. `gen-api` will correctly handle setting up the new folder. For Qiskit SDK, `gen-api` will ensure the release note links are updated correctly for the historical notes, like `1.2.mdx` linking to `/api/qiskit/1.2` rather than the unversioned files.


arnaucasau (approved):
Good idea!!

This script was introduced while still using the web crawler and qiskit.org IIRC. Runtime and Provider didn't publish more than one version, and it was necessary at that time. I remember we decided to keep it afterward to have a safe mechanism to convert a version into a historical one in case something happened to the CI artifacts (like if they expire). See https://github.com/Qiskit/documentation/issues/332

Since we have the artifacts stored in Box, I don't see any other advantage than just the speedup obtained when generating a new version, given that we avoid using the `gen-api` twice. On the other hand, the script has got more complex with all the handling of the release notes and links and I think it would be better to use `gen-api` as you propose. That way we avoid having to update in two places when changing something in the future.

I tested the differences between running `make-historical` and `gen-api` with the `historical` argument for Qiskit and Runtime, and the output was the same but fixing the metadata issue.

```diff
-title: visualization (latest version)
-description: API reference for qiskit_ibm_runtime.visualization in the latest version of qiskit-ibm-runtime
+title: visualization (v0.33)
+description: API reference for qiskit_ibm_runtime.visualization in qiskit-ibm-runtime v0.33
```

PR #2318, opened 2024-11-14
Fix IBM Circuit function optimization level

jyu00:
IBM Circuit function doesn't actually support optimization level 1. This PR fixes that.


Eric-Arellano (commented):
This PR removes optimization level 0, but it still says level 1 is valid. Is this intended?

Screenshot 2024-11-14 at 11 14 03 AM

jyu00:
Yeah it supports level 1, 2, 3; just not 0

Eric-Arellano (approved):

Issue #2330, opened 2024-11-15
Failure to handle abstract property


Eric-Arellano:
The docs in https://github.com/Qiskit/documentation/pull/2328 have an abstract property. In Sphinx, it renders like this:

![Image](https://github.com/user-attachments/assets/d3fa5c80-40c9-4ba0-a932-899a6bd45254)

In our repo, it renders like this:

![Image](https://github.com/user-attachments/assets/3bfbec5d-0963-412f-80c0-0b17d1560a3d)

`abstract property` should no be part of the header name. We instead should call it only the attribute name, and then set something like `Modifiers: abstract`. Probably leave off property, but confirm what we do when it's just a property for an _inline_ class set up with `.. autoclass::` rather than one generated with `.. autosummary::`.

This Sphinx was generated from this code:

```python
    @property
    @abstractmethod
    def evolved_time(self) -> float:
        """Returns the current evolution time."""

    @property
    @abstractmethod
    def conjugate(self) -> bool:
        """Returns whether this time-evolver instance acts on the right-hand side."""
```

Eric-Arellano:
This was fixed in closed source.

PR #2333, opened 2024-11-15
Handle abstract properties

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/2330.

Screenshot 2024-11-15 at 4 40 29 PM

## FYI: attributes don't have modifiers

Our old code would try to remove modifiers for `py.attribute`, but that was not necessary because they cannot be set. (An attribute is the default, like setting a value in `__init__`; a property is when you set `@property` on a function)

Modifiers are only set on properties. This is evidenced by `npm run regen-api` not changing despite removing code to handle `py.attribute`.

## PR does not preserve `property` modifier

For a typical end-user, an attribute and `property` are essentially the same. The only time I think it'd be relevant is if you're subclassing the class; otherwise, I think `property` is distracting.

So, we continue to not preserve `property`, the same as before.

## `abstract` vs `abstract property`

We capture the whole `abstract property` because the phrase is much more meaningful than `abstract`.

### DIFF
--- a/docs/api/qiskit/0.19/qiskit.aqua.algorithms.MinimumEigensolver.mdx
+++ b/docs/api/qiskit/0.19/qiskit.aqua.algorithms.MinimumEigensolver.mdx
@@ -19,7 +19,7 @@ python_api_name: qiskit.aqua.algorithms.MinimumEigensolver
 
   ### aux\_operators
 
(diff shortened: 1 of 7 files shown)


Eric-Arellano:
This is safe to merge even before we teach closed source to handle `modifiers`. The app will simply ignore the value.

frankharkins (approved):
Nice improvement, thank you!

PR #2337, opened 2024-11-18
Link checker: Add a method to ignore non-API docs for objects.inv

### LINKED ISSUE #2296: Improve link checker ignores of objects.inv for non-API docs
We ignore all of these because the docs have non-API docs in their Sphinx build, like tutorials, so the objects.inv has a ton of invalid links.

https://github.com/Qiskit/documentation/blob/6999dcabd9eb3144238aac8ed6fc731ec134bc65/scripts/js/lib/links/ignores.ts#L17-L27

However, it would be better to still check objects.inv and not ignore the API links. We should automatically do that rather than making you specify the non-API links in objects.inv.

arnaucasau:
This PR adds a new mechanism to ignore links where we can add a regex expression to ignore all the URLs that match it. It adds regexes for the non-API docs in the addons' and runtime's objects.inv.

Closes #2296

### DIFF
--- a/scripts/js/lib/links/ignores.ts
+++ b/scripts/js/lib/links/ignores.ts
@@ -98,9 +98,10 @@ function _runtimeObjectsInvRegexes(): string[] {
     "primitives",
     "compare",
     "retired",
-  ].flatMap((path) => [
-    `\/api\/qiskit-ibm-runtime\/(0.16|0.15|0.14)\/${path}(\/.*|#.*|$)`,
-  ]);
+  ].map(
+    (path) =>
+      `\/api\/qiskit-ibm-runtime\/(0.16|0.15|0.14)\/${path}(\/.*|#.*|$)`,
+  );
 }
 
 export const ALWAYS_IGNORED_URL_REGEXES: string[] = [
--- a/scripts/js/lib/links/ignores.ts
+++ b/scripts/js/lib/links/ignores.ts
@@ -76,16 +76,14 @@ function _addonsObjectsInvRegexes(): string[] {
   // Addons have non-API docs in their Sphinx build that translate into invalid links
   // we should ignore
   return ["how-tos", "how_tos", "install", "index", "explanations"]
-    .map(
-      (path) => [
-        // Latest version
-        `\/api\/qiskit-addon-[^\/]+\/${path}(\/.*|#.*|$)`,
-        // Historical versions
-        `\/api\/qiskit-addon-[^\/]+\/[0-9]+\.[0-9]\/${path}(\/.*|#.*|$)`,
-        // Dev version
-        `\/api\/qiskit-addon-[^\/]+\/dev\/${path}(\/.*|#.*|$)`,
-      ],
-    )
+    .map((path) => [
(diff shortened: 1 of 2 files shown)


scripts/js/lib/links/ignores.ts
```diff
@@ -79,6 +68,52 @@ export const ALWAYS_IGNORED_URLS = new Set([
...
+  return ["how-tos", "how_tos", "install", "index", "explanations"]
+    .map((path) => [
+      // Latest version
+      `\/api\/qiskit-addon-[^\/]+\/${path}(\/.*|#.*|$)`,
+      // Historical versions
+      `\/api\/qiskit-addon-[^\/]+\/[0-9]+\.[0-9]\/${path}(\/.*|#.*|$)`,
+      // Dev version
+      `\/api\/qiskit-addon-[^\/]+\/dev\/${path}(\/.*|#.*|$)`,
+    ])
+    .flat();
```
Eric-Arellano, line 87: You can use `flatMap()` instead to avoid this part :)

Eric-Arellano: Also addons don't have dev docs

scripts/js/lib/links/ignores.ts
```diff
@@ -79,6 +68,52 @@ export const ALWAYS_IGNORED_URLS = new Set([
...
+    "primitives",
+    "compare",
+    "retired",
+  ]
+    .map((path) => [
+      `\/api\/qiskit-ibm-runtime\/0.16\/${path}(\/.*|#.*|$)`,
+      `\/api\/qiskit-ibm-runtime\/0.15\/${path}(\/.*|#.*|$)`,
+      `\/api\/qiskit-ibm-runtime\/0.14\/${path}(\/.*|#.*|$)`,
+    ])
+    .flat();
```
Eric-Arellano, line 109: Ditto flatmap

scripts/js/lib/links/ignores.ts
```diff
@@ -79,6 +68,52 @@ export const ALWAYS_IGNORED_URLS = new Set([
...
+    "index",
+    "sessions",
+    "primitives",
+    "compare",
+    "retired",
+  ]
+    .map((path) => [
+      `\/api\/qiskit-ibm-runtime\/0.16\/${path}(\/.*|#.*|$)`,
+      `\/api\/qiskit-ibm-runtime\/0.15\/${path}(\/.*|#.*|$)`,
+      `\/api\/qiskit-ibm-runtime\/0.14\/${path}(\/.*|#.*|$)`,
```
Eric-Arellano, line 107: I think you could use `{14,15,16}` to consolidate these into one rule?

Eric-Arellano (commented):
Nice work!

scripts/js/lib/links/ignores.ts
arnaucasau: I removed the `flat` call in this case given that it's not necessary anymore after combining the three regexes into one

Eric-Arellano (approved):

PR #2344, opened 2024-11-19
Add params to nb-tester

emilkovacev:
Effectively allows the nb-tester script to be run against any backend. Also allows you to pass authentication and configuration directly to QiskitRuntimeService (if preferred).

Here's an example of how you can use the new changes to run a set of notebooks against test_eagle using qiskit_ibm_runtime:

```
test-docs-notebooks --config-path scripts/config/notebook-testing.toml --provider "qiskit_ibm_runtime" --backend "test_eagle2"
```

By default, `--provider` is set to `qiskit_fake_provider`, which runs the same provider code that was present in nb_tester before. By changing the provider, you can also specify "runtime_fake_provider" to use the fake_providers in `qiskit_ibm_runtime.fake_provider`, or "qiskit_ibm_runtime" to rely on real backends, and the credentials provided.

This shouldn't require changes in CI *yet* - it seems like notebook testing CI is able to complete the same number of tests as in main, although it does carry over some errors that I ran into on main too.

### DIFF
--- a/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
+++ b/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
@@ -315,31 +315,30 @@ async def _execute_notebook(filepath: Path, config: Config) -> nbformat.Notebook
     )
 
     await _execute_in_kernel(kernel, PRE_EXECUTE_CODE)
-    # if config.should_patch(filepath):
-    print(vars(config.args))
-    def get_arg(arg, default):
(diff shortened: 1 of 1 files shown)


frankharkins (commented):
Thank you! I've left some small suggestions.

Since you have write access, could you make this PR from a branch in this repo please? It'll mean CI will run this script against our notebooks. You can prefix your branch with your initials.

emilkovacev:
PR migrated to #2350 to run full CI process

PR #2348, opened 2024-11-20
Add check for images' alt text

arnaucasau:
This pull request implements a check to ensure all images have an alt text defined, which is crucial for accessibility. The output of the check shows the file name that contains images without alternative text and the images' names. Ex:
```
Error in file 'docs/guides/custom-transpiler-pass.ipynb':
	- The image '/images/guides/custom-transpiler-pass/DAG.png' does not have an alt text defined.

Some images are missing alt text 💔
```
The PR builds on the work done by @shraddha-aangiras on https://github.com/Qiskit/documentation/pull/1800. I have made some changes to simplify the code and incorporated @eric-arellano's feedback. Thank you both for the work you've done!

Closes https://github.com/Qiskit/documentation/issues/1651


scripts/js/lib/extractMarkdownImages.ts
```diff
@@ -0,0 +1,39 @@
...
+    .use(remarkGfm)
+    .use(() => (tree: Root) => {
+      visit(tree, "image", (node) => {
+        if (!node.alt) {
+          images.add(node.url);
+        }
+      });
+    })
+    .use(remarkStringify)
+    .process(markdown);
```
arnaucasau, line 36: Notice that the plugin will not extract images from HTML tags. I have searched in the repo and we don't have any instance of it. HTML tags can be used for the writers if they want to resize the images in markdown:
```html
<img src="example.jpg" alt="Example" width="200"/>
```
We could implement this by adding another visitor that uses Cheerio as follows:
```ts
visit(tree, "html", (node) => {
  const $ = load(node.value);
  const imgElement = $("img");
  if (imgElement.length) {
    images.add(imgElement.attr("src"));
 }
});
```

Do you think we should take into account this case @eric-arellano?

scripts/js/lib/extractMarkdownImages.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  await unified()
+    .use(remarkParse)
+    .use(remarkGfm)
+    .use(() => (tree: Root) => {
+      visit(tree, "image", (node) => {
+        if (!node.alt) {
+          images.add(node.url);
+        }
+      });
```
arnaucasau, line 33: FYI: In my opinion, the visitor is already very simple and readable, but I discovered that we can add more conditions to the `typeNode` part to make it more concise. Equivalent example:
```ts
 visit(
        tree,
        (node) => node.type === "image" && !node.alt,
        (node) => images.add(node.url),
      );
```
I don't think is useful right now, but we could use it in the future if the logic becomes more complex.

docs/guides/custom-transpiler-pass.ipynb
```diff
@@ -83,7 +83,7 @@
     "dag = circuit_to_dag(qc)\n",
     "dag_drawer(dag)\n",
     "```\n",
-    "![](/images/guides/custom-transpiler-pass/DAG.png)\n",
```
arnaucasau, line 86: This image was missing the alt text. Looking at the notebook (https://docs.quantum.ibm.com/guides/custom-transpiler-pass) I think the alt text of the previous image (the circuit inside the "Background: DAG representation" collapsible) refereed to this one, so I moved it here and added an alt text for the circuit

PR #2349, opened 2024-11-20
Add Image checker for images' alt text and HTML tags

arnaucasau:
This pull request adds an image checker to ensure all images have an alt text defined, which is crucial for accessibility, and that we don't have any `` HTML tag. The output of the check shows the file name that contains invalid images and the images' names. Ex:
```
Error in file 'docs/guides/custom-transpiler-pass.ipynb':
	- The image '/images/guides/custom-transpiler-pass/DAG.png' does not have alt text.

Invalid images found 💔 See https://github.com/Qiskit/documentation#images for instructions.
```
The PR builds on the work done by @shraddha-aangiras on https://github.com/Qiskit/documentation/pull/1800. I have made some changes to simplify the code and incorporated @eric-arellano's feedback. Thank you both for the work you've done!

Closes https://github.com/Qiskit/documentation/issues/1651


docs/guides/custom-transpiler-pass.ipynb
```diff
@@ -83,7 +83,7 @@
     "dag = circuit_to_dag(qc)\n",
     "dag_drawer(dag)\n",
     "```\n",
-    "![](/images/guides/custom-transpiler-pass/DAG.png)\n",
```
arnaucasau, line 86: This image was missing the alt text. Looking at the notebook (https://docs.quantum.ibm.com/guides/custom-transpiler-pass) I think the alt text of the previous image (the circuit inside the "Background: DAG representation" collapsible) refereed to this one, so I moved it here and added an alt text for the circuit

scripts/js/lib/extractMarkdownImages.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  await unified()
+    .use(remarkParse)
+    .use(remarkGfm)
+    .use(() => (tree: Root) => {
+      visit(tree, "image", (node) => {
+        if (!node.alt) {
+          images.add(node.url);
+        }
+      });
```
arnaucasau, line 33: FYI: In my opinion, the visitor is already very simple and readable, but I discovered that we can add more conditions to the `typeNode` part to make it more concise. Equivalent example:
```ts
 visit(
        tree,
        (node) => node.type === "image" && !node.alt,
        (node) => images.add(node.url),
      );
```
I don't think is useful right now, but we could use it in the future if the logic becomes more complex.

scripts/js/lib/extractMarkdownImages.ts
```diff
@@ -0,0 +1,39 @@
...
+    })
+    .use(remarkStringify)
+    .process(markdown);
```
arnaucasau, line 36: Notice that the plugin will not extract images from HTML tags. I have searched in the repo and we don't have any instance of it. HTML tags can be used for the writers if they want to resize the images in markdown:
```html
<img src="example.jpg" alt="Example" width="200"/>
```
We could implement this by adding another visitor that uses Cheerio as follows:
```ts
visit(tree, "html", (node) => {
  const $ = load(node.value);
  const imgElement = $("img");
  if (imgElement.length) {
    images.add(imgElement.attr("src"));
 }
});
```

Do you think we should take into account this case?

scripts/js/commands/checkAltText.ts
```diff
@@ -0,0 +1,43 @@
...
+
+async function main() {
+  const files = await globby(["docs/**/*.{ipynb,mdx}", "!docs/api/**/*.mdx"]);
+  const fileErrors: string[] = [];
+
+  for (const file of files) {
+    const markdown = await readMarkdown(file);
+    const images = await findImagesWithoutAltText(markdown);
+    const imageErrors = [...images].map(
+      (image) => `The image '${image}' does not have an alt text defined.`,
```
Eric-Arellano, line 25:
```suggestion
      (image) => `The image '${image}' does not have alt text.`,
```

scripts/js/commands/checkAltText.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (imageErrors.length) {
+      fileErrors.push(
+        `Error in file '${file}':\n\t- ${imageErrors.join("\n\t- ")}`,
+      );
+    }
+  }
+
+  if (fileErrors.length) {
+    fileErrors.forEach((error) => console.log(error));
+    console.error("\nSome images are missing alt text 💔\n");
```
Eric-Arellano, line 37:
```suggestion
    console.error("\nSome images are missing alt text 💔 See https://github.com/Qiskit/documentation#images for instructions.\n");
```

Eric-Arellano: This will probably make Prettier mad that it's one line

scripts/js/lib/extractMarkdownImages.ts
Eric-Arellano: I like this current approach more.

scripts/js/lib/extractMarkdownImages.ts
Eric-Arellano: We should definitely account for ``, but I think by banning it. We don't want people changing the styling of images and I don't see a good reason to not use a markdown image. If a good reason comes up over time, we can drop the ban.

If there isn't an easy way to find HTML image tags with Unified/remark, you could use a regex match.

Eric-Arellano (commented):
Nice work! This will eventually need to support API docs but that can easily be done in a follow up. Only blocker is the `` tag

scripts/js/lib/extractMarkdownImages.ts
arnaucasau: Good idea, and thanks for the suggestion! As we discussed over Slack I changed the checker to check both the alt text and the HTML tags :+1:

scripts/js/commands/checkImages.ts
```diff
@@ -20,24 +20,23 @@ async function main() {
...
+        `Error in file '${file}':\n\t- ${[...imageErrors].join("\n\t- ")}`,
       );
     }
   }
 
   if (fileErrors.length) {
     fileErrors.forEach((error) => console.log(error));
-    console.error("\nSome images are missing alt text 💔\n");
+    console.error(
+      "\nSome images are missing alt text 💔 See https://github.com/Qiskit/documentation#images for instructions.\n",
```
Eric-Arellano:
```suggestion
      "\n💔 Some images have problems. See https://github.com/Qiskit/documentation#images for instructions.\n",
```

scripts/js/commands/checkImages.ts
```diff
@@ -20,24 +20,23 @@ async function main() {
...
+    );
     process.exit(1);
   }
-  console.log("\nAll images have alt text ✅\n");
+  console.log("All images have alt text ✅\n");
```
Eric-Arellano:
```suggestion
  console.log("✅ All images are valid\n");
```

scripts/js/lib/markdownImages.test.ts
```diff
@@ -28,12 +28,18 @@ test("Test the extraction of the images", async () => {
...
+<img src="/images/HTMLexample2.jpg" alt="Example" width="200"/>
+
 ![And now, our last link](https://ibm.com)
     `;
-  const images = await findImagesWithoutAltText(markdown);
+  const images = await collectInvalidImageErrors(markdown);
   const correct_images = new Set([
-    "/images/invalid_img1.png",
-    "/images/invalid_img2.png",
+    "The image '/images/HTMLexample1.jpg' uses an HTML tag instead of markdown syntax.",
```
Eric-Arellano, line 39: It'd be helpful to say "uses an HTML `` tag instead of markdown syntax"

Eric-Arellano (approved):
Looks great!

PR #2350, opened 2024-11-20
Add params to nb-tester

emilkovacev:
Migrating from #2344

Effectively allows the nb-tester script to be run against any backend. Also allows you to pass authentication and configuration directly to QiskitRuntimeService (if preferred).

Here's an example of how you can use the new changes to run a set of notebooks against test_eagle using qiskit_ibm_runtime:

```
test-docs-notebooks --config-path scripts/config/notebook-testing.toml --provider "qiskit_ibm_runtime" --backend "test_eagle2"
```

By default, `--provider` is set to `qiskit_fake_provider`, which runs the same provider code that was present in nb_tester before. By changing the provider, you can also specify "runtime_fake_provider" to use the fake_providers in `qiskit_ibm_runtime.fake_provider`, or "qiskit_ibm_runtime" to rely on real backends, and the credentials provided.

This shouldn't require changes in CI *yet* - it seems like notebook testing CI is able to complete the same number of tests as in main, although it does carry over some errors that I ran into on main too.

### DIFF
--- a/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
+++ b/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
@@ -315,31 +315,30 @@ async def _execute_notebook(filepath: Path, config: Config) -> nbformat.Notebook
     )
 
     await _execute_in_kernel(kernel, PRE_EXECUTE_CODE)
-    # if config.should_patch(filepath):
-    print(vars(config.args))
-    def get_arg(arg, default):
-        return vars(config.args).get(arg.lstrip("-").replace("-", "_"), default)
+    if config.should_patch(filepath):
+        def get_arg(arg, default):
+            return vars(config.args).get(arg.lstrip("-").replace("-", "_"), default)
 
-    backend = get_arg("--backend", "fake_athens")
-    provider = get_arg("--provider", "qiskit_fake_provider")
-    channel = get_arg("--channel", None)
-    token = get_arg("--token", None)
-    url = get_arg("--url", None)
-    name = get_arg("--name", None)
-    instance = get_arg("--instance", None)
-    
-    # Implements a subset of options from QiskitRuntimeService, but in practice any option
-    # can easily be added here
-    backend_cell = generate_backend_cell(
-        backend_name=backend, 
-        provider=provider,
-        channel=channel,
-        token=token,
-        url=url,
-        name=name,
-        instance=instance,
-    )
-    await _execute_in_kernel(kernel, backend_cell)
+        backend = get_arg("--backend", "fake_athens")
+        provider = get_arg("--provider", "qiskit_fake_provider")
+        channel = get_arg("--channel", None)
+        token = get_arg("--token", None)
+        url = get_arg("--url", None)
+        name = get_arg("--name", None)
+        instance = get_arg("--instance", None)
+
+        # Implements a subset of options from QiskitRuntimeService, but in practice any option
+        # can easily be added here
+        backend_cell = generate_backend_cell(
+            backend_name=backend, 
+            provider=provider,
+            channel=channel,
+            token=token,
+            url=url,
+            name=name,
+            instance=instance,
+        )
+        await _execute_in_kernel(kernel, backend_cell)
 
     notebook_client = nbclient.NotebookClient(
         nb=nb,
--- a/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
+++ b/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
@@ -435,7 +435,7 @@ def get_args() -> argparse.Namespace:
         help=(
             "Specify a backend to run the script against, such as 'fake_athens'"
             "or 'athens'. Only relevant when `--provider` is "
-            "`qiskit_ibm_runtime` or `runtime_fake_provider`."
+            "`qiskit-ibm-runtime` or `runtime-fake-provider`."
         )
     )
     generic_backend_options_group = parser.add_argument_group(
--- a/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
+++ b/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
@@ -38,19 +38,61 @@ matplotlib.set_loglevel("critical")
 """
 
 # If not submitting jobs, we also run this code before notebook execution to mock the real backend
-MOCKING_CODE = """\
+
+# MOCKING_CODE = """\
+# import warnings
+# from qiskit_ibm_runtime import QiskitRuntimeService
+# from qiskit.providers.fake_provider import GenericBackendV2
+# 
+# def patched_least_busy(self, *args, **kwarg):
+#   return GenericBackendV2(num_qubits=6, control_flow=True)
+# 
+# QiskitRuntimeService.least_busy = patched_least_busy
+# 
+# warnings.filterwarnings("ignore", message="Options {.*} have no effect in local testing mode.")
+# warnings.filterwarnings("ignore", message="Session is not supported in local testing mode or when using a simulator.")
+# """
+
+MOCKING_CODE = """
 import warnings
-from qiskit_ibm_runtime import QiskitRuntimeService
-from qiskit.providers.fake_provider import GenericBackendV2
-
-def patched_least_busy(self, *args, **kwarg):
-  return GenericBackendV2(num_qubits=6, control_flow=True)
-
-QiskitRuntimeService.least_busy = patched_least_busy
 
 warnings.filterwarnings("ignore", message="Options {.*} have no effect in local testing mode.")
 warnings.filterwarnings("ignore", message="Session is not supported in local testing mode or when using a simulator.")
+warnings.filterwarnings("ignore", message="Session is not supported in local testing mode or when using a simulator.")
 """
+
+
+def generate_backend_cell(backend_name, fake_provider=False, **kwargs):
+    """
+    generate code for fetching a custom backend to inject into a notebook
+    """
+
+    # Generates a set of arguments for QiskitRuntimeService using kwargs
+    qiskit_runtime_service_args = ", ".join([f"{arg}=\"{val}\"" if val else f"{arg}=None" for arg, val in kwargs.items()])
+
+    cell = """
+from qiskit_ibm_runtime import QiskitRuntimeService"""
+
+    if fake_provider:
+        cell += f"""
+from qiskit_ibm_runtime.fake_provider import FakeProviderForBackendV2
+
+def patched_least_busy(self, *args, **kwargs):
+    provider = FakeProviderForBackendV2()
+    return provider.backend("{backend_name}")"""
+    else: 
+        cell += f"""
+def patched_least_busy(self, *args, **kwargs):
+    service = QiskitRuntimeService({qiskit_runtime_service_args})
+    return service.backend("{backend_name}")"""
+
+    cell += """
+QiskitRuntimeService.least_busy = patched_least_busy"""
+
+    cell += MOCKING_CODE
+    print(cell)
+    return cell
+
 
 def get_package_versions():
     requirements_file = Path("scripts/nb-tester/requirements.txt").read_text()
@@ -260,8 +302,31 @@ async def _execute_notebook(filepath: Path, config: Config) -> nbformat.Notebook
     )
 
     await _execute_in_kernel(kernel, PRE_EXECUTE_CODE)
-    if config.should_patch(filepath):
-        await _execute_in_kernel(kernel, MOCKING_CODE)
+    # if config.should_patch(filepath):
+    print(vars(config.args))
+    def get_arg(arg, default):
+        return vars(config.args).get(arg.lstrip("-").replace("-", "_"), default)
+
+    backend = get_arg("--backend", "fake_athens")
+    fake_provider = get_arg("--fake-provider", False)
+    channel = get_arg("--channel", None)
+    token = get_arg("--token", None)
+    url = get_arg("--url", None)
+    name = get_arg("--name", None)
+    instance = get_arg("--instance", None)
+    
+    # Implements a subset of options from QiskitRuntimeService, but in practice any option
+    # can easily be added here
+    backend_cell = generate_backend_cell(
+        backend_name=backend, 
+        fake_provider=fake_provider,
+        channel=channel,
+        token=token,
+        url=url,
+        name=name,
+        instance=instance,
+    )
+    await _execute_in_kernel(kernel, backend_cell)
 
     notebook_client = nbclient.NotebookClient(
         nb=nb,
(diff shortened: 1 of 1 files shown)


scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -361,6 +427,63 @@ def get_args() -> argparse.Namespace:
...
+        default="fake_athens",
+        help=(
+            "Specify a backend to run the script against"
+        )
+    )
+    parser.add_argument(
+        "--provider",
+        action="store",
+        default="qiskit_fake_provider",
+        choices=["qiskit_ibm_runtime", "qiskit_fake_provider", "runtime_fake_provider"],
```
Eric-Arellano, line 442: Why use `_`? Usually CLIs use `-` as the delimiter

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -361,6 +427,63 @@ def get_args() -> argparse.Namespace:
...
+        help=(
+            "The API URL to submit the notebook against"
+        )
+    )
+    parser.add_argument(
+        "--name",
+        action="store",
+        default=None,
+        help=(
+            "Name of the qiskit account to load"
```
Eric-Arellano, line 476: Can you please add an example?

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -361,6 +427,63 @@ def get_args() -> argparse.Namespace:
...
+        help=(
+            "IBM Cloud API key or IBM Quantum API token"
+        )
+    )
+    parser.add_argument(
+        "--url",
+        action="store",
+        default=None,
+        help=(
+            "The API URL to submit the notebook against"
```
Eric-Arellano, line 468: Any helpful examples?

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -361,6 +427,63 @@ def get_args() -> argparse.Namespace:
...
+        help=(
+            "Specify a channel for running the notebook against"
+        )
+    )
+    parser.add_argument(
+        "--token",
+        action="store",
+        default=None,
+        help=(
+            "IBM Cloud API key or IBM Quantum API token"
```
Eric-Arellano, line 460:
```suggestion
            'IBM Cloud API key or IBM Quantum API token. Warning: for security, you should set this via an environment variable, e.g. `--token="${IQP_API_TOKEN}"'
```

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -37,21 +38,64 @@
...
 
 warnings.filterwarnings("ignore", message="Options {.*} have no effect in local testing mode.")
 warnings.filterwarnings("ignore", message="Session is not supported in local testing mode or when using a simulator.")
+warnings.filterwarnings("ignore", message="Session is not supported in local testing mode or when using a simulator.")
 """
 
+def generate_backend_patch(
+    backend_name, 
+    provider: Literal["qiskit_ibm_runtime", "qiskit_fake_provider", "runtime_fake_provider"] = "qiskit_ibm_runtime", 
+    **kwargs
```
Eric-Arellano, line 52: Rather than using `kwargs`, which is vague/very generic, you can call this `runtime_service_args: dict[str, str | None]`. It will be more clear and keeps `**kwargs` available for future changes to this API.

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -37,21 +38,64 @@
...
-
-QiskitRuntimeService.least_busy = patched_least_busy
 
 warnings.filterwarnings("ignore", message="Options {.*} have no effect in local testing mode.")
 warnings.filterwarnings("ignore", message="Session is not supported in local testing mode or when using a simulator.")
+warnings.filterwarnings("ignore", message="Session is not supported in local testing mode or when using a simulator.")
 """
 
+def generate_backend_patch(
+    backend_name,
```
Eric-Arellano, line 50:
```suggestion
    backend_name: str, 
```

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -37,21 +38,64 @@
...
 warnings.filterwarnings("ignore", message="Options {.*} have no effect in local testing mode.")
 warnings.filterwarnings("ignore", message="Session is not supported in local testing mode or when using a simulator.")
+warnings.filterwarnings("ignore", message="Session is not supported in local testing mode or when using a simulator.")
 """
 
+def generate_backend_patch(
+    backend_name, 
+    provider: Literal["qiskit_ibm_runtime", "qiskit_fake_provider", "runtime_fake_provider"] = "qiskit_ibm_runtime", 
+    **kwargs
+):
```
Eric-Arellano, line 53:
```suggestion
) -> str:
```

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -37,21 +38,64 @@
...
+    """
+    generate code for fetching a custom backend to inject into a notebook
+    """
+
+    # Generates a set of arguments for QiskitRuntimeService using kwargs
+    qiskit_runtime_service_args = ", ".join([f"{arg}=\"{val}\"" if val else f"{arg}=None" for arg, val in kwargs.items()])
```
Eric-Arellano, line 59: No need for a list comprehension - you can use a generator expression. That's more efficient

```suggestion
    qiskit_runtime_service_args = ", ".join(f"{arg}=\"{val}\"" if val else f"{arg}=None" for arg, val in kwargs.items())
```

Eric-Arellano: Also this variable should only be defined in the block for qiskit_ibm_runtime because it's the only one that uses it

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -37,21 +38,64 @@
...
+            service = QiskitRuntimeService({qiskit_runtime_service_args})
+            return service.backend("{backend_name}")
+        """)
+
+    else:
+        raise ValueError(f"Please specify a valid provider. \"{provider}\" is invalid.")
+
+    patch += dedent("""
+    QiskitRuntimeService.least_busy = patched_least_busy
+    """)
```
Eric-Arellano, line 93:
```suggestion
    patch += "\nQiskitRuntimeService.least_busy = patched_least_busy\n"
```

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -261,7 +305,29 @@ async def _execute_notebook(filepath: Path, config: Config) -> nbformat.Notebook
 
     await _execute_in_kernel(kernel, PRE_EXECUTE_CODE)
     if config.should_patch(filepath):
-        await _execute_in_kernel(kernel, MOCKING_CODE)
+        def get_arg(arg, default):
+            return vars(config.args).get(arg.lstrip("-").replace("-", "_"), default)
```
Eric-Arellano, line 309: This shouldn't be necessary? argparse will already handle for you converting `-` into `_` and applying defaults. It's a code smell that you have to specify the default twice atm.

What was the issue with usin `config.args.backend` et al?

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
emilkovacev: For the examples, would it be more helpful to link the QiskitRuntimeService page and explicitly say to reference that page for those args, to use the descriptions from that page, or to come up with unique descriptions?

Eric-Arellano: Linking to https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/qiskit_ibm_runtime.QiskitRuntimeService makes sense. We want to make clear these are args for QiskitRuntimeService

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -37,20 +38,58 @@
...
+
+    warnings.filterwarnings("ignore", message="Options {.*} have no effect in local testing mode.")
+    warnings.filterwarnings("ignore", message="Session is not supported in local testing mode or when using a simulator.")
+    """)
+
+    if provider == "qiskit-fake-provider":
+        patch += dedent("""
+        from qiskit.providers.fake_provider import GenericBackendV2
+        def patched_least_busy(self, *args, **kwargs):
+            return GenericBackendV2(num_qubits=6, control_flow=True)
```
Eric-Arellano, line 62: Maybe make `num_qubits` an option with default of 6 and as a new argument group called for qiskit-fake-provider?

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -361,6 +413,77 @@ def get_args() -> argparse.Namespace:
...
+        )
+    )
+    parser.add_argument(
+        "--backend",
+        action="store",
+        default="fake_athens",
+        help=(
+            "Specify a backend to run the script against, such as 'fake_athens'"
+            "or 'athens'. Only relevant when `--provider` is "
+            "`qiskit_ibm_runtime` or `runtime_fake_provider`."
```
Eric-Arellano, line 432:
```suggestion
            "`qiskit-ibm-runtime` or `runtime-fake-provider`."
```

Eric-Arellano (approved):
This looks great! Please wait for @frankharkins to approve before merging

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -206,7 +251,7 @@ async def execute_notebook(path: Path, config: Config) -> bool:
     Wrapper function for `_execute_notebook` to print status and write result
     """
     if config.should_patch(path):
-        print(f"▶️ Executing {path} (with least_busy patched to return fake backend)")
+        print(f"▶️ Executing {path} (with least_busy patched to {config.args.backend})")
```
frankharkins, line 254: I noticed in CI it says we're patching with `fake_athens` even though we're patching with the Qiskit generic backend. I think easiest is to just remove the backend name. We can always check the command to see how we patched it.

```suggestion
        print(f"▶️ Executing {path} (with least_busy patched)")
```

frankharkins (commented):
Looks great! Just one last thing

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -37,20 +38,64 @@
...
+            provider = FakeProviderForBackendV2()
+            return provider.backend("{backend_name}")
+        """)
+
+    elif provider == "qiskit-ibm-runtime": 
+        # Generates a set of arguments for QiskitRuntimeService using kwargs
+        if not runtime_service_args:
+            runtime_service_args = {}
+        qiskit_runtime_service_args = ", ".join(
+            f"{arg}=\"{val}\"" if isinstance(val, str) else f"{arg}={val}" for arg, val in generic_backend_args.items()
```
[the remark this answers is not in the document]
emilkovacev: The only issue with this example (which I ran into while trying this out) is if the args aren't given, then config.args.channel is "None", which is invalid - which either means we write the following for each option:
```
f'channel="{config.args.channel}"' if config.args.channel else "channel=None"
```

Or, I could instead make that initial long join statement into a generic function since we would use it multiple times (and are likely to make use of it again):
```
def render_kwarg(arg, val):
    if isinstance(val, str):
        return f"{arg}=\"{val}\""
    return f"{arg}={val}"
```

and if there are any more use cases we can expand the usage of the function.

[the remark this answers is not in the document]
Eric-Arellano: > render_kwarg

Sounds good to me.

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -37,20 +38,70 @@
...
+    import warnings
+
+    warnings.filterwarnings("ignore", message="Options {.*} have no effect in local testing mode.")
+    warnings.filterwarnings("ignore", message="Session is not supported in local testing mode or when using a simulator.")
+    """)
+
+    if provider == "qiskit-fake-provider":
+        if not generic_backend_kwargs:
+            generic_backend_kwargs = {}
+        qiskit_fake_provider_args = render_kwargs(generic_backend_kwargs)
```
Eric-Arellano, line 70:
```suggestion
        qiskit_fake_provider_args = render_kwargs(generic_backend_kwargs or {})
```

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -37,20 +38,70 @@
...
+        def patched_least_busy(self, *args, **kwargs):
+            provider = FakeProviderForBackendV2()
+            return provider.backend("{backend_name}")
+        """)
+
+    elif provider == "qiskit-ibm-runtime": 
+        # Generates a set of arguments for QiskitRuntimeService using kwargs
+        if not runtime_service_kwargs:
+            runtime_service_kwargs = {}
+        qiskit_runtime_service_args = render_kwargs(runtime_service_kwargs)
```
Eric-Arellano, line 90: We prefer self-documenting code over comments, unless when the comment adds something you can't easily infer

```suggestion
        qiskit_runtime_service_args = render_kwargs(runtime_service_kwargs or {})
```

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -361,6 +429,104 @@ def get_args() -> argparse.Namespace:
...
+        "--backend",
+        action="store",
+        default="fake_athens",
+        help=(
+            "Specify a backend to run the script against, such as 'fake_athens'"
+            "or 'athens'. Only relevant when `--provider` is "
+            "`qiskit-ibm-runtime` or `runtime-fake-provider`."
+        )
+    )
+    generic_backend_options_group = parser.add_argument_group(
```
Eric-Arellano, line 451: Nit to use whitespace to make the code more readable

```suggestion
    )

    generic_backend_options_group = parser.add_argument_group(
```

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -361,6 +429,104 @@ def get_args() -> argparse.Namespace:
...
+        )
+    )
+    generic_backend_options_group.add_argument(
+        "--control-flow",
+        action="store_true",
+        help=(
+            "Specify if the qiskit generic backend should enable control flow"
+            "directives on the target"
+        )
+    )
```
Eric-Arellano, line 477:
```suggestion
    )

```

Eric-Arellano (approved):
Looks great! Feel free to merge after those last suggestions. Thanks for iterating on this!

PR #2383, opened 2024-11-25
Execute each notebook in its own temporary directory

frankharkins:
We have some notebooks that write files to disk. At the moment, running these notebooks through `tox` means these notebooks use the docs repo as their working directory and will write files there. This is problematic for a couple of reasons:

1. Notebooks that write to disk can interfere with eachother. For example, two notebooks may try to write/read from the same file.
2. Notebook failures (or killing the script) will leave working files in the repo.

This PR solves the problem by modifying the script to execute each notebook in its own temporary directory.

### DIFF
--- a/docs/guides/save-circuits.ipynb
+++ b/docs/guides/save-circuits.ipynb
@@ -113,23 +113,6 @@
     "\n",
     "qc[0].draw(\"mpl\")"
    ]
-  },
-  {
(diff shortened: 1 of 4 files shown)


Eric-Arellano (approved):
Good catch! Nice implementation also and thanks for removing the pathlib code

PR #2382, opened 2024-11-25
Update notebook requirements

frankharkins:
Closes #2374

This PR makes a few changes:

1. Update our Python dependencies to the latest versions (0864f5a9bea625fc15cb0dbdb928df04343e0a2b)
2. Fix the serverless notebooks (f686f24329834995ff2d59542841c9c9f190ad82, 6405a0ee80c468d1b94c76ecfd7c84a30c254f11). These were actually broken but we didn't catch it because we only test that the program uploads correctly, not that it runs. I thought I'd added tests for this but I'm not sure what happened to them.
3. Re-run all noteboooks with the new requirements (77f0c108cda3879f5473ace930add96e808323b5, 4af95dd75a412ae2ee617e316325c27d61516ded).

PR #2386, opened 2024-11-25
Reduce chance of `save-jobs.ipynb` failing in CI

frankharkins:
The problem is `service.jobs()` by default only gets the past 10 jobs. Usually this is fine, but there's a chance the past 10 jobs we've submitted have all been cancelled or have failed.

This PR reduces the change of CI failing by increasing the number of jobs we collect. We can remove the limit by setting `limit=None`, but this takes too long to execute. I've picked 1000 as a compromise.

***

This is ready for review, but don't merge until after #2382 is merged. When that happens, I'll pull main and re-run this notebook to update it.


Eric-Arellano (approved):
Would 500 be reasonable? Idk how much of a difference it would make in performance

frankharkins:
I actually didn't really notice much difference between 10 and 1000, just setting it to `None` took a long time (I never actually let it complete; just CTRL+C'd it). I could probably increase it more.

Eric-Arellano:
Sg, 1000 makes sense then.

PR #2389, opened 2024-11-25
Fix changed files check for APIs in CI

Eric-Arellano:
There were two issues:

1. We unintentionally ignored API docs because `--from-file` was being combined with the other args.
2. Our call to `echo` the changed files to save them could blow up the overall GitHub Action when there were too many files. Instead, we can use the changed-files action to write the output directly.


Eric-Arellano:
You can see a successful run at https://github.com/Qiskit/documentation/actions/runs/12018028777/job/33501631056#step:19:7

scripts/js/commands/checkPagesRender.ts
```diff
@@ -39,7 +39,9 @@ const readArgs = (): Arguments => {
       type: "string",
       normalize: true,
       description:
-        "Read the file path for file paths and globs to check, like `docs/start/index.md`. Entries should be separated by a newline and should be valid file types (.md, .mdx, .ipynb).",
+        "Read the file path for file paths and globs to check, like `docs/start/index.md`. " +
+        "Entries should be separated by a newline and should be valid file types (.md, .mdx, .ipynb)." +
+        "Mutually exclusive with the other args.",
```
arnaucasau, line 44: Nice comment! Maybe in the future, we can refactor it and use `.conflcts()` to error if `--from-file` is used with another argument to avoid setting them unintentionally.

https://yargs.js.org/docs/#api-reference-conflictsx-y

arnaucasau (approved):
Awesome! Thank you!

scripts/js/commands/checkPagesRender.ts
arnaucasau: I think, in https://github.com/Qiskit/documentation/pull/2389/commits/5b577010386927e99b50c97fee46f4c2a5f288a0,  we also need to remove the `default: false` from the other arguments and let them be `undefined`. Otherwise, they count as set by `conflicts` and the option `from-files` will always error

PR #2396, opened 2024-11-26
Add tests to serverless notebooks

frankharkins:
This PR adds some testing cells to two of the serverless notebooks. Most important is `serverless-first-program`, which creates and uploads a program but doesn't run it. This means we won't spot broken code until we try to run the same program in `serverless-run-first-workload`. To fix this, I've added a test cell that runs the program and checks it completes without error.


docs/guides/serverless-first-program.ipynb
```diff
@@ -329,7 +327,121 @@
     }
    ],
    "source": [
-    "serverless.list()[:3]  # Limit to first three functions"
+    "# Get program from serverless.list() that matches the title of the one we uploaded\n",
+    "next(\n",
+    "    program for program in serverless.list()\n",
+    "    if program.title == \"transpile_remote_serverless\"\n",
+    ")"
```
frankharkins, line 334: I changed this to better fit the prose. We're trying to show the program uploaded successfully, but the output actually only shows other programs.

docs/guides/serverless-run-first-workload.ipynb
```diff
@@ -118,7 +119,7 @@
    "source": [
     "from qiskit.circuit.random import random_circuit\n",
     "\n",
-    "qc_random = [(random_circuit(4, 4, measure=True)) for _ in range(10)]\n",
+    "qc_random = [(random_circuit(4, 4, measure=True, seed=i)) for i in range(10)]\n",
```
frankharkins, line 122: Setting this seed means re-running the notebook won't change the output.

arnaucasau (approved):
Looks good to me! Nice work!

PR #2399, opened 2024-11-26
remove deprecated tag

### LINKED ISSUE #2363: [QDC Feedback] erroneous mention of `noise_adaptive` option
In one (or maybe more) of the transpile guides we reference `noise_adaptive` as an option someone can set, however this option is now deprected in Qiskit (it may have even been removed already) so we should remove it from the guides. It's not mentioned in the code but it is mentioned somewhere in the text where we talk about all the different options users can set.

beckykd:
closes #2363


beckykd:
Changed file:  https://qiskit.github.io/documentation/pr-2399/guides/common-parameters

abbycross (approved):

PR #2402, opened 2024-11-27
Increase timeout in serverless tests

frankharkins:
This notebook was regularly timing out in CI. This PR removes the manual polling/timeout and just polls until the cell times out.

* In CI, it'll time out after 300s (an increase from 100s).
* In the fortnightly cron job, it'll wait until the action timesout at 6hrs.

If this is still causing problems, we can remove it from CI.


Eric-Arellano (commented):
This makes sense, but it might be worth logging because otherwise people won't know why the program appears hung. The logs are helpful in CI debugging, and especially when running locally. 

Maybe log every 15 seconds the result of the `status`, or even every 20? Don't log the first time. Mention the notebook name.

frankharkins:
Any logs we add in the notebook won't be seen by users executing through the script. We don't echo things printed in notebooks to the tester script as it's not trivial with nbformat and it would be super noisy. The only things we do pass along are exceptions because nbformat makes that easy, and warnings which we find by inspecting the notebook after its finished running.

Eric-Arellano (approved):

PR #2403, opened 2024-11-27
[Do not merge] Diff after migrating away from jupyter-execute

arnaucasau:
This is just to show the git diff of the generation of the runtime dev docs after migrating away from` jupyter-execute`


docs/api/qiskit-ibm-runtime/dev/qiskit_ibm_runtime.transpiler.passes.scheduling.PadDynamicalDecoupling.mdx
```diff
@@ -36,52 +36,16 @@ python_api_name: qiskit_ibm_runtime.transpiler.passes.scheduling.PadDynamicalDec
...
-  ```
 
-  ```python
   # balanced X-X sequence on all qubits
   dd_sequence = [XGate(), XGate()]
   pm = PassManager([ALAPScheduleAnalysis(durations),
                     PadDynamicalDecoupling(durations, dd_sequence)])
   circ_dd = pm.run(circ)
-  circ_dd.draw()
+  circ_dd.draw('mpl', style="iqp")
```
Eric-Arellano, line 45: If we want these to be text based, we need to save the output in a `.. code:: text` block. But I think the image is fine. This text is not very readable to a screen-reader, so the image with alt text will be more useful to non-sighted users.

arnaucasau: Agreed! I like having an image there rather than the text much more too, but that's nice! TIL

PR #2405, opened 2024-11-28
Redesign nb-tester interface

### LINKED ISSUE #2401: [nb-tester]: Improve config
### Context

We currently organise all notebooks into categories based on how and when we can test them. To test functions notebooks, we're gonna need two types of patch:
1. Local simulator: This is what we currently use for notebooks that run small circuits. It runs locally and returns realistic results.
2. Test eagle: This runs on IBM Quantum and returns nonsense results. The functions notebooks introduce the need for this backend because functions only accept a backend _name_ (string) rather than a backend _object._

During CI, we want to be able to run some notebooks with the local sim and others with test-eagle. We also want to disable all patching in our fortnightly cron job, and skip some notebooks in all cases.

This issue aims to add this ability, but also to try and reign in some of the complexity of all these branching paths.

### Idea

The config file currently organizes notebooks into categories, maybe we can include the type of patch in the config file too? The new format would look something like this:

```toml
[group-name]
notebooks = [
  "path/to/notebook.ipynb"
]
patches = {\
  patch_a = { provider="qiskit-fake-provider" },\
  patch_b = { provider="qiskit-ibm-runtime", backend_name="test-eagle" }\
}
```

The CLI would then just take one patch-related arg: The type of patch to apply. For example, the following command would run all groups with their specification for "patch_a". Different groups of notebooks can have different definitions of `patch_a`.

```
--patch patch_a
```

If a notebook group does not have the specified patch, we skip that group. If the patch is an empty dict, we run the group without patching.

<details><summary>Example of what the Qiskit/documentation config would look like</summary>

```toml
[normal-test]
notebooks = ["..."]
patches = { ci = {}, hardware = {} }

[mock-local-sim]
notebooks = ["..."]
patches = {\
  ci = { provider="qiskit-fake-provider" },\
  hardware = {}\
}

[mock-test-eagle]
notebooks = ["..."]
patches = {\
  ci = { provider="qiskit-ibm-runtime", backend-name="test-eagle" },\
  hardware = {}\
}

[only-submit-jobs]
notebooks = ["..."]
patches = { hardware = {} }

[exclude]
notebooks = ["..."]
patches = {}
```

In CI, we'd pass the argument `--patch ci`. For our fortnightly cron job, we'd pass the arg `--patch hardware`.

</details>

frankharkins:
Attempt to simplify the way we control `nb-tester`.

This PR includes the patch information in the config file. Like before, each notebook is part of a group. The new aspect is each group can define how to patch each notebook in different situations.

This PR also adds tests and documentation for the config part of the script. If you exclude the new README and tests, this PR reduces the script size by ~100 LOC.

### Config example

Take the following config file as an example.

```toml
[my-group]
test-strategy.ci = { patch="qiskit_ibm_runtime", backend="test-eagle" }
test-strategy.hardware = {}
notebooks = ["..."]
```

If we run nb-tester against this file with `--test-strategy=ci`, each notebook in the group will be patched with the `test-eagle` device. If we run it with `--test-strategy=hardware`, the notebooks will be run with no patch. Running with a test strategy not defined for that group (e.g. `--test-strategy=new-strat`) means that group will be skipped.

### CLI example

You can also run the script without a config file with the `--patch` arg. In this case, all files listed after the `--patch` argument are part of the same "group". You must include a patch for this group using TOML syntax. For example:

```sh
test-docs-notebooks --patch '{ patch="qiskit-fake-provider", num_qubits=6 }' path/to/file.ipynb
```

***

Closes #2401

### DIFF
--- a/.github/workflows/notebook-test.yml
+++ b/.github/workflows/notebook-test.yml
@@ -47,7 +47,7 @@ jobs:
           """.strip().split("\n")
           import os
           github_output = os.getenv("GITHUB_OUTPUT")
-          all_files = "${{ steps.all-changed.outputs.all_changed_files }}".split("\n")
+          all_files = """${{ steps.all-changed.outputs.all_changed_files }}""".split("\n")
 
           config_changed = any(path.startswith("scripts/") for path in all_files)
           extra_deps = config_changed or any(path in EXTRA_DEPS_NOTEBOOKS for path in all_files)
--- a/scripts/config/notebook-testing.toml
+++ b/scripts/config/notebook-testing.toml
@@ -1,12 +1,12 @@
-default-strategy="ci"
+default-strategy = "ci"
 
 [test-strategies]
 ci = { timeout = 300 }
 
 # For notebooks to be tested in CI "normally" (no mocking)
 [groups.normal]
-test-strategies.ci={}
-test-strategies.hardware={}
+test-strategies.ci = {}
+test-strategies.hardware = {}
 notebooks = [
     "docs/guides/build-noise-models.ipynb",
     "docs/guides/circuit-library.ipynb",
@@ -46,8 +46,8 @@ notebooks = [
 
 # The following notebooks submit jobs that can be mocked with a simulator
 [groups.mockable]
-test-strategies.ci={ provider="qiskit-fake-provider", generic_backend_kwargs={ num_qubits=6 } }
-test-strategies.hardware={}
+test-strategies.ci = { provider="qiskit-fake-provider", generic_backend_kwargs={ num_qubits=6 } }
+test-strategies.hardware = {}
 notebooks = [
     "docs/guides/debug-qiskit-runtime-jobs.ipynb",
     "docs/guides/primitive-input-output.ipynb",
@@ -57,7 +57,7 @@ notebooks = [
 # A job is "too big" if a cell can't run in under 5 mins, or we run out of
 # memory on a reasonable device.
 [groups.cron-job-only]
-test-strategies.hardware={}
+test-strategies.hardware = {}
 notebooks = [
     "docs/guides/get-started-with-primitives.ipynb",
     "docs/guides/hello-world.ipynb",
--- a/scripts/config/notebook-testing.toml
+++ b/scripts/config/notebook-testing.toml
@@ -1,68 +1,83 @@
-# For notebooks to be tested "normally" (no mocking)
-notebooks_normal_test = [
-    "docs/guides/construct-circuits.ipynb",
+default-strategy="ci"
+
+[test-strategies]
+ci = { timeout = 300 }
+
+# For notebooks to be tested in CI "normally" (no mocking)
+[groups.normal]
+test-strategies.ci={}
+test-strategies.hardware={}
+notebooks = [
+    "docs/guides/build-noise-models.ipynb",
     "docs/guides/circuit-library.ipynb",
-    "docs/guides/visualize-circuits.ipynb",
     "docs/guides/classical-feedforward-and-control-flow.ipynb",
-    "docs/guides/operators-overview.ipynb",
-    "docs/guides/pulse.ipynb",
-    "docs/guides/save-circuits.ipynb",
-    "docs/guides/save-jobs.ipynb",
-    "docs/guides/dynamic-circuits-considerations.ipynb",
-    "docs/guides/get-qpu-information.ipynb",
-    "docs/guides/visualize-results.ipynb",
     "docs/guides/common-parameters.ipynb",
+    "docs/guides/construct-circuits.ipynb",
     "docs/guides/create-transpiler-plugin.ipynb",
     "docs/guides/custom-backend.ipynb",
     "docs/guides/custom-transpiler-pass.ipynb",
     "docs/guides/defaults-and-configuration-options.ipynb",
+    "docs/guides/dynamic-circuits-considerations.ipynb",
     "docs/guides/dynamical-decoupling-pass-manager.ipynb",
-    "docs/guides/represent-quantum-computers.ipynb",
-    "docs/guides/set-optimization.ipynb",
-    "docs/guides/transpile-with-pass-managers.ipynb",
-    "docs/guides/transpiler-plugins.ipynb",
-    "docs/guides/transpiler-stages.ipynb",
-    "docs/guides/build-noise-models.ipynb",
-    "docs/guides/local-testing-mode.ipynb",
-    "docs/guides/plot-quantum-states.ipynb",
-    "docs/guides/simulate-with-qiskit-aer.ipynb",
-    "docs/guides/simulate-stabilizer-circuits.ipynb",
-    "docs/guides/operator-class.ipynb",
     "docs/guides/error-mitigation-and-suppression-techniques.ipynb",
+    "docs/guides/get-qpu-information.ipynb",
+    "docs/guides/local-testing-mode.ipynb",
+    "docs/guides/operator-class.ipynb",
+    "docs/guides/operators-overview.ipynb",
+    "docs/guides/plot-quantum-states.ipynb",
+    "docs/guides/pulse.ipynb",
+    "docs/guides/qiskit-addons-aqc-get-started.ipynb",
+    "docs/guides/represent-quantum-computers.ipynb",
+    "docs/guides/save-circuits.ipynb",
+    "docs/guides/save-jobs.ipynb",
     "docs/guides/serverless-first-program.ipynb",
     "docs/guides/serverless-manage-resources.ipynb",
     "docs/guides/serverless-run-first-workload.ipynb",
+    "docs/guides/set-optimization.ipynb",
+    "docs/guides/simulate-stabilizer-circuits.ipynb",
+    "docs/guides/simulate-with-qiskit-aer.ipynb",
     "docs/guides/specify-observables-pauli.ipynb",
-    "docs/guides/qiskit-addons-aqc-get-started.ipynb",
+    "docs/guides/transpile-with-pass-managers.ipynb",
+    "docs/guides/transpiler-plugins.ipynb",
+    "docs/guides/transpiler-stages.ipynb",
+    "docs/guides/visualize-circuits.ipynb",
+    "docs/guides/visualize-results.ipynb",
 ]
 
-# Don't test the following notebooks (this section can include glob patterns)
-notebooks_exclude = [
-    "docs/guides/qedma-qesem.ipynb",
-    "docs/guides/q-ctrl-optimization-solver.ipynb",
-    "docs/guides/q-ctrl-performance-management.ipynb",
-    "docs/guides/algorithmiq-tem.ipynb",
-    "docs/guides/functions.ipynb",
-    "docs/guides/qunasys-quri-chemistry.ipynb",
-    "docs/guides/multiverse-computing-singularity.ipynb",
-    "docs/guides/ibm-circuit-function.ipynb",
-    "docs/guides/qiskit-addons-sqd-get-started.ipynb",
-    "docs/guides/fractional-gates.ipynb",
-]
-
-# The following notebooks submit jobs that can be mocked with a 5Q simulator
-notebooks_that_submit_jobs = [
-    "docs/guides/primitive-input-output.ipynb",
+# The following notebooks submit jobs that can be mocked with a simulator
+[groups.mockable]
+test-strategies.ci={ provider="qiskit-fake-provider", generic_backend_kwargs={ num_qubits=6 } }
+test-strategies.hardware={}
+notebooks = [
     "docs/guides/debug-qiskit-runtime-jobs.ipynb",
+    "docs/guides/primitive-input-output.ipynb",
 ]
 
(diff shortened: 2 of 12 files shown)


scripts/nb-tester/qiskit_docs_notebook_tester/config.py
```diff
@@ -59,159 +59,146 @@ def get_notebook_jobs(args: argparse.Namespace) -> Iterator[NotebookJob]:
...
+    if not patch:
+        return {}
+    try:
+        patch_info = tomllib.loads(f"patch={patch}")["patch"]
+    except tomllib.TOMLDecodeError as err:
+        raise ValueError(
+            "Problem parsing your --patch argument; "
+            "Make sure it's of the form '{ string=\"value\", number=4, ... }'."
+        ) from err
+    return patch_info
```
frankharkins, line 181: @emilkovacev this is a potentially disruptive change for you, does this new interface meet your needs?

Tl;dr: Rather than having separate CLI arguments for `--provider`/`--backend`/etc., you'd specify the whole patch as a TOML string, e.g.

```
--patch '{ provider="qiskit-fake-provider", backend="fake_athens" }'
```

This patch applies to all filenames passed to the script.

emilkovacev: I would also propose moving patches into files (like backend.patch), then having an extra required param in the toml for "filename" or "patch_file" that would point to the patch you want to apply. Then you can take the rest of the params in the dict and use `.format()` to inject variables into the patch.

The command would look something like this:
```
--patch '{ file="generic_backend.patch", backend="fake_athens" }'
```

And generic_backend.patch would look like this:
```
from qiskit.providers.fake_provider import GenericBackendV2
def patched_least_busy(self, *args, **kwargs):
    return GenericBackendV2(backend={backend}, ...)
```

In this case, all that would be needed to create a custom patch would be to create a new patch file, no new code is needed.

.github/workflows/main.yml
```diff
@@ -72,6 +72,32 @@ jobs:
         uses: ./.github/actions/set-up-notebook-testing
       - name: nb-tester tests
         run: tox -e nb-tester
+      - name: Check all notebooks are listed in config file
```
Eric-Arellano, line 75: Now that this test is a direct workflow, we could move the Python tests to be in the API check. That should speed up CI for content-only changes because we don't need to run tox. It's fine if that is changed in a follow up.

Also, I'd recommend this script be a dedicated file in `scripts/`. Even though it's pretty simple, it is more maintainable and discoverable living in a dedicated Python file. It keeps this YAML file simpler and allows us to do things like run Ruff on the file.

You could have a folder called `scripts/ci`?

.github/workflows/notebook-test.yml
```diff
@@ -68,9 +68,25 @@ jobs:
           args = ["tox", "--", "--write"]
```
Eric-Arellano, line 68: This has gotten complex enough that I think it could also benefit from being a script under `scripts/ci`.

Btw, you can have the changed files action write its output to a file. Imo it's more maintainable than this string interpolation. See https://github.com/Qiskit/documentation/pull/2389

scripts/config/notebook-testing.toml
```diff
@@ -1,68 +1,83 @@
-# For notebooks to be tested "normally" (no mocking)
-notebooks_normal_test = [
-    "docs/guides/construct-circuits.ipynb",
+default-strategy="ci"
```
Eric-Arellano, line 1: Style nit to use space between these, along with lines 9-10 and all

scripts/config/notebook-testing.toml
```diff
@@ -1,68 +1,83 @@
...
+
+[test-strategies]
+ci = { timeout = 300 }
+
+[groups]
```
Eric-Arellano, line 6: I don't think this is necessary since you're not setting anything underneath the key?

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
```diff
@@ -59,159 +59,146 @@ def get_notebook_jobs(args: argparse.Namespace) -> Iterator[NotebookJob]:
...
+                if matches(path, config.cli_filenames):
+                    # User explicitly defined this path so we'll explain why we're skipping it
+                    print(
+                        f'ℹ️ Skipping {path} (no test strategy "{config.test_strategy}" defined for this notebook)'
+                    )
+                continue
+
+            patch = config.get_patch_for_group(group)
+
+            def should_write(config: Config, patch: str | None):
```
Eric-Arellano, line 76: Why define a function, given that it's only invoked one time?

Eric-Arellano (commented):
I like this interface! It's a good design to only allow specifying one test strategy at a time. If you want to run with multiple test strategies, then run once per test strategy. That keeps the interface and code more understandable.

Nit in the PR title that this is not a "refactor" but a "revamp" or "redesign".

Consider splitting out the changes I proposed with `scripts/ci` as a prefactor. Fwict, they can be separated out from everything else. We've messed up CI config before, so it's worth having close review on them.

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
frankharkins: I thought it'd be more readable. Alternative is something like the following.

```python
write = None
if patch:
    write = Result(False, ...)
elif not config.write:
    write = Result(False, ...)
else:
   write = Result(True)
```

We should go with whichever is clearer to you; I've seen the code too much to make a good judgement.

Eric-Arellano: Python is really weird with variable scope. You don't need the initial `write = None`, unlike JS and most languages. 
Screenshot 2024-12-03 at 12 50 41 PM

I find that style the most readable.

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
frankharkins: > how would passing multiple patches look?

I hadn't planned to have that functionality through CLI arguments. At the moment, it's one patch for all filepaths passed to the CLI. Do you need the ability to apply different patches to different notebooks without a config file? I'm not sure how that'd look CLI-wise but happy to explore it.

> I would propose moving patches into files...

This is a great idea! I didn't fully understand it when you mentioned it the other day but I do now. I think this could reduce our code complexity a lot and make the interface easier to use.

scripts/config/notebook-testing.toml
frankharkins: Nice! I assumed you'd need to define the table before adding fields but it turns out you don't

```suggestion
```

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
emilkovacev: >Do you need the ability to apply different patches to different notebooks without a config file?

Once we introduce more patches for more classes (like SamplerV2 for instance) this would be very useful. 

For instance, if the interface looked like this:
```
--patch '{ file="generic_backend.patch", backend="fake_athens" }' \
--patch '{ file="samplerv2.patch", options={ default_shots=100 } }'
```

Then we could apply layers of patches very easily.

frankharkins: @emilkovacev added in 2c25a1844c3787317c38912c63a111e94885468e

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
```diff
@@ -30,6 +30,50 @@
 _set_mpl_loglevel("critical")
 """
 
+BUILT_IN_PATCHES = {
+    "qiskit-fake-provider": dedent(
+        """
+        import warnings
+        from qiskit.providers.fake_provider import GenericBackendV2
+        from qiskit_ibm_runtime import QiskitRuntimeService
```
emilkovacev, line 38: Can you move the builtin patches to files internal to the project for consistency? This will make it easier to edit/add builtin patches

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
```diff
@@ -59,143 +103,162 @@ def get_notebook_jobs(args: argparse.Namespace) -> Iterator[NotebookJob]:
...
+
+        patch_name = patch_config.get("patch", None)
+        if patch_name is None:
+            return None
+        if Path(patch_name).exists():
+            return Path(patch_name).read_text().format(**patch_config)
+        if patch_name in BUILT_IN_PATCHES:
+            return BUILT_IN_PATCHES[patch_name].format(**patch_config)
+
+        valid_patch_names = list(BUILT_IN_PATCHES.keys())
```
emilkovacev, line 220: Same comment as above, if we can unify the interface to use files, then we make it easier to add built-in patches

scripts/nb-tester/README.md
```diff
@@ -0,0 +1,168 @@
...
+variables as extra arguments in the TOML dict.
+
+```
+{ patch="qiskit-fake-provider", num_qubits=3 }
+```
+
+There are two ways to provide patches: Through the CLI or through a config file.
+You cannot use both at the same time.
+
+### Set patches through the CLI
```
emilkovacev, line 63: Not sure why, but looks like this line isn't rendered properly

emilkovacev (changes requested):
Great PR! Thank you for implementing the file patch interface, I think it will be really powerful. I included a couple of suggestions for how to utilize patches, and a small formatting error.

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
frankharkins: Thanks for reviewing!

I could put them in separate files and use [importlib](https://docs.python.org/3/library/importlib.resources.html) to read them, but we'd still need the aliases (e.g. `qiskit-fake-provider`) so users can refer to them. I feel like that would make things _more_ complex?

emilkovacev: From my perspective, you don't need aliases as long as the different options are documented. But if you do choose to use aliases, I think having the interface for adding a new patch be exclusively via a new file is easier to follow - i.e. if I want to add a bunch of patches into the repo I can just write the patch file, the same way I would make a custom patch. This would clean up the code for default patches as well. For instance, you could look for the patch name as a file, and if it doesn't work, check all patches in `qiskit_docs_notebook_tester/patches/`.

For instance:
`--patch { name="defaultpatch" }`
Would first search `./defaultpatch`, then `qiskit_docs_notebook_tester/patches/defaultpatch.patch`

This way I can also go to the repo and very easily see what default patches we have by going to `qiskit_docs_notebook_tester/patches/`.

Ideally I would like to be able to create new patch files as a PR into nb-tester in the future, but if it makes things too complex then I'm ok merging without it.

scripts/nb-tester/README.md
frankharkins: Weird, must have somehow typed a different unicode space character 😂 
I re-typed it into GitHub and that fixed it (bce4f675a2311a9a2c4738b0d01c8a9f7a6c9314)

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
frankharkins: Fair enough! Added in 938dea027450f1c606881928c14dd977d1fabf20

scripts/nb-tester/README.md
```diff
@@ -0,0 +1,168 @@
...
+specific backend so we can test them without waiting in a queue. We do this by
+monkey-patching `QiskitRuntimeService.least_busy` to return the backend of our
+choosing. This technique requires notebooks always use `least_busy` to select a
+backend.
+
+A patch is just Python code that we run in the kernel before executing your
+notebook's code. You can provide your own patch code, or use one of our
+built-in patches.
+
+You specify patch information as a TOML dict string. This dict _must_ inculde a
```
arnaucasau, line 31:
```suggestion
You can specify patch information as a TOML dict string. This dict _must_ include a
```

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
```diff
@@ -59,143 +60,166 @@ def get_notebook_jobs(args: argparse.Namespace) -> Iterator[NotebookJob]:
...
 
     @classmethod
     def from_args(cls, args: argparse.Namespace) -> Config:
         """
-        Create config from args, including loading the globs from the TOML file
+        Create config from args, including loading the paths from the TOML file if needed.
         """
-        path = Path(args.config_path)
+        if (args.config_path or args.test_strategy) and args.patch:
+            raise ValueError("Can't set --patch with --config-file or --test-strategy")
```
arnaucasau, line 107: I think it's okay to keep this check here with the custom message, but I recently found that we could use the mutual exclusion feature from `argparse` for these kinds of things: https://docs.python.org/3/library/argparse.html#mutual-exclusion

arnaucasau (approved):
Amazing work Frank! I left a small typo fix suggestion and a comment about the args that I think it's cool to know. I like a lot the new interface and the code looks good, so I think we can merge it once @emilkovacev's suggestions are implemented :)

PR #2410, opened 2024-11-29
[Refactor] Decouple config and execution in nb-tester 

frankharkins:
Pre-work for #2405.

This PR refactors nb-tester to decouple the config logic, which decides which notebooks to run and how to patch them, from the notebook execution logic, which runs the notebooks.

The config logic is already the most complicated part of the program. This PR decouples config and execution by moving them to separate files and having them communicate only by a single `NotebookJob` interface. Each `NotebookJob` contains a notebook path and instructions on how to run that notebook. This means the config logic can be scoped to creating a list of `NotebookJob` objects. This also makes the config logic easier to test.

There _should_ be no behaviour changes in this PR.


scripts/nb-tester/qiskit_docs_notebook_tester/config.py
```diff
@@ -0,0 +1,448 @@
...
+    return nb
+
+
+@dataclass
+class Result:
+    ok: bool
+    reason: str | None = None
+
+    def __bool__(self) -> bool:
+        return self.ok
```
frankharkins, line 76: Aside from separating the code and introducing the `NotebookJob` interface, this is the only other change. It's useful to know "why" a bool is set so we can give feedback to users. This was introduced for the `write` option but I've switched most functions that return a `bool` to return a `Result` object.

scripts/nb-tester/qiskit_docs_notebook_tester/execute.py
```diff
@@ -0,0 +1,140 @@
...
+            continue
+        for output in cell.outputs:
+            if hasattr(output, "name") and output.name == "stderr":
+                notebook_warnings.append(
+                    NotebookWarning(cell_index=cell_index, msg=output.text)
+                )
+    return notebook_warnings
+
+
+async def execute_notebook(job: NotebookJob) -> Result:
```
frankharkins, line 64: This function is logically the same, but it gets its information from the `NotebookJob` rather than the separate arguments.

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -373,210 +65,13 @@ def cancel_trailing_jobs(start_time: datetime, config_path: str) -> bool:
...
     ]
     if not jobs:
-        return True
+        return Result(True)
 
     print(
         f"⚠️ Cancelling {len(jobs)} job(s) created after {start_time}.\n"
         "Add any notebooks that submit jobs to `notebooks-that-submit-jobs` in "
-        f"`{config_path}`."
+        f"`scripts/config/notebook-testing.toml`."
```
frankharkins, line 73: Hardcoded for now but will make more general in a follow-up

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
```diff
@@ -0,0 +1,448 @@
...
+            return "hardware was mocked"
+        return False
+
+
+def matches(path: Path, glob_list: list[str]) -> bool:
+    return any(path.match(glob) for glob in glob_list)
+
+
+
+def get_args() -> argparse.Namespace:
```
arnaucasau, line 226: Maybe we can keep this function in `__init__.py` to reduce the complexity of the `config.py` script. Although, I think it's fine either way if you like it more here :)

arnaucasau (commented):
Awesome work Frank! I like the communication between scripts using `NotebookJob` a lot!

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
Eric-Arellano: I think it makes sense here since it's about config

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
```diff
@@ -0,0 +1,448 @@
...
+We recommend using these versions or newer.
+
+```
+{get_package_versions()}
+```
+</details>
+"""
+
+
+def post_process_notebook(nb: nbformat.NotebookNode) -> nbformat.NotebookNode:
```
Eric-Arellano, line 59: Nit: should this go in execute? It's not configurable. Or even a new `version_info.py` or `post_process.py`?

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -13,352 +13,44 @@
...
-    nb, _ = clean_notebook(nb)
-    return nb
+def main():
+    if platform.system() == "Windows":
+        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+    asyncio.run(_main())
 
 
-def cancel_trailing_jobs(start_time: datetime, config_path: str) -> bool:
+def cancel_trailing_jobs(start_time: datetime) -> Result:
```
Eric-Arellano, line 51: I'd probably expect this to be in `execute.py` or a `post_process.py` so that `__init__.py` is only the very high-level business logic.

Eric-Arellano (approved):
Thanks!

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -373,210 +65,13 @@ def cancel_trailing_jobs(start_time: datetime, config_path: str) -> bool:
...
-            "If you have write access to Qiskit/documentation, push to a new "
-            "branch there and make your pull request from that branch instead."
-            "\n\n"
-            "If you don't have write access, you should locally test out the notebooks you're modifying "
-            "by using the instructions in "
-            "https://github.com/Qiskit/documentation#execute-notebooks. "
-            "When this PR is approved, a maintainer will merge it to a new "
-            "branch in Qiskit/documentation, then make a PR from that branch "
-            "into main so it can pass CI.\n",
-        )
```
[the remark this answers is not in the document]
Eric-Arellano: Good catch. It'd be helpful to keep this.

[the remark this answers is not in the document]
frankharkins: Good spot! I plan to move that functionality to our GitHub action because it's very specific to our repo, but it shouldn't have been removed in this PR.

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
frankharkins: Yeah that's simpler, also means we're only passing serializable data between the two which is nicer.

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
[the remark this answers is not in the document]
Eric-Arellano: It used to live there, but we decided to move it to the script in https://github.com/Qiskit/documentation/pull/2062. I'm fine with moving this back to the GH action, so long as it preserves the same prior functionality of not failing when the notebooks couldn't have been tested anyways.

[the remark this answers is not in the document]
frankharkins: I remember that, here's what I had in mind. It is the kind of thing I'd forget so thanks for reminding me.

https://github.com/Qiskit/documentation/blob/9dbd215c8e31db1c69e7dcd1b069c6bf47fe2b04/.github/workflows/notebook-test.yml#L71-L88

PR #2418, opened 2024-12-02
Regen Qiskit 1.3 with Sphinx improvements

Eric-Arellano:
Applies https://github.com/Qiskit/qiskit/pull/13483


arnaucasau (approved):
Nice improvements! Thanks!

Eric-Arellano:
Note that results in a new bug https://github.com/Qiskit/documentation/issues/2419. I don't think that bug should block this PR - it's still an improvement.

Issue #1607, opened 2024-06-27
Improve Windows support for local previews


Eric-Arellano:
We got PR previews working for a Windows-user! But they had issues using Windows Subsystem for Linux and had to use Powershell to get Docker (from Rancher Desktop) to load.

How can Powershell users run `./start`?

PR #2421, opened 2024-12-02
Improve `./start` to work on Windows

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/1607. There are two issues:

1. The shebang will be ignored, so you need to run `python ./start` instead of `./start`
2. Volume mounting with Docker on Windows uses Unix-style path, per https://medium.com/@kale.miller96/how-to-mount-your-current-working-directory-to-your-docker-container-in-windows-74e47fa104d7

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ You can preview the docs locally by following these two steps:
 
 1. Ensure Docker is running. For example, open Rancher Desktop.
 2. Run `./start` in your terminal, and open http://localhost:3000 in your browser.
-   - On Windows, run `python start` instead.
+   - On Windows, run `python start` instead. Alternatively, use Windows Subsystem for Linux and run `./start`.
 
 The preview application does not include the top nav bar. Instead, navigate to the folder you want with the links in the home page. You can return to the home page at any time by clicking "IBM Quantum Documentation Preview" in the top-left of the header.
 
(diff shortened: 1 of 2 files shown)


Eric-Arellano:
TODO: confirm with @beckykd this fixes things.

Eric-Arellano:
@frankharkins @arnaucasau before landing this, please see if the status quo works with Git Bash for Becky. If it does, please confirm this change does not break Git Bash and it also fixes PowerShell.

Imo supporting PowerShell is not essential, so long as Git Bash works on Windows.

README.md
```diff
@@ -77,6 +77,7 @@ You can preview the docs locally by following these two steps:
 
 1. Ensure Docker is running. For example, open Rancher Desktop.
 2. Run `./start` in your terminal, and open http://localhost:3000 in your browser.
+   - On Windows, run `python start` instead.
```
Eric-Arellano, line 80:
```suggestion
   - On Windows, run `python start` instead. Alternatively, use Windows Subsystem for Linux and run `./start`.
```

Eric-Arellano:
We got it working on both Powershell and Git Bash 🚀

Eric-Arellano:
@arnaucasau @frankharkins bump on reviewing this please

arnaucasau (approved):
Awesome! I tested the script on a Windows machine using PowerShell and Git Bash and worked well! Thanks :rocket:

PR #2426, opened 2024-12-04
Extend the images' checker to check API docs

arnaucasau:
This PR extends the images' checker introduced in https://github.com/Qiskit/documentation/pull/2349 to check the API docs when the new argument `--apis` is set.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,7 +29,7 @@ jobs:
       - name: File metadata
         run: npm run check:metadata
       - name: Check images
-        run: npm run check:images
+        run: npm run check:images -- --apis
       - name: Spellcheck
         run: npm run check:spelling
       - name: Check Qiskit bot config
--- a/scripts/js/commands/checkImages.ts
+++ b/scripts/js/commands/checkImages.ts
@@ -11,11 +11,43 @@
 // that they have been altered from the originals.
 
 import { globby } from "globby";
+import yargs from "yargs/yargs";
+import { hideBin } from "yargs/helpers";
 import { collectInvalidImageErrors } from "../lib/markdownImages.js";
 import { readMarkdown } from "../lib/markdownReader.js";
 
+// Known APIs containing images without alt text that will be skipped
+// by the checker.
+const APIS_TO_IGNORE__SHOULD_FIX: string[] = [
+  "qiskit",
+  "qiskit-ibm-runtime",
+  "qiskit-addon-cutting",
+  "qiskit-addon-mpf",
+  "qiskit-addon-obp",
+  "qiskit-addon-utils",
+];
+
+interface Arguments {
+  [x: string]: unknown;
+  apis: boolean;
+}
+
+const readArgs = (): Arguments => {
+  return yargs(hideBin(process.argv))
+    .version(false)
+    .option("apis", {
+      type: "boolean",
+      default: false,
+      description:
+        "Check the images in the current and dev versions of the API docs.",
+    })
+    .parseSync();
+};
+
 async function main() {
-  const files = await globby(["docs/**/*.{ipynb,mdx}", "!docs/api/**/*.mdx"]);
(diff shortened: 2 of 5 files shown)


scripts/js/commands/checkImages.ts
```diff
@@ -11,11 +11,43 @@
...
+}
+
+const readArgs = (): Arguments => {
+  return yargs(hideBin(process.argv))
+    .version(false)
+    .option("apis", {
+      type: "boolean",
+      default: false,
+      description:
+        "Check the images in the current and dev versions of the API docs.",
```
frankharkins, line 42:
```suggestion
        "Check the images in the current and dev versions of the API docs have alt text.",
```

scripts/js/commands/checkImages.ts
```diff
@@ -24,19 +56,38 @@ async function main() {
...
+  // historical version that wasn't in the original release, that's very unlikely.
+  // If it happens, it would probably be a backport from latest or dev, and the linter in
+  // the API repo should catch it.
+  //
+  // If an image is missed by the API repo's linter, it will still have an alt text defined,
+  // although it won't be very useful.
+  const globs = [
+    "docs/**/*.{ipynb,mdx}",
+    "!docs/api/qiskit/release-notes/*.mdx",
+    args.apis ? "!docs/api/*/([0-9]*)/*.mdx" : "!docs/api/**/*.mdx",
```
frankharkins, line 86: Should we check release notes with `--apis`? If it's slow then we could ignore anything < 1.0

```
!docs/api/qiskit/release-notes/0.*.mdx
```

scripts/js/lib/markdownImages.ts
```diff
@@ -28,7 +29,10 @@ export async function collectInvalidImageErrors(
     .use(remarkGfm)
     .use(() => (tree: Root) => {
       visit(tree, "image", (node) => {
-        if (!node.alt) {
+        // Sphinx uses the image path as alt text if it wasn't defined using the
+        // :alt: option.
+        const imageName = last(split(node.url, "/"));
+        if (!node.alt || node.alt.endsWith(imageName!)) {
```
frankharkins, line 35: Good call

frankharkins (approved):

scripts/js/commands/checkImages.ts
arnaucasau: Yeah, I think it's a good idea to check the release notes. I've checked and it's not adding a lot of extra time so it's probably okay to check all of them. Thanks!

.github/workflows/main.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: File metadata
         run: npm run check:metadata
       - name: Check images
-        run: npm run check:images
+        run: npm run check:images -- --apis
```
Eric-Arellano, line 32: Insted, this should go into https://github.com/Qiskit/documentation/blob/main/.github/workflows/api-checks.yml. We want the main job to be as fast as possible for non-API changes. We're okay with API changes being slightly slower (repeating the check of non-API)

scripts/js/commands/checkImages.ts
```diff
@@ -24,19 +56,37 @@ async function main() {
...
 
+async function determineTocFiles(args: Arguments): Promise<string[]> {
+  // We always skip historical versions to simplify the code and to have a faster script.
+  // Even though it is possible for someone to add a new image without alt text to a
+  // historical version that wasn't in the original release, that's very unlikely.
+  // If it happens, it would probably be a backport from latest or dev, and the linter in
+  // the API repo should catch it.
+  //
+  // If an image is missed by the API repo's linter, it will still have an alt text defined,
+  // although it won't be very useful.
```
Eric-Arellano, line 82: It'd be helpful to add

> That's because Sphinx auto-generates alt text.

Eric-Arellano (commented):
Nice work!

PR #2437, opened 2024-12-06
Add additional backends for gen3-turbo

jyu00:
Add `ibm_brisbane` and `ibm_strasbourg` to the gen3-turbo enabled backend list.


abbycross (approved):
👍

PR #2445, opened 2024-12-09
Add the sphinx-alt-text-validator tool to VCS install it in each API repo

arnaucasau:
This PR adds a tool that all the APIs should use to avoid having images without alt text. By having the script in a central repository, we could use the VCS method to pip install it everywhere instead of duplicating the same script 9 times. 

The tool is currently being used in qiskit, runtime, and transpiler. Code -> https://github.com/Qiskit/qiskit/blob/main/tools/verify_images.py

The code added in this PR is the same but split into two files (`__init__.py` and `verify_images.py`) and with the possibility of using two new arguments to specify the folder where the linter will check their files and an argument to enumerate all the files the user wants to skip, replacing the `ALLOWLIST_MISSING_ALT_TEXT` list that was defined at the beginning of the script.

To install and test the tool we use

```bash
pip install -e "git+https://github.com/arnaucasau/documentation.git@AC/sphinx-alt-text-validator#egg=sphinx-alt-text-validator&subdirectory=scripts/image-tester"
```
and run `sphinx-alt-text-validator -f [FolderToCheck]`

> [!NOTE]  
> This PR only copies the script to the repo and adds the necessary files to make it installable from the API repos. In a follow-up, I'm planning to add some tests and maybe do a little refactoring to the verify_image.py


scripts/image-tester/sphinx_alt_text_validator/verify_images.py
```diff
@@ -0,0 +1,68 @@
...
+    """Validate one single image"""
+    alt_exists = any(option.strip().startswith(":alt:") for option in options)
+    nofigs_exists = any(option.strip().startswith(":nofigs:") for option in options)
+
+    # Only `.. plot::`` directives without the `:nofigs:` option are required to have alt text.
+    # Meanwhile, all `.. image::` directives need alt text and they don't have a `:nofigs:` option.
+    return alt_exists or nofigs_exists
+
+
+def validate_image(file_path: str) -> tuple[str, list[str]]:
```
arnaucasau, line 39: Same function but removing this conditional:
```python
if file_path in ALLOWLIST_MISSING_ALT_TEXT:
    return [file_path, []]
```

Now the function gets the files filtered from `__init__.py` so no need to verify any allowlist anymore

scripts/image-tester/sphinx_alt_text_validator/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+
+def main() -> None:
+    parser = argparse.ArgumentParser(prog="verify_images.py")
+    parser.add_argument("-f", "--folder", required=True)
+    parser.add_argument("-s", "--skip", nargs="+")
+    args = parser.parse_args()
+
+    skip_list = args.skip if args.skip is not None else []
+    files = glob.glob(f"{args.folder}/**/*.py", recursive=True)
+    filtered_files = [file for file in files if file not in skip_list]
```
arnaucasau, line 29: This is the only new code added in this PR

scripts/image-tester/sphinx_alt_text_validator/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+import argparse
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(prog="verify_images.py")
+    parser.add_argument("-f", "--folder", required=True)
+    parser.add_argument("-s", "--skip", nargs="+")
+    args = parser.parse_args()
+
+    skip_list = args.skip if args.skip is not None else []
```
frankharkins, line 27: You might like this:

```suggestion
    skip_list = args.skip or []
```

scripts/image-tester/sphinx_alt_text_validator/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+        for image_error in image_errors:
+            print(image_error, file=sys.stderr)
+
+    print(
+        "\nAlt text is crucial for making documentation accessible to all users.",
+        "It should serve the same purpose as the images on the page,",
+        "conveying the same meaning rather than describing visual characteristics.",
+        "When an image contains words that are important to understanding the content,",
+        "the alt text should include those words as well.",
+        file=sys.stderr,
```
frankharkins, line 56: Great message!

frankharkins (approved):
Great job! I did a quick test and seems to work well :)

When this is merged to main, make sure to pin to the commit when installing in other repos. This means we can make changes to the script without worrying about accidentally breaking their CI.

```sh
pip install -e "git+https://github.com/Qiskit/documentation.git@<COMMIT-HASH>#egg=sphinx-alt-text-validator&subdirectory=scripts/image-tester"
```

scripts/image-tester/sphinx_alt_text_validator/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+import argparse
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(prog="verify_images.py")
+    parser.add_argument("-f", "--folder", required=True)
+    parser.add_argument("-s", "--skip", nargs="+")
+    args = parser.parse_args()
+
+    skip_list = args.skip or []
```
Eric-Arellano, line 27: Nit: using a set will be faster when the skip list is larger

Eric-Arellano (commented):
Nice work! Can you please add in a follow up a simple Pytest test for `validate_image`? Now that it's a public API, it would be good to ensure we don't break it in the future. 

- You can reuse the Pytest infrastructure we have for nb-tester. I'd probably rename the Tox env to be something like `tests`
- You'll probably want to change `validate_image` to be a pure function and have the text be passed in directly rather than using Pathlib

PR #2447, opened 2024-12-09
Test notebooks with `test-eagle`

frankharkins:
### New test strategy

This PR adds a new "extended" test that runs the notebooks on IBM Quantum's `test-eagle` device. This device accepts jobs and returns random bit strings in a very short time period, which is great for testing notebooks that submit jobs too large for a local simulator. This also allows us to test functions notebooks, which take a backend name (string) of an IBM Quantum device, rather than the backend object itself.

I originally hoped we could test notebooks with `test-eagle` in CI, but the jobs can be flaky and take just a little too long to complete. Running these jobs on every commit is too expensive and could block writers.

Instead, I've added a new "extended" test that we can trigger manually on branches when we need it. Examples could be dependabot PRs, or PRs to functions notebooks; when the PR is ready, we'll trigger the workflow on the PR branch to check the notebooks run correctly. To trigger the workflow, go to Actions > Test notebook (extended) > Run workflow, then select the PR branch and click "Run workflow".

### Re-categorise notebooks

With the new "extended" test strategy available, I've recategorised the notebooks so we can test them. I've also modified some notebooks slightly so they'll run from start to finish.

### New patch type

This PR adds a new type of patch: `qiskit-ibm-runtime-open`, which patches `least_busy` to only return backends in the open plan. This is to stop `least_busy` from always returning `test_eagle_us-east`, which is always the least busy backed in our testing instance.

### Allow writing some patches to disk

Finally, this PR adds a feature to allow writing some patched notebooks to disk. We haven't allowed this in the past, but it makes sense with the new `qiskit-ibm-runtime-open` patch.

***

After writing this all out, I realise I should split this into smaller PRs. Stay tuned.


docs/guides/ibm-circuit-function.ipynb
```diff
@@ -342,7 +342,11 @@
    "cell_type": "code",
    "execution_count": 8,
    "id": "4070e592",
-   "metadata": {},
+   "metadata": {
+     "tags": [
+       "raises-exception"
+     ]
```
frankharkins, line 348: Nice feature of most Jupyter execution tools. The exception in this cell is intentional.

docs/guides/q-ctrl-optimization-solver.ipynb
```diff
@@ -476,7 +491,9 @@
    "source": [
     "# Solve the problem\n",
     "mvc_job = solver.run(\n",
-    "    problem=srepr(cost_function), instance=hub + \"/\" + group + \"/\" + project\n",
+    "    problem=srepr(cost_function),\n",
+    "    instance=instance,\n",
```
frankharkins, line 495:
```suggestion
    "    instance=instance,  # E.g. \"ibm-q/open/main\"\n",
```

docs/guides/q-ctrl-optimization-solver.ipynb
```diff
@@ -261,7 +275,8 @@
     "maxcut_job = solver.run(\n",
     "    problem=problem_as_str,\n",
     "    problem_type=\"maxcut\",\n",
-    "    instance=hub + \"/\" + group + \"/\" + project,\n",
+    "    instance=instance,  # E.g. \"ibm-q/open/main\"\n",
+    "    backend_name=backend_name,  # E.g. \"ibm_kyiv\"\n",
```
frankharkins, line 279: This is the main change to most notebooks: We load the `backend_name` and `instance` variables in a hidden cell, then use them in the function call with a comment giving an example of the value.

scripts/nb-tester/patches/qiskit-ibm-runtime-open
frankharkins, line 1: I've added a new patch. This is because we've added the `test_eagle_us-east` device to the instance we use for testing. The problem is that this device is always the least busy, so it'll be selected when we're doing our hardware runs. We can't just switch to the open instance for this test because I'm regularly out of credits and the priority is too low to run all the jobs in our time window.

This new patch picks the least busy backend that's also available on the open plan, regardless of which instance you're using.

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
```diff
@@ -74,7 +74,7 @@ def get_notebook_jobs(args: argparse.Namespace) -> Iterator[NotebookJob]:
 
             patch = config.get_patch_for_group(group)
 
-            if patch:
+            if patch and not "# nb-tester: allow-write" in patch:
```
frankharkins, line 77: New minor feature: We can allow writing notebooks to disk with specific patches, see the new README entry for more information.

frankharkins:
Closing because I'm going to split this into smaller PRs

PR #2456, opened 2024-12-10
Allow writing nb-tester results with certain patches

frankharkins:
By default, nb-tester will **not** allow writing patched notebooks to disk. This is to reduce the risk of accidentally displaying results from patched notebooks to users. For example, if we write the output of a notebook patched with `qiskit-fake-provider`, it will appear the results in the notebook come from the least busy IBM Quantum backend when they actually come from a noiseless simulator. This could be considered false advertising.

However, we do have patches with which it's safe to generate user-facing results. This PR adds the ability to allow writing results from these patches by adding the comment `# nb-tester: allow-write` anywhere in the patch file.

The built-in `qiskit-ibm-runtime-open` patch is an example of such a patch: It only selects a real hardware backend that's included in the open plan. We use this to select open-access backends but submit jobs using our dedicated testing instance.

### DIFF
--- a/scripts/nb-tester/README.md
+++ b/scripts/nb-tester/README.md
@@ -166,3 +166,21 @@ Here's a few different commands you could run:
   wasn't passed as a filename arg.
 
 </details>
+
+### Patches and writing to disk
+
+By default, this tool will **not** allow writing patched notebooks to disk.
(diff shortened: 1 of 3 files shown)


arnaucasau (approved):
Nice! Thanks!

PR #2463, opened 2024-12-11
Test notebooks with `test-eagle`

### LINKED ISSUE #2198: `qedma-qesem.ipynb` has undefined variables
There are quite a few undefined variables in this notebook.

We should discuss if this is something we want to allow. If not, then we need to fix the notebook.

### LINKED ISSUE #2239: Test functions notebooks in fortnightly cron job
Investigate if we can test any of the functions notebooks in the fortnightly cron job.

**Todo:**
- [ ] Check which notebooks we can already test
- [ ] (Optional) Add feature to find/replace standardized strings like `<YOUR_API_TOKEN>` and `<YOUR_IQP_GROUP>` when testing
- [ ] Add testable notebooks to cron job

Large notebook changes are out of scope of this issue.

### LINKED ISSUE #2240: Use testing backends with functions notebooks
We can test functions that don't post-process results using the IBM test backends. This will submit jobs that should run very quickly and return nonsense results. We'll need to add some more patching code, but it'll hopefully mean we can test these notebooks on PR, which would be great.

**Todo:**

- [ ] Identify candidate notebooks (that is, any notebooks that don't post-process results)
- [ ] Add feature to patch these notebooks and prevent writing results
- [ ] Add candidate notebooks to PR checks

frankharkins:
This PR adds a new "extended" test that runs the notebooks on IBM Quantum's `test-eagle` device. This device accepts jobs and returns random bit strings in a very short time period, which is great for testing notebooks that submit jobs too large for a local simulator. This also allows us to test functions notebooks, which take a backend name (string) of an IBM Quantum device, rather than the backend object itself.

I originally hoped we could test notebooks with test-eagle in CI, but the jobs can be flaky and take just a little too long to complete. Running these jobs on every commit is too expensive and could block writers.

Instead, I've added a new "extended" test that we can trigger manually on branches when we need it. Examples could be dependabot PRs, or PRs to functions notebooks; when the PR is ready, we'll trigger the workflow on the PR branch to check the notebooks run correctly. To trigger the workflow, go to Actions > Test notebook (extended) > Run workflow, then select the PR branch and click "Run workflow".

This PR also modifies some of the notebooks to make them executable from start to finish.

***

Fixes #2240, fixes #2239, and fixes #2198.


docs/guides/algorithmiq-tem.ipynb
```diff
@@ -175,7 +175,9 @@
     "service = QiskitRuntimeService()\n",
     "backend_name = service.least_busy(operational=True).name\n",
     "\n",
-    "instance = \"<IQP_HUB/IQP_GROUP/IQP_PROJECT>\"\n",
+    "# Instance is a string of the form \"<IQP_HUB/IQP_GROUP/IQP_PROJECT>\".\n",
+    "# The following line gets the instance of the active QiskitRuntimeService account.\n",
+    "instance = service.active_account()[\"instance\"]\n",
```
arnaucasau, line 180: I see that in the rest of the notebooks we hide this line from the users. Do we want to show it here?

frankharkins: No strong preference, I think I left this one in because it's already using `QiskitRuntimeService`, whereas the other notebooks didn't.

frankharkins: Alternatively, we could un-hide the other cells and add a similar comment

arnaucasau: Oh, I see. Then, I think we can leave it as is :+1:

arnaucasau (approved):
Thanks!

PR #2497, opened 2024-12-17
Categorize fractional gates guide as cron-job only

### LINKED ISSUE #2454: Non-open-access backend used in `fractional-gates.ipynb`
The fractional gates notebook uses a backend that is not available on `ibm-q/open/main`. We should switch this notebook to use an open-access backend so all users can run the code examples on the page.

frankharkins:
Switches the notebook to use an open-access backend. Fixes #2454.


docs/guides/fractional-gates.ipynb
```diff
@@ -28,7 +28,7 @@
     "\n",
```
frankharkins, line 28: <p>@kaelynj these outputs look quite a bit different? Is this ok?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Qiskit/documentation/pull/2497/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

frankharkins: <p>Looking at the page, it says:</p><blockquote>These <em>fractional</em> gates are supported on <a href="https://qiskit.github.io/documentation/pr-2497/guides/processor-types#heron" target="_blank">Heron QPUs</a> in the form...</blockquote><p>But there are no open-access heron backends, only eagle. Does that mean this feature is only available to non-open-plan users?</p>

kaelynj: <p>Ah yeah, then this feature isn't available to non-open plan users at this point. There we some Heron QPUs which were part of the open plan that have been retired.</p>

Eric-Arellano (approved):

PR #2499, opened 2024-12-17
Add docs on installing nb-tester

frankharkins:
We already include docs on the config file and CLI, just need to explain how to install.


Eric-Arellano (approved):
🚀

PR #2425, opened 2024-12-04
Separate repo-specific functionality from nb-tester

frankharkins:
Splits out part of #2405

This PR pulls some of our CI functionality out of the nb-tester package and into separate scripts.

PR #2502, opened 2024-12-18
Add tests to sphinx-alt-text-validator

arnaucasau:
This PR creates tests for the sphinx-alt-text-validator and recovers the tests of the nb-tester in CI. See https://github.com/Qiskit/documentation/pull/2445#pullrequestreview-2507332120 and https://github.com/Qiskit/documentation/pull/2425#discussion_r1887629503

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,8 +23,9 @@ commands =
     fix: ruff format {posargs:docs}
     fix: ruff check --fix {posargs:docs}
 
-[testenv:nb-tester]
+[testenv:tests]
 deps =
   -e scripts/nb-tester
+  -e scripts/image-tester
   pytest
-commands = pytest scripts/nb-tester
+commands = pytest scripts/nb-tester scripts/image-tester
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -23,8 +23,9 @@ commands =
...
     fix: ruff check --fix {posargs:docs}
 
-[testenv:nb-tester]
+[testenv:tests]
 deps =
   -e scripts/nb-tester
+  -e scripts/image-tester
   pytest
-commands = pytest scripts/nb-tester
+commands = pytest scripts/nb-tester scripts/image-tester
```
arnaucasau, line 31: Interestingly this seems to work fine but the output shows the tests as if they were all in the `scripts/nb-tester` folder :thinking: 

```
scripts/nb-tester/test/test_get_notebook_jobs.py .........               [ 90%]
scripts/nb-tester/test/test_validate_image.py .                          [100%]
```

I think we can solve it by just running `pytest` instead of `pytest scripts/nb-tester scripts/image-tester` but perhaps there's a better approach.

frankharkins: Just running `pytest` is fine by me

Eric-Arellano (approved):
🙏

PR #2503, opened 2024-12-18
Generate qiskit-ibm-transpiler 0.9.3 API docs

arnaucasau:
This PR generates the docs for qiskit-ibm-transpiler v0.9.3

```bash
npm run gen-api -- -p qiskit-ibm-transpiler -v 0.9.3
```


docs/api/qiskit-ibm-transpiler/qiskit_ibm_transpiler.ai.AICliffordSynthesis.mdx
```diff
@@ -76,7 +76,7 @@ python_api_name: qiskit_ibm_transpiler.ai.AICliffordSynthesis
 
     **Parameters**
 
-    **dag** ([*DAGCircuit*](/api/qiskit/qiskit.dagcircuit.DAGCircuit "(in Qiskit v1.3)")) – the dag on which the pass is run.
```
frankharkins, line 79: Do we know why these links disappeared?

frankharkins (approved):
Thanks!

docs/api/qiskit-ibm-transpiler/qiskit_ibm_transpiler.ai.AICliffordSynthesis.mdx
arnaucasau: I think it's because of a change in the requirements (this [line](https://github.com/Qiskit/qiskit-ibm-transpiler/blob/e349523bb6066060659ff4adb2acb8c7de246f0f/requirements-dev.txt#L1)), but I'm not completely sure.

As I understand, before, we were setting this dependency `qiskit~=1.0`, and that meant we wanted the latest version of qiskit greater or equal to 1.0, which is 1.3. 

Now, by using `qiskit>=1.0,<1.3`, there's more than one version allowed, and I guess the link is not generated for a specific version as before.

Does that make sense @Eric-Arellano?

frankharkins: That makes sense, thanks.

Not related to this PR, but I'm slightly concerned about them pinning Qiskit < 1.3. We're currently using Qiskit 1.3 so this means we won't be able to upgrade to the latest version of the transpiler service

arnaucasau: Agreed! I think they are already working on it -> https://github.com/Qiskit/qiskit-ibm-transpiler/pull/144

PR #2504, opened 2024-12-18
Hamiltonian simulation function: Display entire source file at bottom

garrison:
I originally wanted to be able to validate that this is the same as the code displayed above.  I had planned to do this using a cell with a `%%writefile` directive, like is above in this file.  Given this, I could have asserted that the two files are equivalent.  Unfortunately I don't think it is possible to use Jupyter directives like this within an HTML block in a notebook.


jenglick:
Looks good, thanks Jim!

PR #2512, opened 2024-12-20
fixed typo

ne7ermor3:
cleaned up some instances of typo throughout the documentation, namely "each each" instead of "each".


Eric-Arellano:
Hey @ne7ermor3 , thank you! Pardon we didn't make this more clear. To fix https://github.com/Qiskit/qiskit/issues/13576, you'll need to open PRs in https://github.com/Qiskit/qiskit and https://github.com/Qiskit/qiskit-ibm-runtime to fix the Python docstring. That's because the API docs are generated from Sphinx in those two repositories and converted into the MDX files you see here. So, your current PR would be reverted the next time we re-generate the API docs.

Issue #2517, opened 2024-12-20
Make API dev docs images deterministic


Eric-Arellano:
I'm [reading](https://www.atlassian.com/git/tutorials/big-repositories) that binary files are particularly bad for Git repo size. Beyond Git repo size, the diff is annoying that ever time we generate dev API docs, we have unrelated image changes, like https://github.com/Qiskit/documentation/pull/2509.

So, see if we can make the images deterministic, even a few of them.

PR #2524, opened 2024-12-23
[Suggestion] Don't commit images with low visual change in dev docs sync

frankharkins:
Experiment to avoid diffs for images with basically no change. Threshold was chosen by eyeballing. Here's two examples:

**Example below similarity threshold**
below-threshold

**Example above similarity threshold**
above-threshold


frankharkins:
Closing for now after discussion with team.

Eric-Arellano:
> Closing for now after discussion with team.

For posterity, it was closed because of concerns that we would miss changes like the label for a Gate changing. It's tricky to get the threshold right, and we found it tricky in qiskit-sphinx-theme with visual regression testing.

We may want to make this change, but we should start with https://github.com/Qiskit/documentation/issues/2517

Issue #2518, opened 2024-12-20
Fix performance of checkStaleImages.ts


Eric-Arellano:
It is no longer viable to use because we have so many images. Ideas to improve:

* Use ripgrep
* Batch by non-API then each API

PR #2528, opened 2024-12-23
Fix performance of stale image check

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/2518.

We were before checking every image inside `await Promise.all()` and only logging results at the end. This is too slow to keep that much in memory, and it results in the experience feeling laggy to not get results until the very end.

Instead, it makes more sense to check one image at a time and eagerly report violations. 

Other improvements in this PR:

* Use ripgrep rather than git grep. The downside is requiring users have ripgrep installed, but this is a manual script we run with `npx tsx`, so it should be fine.
* Log progress every 20 images
* Add an allowlist.

<details><summary>example run</summary>

```
❌ image is unused: public/images/migration/dedicated_queue_no_jobs1.jpg
❌ image is unused: public/images/migration/switcher-small2.png
❌ image is unused: public/images/guides/normal_queue_with_providers1.jpg
❌ image is unused: public/images/guides/paulibasis.png
⏳ Checked 20 / 3519 images
❌ image is unused: public/images/qiskit-ibm-runtime/noisy-sim-circuit.png
❌ image is unused: public/images/qiskit-ibm-runtime/noisy-sim-estimator-circuit.png
❌ image is unused: public/images/qiskit-ibm-runtime/noisy-sim-estimator-ideal.png
❌ image is unused: public/images/qiskit-ibm-runtime/noisy-sim-estimator-noisy.png
❌ image is unused: public/images/qiskit-ibm-runtime/noisy-sim-sampler-ideal.png
⏳ Checked 40 / 3519 images
❌ image is unused: public/images/qiskit-ibm-runtime/noisy-sim-sampler-noisy.png
❌ image is unused: public/images/api/qiskit-addon-cutting/qiskit_addon_cutting-instructions-1.svg
```
</details>

### DIFF
--- a/scripts/js/commands/checkStaleImages.ts
+++ b/scripts/js/commands/checkStaleImages.ts
@@ -12,42 +12,60 @@
 
 import { $ } from "zx";
 import { globby } from "globby";
-import { zip } from "lodash-es";
 
 import { zxMain } from "../lib/zx.js";
 
+// Files that are used in closed source, so should not be removed.
+// Format: full file path, e.g. "public/images/guides/paulibasis.png"
+const ALLOW_LIST = new Set<string>([]);
+
 zxMain(async () => {
   const paths = await getStrippedImagePaths();
-  const results = await Promise.all(paths.map((fp) => isImageUnused(fp)));
-  const unused = zip(paths, results)
-    .filter(([_fp, result]) => result)
-    .map(([fp]) => `public/${fp}`);
-  if (unused.length === 0) {
+  let allGood = true;
+  let numFilesChecked = 0;
+  for (const fp of paths) {
+    numFilesChecked++;
+    const isUnused = await isImageUnused(fp);
+    if (isUnused) {
+      console.error(`❌ image is unused: public/${fp}`);
+      allGood = false;
+    }
+
+    // This script can be slow, so log progress every 20 files.
+    if (numFilesChecked % 20 == 0) {
+      console.log(`⏳ Checked ${numFilesChecked} / ${paths.size} images`);
+    }
+  }
+
+  if (allGood) {
     console.log("✅ all images used");
   } else {
     console.error(
-      `🙅 ${
-        unused.length
-      } stale images found. These images should usually be deleted:\n\n${unused.join(
-        "\n",
-      )}`,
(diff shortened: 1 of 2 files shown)


scripts/js/commands/checkStaleImages.ts
```diff
@@ -12,42 +12,60 @@
...
     console.error(
-      `🙅 ${
-        unused.length
-      } stale images found. These images should usually be deleted:\n\n${unused.join(
-        "\n",
-      )}`,
+      "\n\n❌ Some images are unused. They should usually be deleted to reduce our repository size.",
+    );
+    console.warn(
+      "⚠️ Be careful that some of these images be used in close source. Before deleting files, check for their usage there. If they're unused, add it to the allowlist.",
```
frankharkins, line 47:
```suggestion
      "⚠️ Be careful that some of these images may be used in close source. Before deleting files, check for their usage there. If they're unused, add it to the allowlist.",
```

scripts/js/commands/checkStaleImages.ts
```diff
@@ -12,42 +12,60 @@
...
 zxMain(async () => {
   const paths = await getStrippedImagePaths();
-  const results = await Promise.all(paths.map((fp) => isImageUnused(fp)));
-  const unused = zip(paths, results)
-    .filter(([_fp, result]) => result)
-    .map(([fp]) => `public/${fp}`);
-  if (unused.length === 0) {
+  let allGood = true;
+  let numFilesChecked = 0;
+  for (const fp of paths) {
```
frankharkins, line 26: The `await Promise.all` approach should work fine here. There won't be a race condition with `numFilesChecked` since Node is single-threaded. I'd expect it to make a difference since there's a lot of time waiting for commands to run. Just move L27 -> L34 so there's no `await` between incrementing and logging.

frankharkins (approved):
Thanks!

scripts/js/commands/checkStaleImages.ts
Eric-Arellano: I don't think this would work well because we could trigger too many concurrent calls to ripgrep, which can peg the system.

PR #2529, opened 2024-12-23
Delete stale images

Eric-Arellano:
Found with `npx tsx scripts/js/commands/checkStaleImages.ts`. CI's link checker will validate that no images were removed that shouldn't have been.


arnaucasau (approved):
Nice, thanks!

scripts/js/lib/api/saveImages.ts
```diff
@@ -18,6 +18,23 @@ import { Pkg } from "./Pkg.js";
...
+  if (!isReleaseNote) return false;
+
+  if (pkg.hasSeparateReleaseNotes()) {
+    // If the pkg has dedicated release notes per release, we should copy its images
+    // unless it's the dev version. We don't (yet) support release notes for dev versions:
+    // https://github.com/Qiskit/documentation/issues/1296
+    return pkg.isDev();
+  } else {
+    // Else, if the pkg has only a single release note file, only 'latest' should copy images.
+    return pkg.isLatest();
```
arnaucasau, line 34: Shouldn't this return true when pkg is not latest?
```suggestion
    return !pkg.isLatest();
```

arnaucasau (commented):
Good call early returning when downloading the release notes' images for the dev version!

scripts/js/lib/api/saveImages.ts
Eric-Arellano: Oof, good catch!

arnaucasau (approved):

PR #2539, opened 2025-01-02
Document Git clone tip

Eric-Arellano:
See https://git-scm.com/docs/git-clone#Documentation/git-clone.txt-code--filtercodeemltfilter-specgtem. This option means we lazily download blobs. We use this technique in closed source's sync very well. It should not break Git history like git blame.


arnaucasau (approved):
Thank you!!

Issue #2540, opened 2025-01-02
Qiskit Box artifacts are missing images from 0.19-0.44


Eric-Arellano:
Re-upload the box artifacts with the images from qiskit/documentation. 

Note that objects.inv is missing in the Box folder, too, but we don't have access to it anywhere so those will always be missing.

Issue #2538, opened 2025-01-02
Why does API pipeline not delete all images before copying them?


Eric-Arellano:
See here:

https://github.com/Qiskit/documentation/blob/be6687be304ffc3262c6c21bb6117d21531821f6/scripts/js/lib/api/saveImages.ts#L44-L63

Why don't we always remove the image? That is the only way to delete stale images that are no longer in use.

We could adapt this code:

https://github.com/Qiskit/documentation/blob/be6687be304ffc3262c6c21bb6117d21531821f6/scripts/js/commands/api/updateApiDocs.ts#L130-L137

Eric-Arellano:
The reason why is https://github.com/Qiskit/documentation/issues/2540. We should fix this.

Issue #2530, opened 2024-12-23
Speed up stale image checker


Eric-Arellano:
https://github.com/Qiskit/documentation/pull/2528 improved the stale image checker to be "fast enough" to run occasionally, but the program is still quite slow due to starting up ripgrep 3450 times. It takes several minutes on my mac.

I suspect a faster approach would be to read the markdown to extract all used images, such as by using the link checker, then compare that to the images we do have on disk. It would avoid subprocess calls. However, it's possible this solution would overwhelm system memory.

arnaucasau:
What do you think about integrating the stale image checker with our alt text image checker in https://github.com/Qiskit/documentation/blob/main/scripts/js/commands/checkImages.ts? I think we could use the unified plugin to find the stale images:

https://github.com/Qiskit/documentation/blob/be6687be304ffc3262c6c21bb6117d21531821f6/scripts/js/lib/markdownImages.ts#L31-L39

Eric-Arellano:
That might be helpful to use the stale image checker. However, the stale image checker doesn't check historical API docs. Also, I realize we could probably implement https://github.com/Qiskit/documentation/issues/2538 such that we no longer even need to worry about checking API docs. If true, that would speed up this overall program enough automatically that we could avoid spending more dev resources on this ticket.

Eric-Arellano:
Indeed, https://github.com/Qiskit/documentation/pull/2541 fixes the performance problem 🚀

It's blocked by https://github.com/Qiskit/documentation/issues/2540, which I want to fix because that is a total gotcha for future contributors.

PR #2541, opened 2025-01-02
Ensure API docs don't have stale images

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/2538. We only didn't delete the API images due to https://github.com/Qiskit/documentation/issues/2540. 

To fix https://github.com/Qiskit/documentation/issues/2540, we have to fix our skip of release note images. We also had to manually move 4 release note images for legacy release note files from the latest Qiskit images folder to instead the 0.40 and 0.43 Box artifacts, then manually update those release notes to point to the paths. With those fixes and the re-uploaded Box artifacts, `npm run regen-apis` doesn't make changes.

This change allows us to have checkStaleImages.ts no longer worry about API docs images, which dramatically speeds up the program. It's now fast enough to run on every PR. Closes https://github.com/Qiskit/documentation/issues/2530.

With ripgrep:

```
  Time (mean ± σ):      1.705 s ±  0.022 s    [User: 1.143 s, System: 1.085 s]
  Range (min … max):    1.660 s …  1.735 s    10 runs
```

With git grep (CI):

```
  Time (mean ± σ):      4.118 s ±  0.054 s    [User: 11.290 s, System: 1.092 s]
  Range (min … max):    4.044 s …  4.200 s    10 runs
```


scripts/js/lib/api/saveImages.ts
```diff
@@ -44,10 +48,6 @@ export async function saveImages(
   const destFolder = pkg.outputDir(`${publicBaseFolder}/images`);
   if (!(await pathExists(destFolder))) {
     await mkdirp(destFolder);
-  } else if (pkg.isDev()) {
-    // We don't want to store images from other versions when we generate a
-    // different dev version
-    await rmFilesInFolder(destFolder);
```
Eric-Arellano, line 50: Not necessary because updateApiDocs.ts will have already removed all images as part of removing Markdown.

Eric-Arellano:
Hold off on merging. `npm run regen-apis` shows that these images are lost due to this PR:

```
docs/api/qiskit/release-notes/0.40.mdx
59:>     ![\_images/legacy\_release\_notes-2.png](/images/api/qiskit/legacy_release_notes-2.png)
349:    ![\_images/legacy\_release\_notes-3.png](/images/api/qiskit/legacy_release_notes-3.png)
599:    ![\_images/legacy\_release\_notes-4.png](/images/api/qiskit/legacy_release_notes-4.png)

docs/api/qiskit/release-notes/0.43.mdx
595:    ![\_images/legacy\_release\_notes-1.png](/images/api/qiskit/legacy_release_notes-1.png)
```

I need to figure out the cleanest way to fix this problem.

scripts/js/lib/zx.ts
```diff
@@ -29,3 +29,12 @@ export function enableCliColors() {
 export function disableCliColors() {
   process.env.FORCE_COLOR = "0";
 }
+
+export async function isInstalled(binary: string): Promise<boolean> {
+  try {
+    await $`which ${binary}`.quiet();
```
arnaucasau, line 35: Probably it's not a problem right now given that `checkStaleImages` is used in CI and we don't even have an entry for `check:stale-images` on the readme, but I think the `which` command won't work on Windows. We should be careful if we use this function elsewhere. I think the equivalent would be the `where` command.

scripts/js/commands/api/updateApiDocs.ts
```diff
@@ -127,13 +127,18 @@ async function prepareSphinxFolder(pkg: Pkg, args: Arguments): Promise<string> {
...
+
+  const imagesDir = pkg.outputDir("public/images");
+  if (!(await pathExists(imagesDir))) return;
+  await rmFilesInFolder(imagesDir);
+
   console.log(
-    `Deleting existing markdown for ${pkg.name}:${pkg.versionWithoutPatch}`,
+    `Deleted existing markdown & images for ${pkg.name}:${pkg.versionWithoutPatch}`,
   );
-  await rmFilesInFolder(markdownDir);
```
arnaucasau, line 136: Maybe we can split the print for files and images or refactor the code to not early return? If there are no images, we won't have any print for files either

arnaucasau (commented):
Nice work!! Looking at the lost images, they were manually added to the current version folder in https://github.com/Qiskit/documentation/pull/346. Given that only those two versions use the legacy release note images, maybe we could store the images in their folders and manually modify their links in `0.40.mdx` and `0.43.mdx` to point to the new location.

Eric-Arellano:
> maybe we could store the images in their folders and manually modify their links in 0.40.mdx and 0.43.mdx to point to the new location.

Thanks for finding that. I tried that approach but we still don't copy the images because this code only copies images we actually use:

https://github.com/Qiskit/documentation/blob/acdcf5b243178bd21edd19562695658ee88b7420/scripts/js/lib/api/conversionPipeline.ts#L138-L143

So, I think the best approach is to instead save the images in a non-auto-generated folder.

arnaucasau (approved):
Nice work! Tested locally and works well! :rocket:

PR #2556, opened 2025-01-09
Convert images to AVIF during API docs generation

frankharkins:
Includes a feature flag as requested. This requires the user have ImageMagick installed.

### DIFF
--- a/scripts/js/lib/api/processHtml.ts
+++ b/scripts/js/lib/api/processHtml.ts
@@ -11,6 +11,7 @@
 // that they have been altered from the originals.
 
 import path from "node:path";
+
 import { CheerioAPI, Cheerio, load, Element } from "cheerio";
 
 import { Image } from "./HtmlToMdResult.js";
@@ -97,6 +98,8 @@ export function loadImages(
       const fileName = $img.attr("src")!.split("/").pop()!;
       const fileExtension = path.extname(fileName);
 
+      // We convert PNG and JPG to AVIF for reduced file size. The image-copying
+      // logic detects changed extensions and converts the files.
       // Remove `false &&` when ready to switch to AVIF
       let dest =
         false && [".png", ".jpg", ".jpeg"].includes(fileExtension)
--- a/scripts/js/lib/api/processHtml.ts
+++ b/scripts/js/lib/api/processHtml.ts
@@ -10,6 +10,7 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
+import path from "node:path";
 import { CheerioAPI, Cheerio, load, Element } from "cheerio";
 
 import { Image } from "./HtmlToMdResult.js";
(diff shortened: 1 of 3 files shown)


scripts/js/lib/api/processHtml.ts
```diff
@@ -94,8 +95,14 @@ export function loadImages(
       const $img = $(img);
 
       const fileName = $img.attr("src")!.split("/").pop()!;
+      const fileExtension = path.extname(fileName);
+
+      // Remove `false &&` when ready to switch to AVIF
+      let dest =
+        false && [".png", ".jpg", ".jpeg"].includes(fileExtension)
+          ? `${imageDestination}/${path.basename(fileName, fileExtension)}.avif`
+          : `${imageDestination}/${fileName}`;
```
Eric-Arellano, line 104: Feel free to ignore. Motivations for this suggestion:
- make the feature gate more explicit
- add a comment about why we convert to AVIF. It'd be good to have this comment even after removing the feature gate

```suggestion
      // Remove `false &&` when ready to switch to AVIF
      const avif_feature_gate_enabled = false
      let dest;
      if (avif_feature_gate_enabled) {
         // We convert PNG and JPG to AVIF for reduced file size.
        dest = [".png", ".jpg", ".jpeg"].includes(fileExtension)
          ? `${imageDestination}/${path.basename(fileName, fileExtension)}.avif`
          : `${imageDestination}/${fileName}`;
      } else {
       dest = `${imageDestination}/${fileName}`;
      }
```

The formatting is probably wrong

scripts/js/lib/api/saveImages.ts
```diff
@@ -13,6 +13,8 @@
 import { mkdirp } from "mkdirp";
 import pMap from "p-map";
 import { copyFile } from "fs/promises";
+import { extname } from "node:path";
```
Eric-Arellano, line 16: Nit: could you please group the imports in this file by std lib and third-party?

scripts/js/lib/api/saveImages.ts
```diff
@@ -54,10 +56,20 @@ export async function saveImages(
...
-      `${originalImagesFolderPath}/${img.fileName}`,
-      `${publicBaseFolder}/${img.dest}`,
-    );
+    if (extname(source) === extname(dest)) {
+      await copyFile(source, dest);
+    } else {
+      try {
+        await $`magick ${source} ${dest}`;
+      } catch (err) {
+        console.log(
```
Eric-Arellano, line 68:
```suggestion
        console.error(
```

scripts/js/lib/api/saveImages.ts
```diff
@@ -54,10 +56,20 @@ export async function saveImages(
...
 
-    await copyFile(
-      `${originalImagesFolderPath}/${img.fileName}`,
-      `${publicBaseFolder}/${img.dest}`,
-    );
+    if (extname(source) === extname(dest)) {
+      await copyFile(source, dest);
+    } else {
+      try {
+        await $`magick ${source} ${dest}`;
```
Eric-Arellano, line 66: Does this have output? Should it be quiet?

```suggestion
        await $`magick ${source} ${dest}`.quiet();
```

Eric-Arellano (commented):
Thank you! Could you please document in the README prerequisites section that you need to install imagemagick and link to its install guide?

scripts/js/lib/api/processHtml.ts
frankharkins: I would usually agree with the named bools, but I think since the feature gate code will only exist for a few days it's better to avoid the extra nesting. Comment is a good idea though.

Eric-Arellano:
> Soft push back on this: Most users will have it installed already, and if they don't they'll immediately see the message explaining that they need it. It's also likely to become outdated if we ever change how the pipeline works.

1. We will need it when you contribute to guides if you add new images, not only API docs. Fwiw, I'm fine with saying "You'll need imagemick if adding new images"
2. Why would most people have it installed already? I had to install it and it's not default on at least macOS

Eric-Arellano (approved):
My README suggestion isn't blocking. Thanks for adding the install to the console log!

frankharkins:
> I had to install it and it's not default on at least macOS

My mistake; I didn't need to install it so I assumed it came pre-installed with OSX, it must have been a dependency of something else.

PR #2557, opened 2025-01-09
Add lint rule to detect non-AVIF images

frankharkins:
Includes instructions on how to convert images to AVIF.

### DIFF
--- a/scripts/js/lib/markdownImages.ts
+++ b/scripts/js/lib/markdownImages.ts
@@ -35,7 +35,8 @@ export async function collectInvalidImageErrors(
         if (!node.alt || node.alt.endsWith(imageName!)) {
           imagesErrors.add(`The image '${node.url}' does not have alt text.`);
         }
-        if (node.url.match(/\.(png|jpe?g)$/)) {
+        // Remove `false &&` when ready to switch to AVIF
+        if (false && node.url.match(/\.(png|jpe?g)$/)) {
           const urlWithAvifExtension = node.url.replace(
             /\.(png|jpe?g)$/,
             ".avif",
--- a/scripts/js/lib/markdownImages.ts
+++ b/scripts/js/lib/markdownImages.ts
@@ -35,6 +35,16 @@ export async function collectInvalidImageErrors(
         if (!node.alt || node.alt.endsWith(imageName!)) {
           imagesErrors.add(`The image '${node.url}' does not have alt text.`);
         }
+        if (node.url.match(/\.(png|jpe?g)$/)) {
+          const urlWithAvifExtension = node.url.replace(
+            /\.(png|jpe?g)$/,
+            ".avif",
+          );
+          imagesErrors.add(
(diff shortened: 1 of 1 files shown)


scripts/js/lib/markdownImages.ts
```diff
@@ -35,6 +35,17 @@ export async function collectInvalidImageErrors(
...
+        // Remove `false &&` when ready to switch to AVIF
+        if (false && node.url.match(/\.(png|jpe?g)$/)) {
+          const urlWithAvifExtension = node.url.replace(
+            /\.(png|jpe?g)$/,
+            ".avif",
+          );
+          imagesErrors.add(
+            `Convert '${imageName}' to AVIF. You can use the command \`magick public${node.url} public${urlWithAvifExtension}\`. ` +
+              `If ImageMagick isn't preinstalled, you can get it from https://imagemagick.org/script/download.php.`,
+          );
```
arnaucasau, line 47: What do you think about moving the `ImageMagick` instructions to the README and adding a small comment in the https://github.com/Qiskit/documentation#images section? Otherwise, it could add a lot of noise to the error message (it will be duplicated for each invalid image).

The good thing is that we already link to the README here:

https://github.com/Qiskit/documentation/blob/f861330adb571bc1538c27dc629fb6afd0a0c99b/scripts/js/commands/checkImages.ts#L60

Eric-Arellano (approved):
Great that this is so simple! Good job @arnaucasau with the foundation, and thanks @frankharkins for figuring this out

arnaucasau (commented):
Thanks for working on this Frank! :rocket:

scripts/js/lib/markdownImages.ts
frankharkins: Thanks! Personally I prefer it as an error message for a few reasons:
* users will see the advice only when they need it, avoiding README bloat
* in the script, we can use the filename to provide the exact command they need
* they don't need to jump between pages to work out what to do

I think it shouldn't be too noisy because writers are unlikely to add more than a handful of images in a PR.

scripts/js/lib/markdownImages.ts
```diff
@@ -35,6 +35,17 @@ export async function collectInvalidImageErrors(
...
         }
+        // Remove `false &&` when ready to switch to AVIF
+        if (false && node.url.match(/\.(png|jpe?g)$/)) {
+          const urlWithAvifExtension = node.url.replace(
+            /\.(png|jpe?g)$/,
+            ".avif",
+          );
+          imagesErrors.add(
+            `Convert '${imageName}' to AVIF. You can use the command \`magick public${node.url} public${urlWithAvifExtension}\`. ` +
+              `If ImageMagick isn't preinstalled, you can get it from https://imagemagick.org/script/download.php.`,
```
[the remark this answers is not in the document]
frankharkins: Good call, added in 028db3f5194eb80cc8dac78c421f83e5b6740044

arnaucasau (approved):

Eric-Arellano (approved):

PR #2586, opened 2025-01-21
Add external link redirect mechanism

### LINKED ISSUE #2579: Extend link checker to handle redirects
Today's scheduled extended checks failed.

Please [check the logs](https://github.com/Qiskit/documentation/actions/runs/12848564515/).

> [!NOTE]
> This issue was created by a GitHub action.

frankharkins:
Fixes #2579 by adding a mechanism to redirect links without needing to modify the artifact. This is less work for maintainers, and less risky as it means we don't need to edit the artifact as much.

A future PR could have the link checker push to `external-redirects.ts` when it detects a 301, but this should be fine for now.


Eric-Arellano (commented):
Thanks! This is a good idea.

scripts/js/lib/api/htmlToMd.ts
```diff
@@ -39,10 +40,15 @@ export async function sphinxHtmlToMarkdown(options: {
...
   const processedHtml = await processHtml(options);
-  const markdown = await generateMarkdownFile(
+  let markdown = await generateMarkdownFile(
     processedHtml.html,
     processedHtml.meta,
   );
+
+  for (const redirect of externalRedirects) {
+    markdown = markdown.replaceAll(redirect.old, redirect.new);
+  }
```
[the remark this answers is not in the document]
arnaucasau: I agree with Eric that maybe this is better to implement in `processHtml.ts` :+1:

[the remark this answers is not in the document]
frankharkins: Ok done, the catch with `<a>` is that it'll only replace the `href`, not the link text. Is that the intended behaviour?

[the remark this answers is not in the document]
Eric-Arellano: It'd be good to replace the text as well if it's a match. You can use `.text()` to do that

https://github.com/Qiskit/documentation/blob/12a981b8c40836db939c16765e13d09836c74392/scripts/js/lib/api/processHtml.ts#L409-L427

scripts/js/lib/api/htmlToMd.ts
```diff
@@ -39,10 +39,11 @@ export async function sphinxHtmlToMarkdown(options: {
   hasSeparateReleaseNotes: boolean;
 }): Promise<HtmlToMdResult> {
   const processedHtml = await processHtml(options);
-  const markdown = await generateMarkdownFile(
+  let markdown = await generateMarkdownFile(
```
Eric-Arellano, line 42:
```suggestion
  const markdown = await generateMarkdownFile(
```

scripts/js/lib/api/processHtml.test.ts
```diff
@@ -313,6 +314,25 @@ test("addLanguageClassToCodeBlocks()", () => {
...
 
+test("updateRedirectedExternalLinks()", () => {
+  const redirects = {
+    "https://ibm.com/old": "https://ibm.com/new",
+    "https://basename.example-website.com/old#anchor?and=query":
+      "https://newbasename.example.com/new#newanchor?query=new",
+  };
+  const doc = CheerioDoc.load(
+    `<a href="https://ibm.com/old">https://ibm.com/old</a>
+    <a href="https://basename.example-website.com/old#anchor?and=query"></a>
```
Eric-Arellano, line 325: Can you please add a URL that is not redirected? Ideally it looks very similar to an old URL, such as `https://ibm.com/old-but-not-redirected`. We want to only change exact matches.

scripts/js/lib/api/processHtml.ts
```diff
@@ -206,6 +208,23 @@ export function addLanguageClassToCodeBlocks(
...
+  redirects: { [old: string]: string },
+): void {
+  $main.find("a").each((_, a) => {
+    const $a = $(a);
+    const href = $a.attr("href");
+    const replacement = href && redirects[href];
+    if (replacement) {
+      $a.attr("href", replacement);
+      const newText = $a.text().replaceAll(href, replacement);
+      $a.text(newText)
```
Eric-Arellano, line 223: This should require an exact match, not partial

frankharkins: Really? What if the link text was `https://google.com (opens in new tab)` or similar?

scripts/js/lib/api/processHtml.test.ts
```diff
@@ -313,6 +314,29 @@ test("addLanguageClassToCodeBlocks()", () => {
...
+    `<a href="https://ibm.com/old">https://ibm.com/old</a>
+    <a href="https://ibm.com/old">https://ibm.com/old (v3)</a>
+    <a href="https://ibm.com/old-but-not-redirected"></a>
+    <a href="https://basename.example-website.com/old#anchor?and=query"></a>
+    <p>https://ibm.com/old</p>`,
+  );
+  updateRedirectedExternalLinks(doc.$, doc.$main, redirects);
+  doc.expectHtml(
+    `<a href="https://ibm.com/new">https://ibm.com/new</a>
+    <a href="https://ibm.com/new">https://ibm.com/new (v3)</a>
```
Eric-Arellano, line 333: Nice! I like this.

scripts/js/lib/api/processHtml.test.ts
```diff
@@ -313,6 +314,29 @@ test("addLanguageClassToCodeBlocks()", () => {
...
+    <a href="https://ibm.com/old">https://ibm.com/old (v3)</a>
+    <a href="https://ibm.com/old-but-not-redirected">https://ibm.com/old-but-not-redirected (v3)</a>
+    <a href="https://basename.example-website.com/old#anchor?and=query"></a>
+    <p>https://ibm.com/old</p>`,
+  );
+  updateRedirectedExternalLinks(doc.$, doc.$main, redirects);
+  doc.expectHtml(
+    `<a href="https://ibm.com/new">https://ibm.com/new</a>
+    <a href="https://ibm.com/new">https://ibm.com/new (v3)</a>
+    <a href="https://ibm.com/old-but-not-redirected">https://ibm.com/old-but-not-redirected (v3)</a>
```
Eric-Arellano, line 334: Given how your regex has `(?=$|[\\s<'"])`, it would maybe help to also test 

a) text is the exact non-replaced URL
b) text is the exact non-replaced URL with a query parameter

Eric-Arellano (approved):
Thanks for iterating on this!

scripts/js/lib/api/processHtml.test.ts
frankharkins: Is this what you mean? da6b62efa64d4951027cd35a0241da54f6d15774

PR #2588, opened 2025-01-22
Shorten and use kebab-case for qiskit-ibm-transpiler URLs

arnaucasau:
This PR is a continuation of https://github.com/Qiskit/documentation/pull/2177 for all the versions of qiskit-ibm-transpiler.

Once this PR is reviewed and accepted, I'll also set the redirects in closed source.


frankharkins (approved):

Eric-Arellano (approved):
@arnaucasau thoughts on removing the prefix `qiskit-transpiler-service` for the historical versions? Benefits:

a) Shorter URL for those versions
b) No need for the historical redirects
c) Maybe clearer code? Including simpler redirects in closed source?

Maybe add to `Pkg.ts` an argument for `truncate-url-prefix: string[]`. No worries if you think that is overly complex. These are old versions we don't care very much about and it's likely what you have is simpler

--

We should do an open source sync in closed source before merging this PR because this PR needs to be synced with care.

arnaucasau:
I think I'm in favor of keeping `qiskit-transpiler-service` because:

- It only affects to 2 out of 7 historical versions
- The redirects will be temporary for those versions
- Simpler code in this repo

> We should do an open source sync in closed source before merging this PR because this PR needs to be synced with care.

Good idea, I'll do it! Afterwards, I was thinking of adding the second OS sync to the redirects' PR to make it atomic too

PR #2589, opened 2025-01-22
Set seeds to make notebook outputs deterministic

frankharkins:
By setting seeds in a few places, we can make these two notebooks almost completely deterministic. This reduces diff noise, git bloat, and reduces chance of a merge conflict.

I spotted this opportunity in the recent fortnightly notebook run PR.


docs/guides/visualize-results.ipynb
```diff
@@ -197,8 +199,8 @@
...
...data': [{'num_bind_params': 1, 'runtime_parameter_bind': False, 'parallel_state_update': 16, 'parallel_shots': 1, 'sample_measure_time': 0.000200758, 'noise': 'ideal', 'batched_shots_optimization': False, 'remapped_qubits': False, 'active_input_qubits': [0, 1], 'device': 'CPU', 'time_taken': 0.000581137, 'measure_sampling': True, 'num_clbits': 2, 'max_memory_mb': 32768, 'input_qubit_map': [[1, 1], [0, 0]], 'num_qubits': 2, 'method': 'statevector', 'required_memory_mb': 1, 'fusion': {'enabled': True, 'threshold': 14, 'applied': False, 'max_fused_qubits': 5}}]}, {'shots': 1024, 'variance': 0.4433441162109375, 'simulator_metadata': [{'num_bind_params': 1, 'runtime_parameter_bind': False, 'parallel_state_update': 16, 'parallel_shots': 1, 'sample_measure_time': 0.000214086, 'noise': 'ideal', 'batched_shots_optimization': False, 'remapped_qubits': False, 'active_input_qubits': [0, 1], 'device': 'CPU', 'time_taken': 0.000600807, 'measure_sampling': True, 'num_clbits': 2, 'max_memory_mb': 32768, 'input_qubit_map': [[1, 1], [0, 0]], 'num_qubits': 2, 'method': 'statevector', 'required_memory_mb': 1, 'fusion': {'enabled': True, 'threshold': 14, 'applied': False, 'max_fused_qubits': 5}}]}])\n"
```
frankharkins, line 203: Unfortunately this one output is still not deterministic, even with a seed. This is the only case across the two notebooks, and it's just small changes in the floats.

docs/guides/transpiler-stages.ipynb
```diff
@@ -128,7 +128,7 @@
     {
```
frankharkins, line 128: <p>Ugh, just noticed this chart doesn't really make sense with the following sentence:</p><blockquote>This wide distribution demonstrates how difficult it is for the SWAP mapper to compute the best mapping.</blockquote><p>I'll try with a different seed</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Qiskit/documentation/pull/2589/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

frankharkins: The problem was that the seed was the same for each transpilation. Fixed by moving the pass manager creation inside the loop.

arnaucasau (approved):
Awesome, thanks for doing this!

PR #2593, opened 2025-01-23
Shorten and use kebab-case for qiskit-ibm-runtime URLs

arnaucasau:
This PR is a continuation of https://github.com/Qiskit/documentation/pull/2177 for all the versions of qiskit-ibm-runtime.

Once this PR is reviewed and accepted, I'll also set the redirects in closed source.


Eric-Arellano:
Please don't merge until a) ibm-transpiler is 100% landed in prod, and b) you have the redirect code set up in closed source and we've reviewed it

arnaucasau:
Thank you both for the review!

> Please don't merge until...

Yes! That was my plan :+1:

----

In the last commit, I changed the `normalizeUrl` function to update the links that point to the same API (mainly for the release-notes). In , I updated them with another script I wrote, so this change is needed to avoid reverting the changes when regenerating the docs.

I think the change is safe, and I tested it by regenerating qiskit (latest, and one historical version), qiskit-addon-obp (currently using the new URLs), and runtime (latest, and one historical version). None of them showed any change in the docs.

Once all versions use the new URLs, we can change it to update links from `/api` in general. That way we will fix cross-reference between APIs.

Eric-Arellano:
I find it really annoying that V2 becomes v-2. Like SamplerV2 is sampler-v-2. Can you please confirm with `@javabster` this is okay?

scripts/js/lib/api/updateLinks.ts
```diff
@@ -62,7 +62,10 @@ export function normalizeUrl(
   kwargs: { kebabCaseAndShorten: boolean; pkgName: string },
 ): string {
   if (isAbsoluteUrl(url)) return url;
-  if (url.startsWith("/")) return url;
+  // We only want to normalize the API URLs to transform them to kebab-case
+  // Todo: update URLs from other APIs, when they all use kebab-case.
+  if (url.startsWith("/") && !url.startsWith(`/api/${kwargs.pkgName}`))
+    return url;
```
Eric-Arellano, line 68: I don't think I understand this? Your PR comment makes it sound necessary and good, but I don't think I understand why we need this? Notably, a lot of the URLs we do normalize are links to the API docs!

Is this maybe more accurate?

> If the URL is already pointing to `/api`, we assume that it was already normalized. This can happen, for example, when .... We need to be careful to ... because otherwise ...

Please make sure @frankharkins understands very well why we need this code. A good test is for him to explain to you why we can't delete that code, based only on him reading this section of the code (not your PR review).

arnaucasau:
I updated the script to special case URLs with a version (e.g. v1 or v2) in [1cbca72](https://github.com/Qiskit/documentation/commit/1cbca721e2364433a6b11d34c1a0acaf6a12e9dd). To help with the review of that change, I regenerated every version in a different commit again. That way, we can check that the change didn't affect any other link.

scripts/js/lib/api/normalizeResultUrls.ts
```diff
@@ -16,7 +16,7 @@ import { HtmlToMdResultWithUrl } from "./HtmlToMdResult.js";
 import { removePart } from "../stringUtils.js";
 
 export function kebabCaseAndShortenPage(page: string, pkgName: string): string {
-  const kebab = kebabCase(page);
+  const kebab = kebabCase(page).replace(/-v-([0-9]+)/g, `-v$1`);
```
Eric-Arellano: Nit to maybe add a comment:

// We also replace values like `-v-2` with `-v2`.

PR #2604, opened 2025-01-27
Remove meas level 1 limit on gen3 engine

jyu00:
Gen3 execution path now supports measurement level 1. Removing the restriction from `gen3-turbo` list.


abbycross (approved):
Thanks!

PR #2608, opened 2025-01-30
Update previews to set the correct docs' path

arnaucasau:
This PR updates the preview to use the new `docs/` structure where we need to specify that the content corresponds to the English locale.

### DIFF
--- a/scripts/pr-previews/builder.py
+++ b/scripts/pr-previews/builder.py
@@ -151,7 +151,7 @@ def _extract_docker_files(root_dir: Path) -> None:
     logger.info("Docker contents extracted")
 
 def _add_locale_to_docs(dir: Path) -> Path:
-    return dir.replace("docs/", "docs/en/")
+    return Path(str(dir).replace("docs/", "docs/en/"))
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


Eric-Arellano (commented):
You'll also need to update https://github.com/Qiskit/documentation/blob/264197059730c93ed4ffffb13037ac1375462395/scripts/pr-previews/builder.py#L117

scripts/pr-previews/builder.py
```diff
@@ -130,14 +130,14 @@ def _copy_local_content(root_dir: Path) -> None:
             if dir.startswith("public")
             else root_dir / dir
         )
-        shutil.copytree(dir, dest)
+        shutil.copytree(dir, _add_locale_to_docs(dest))
```
Eric-Arellano, line 133: You should instead call this on line 131 so that the expected type for add_locale is `str`. Note that in 140 you are passing it a string, which makes sense.

This way you also avoid the awkward `Path -> str -> Path` conversion

Eric-Arellano (approved):
Assuming CI passes, looks good. Thanks!

PR #2629, opened 2025-02-07
method get_job_by_id() missing in Qiskit Function Intro

### LINKED ISSUE #2627: method get_job_by_id() missing in Qiskit Function Intro
### URL, if applicable

https://docs.quantum.ibm.com/guides/functions#list-previously-run-jobs-run-with-qiskit-functions

A Qiskit Function user notified us via the startup providing the related Function that https://docs.quantum.ibm.com/guides/functions#list-previously-run-jobs-run-with-qiskit-functions is missing one or two lines on the functionality to retrieve a job by id, namely `catalog.get_job_by_id()`

abbycross:
Closes #2627 which was opened by @johannesgreiner. Johannes, can you provide more info on the fix please?

### DIFF
--- a/docs/guides/functions.ipynb
+++ b/docs/guides/functions.ipynb
@@ -363,6 +363,8 @@
    "source": [
     "### List previously run jobs run with Qiskit Functions\n",
     "\n",
+    "PLACEHOLDER: need to add catalog.get_job_by_id()\n",
+    "\n",
     "You can use `jobs()` to list all jobs submitted to Qiskit Functions:"
    ]
   },


johannesgreiner:
So basically we'd want a few lines showing how to take an existing job_id and pass it to `catalog.get_job_by_id()` to load a job

johannesgreiner:
e.g. we can use the existing `old_jobs = catalog.jobs()` and take `jobid=old_jobs[0].job_id` to get `job_by_id = catalog.get_job_by_id(jobid)`

docs/guides/functions.ipynb
```diff
@@ -397,6 +397,46 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d43b0824-bcf1-4f44-8eb5-1f9e5f4ee5a9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print(job_by_id)"
```
johannesgreiner, line 427: I like this suggestion by @Eric-Arellano 
```suggestion
   assert job_by_id.job_id == latest_job.job_id
```

docs/guides/functions.ipynb
```diff
@@ -397,6 +397,46 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d43b0824-bcf1-4f44-8eb5-1f9e5f4ee5a9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+   "assert job_by_id.job_id == latest_job.job_id"
```
abbycross, line 427:
```suggestion
   "assert job_by_id.job_id == old_jobs.job_id"
```

docs/guides/functions.ipynb
```diff
@@ -397,6 +397,46 @@
...
+   "assert job_by_id.job_id == old_jobs.job_id"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "eb869150-0942-4640-a8c7-7b73c19173ba",
+   "metadata": {},
+   "source": [
+    "```\n",
+    "<Job | b1486233-3f59-4c19-b52d-6ad729b891cc>\n",
```
abbycross, line 436: @Eric-Arellano when there is no print statement, would the assert statement still give an output of any kind? (What is here is left over from when there was a print statement, so I need to update it)

Eric-Arellano: No, the `assert` does not give output. Rather than hardcoding the output as markdown like this, it's better to update the `code` block to directly call `print()` and rely on `nb-tester` (i.e. run `tox`) to generate the output.

docs/guides/functions.ipynb
```diff
@@ -440,7 +414,49 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 20,
+   "id": "18d87f07-8a8e-4b93-80d0-7c07e49272e2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Get the job_id for the job at index position 1\n",
+    "latest_job = old_jobs[1]\n",
```
Eric-Arellano, line 431: I'm not sure why we're using index 1? The latest job would be index 0

```suggestion
    "# Get the last job that's executed1\n",
    "latest_job = old_jobs[0]\n",
```

docs/guides/functions.ipynb
abbycross: I've been running the notebook locally to get outputs. Should I be leaving them blank and leave it to tox to supply the outputs?

docs/guides/functions.ipynb
```diff
@@ -440,7 +414,49 @@
...
+    "job_by_id = catalog.get_job_by_id(latest_job.job_id)\n",
+    "\n",
+    "# verify that the job ID is the same using both retrieval methods\n",
```
Eric-Arellano, line 434: Comments should be capitalized and sentences

```suggestion
    "# Verify that the job ID is the same using both retrieval methods.\n",
```

docs/guides/functions.ipynb
```diff
@@ -440,7 +414,49 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "<Job | 8a21f700-b42f-4676-846d-4d4ce37da987>\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Print the job_id for the job at index position 1\n",
+    "print(latest_job)"
```
Eric-Arellano, line 454: This can be part of the same code block that you had above. There's no need to split them out and it may be more confusing to split them. The code is tightly "coupled"

docs/guides/functions.ipynb
Eric-Arellano: Running it locally is fine too

docs/guides/functions.ipynb
abbycross: oh right - I can change that. I had wanted to do index 1 in my previous iterations, but now that we're using latest_job as the variable, I'll change it

docs/guides/functions.ipynb
```diff
@@ -272,21 +265,10 @@
...
-      "text/plain": [
-       "'d76b6abf-3a16-4dd6-a9ac-a7ffa780d2f7'"
-      ]
-     },
-     "execution_count": 5,
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
+   "outputs": [],
```
Eric-Arellano, line 271: This PR makes a bad revert

```suggestion
   "outputs": [
    {
     "data": {
      "text/plain": [
       "'d76b6abf-3a16-4dd6-a9ac-a7ffa780d2f7'"
      ]
     },
     "execution_count": 5,
     "metadata": {},
     "output_type": "execute_result"
    }
   ],
```

docs/guides/functions.ipynb
```diff
@@ -440,7 +414,33 @@
   },
   {
    "cell_type": "markdown",
-   "id": "9137071a-e39c-4b66-afbd-697db38cc670",
+   "id": "af7fa72a-f272-43ff-800d-5ae37dd42b2b",
+   "metadata": {},
+   "source": [
+    "To retrieve a job's job_id by its position in the list, use `catalog.get_job_by_id()`:"
```
Eric-Arellano, line 420: @johannesgreiner this seems wrong to me to say "by its position in the list"? I think this should instead read as something like:

> If you already have the job ID for a certain job, you can retrieve it with `catalog.get_job_by_id()`:

docs/guides/functions.ipynb
```diff
@@ -440,7 +414,33 @@
...
+   "source": [
+    "# Get the most recent job that has been executed.\n",
+    "latest_job = old_jobs[0]\n",
+    "job_by_id = catalog.get_job_by_id(latest_job.job_id)\n",
+    "\n",
+    "# Verify that the job ID is the same using both retrieval methods.\n",
+    "assert job_by_id.job_id == latest_job.job_id\n",
+    "\n",
+    "# Print the job_id for the latest job.\n",
+    "print(latest_job)"
```
Eric-Arellano, line 438: job_by_id and latest_job should be the same thing. It might be clearer if we use job_by_id since that is what we are demoing

```suggestion
    "# Print the job_id for this job.\n",
    "print(job_by_id.job_id)"
```

docs/guides/functions.ipynb
```diff
@@ -440,7 +414,33 @@
...
+   "id": "18d87f07-8a8e-4b93-80d0-7c07e49272e2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Get the most recent job that has been executed.\n",
+    "latest_job = old_jobs[0]\n",
+    "job_by_id = catalog.get_job_by_id(latest_job.job_id)\n",
+    "\n",
+    "# Verify that the job ID is the same using both retrieval methods.\n",
+    "assert job_by_id.job_id == latest_job.job_id\n",
```
Eric-Arellano, line 435: I think this will actually be more clear. It demonstrates that the entire Job is the same, not just the ID.

```suggestion
    "# Verify that the job is the same using both retrieval methods.\n",
    "assert job_by_id == latest_job\n",
```

docs/guides/functions.ipynb
```diff
@@ -440,7 +414,33 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "18d87f07-8a8e-4b93-80d0-7c07e49272e2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Get the most recent job that has been executed.\n",
+    "latest_job = old_jobs[0]\n",
+    "job_by_id = catalog.get_job_by_id(latest_job.job_id)\n",
```
Eric-Arellano, line 432:
```suggestion
    "latest_job = old_jobs[0]\n",
    "\n",
    "# We can also get that same job with get_job_by_id\n",
    "job_by_id = catalog.get_job_by_id(latest_job.job_id)\n",
```

docs/guides/functions.ipynb
```diff
@@ -440,7 +414,33 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "18d87f07-8a8e-4b93-80d0-7c07e49272e2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Get the most recent job that has been executed.\n",
```
Eric-Arellano, line 430:
```suggestion
    "# First, get the most recent job that has been executed.\n",
```

docs/guides/functions.ipynb
abbycross: Isn't the job id the thing you're retrieving, and rather you're trying to find the nth job id in a list?

Eric-Arellano: No, `get_job_by_id` is when you already have a particular job ID (some string) and you are getting a job (I think https://docs.quantum.ibm.com/api/qiskit-ibm-runtime/runtime-job-v2?)

> and rather you're trying to find the nth job id in a list?

That's what you're doing with `old_jobs[0]`: you're getting the nth job where `old_jobs` is all the jobs that have run, sorted by date.

scripts/config/cspell/dictionaries/qiskit.txt
```diff
@@ -132,6 +132,7 @@ ipywidgets
 iqft
 iswap
 ISYM
+jobid
```
frankharkins, line 135: Not sure this should be in the dictionary, I think we should either say " job ID" in a sentence, or use code stying (`jobid`) if referring to the method / variable.

abbycross, line 135:
```suggestion
```

docs/guides/functions.ipynb
```diff
@@ -430,7 +431,35 @@
   },
   {
    "cell_type": "markdown",
-   "id": "9137071a-e39c-4b66-afbd-697db38cc670",
+   "id": "af7fa72a-f272-43ff-800d-5ae37dd42b2b",
+   "metadata": {},
+   "source": [
+    "To retrieve a job's job_id by its position in the list, use `catalog.get_job_by_id()`:"
```
abbycross, line 437:
```suggestion
    "If you already have the job ID for a certain job, you can retrieve the job with `catalog.get_job_by_id()`:"
```

Eric-Arellano (approved):
This looks great to me now!

PR #2636, opened 2025-02-14
Fix qubit number convention in bit-ordering guide

garrison:
The page at https://docs.quantum.ibm.com/guides/bit-ordering says

> If you have a set of $n$ bits (or qubits), you'll usually label each bit $0 \rightarrow n-1$.

However, thereafter, it refers to the most significant qubit as qubit $n$.  Since $n$ is presumably the number of qubits, this PR updates the guide to be consistent by making $n-1$ the largest qubit index.

### DIFF
--- a/docs/guides/bit-ordering.mdx
+++ b/docs/guides/bit-ordering.mdx
@@ -31,7 +31,7 @@ Qubit(QuantumRegister(2, 'q'), 0)
 
 ### Circuit diagrams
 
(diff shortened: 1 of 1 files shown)


jakelishman (approved):
This seems both correct and an improvement in clarity to me - thanks Jim. (I don't remember this file, though!)

frankharkins (approved):
Thanks!

PR #2643, opened 2025-02-17
Add `networkx` requirement

frankharkins:
Parital fix for #2638.

We've resisted adding this requirement for a while, but `docs/guides/q-ctrl-optimization-solver.ipynb` uses some features of networkx that rustworkx doesn't have (e.g. `random_regular_graph`). It's possible there are suitable replacements, but I don't know enough about the domain or graph theory to work that out. I think this is the best solution for now.

### DIFF
--- a/scripts/nb-tester/requirements.txt
+++ b/scripts/nb-tester/requirements.txt
@@ -13,3 +13,4 @@ qiskit-addon-aqc-tensor~=0.1.2
 qiskit-addon-obp~=0.1.0
 scipy~=1.15.1
 pyscf~=2.8.0; sys.platform != 'win32'
+networkx~=3.4.2


Eric-Arellano:
This is good to go once https://github.com/Qiskit/documentation/pull/2661 is merged.

Eric-Arellano (approved):

mtreinish:
Fwiw, it would be good if there were open issues for the functionality that rustworkx is missing so we can prioritize adding it. Not just for the tutorials but to make it more useful library.

frankharkins:
I've opened https://github.com/Qiskit/rustworkx/issues/1391, the other feature was the `kamada_kawai_layout` which already has an open issue, although I don't think that's essential anyway.

On closer inspection of the notebook, it looks like the function specifically wants output of Networkx's [`readwrite.json_graph.adjacency_data`](https://networkx.org/documentation/stable/reference/readwrite/generated/networkx.readwrite.json_graph.adjacency_data.html). I think Rustworkx's [`node_link_json`](https://www.rustworkx.org/apiref/rustworkx.node_link_json.html) might do the same thing, but we could risk breaking things if the function is somehow tightly coupled to the Networkx output.

PR #2651, opened 2025-02-18
Fix dev docs sync

Eric-Arellano:


arnaucasau (approved):
Thanks!

.github/workflows/dev-docs-sync.yml
```diff
@@ -26,8 +26,11 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 18
-      - name: Install Node.js dependencies
-        run: npm ci
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y imagemagick
```
arnaucasau, line 32: It might not be that easy to install the `imagemagick` package. In my Debian I got some errors because of the stable version of the package available in the repos and because some missing encoding format like:

``
magick: no encode delegate for this image format `AVIF' @ warning/constitute.c/WriteImage/1403.
```

I still have some problems when generating Qiskit (not in macos), but I think it's good to test this simple install command in CI first

Comment (frankharkins) [reply .github/workflows/dev-docs-sync.yml:32]: I think the job is failing because it's using magick 6 rather than 7, which changed the name of the command. We could instead try

```sh
curl https://imagemagick.org/archive/binaries/magick --output /usr/bin/magick
```

arnaucasau: > I think the job is failing because it's using magick 6 rather than 7

Yes, that's the main reason. In ImageMagick 7, the `convert` command was replaced by `magick`.

For me, the only thing that worked locally was to create a symlink of `/usr/bin/convert` to `/usr/bin/magick` after the `apt install`:

```bash
ln -s /usr/bin/convert /usr/bin/magick
```

Otherwise, I always missed the AVIF delegate

frankharkins: That worked, thanks!

frankharkins:
This seems to work now thanks to @arnaucasau (see https://github.com/Qiskit/documentation/actions/runs/13457907119/job/37605876981).

Eric-Arellano:
I can't approve because it's my PR, but thank you both for figuring this out!

Issue #2387, opened 2024-11-25
Kernel sometimes dies with notebook executor


Eric-Arellano:
For example https://github.com/Qiskit/documentation/actions/runs/12015838807/job/33494742907

```
task: <Task finished name='Task-35' coro=<execute_notebook() done, defined at /home/runner/work/documentation/documentation/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py:253> exception=DeadKernelError('Kernel died')>
Traceback (most recent call last):
  File "/home/runner/work/documentation/documentation/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py", line 268, in execute_notebook
    nb = await _execute_notebook(path, config, working_directory.name)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/documentation/documentation/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py", line 346, in _execute_notebook
    await notebook_client.async_execute()
  File "/home/runner/work/documentation/documentation/.tox/py311/lib/python3.11/site-packages/nbclient/client.py", line 709, in async_execute
    await self.async_execute_cell(
```

frankharkins:
We can't tell which notebook it is from the logs as the error doesn't contain any defining information and the notebooks all run asynchronously. We can whittle it down if jobs fail while only running a subset of notebooks.

frankharkins:
I managed to reproduce a similar problem locally when I added more notebooks to the script.

```
zmq.error.ZMQError: Too many open files
```

I fixed this locally by increasing my ulimit to 6000 (`ulimit -n 6000`). Hopefully we can set this in our action too.

Eric-Arellano:
@frankharkins let's keep this open until it's been a few weeks of not seeing it to confirm #2464 did fix the issue.

Eric-Arellano:
We haven't noticed this in a while. I think https://github.com/Qiskit/documentation/pull/3143 helped a lot. Thanks @frankharkins!

PR #2661, opened 2025-02-21
Filter packages listed in notebooks

frankharkins:
At the top of each notebook, we list the packages needed to run every notebook in this repo. This PR analyses the notebook filter out any packages not required by that notebook.

I've also re-run three notebooks to demonstrate, the rest will be updated in the fortnightly cron job.


scripts/nb-tester/qiskit_docs_notebook_tester/post_process.py
```diff
@@ -37,15 +43,43 @@ def post_process_notebook(nb: nbformat.NotebookNode) -> nbformat.NotebookNode:
...
+      if module not in sys.stdlib_module_names
+    ])
     package_versions = "\n".join(
-        line for line in requirements_file.split("\n") if not line.startswith("#")
+        line for line in requirements_txt.split("\n")
+        if re.split('[\\[~=]', line)[0].strip() in packages
     )
     return package_versions.strip()
+
+def get_used_modules(source: str) -> Iterable[str]:
```
Eric-Arellano, line 70: You could instead take the arg as a `list[str]`. Note that you joined it with `\n` up above, then split, then rejoin again.

scripts/nb-tester/qiskit_docs_notebook_tester/post_process.py
```diff
@@ -37,15 +43,43 @@ def post_process_notebook(nb: nbformat.NotebookNode) -> nbformat.NotebookNode:
...
+    # Remove Jupyter magics
+    source = "\n".join([
```
Eric-Arellano, line 72: It's best practice to not shadow the argument with the variable name.

scripts/nb-tester/qiskit_docs_notebook_tester/post_process.py
```diff
@@ -37,15 +43,43 @@ def post_process_notebook(nb: nbformat.NotebookNode) -> nbformat.NotebookNode:
...
+        line for line in source.split("\n")
+        if not line.strip().startswith("%")
+    ])
+    for node in ast.iter_child_nodes(ast.parse(source=source, filename="_.py")):
+        if isinstance(node, ast.Import):
+            for module in node.names:
+                yield module.name.split(".")[0]
```
Eric-Arellano, line 79: FYI there are certain packages that expose a module like `overall_namespace.my_pkg_namespace`, and another dep exposes `overall_namespace.another_pkg_namespace`. We don't have any deps that do that though, so I am fine with this simpler approach.

I doubt we will encounter this problem in the future; it's not a common pattern. So I think this is fine to not have a comment.

scripts/nb-tester/qiskit_docs_notebook_tester/post_process.py
```diff
@@ -37,15 +43,43 @@ def post_process_notebook(nb: nbformat.NotebookNode) -> nbformat.NotebookNode:
...
     )
     return package_versions.strip()
+
+def get_used_modules(source: str) -> Iterable[str]:
+    # Remove Jupyter magics
+    source = "\n".join([
+        line for line in source.split("\n")
+        if not line.strip().startswith("%")
+    ])
+    for node in ast.iter_child_nodes(ast.parse(source=source, filename="_.py")):
```
Eric-Arellano, line 76: The AST parsing can fail if there are syntax errors:

```
>>> ast.parse("imprtttttt sys", filename="_.py")
  File "<stdin>", line 1, in <module>
  File "/Users/arellano/.pyenv/versions/3.9.17/lib/python3.9/ast.py", line 50, in parse
    return compile(source, filename, mode, flags,
  File "_.py", line 1
    imprtttttt sys
               ^
SyntaxError: invalid syntax
```

Generally, I'd say that's fine, but I think it would break nb-tester for every other notebook that is valid because it would crash the overall nb-tester program. So, we should probably catch it. However, we do want to fail the overall run, only not crash it. So maybe return a `Result` that preserves the exception?

I do think this is probably worth fixing because the error would be cryptic otherwise.

It'd be good to test this edge case.

--

Another small tip is to make the call `ast.parse(source=source, filename="_.py")` a standalone line and save as a variable like `ast_tree`. That will make the stacktrace more clear that that specific function call was the problem.

scripts/nb-tester/qiskit_docs_notebook_tester/post_process.py
```diff
@@ -37,15 +43,43 @@ def post_process_notebook(nb: nbformat.NotebookNode) -> nbformat.NotebookNode:
...
+        if isinstance(node, ast.Import):
+            for module in node.names:
+                yield module.name.split(".")[0]
+        elif isinstance(node, ast.ImportFrom):
+            if node.module is not None:
+                yield node.module.split(".")[0]
+
+def flatten_to_list(i: Iterable[Iterable]) -> list:
```
Eric-Arellano, line 84: Imo it's fine to call this `flatten` because the type hint already makes clear it's a list and it's a sensible default. I wish Python std lib implemented it.

scripts/nb-tester/qiskit_docs_notebook_tester/post_process.py
```diff
@@ -10,7 +10,13 @@
 # notice, and modified files need to carry a notice indicating that they have
 # been altered from the originals.
 
+import ast
+import re
+import sys
+import importlib
+import itertools
 from pathlib import Path
+from typing_extensions import Iterable
```
Eric-Arellano, line 19: This should be `collections.abc`.

scripts/nb-tester/qiskit_docs_notebook_tester/post_process.py
Eric-Arellano: Also FYI no need to use the list comprehension with the outer `[]`. You can use a generator expression by removing the `[]`. That is lazy (uses an iterator), rather than storing the whole memory in list.

Eric-Arellano (commented):
Awesome job, Frank! Good job taking initiative and very readable code :) 

This is a fun PR for me because you re-created dependency inference, a key feature of Pantsbuild I worked on.

--

Technically, it would be more robust to have a distinct venv per notebook execution to be confident that we didn't miss any dependencies. With the current approach, it is theoretically possible for a dep to be necessary but not inferred.

However, it is not necessary to make that change. I'm confident in your dependency inference code and it's not worth the complexity to have multiple venvs (Pantsbuild would do it for us, but that's out of scope).

scripts/nb-tester/qiskit_docs_notebook_tester/post_process.py
frankharkins: I agree it's unlikely, however, if that does happen, we'll list both packages in the requirements as we include all packages from [`importlib.metadata.package_distributions()`](https://docs.python.org/3/library/importlib.metadata.html#package-distributions). I left a comment about it in `get_package_versions`.

frankharkins:
Thanks!

> Technically, it would be more robust to have a distinct venv per notebook execution...

This was my first thought, however, it's gonna be messy making that work (listing problems for posterity):

* AFAIK there's no Python API to create a venv, so we'd have to call out to subprocesses to create the venvs and link them to IPython kernels. We'd also need to delete each kernel after running.
* nbconvert isn't super robust and we've already hit weird flakes when running lots of notebooks (e.g. #2387). Running ~60 different kernels is likely to introduce more strange bugs.
* We'd have to list the individual requirements for each notebook and find a way to update them all at the same time (e.g. with dependabot).

scripts/nb-tester/qiskit_docs_notebook_tester/post_process.py
frankharkins: We can assume the code is valid as we only post-process notebooks after successful execution. However, I've added `SyntaxError` to the list of expected exceptions so we don't crash the whole script if something unexpected happens (see d7b618bb0262fb9563851baf48f8a1deead18b70).

scripts/nb-tester/qiskit_docs_notebook_tester/post_process.py
frankharkins: The first time we join (L47), we're joining cells, which can contain many lines. We could split those cells and flatten the list, but I think joining there is better because I'd expect "source code" to be a single string rather than a list.

PR #2679, opened 2025-02-26
Sync some of the new cloud docs

Eric-Arellano:
This brings over about half of the new changes for the docs at https://quantum.cloud.ibm.com/docs/en. The other half will be brought over in a follow up (split up to make review & green CI easier).

This brings over:
- all top level pages
- all migration guides
- about half of guides/

The PR adds a mechanism to have duplicate entries in a Patterns index page.

We also have to turn off testing of `get-qpu-information` for now because the migration broke the guide.


Eric-Arellano:
Review tips:

* I copied over MDX files wholesale
* For Jupyter notebooks, I only copied over relevant sections. Otherwise, we would have copied over incorrectly how images get extracted to AVIF in closed source
* No need to copy-edit, but it's good to check that every file's changes look sensible:
    - We didn't revert good changes from qiskit/documentation

arnaucasau (approved):
All the docs look good to me, but probably it's worth having the okay from Abby and Becky for them. Thanks!

PR #2683, opened 2025-02-27
nb-tester: Module Not Found fix

emilkovacev:
When invoking nb-tester from a directory outside of nb-tester (after pip installing nb-tester) fails with the error `ModuleNotFound: "patches"`. This PR replaces the importlib string "patches" with a relative import to the patches directory, and places patches within the source for nb-tester, ensuring it is always packaged when installing with pip.


frankharkins (approved):
Thank you!

Issue #2704, opened 2025-03-03
Add migration guide from BackendV1 to BackendV2?


ElePT:
### Describe the new content you are requesting.

With the removal of `BackendV1` in Qiskit 2.0, the section of the provider API docs indicating how to migrate from `BackendV1` to `BackendV2` will be lost. I suggest keeping it as a standalone migration guide during the transition period to 2.0 so that users can have an easier time adjusting to the change.

https://github.com/Qiskit/qiskit/blob/main/qiskit/providers/__init__.py#L670-L772

### If this new content request is accepted, do you want to write the content?

I will write a draft of the proposed content

PR #2705, opened 2025-03-03
Add Migration guide from BackendV1 to BackendV2

ElePT:
This PR addresses #2704. As described in the issue, this was part of the API reference of Qiskit. Given the extended support period of 1.4, which still supports BackendV1, I think it would make sense to keep this content alive a bit longer than the BackendV1 interface to help users transition to the 2.0 standard.

This migration guide can also be linked in the general 2.0 migration guide. 

## Notes

I have copy-pasted the original content and reformatted to mdx, but I also want to update the text a bit to make it more guide-y. This is why the RP is currently WIP. 

preview: https://qiskit.github.io/documentation/pr-2705/migration-guides/qiskit-backendv1-to-v2


beckykd:
I'm putting the new migration guide in the TOC to remove the lint error.  Do you think it makes sense to put it in a section for Qiskit 2.0?  @kaelynj is working on that guide.

Eric-Arellano:
> I'm putting the new migration guide in the TOC to remove the lint error. Do you think it makes sense to put it in a section for Qiskit 2.0? @kaelynj is working on that guide.

I think Elena's suggestion in the original post is great: have a dedicated guide because this is niche, but link to it from the general 2.0 guide. Becky, I like how you set up the ToC

beckykd:
FYI, @ElePT  links to APIs automatically go to the newest release (1.4), so actually putting 1.4 in the link breaks it unless 1.4 is not the current release :)

Eric-Arellano:
> links to APIs automatically go to the newest release (1.4), so actually putting 1.4 in the link breaks it unless 1.4 is not the current release :)

Generally, yes. But these links are for APIs that are going to be deleted so it's good that Elena is putting 1.4.

beckykd:
> > links to APIs automatically go to the newest release (1.4), so actually putting 1.4 in the link breaks it unless 1.4 is not the current release :)
> 
> Generally, yes. But these links are for APIs that are going to be deleted so it's good that Elena is putting 1.4.

I got lint errors on the links with 1.4 in them.  Should I have left them?

beckykd:
@ElePT, I added a second version of the table to try to make it skinnier, so the Notes column shows up.  That outlier "instructionDurations" messed it up though, so I pulled that into a different table.  Thoughts?  I also tried to reorganize it to make it easier to follow.

ElePT:
Thanks for the edit @beckykd! I really like the warning to scroll for notes, but I am not sure about the "skinnier" table. These are my thoughts:

1. I feel like seeing `backend.whatever()` is a bit easier to contextualize and copy-paste than just `whatever()`
2. In this table we are mixing a bit methods and attributes. For example:  `backend.configuration().coupling_map` is calling a method and then getting an attribute from the backend configuration class, the alternative for V2 is `backend.coupling_map` which is retrieving an attribute directly (no method). So having titles that say "BackendV1 method" isn't the most accurate.
3. I would be ok with the other 2 points if the change allowed the table to fit in the page without scrolling, but I think it doesn't, so maybe it's not so bad to have it be a bit wider and keep the original?

docs/migration-guides/qiskit-backendv1-to-v2.mdx
```diff
@@ -0,0 +1,72 @@
+---
+title: Migrate from BackendV1 to BackendV2
+description: How to update your code to use `BackendV2` instead of `BackendV1`.
+---
+
+# Migrate from `BackendV1` to `BackendV2`
+
+The Qiskit [`BackendV1`](/api/qiskit/1.4/qiskit.providers.BackendV1) class has been deprecated and will be removed from service. This migration guide describes the small adjustments you need to make if your are a user of a provider that upgraded from [`BackendV1`](/api/qiskit/1.4/qiskit.providers.BackendV1) to [`BackendV2`](/api/qiskit/qiskit.providers.BackendV2).
+
```
jyu00, line 9: I think it's worth mentioning that `qiskit_ibm_runtime` has always used `BackendV2` and `qiskit_aer` has also been using V2 since 0.13. So no actions are needed for users that use exclusively backends from those 2 packages.

docs/migration-guides/qiskit-backendv1-to-v2.mdx
```diff
@@ -0,0 +1,76 @@
...
+title: Migrate from BackendV1 to BackendV2
+description: How to update your code to use `BackendV2` instead of `BackendV1`.
+---
+
+# Migrate from `BackendV1` to `BackendV2`
+
+The Qiskit [`BackendV1`](/api/qiskit/1.4/qiskit.providers.BackendV1) class has been deprecated and will be removed from service. This migration guide describes the small adjustments you need to make if you use a provider that upgraded from [`BackendV1`](/api/qiskit/1.4/qiskit.providers.BackendV1) to [`BackendV2`](/api/qiskit/qiskit.providers.BackendV2).
+
+<Admonition type="note">
+ If you exclusively use `qiskit_ibm_runtime` and `qiskit_aer`, no action is needed.  The `qiskit_ibm_runtime` package has always used `BackendV2`, and `qiskit_aer` has used `BackendV2` 0.13.
```
jyu00, line 11:
```suggestion
 If you exclusively use `qiskit_ibm_runtime` and `qiskit_aer`, no action is needed.  The `qiskit_ibm_runtime` package has always used `BackendV2`, and `qiskit_aer` has used `BackendV2` since version 0.13.
```

jyu00 (approved):
One small comment but the rest looks good.

beckykd (approved):

beckykd:
Approving this, but it can't be merged until the infra team fixes the issue with the 1.4 links.

arnaucasau:
Thanks for the PR! I merged main to fix the link checker 👍

PR #2720, opened 2025-03-04
Update qiskit_bot.yaml for new guides

### LINKED ISSUE #1602: Revise qiskit_bot file post-reorg
Update the list of people to auto-tag in PRs to ensure we aren't creating noise for anyone working exclusively on the enablement side.

abbycross:
Closes #1602

### DIFF
--- a/qiskit_bot.yaml
+++ b/qiskit_bot.yaml
@@ -30,6 +30,7 @@ notifications:
     - "`@jakelishman`"
   "docs/guides/cloud-setup":
     - "@beckykd"
+    - "@abbycross"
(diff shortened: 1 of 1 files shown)


Eric-Arellano (approved):
Thanks! We don't need any SMEs like Fran for new files like `calibration-jobs`?

abbycross:
> Thanks! We don't need any SMEs like Fran for new files like `calibration-jobs`?

I will check with him. It came from the old admin directory, and the author of that file is unknown to me... I don't think it was him but he may know who.

Issue #2685, opened 2025-02-27
Decide if we want the pages render check for legacy docs


Eric-Arellano:
checkPagesRender.ts is being run with the cloud platform. This means that we don't check the page actually works for the legacy platform, and we could have syntax errors that cause the page to crash.

I worked around this in https://github.com/Qiskit/documentation/pull/2684 by skipping pages marked `platform: legacy`. But probably we should test it instead.

This looks like running the check with both platforms in CI. However, we shouldn't check API docs with both platforms because those are platform-agnostic so it's a waste of time.

PR #2733, opened 2025-03-06
Render legacy only pages

arnaucasau:
Closes https://github.com/Qiskit/documentation/issues/2685

This PR extends the page render checker to verify the legacy only pages. When the new `--legacy` argument is used, the checker will only run for non-API pages that are not cloud only.

#### Tests

- In https://github.com/Qiskit/documentation/pull/2733/commits/c43afa420c68eb148a94f5861056e3f4a563b904 I changed 1 common page and 1 legacy only page. This results in the checker verifying 1 page when running in cloud mode and 2 pages when running in legacy mode
- In https://github.com/Qiskit/documentation/pull/2733/commits/6c205b62f2ef80fef68b48535f9292e814c22958 2 extra cloud only pages. This results in the checker verifying 3 page when running in cloud mode and 2 pages when running in legacy mode
- In https://github.com/Qiskit/documentation/pull/2733/commits/e599a5e130fccff0570cc2739f9c1fedaa39d1dc I converted a legacy page to cloud only to make CI fail.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,15 +60,24 @@ jobs:
       - name: Pull preview image
         if: steps.all-changed-files.outputs.any_changed == 'true'
         run: ./start --pull-only
-      - name: Start local Docker preview
+      - name: Start local Docker preview (cloud app)
         if: steps.all-changed-files.outputs.any_changed == 'true'
         run: |
           ./start --apis &
           sleep 1
-      - name: Check that pages render
+      - name: Check that pages render (cloud app)
         if: steps.all-changed-files.outputs.any_changed == 'true'
         run: |
           npm run check:pages-render -- --from-file .github/outputs/all_changed_files.txt
+      - name: Start local Docker preview (legacy app)
+        if: steps.all-changed-files.outputs.any_changed == 'true'
+        run: |
+          ./start --legacy &
+          sleep 1
+      - name: Check that non-API pages render (legacy app)
+        if: steps.all-changed-files.outputs.any_changed == 'true'
+        run: |
+          npm run check:pages-render -- --legacy --from-file .github/outputs/all_changed_files.txt
 
       - name: Setup Python environment
         uses: ./.github/actions/set-up-notebook-testing
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,7 +68,11 @@ jobs:
       - name: Check that pages render (cloud app)
         if: steps.all-changed-files.outputs.any_changed == 'true'
         run: |
-          npm run check:pages-render -- --from-file .github/outputs/all_changed_files.txt
+          npm run check:pages-render -- --from-file .github/outputs/all_changed_files.txt &
+          sleep 1
+      - name: Stop Docker preview (cloud app)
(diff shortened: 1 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -60,15 +60,27 @@ jobs:
...
           ./start --apis &
           sleep 1
-      - name: Check that pages render
+      - name: Check that pages render (cloud app)
         if: steps.all-changed-files.outputs.any_changed == 'true'
         run: |
           npm run check:pages-render -- --from-file .github/outputs/all_changed_files.txt
+      - name: Stop Docker preview (cloud app)
+        if: steps.all-changed-files.outputs.any_changed == 'true'
+        run: docker ps -q | xargs docker stop
```
arnaucasau, line 74: This is necessary to avoid trying to run two web servers in the same port when checking the legacy app

.github/workflows/weekly-checks.yml
```diff
@@ -28,18 +28,27 @@ jobs:
...
           --current-apis
           --dev-apis
           --historical-apis
+      - name: Stop Docker preview (cloud app)
+        run: docker ps -q | xargs docker stop
+      - name: Start local Docker preview (legacy app)
+        run: |
+          ./start --legacy &
+          sleep 20
+      - name: Check all pages render (legacy app)
```
Eric-Arellano, line 49:
```suggestion
      - name: Check all non-API pages render (legacy app)
```

scripts/js/commands/checkPagesRender.ts
```diff
@@ -74,6 +85,16 @@ const readArgs = (): Arguments => {
       type: "boolean",
       description: "Check the pages in the `api/qiskit/release-notes` folder.",
     })
+    .option("legacy", {
+      type: "boolean",
+      description: "Check that the non-API pages render in the legacy app.",
+      conflicts: [
```
Eric-Arellano, line 91: TIL. Very cool!

scripts/js/commands/checkPagesRender.ts
```diff
@@ -188,5 +209,13 @@ async function determineFilePaths(args: Arguments): Promise<string[]> {
     globs.push(`${prefix}${glob}`);
   }
   const result = await globby(globs);
-  return result.filter((fp) => !ALLOWLIST.has(fp));
+  return result.filter((fp) => filterPlatformSpecificPage(fp, args.legacy));
+}
+
+function filterPlatformSpecificPage(page: string, legacy?: boolean) {
+  if (legacy) {
+    return !CLOUD_ONLY_PAGES.has(page) && !page.startsWith("/docs/api");
```
Eric-Arellano, line 217: I think the `/` prefix will mess things up with `/docs/api`?

```suggestion
    // API docs should never be checked with the legacy app because they render
    // identically in the cloud app.
    return !CLOUD_ONLY_PAGES.has(page) && !page.startsWith("docs/api");
```

Eric-Arellano (approved):
Excellent job 👏 Nice attention to detail with how you handle API pages

frankharkins (approved):
Awesome! Very clean implementation

PR #2735, opened 2025-03-06
Convert directly mappable Doxygen XML output into markdown

frankharkins:
We're currently planning on parsing the Doxygen's XML output to create the markdown for our C API docs. We decided not to use Breathe to generate Sphinx code as it'd add extra complication to the pipeline and our Sphinx-parsing code is already quite complicated. Breathe uses Doxygen's XML output, so the idea is to go straight from the Doxygen XML to our custom MDX.

This PR is the first step towards that. Parts of the Doxygen XML should map (almost) 1:1 to markdown. This PR adds a function to handle these parts of the XML. Here's an example of such XML (see also the [source docstring](https://github.com/Qiskit/qiskit/blob/2ec8f8f66650d84159f3bc058b9f38a1285175cb/crates/cext/src/sparse_observable.rs#L585-L595)):

```xml
<title>Example</title><para><verbatim>QkSparseObservable *obs = qk_obs_identity(100);
QkSparseObservable *result = qk_obs_multiply(obs, 2);
</verbatim></para>
<title>Safety</title><para>Behavior is undefined if any of the following is violated<itemizedlist>
<listitem><para><computeroutput>obs</computeroutput> is a valid, non-null pointer to a <computeroutput>QkSparseObservable</computeroutput></para>
</listitem><listitem><para><computeroutput>coeff</computeroutput> is a valid, non-null pointer to a <computeroutput>complex double</computeroutput> </para>
</listitem></itemizedlist>
</para>
```

And the markdown it converts to:

````markdown
# Example

```C
QkSparseObservable *obs = qk_obs_identity(100);
QkSparseObservable *result = qk_obs_multiply(obs, 2);
```

# Safety

Behavior is undefined if any of the following is violated

*   `obs` is a valid, non-null pointer to a `QkSparseObservable`
*   `coeff` is a valid, non-null pointer to a `complex double`
````

I did this first as I wasn't sure how complicated it'd be, it's also nicely separable from the rest of the parsing.

### XML parser alternatives considered

* https://github.com/libxmljs/libxmljs – Last commit two years ago, has CVEs in dependencies
* https://github.com/jameslan/libxml2-wasm – WASM-based implementation of the previous library, I tried this but it's quite clunky as it requires traversing the tree manually.

### CODE REVIEW

File: scripts/js/lib/api/xmlToMdx.ts [hunk 1]
```diff
@@ -0,0 +1,27 @@
...
+  preserveOrder: true,
+  alwaysCreateTextNode: true,
+  trimValues: false,
+});
+
+// Extra whitespace in the XML can appear in the MDX output.
+// This shouldn't affect how it renders on the site so we trim it to reduce noise.
+export function collapseWhitespace(mdx: string): string {
+  return mdx
+    .replaceAll("&#x20;", " ")
```
frankharkins, line 24: I couldn't work out why, but encoded spaces were appearing in the output. I think this solution is OK.

scripts/js/lib/api/directMapXmlToMdx.ts
```diff
@@ -0,0 +1,127 @@
...
+
+import type {
+  Root,
+  BlockContent,
+  Heading,
+  Text,
+  PhrasingContent,
+  List,
+  ListItem,
+  Paragraph,
```
Eric-Arellano, line 21: Consider renaming all these types to have Mdast in them, like `TextMdast` or `MdastList`. It can be confusing what is XML versus Mdast. (You'd use `as` keyword)

scripts/js/lib/api/directMapXmlToMdx.ts
```diff
@@ -0,0 +1,127 @@
...
+ */
+export function directMapXmlToMdx(nodes: MdxMappableXmlNode[]): Root {
+  return {
+    type: "root",
+    children: nodes.flatMap((n) => xmlToBlockNodes(n, 1)),
+  };
+}
+
+/*
+ * Map an XML node (and it's children) to an array of mdast `Block` nodes
```
Eric-Arellano, line 52:
```suggestion
/**
 * Map an XML node (and its children) to an array of mdast `Block` nodes
```

scripts/js/lib/api/directMapXmlToMdx.ts
```diff
@@ -0,0 +1,127 @@
...
+    return [
+      {
+        type: "code",
+        lang: "C",
+        value: extractText(node.verbatim).trimEnd(),
+      },
+    ];
+  if ("itemizedlist" in node) return [xmlListToMdx(node)];
+  // If the node isn't a recognised block node, try wrapping in a paragraph and
+  // converting to a phrasing node.
```
Eric-Arellano, line 77: Would it be safer to error for now? I imagine we'll go through a big period of discovery with this script: new edge cases we haven't yet encountered. So defensive programming is highly valuable here

scripts/js/lib/api/directMapXmlToMdx.ts
```diff
@@ -0,0 +1,127 @@
...
+    return { type: "inlineCode", value: extractText(node.computeroutput) };
+  throw new Error(`Can't map XML node: ${JSON.stringify(node)}`);
+}
+
+function xmlParagraphToBlockNode(
+  node: { para: MdxMappableXmlNode[] },
+  headingDepth: Heading["depth"],
+): BlockContent[] {
+  // Doxygen XML sometimes nests block nodes (such as lists) in paragraphs, which
+  // markdown doesn't allow. If we encounter a When processing a block node in a
```
Eric-Arellano, line 97: Is there a missing word in the second sentence?

scripts/js/lib/api/directMapXmlToMdx.test.ts
```diff
@@ -0,0 +1,128 @@
...
+import { toMarkdown } from "mdast-util-to-markdown";
+
+import { directMapXmlToMdx } from "./directMapXmlToMdx.js";
+import { xmlParser, collapseWhitespace } from "./xmlToMdx.js";
+
+// Helper function to parse string input and stringify output
+// This is easier than writing and reading the XML and MDX ASTs.
+function directMapXmlToMdxString(xml: string): string {
+  const parsedXml = xmlParser.parse(xml);
+  // console.log(JSON.stringify(parsedXml));
```
Eric-Arellano, line 23: Nit to remove all the logs before landing the code

scripts/js/lib/api/directMapXmlToMdx.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  const mdastRoot = directMapXmlToMdx(parsedXml);
+  // console.log(JSON.stringify(mdastRoot, null, 2));
+  return collapseWhitespace(toMarkdown(mdastRoot));
+}
+
+function dedentAndTrim(s: string): string {
+  return s.replaceAll("\n        ", "\n").trim();
```
Eric-Arellano, line 30: Nbd, more an FYI because I was curious how to do it in JS. You can use `" ".repeat(n)`, like how Python has `" " * 10`.

Eric-Arellano (commented):
This looks great! 

* Good job splitting this out into a smaller PR - This was easy to review :) 
* Nice tests. Very readable with how you set up the strings
* Nice use of type hints!
* Great job using early returns so much. It makes the code very readable

My only doubt is how confident you are that certain nodes never have unexpected descendants? For example, can a `<title>` node ever have a descendent even though we're not expecting it? I'm wondering how feasible it would be to use defensive programming to try to detect those edge cases and eagerly error because it violates our core assumptions. We had a lot of edge cases in the Sphinx pipeline that no one expected and we didn't find them until visually reading the output, which is not scalable.

scripts/js/lib/api/directMapXmlToMdx.ts
frankharkins: Oops, yes

scripts/js/lib/api/directMapXmlToMdx.ts
frankharkins: Good call

scripts/js/lib/api/directMapXmlToMdx.ts
```diff
@@ -0,0 +1,83 @@
...
+  if ("title" in node)
+    return { type: "heading", children: node.title.map(directMapNode) };
+  if ("verbatim" in node)
+    return {
+      type: "code",
+      lang: "C",
+      value: extractText(node.verbatim).trimRight(),
+    };
+  if ("computeroutput" in node)
+    // TODO: Handle xrefs in computeroutput nodes
```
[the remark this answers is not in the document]
frankharkins: Good call, I went with all three for now (see #2762). I might adapt (1) to ignore the refs when I get to parsing full pages.

scripts/js/lib/api/directMapXmlToMdx.ts
frankharkins: Generally, yes, but I think in this case it's nice to be able to point the function at any XML and have it return a valid block node. Especially as the XML markup doesn't have the same concepts of block vs phrasing nodes.

frankharkins:
> My only doubt is how confident you are that certain nodes never have unexpected descendants?

This is a valid concern, the [XML Schema](https://github.com/doxygen/doxygen/blob/master/templates/xml/compound.xsd) is a bit hard to read, but it does give some guarantees such as [`verbatim` always being a string](https://github.com/doxygen/doxygen/blob/b4437b34779f8e8eafa2a12972aeba956809b6c3/templates/xml/compound.xsd#L591) and [`title` only containing phrasing-like nodes](https://github.com/doxygen/doxygen/blob/b4437b34779f8e8eafa2a12972aeba956809b6c3/templates/xml/compound.xsd#L509-L542) (although we don't support them all yet). However, I was imagining we'd have to make a few adaptions in the early stages.

Eric-Arellano (approved):
Looks great! Although it may make more sense to wait to merge until we confirm the direction.

frankharkins:
I'm going to merge because it'll make it easier to move forward if we do decide to continue, and it's quite easy to remove if we decide not to use it (just delete the folder – no untangling needed).

PR #2759, opened 2025-03-11
Add 'ignore-trailing-jobs' as optional flag

emilkovacev:
nb-tester by default uses QiskitRuntimeService to cancel trailing jobs. This assumes that jobs are run using that same configuration, as opposed to being passed in via a patch. For notebooks run with fake backends, this should not be a requirement. The flag '--ignore-trailing-jobs' allows users of nb-tester to skip the step to cancel trailing jobs.

### DIFF
--- a/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
+++ b/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
@@ -33,8 +33,10 @@ async def _main() -> None:
     start_time = datetime.now()
     print("Executing notebooks:")
     results = await asyncio.gather(*(execute_notebook(job) for job in jobs))
-    print("Checking for trailing jobs...")
-    results.append(cancel_trailing_jobs(start_time))
+
+    if args.cancel_trailing_jobs:
+        print("Checking for trailing jobs...")
+        results.append(cancel_trailing_jobs(start_time))
     if not all(results):
         sys.exit(1)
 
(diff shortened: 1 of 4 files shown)


scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -33,8 +33,10 @@ async def _main() -> None:
     start_time = datetime.now()
     print("Executing notebooks:")
     results = await asyncio.gather(*(execute_notebook(job) for job in jobs))
-    print("Checking for trailing jobs...")
-    results.append(cancel_trailing_jobs(start_time))
+
+    if not args.ignore_trailing_jobs:
+        print("Checking for trailing jobs...")
+        results.append(cancel_trailing_jobs(start_time))
```
Eric-Arellano, line 39: Nit that the blank line was significant
```suggestion
        results.append(cancel_trailing_jobs(start_time))

```

scripts/nb-tester/qiskit_docs_notebook_tester/config.py
```diff
@@ -253,4 +253,11 @@ def get_parser() -> argparse.ArgumentParser:
             "This affects which notebooks are run and how they're patched."
         ),
     )
+    parser.add_argument(
+        "--ignore-trailing-jobs",
+        action="store_true",
+        help=(
+            "Ignores cancellation of trailing jobs with QiskitRuntimeService"
```
Eric-Arellano, line 260:
```suggestion
            "Do not cancel trailing jobs with QiskitRuntimeService. This is "
            "useful to set if your patch does not use QiskitRuntimeService"
```

Eric-Arellano (commented):
Thanks for making that change!

Eric-Arellano:
PR title and description need updates, please

Eric-Arellano (approved):
Going to let @frankharkins review and merge, but lgtm :)

frankharkins (approved):
Thank you!

PR #2766, opened 2025-03-11
change backend name to Torino

chongyouquan:
- Change backend name used in [q-ctrl-performance-management.ipynb](https://github.com/Qiskit/documentation/blob/eb4ba1b4db0735849d1c1471769d91ab24048439/docs/guides/q-ctrl-performance-management.ipynb) to Torino


frankharkins (changes requested):
Thanks! This approach should work fine. However, I don't think we have access to Fez from our testing instance, can we use `ibm_torino` instead?

Then it's probably easiest if I run the notebook and push to this branch.

chongyouquan:
> Thanks! This approach should work fine. However, I don't think we have access to Fez from our testing instance, can we use `ibm_torino` instead?
> 
> 
> 
> Then it's probably easiest if I run the notebook and push to this branch.

Ok done, changed the backend name to Torino. Could you take a look again? Thanks!

frankharkins:
@chongyouquan sorry for the delay, I'm trying to run the notebook but am getting an error with Qiskit Serverless that I'm struggling to debug. If you want to try it yourself, you can run this command from the root of this repo, then commit the results and push.

```
tox exec -- test-docs-notebooks docs/guides/q-ctrl-performance-management.ipynb --write
```

I'll continue looking into the serverless error.

frankharkins:
@eginez thanks, I'm running now with those requirements but it's taking quite a long time. I'll push the results as soon as it's done. In the meantime, I've pushed a couple of small changes to the notebook, please let me know if this is ok:

1. I've added a `version-info` cell to the notebook: This is just an empty cell for now, but when we run the notebook with our script, it'll populate with the package versions used so users (and ourselves) can reproduce the results in the future. All our notebooks should have one of these, we just overlooked it when adding this guide.
2. Rather than printing the full output of `get_counts()`, I've changed it to print the first 10 items. This is just because the full counts are super long.
3. Finally, I've added a hidden cell that asserts the hidden bitstring is the most frequent result. This means the notebook execution will fail if the device results are bad, which will stop us committing them. This should prevent the issue described in #2775.

Also, here's the stack trace, although it's not very helpful:

<details><summary>Stack trace</summary>

```pytb
---------------------------------------------------------------------------
QiskitServerlessException                 Traceback (most recent call last)
Cell In[8], line 2
      1 # Retrieve the counts from the result list
----> 2 result = qctrl_estimator_job.result()

File ~/Documents/documentation/.venv/lib/python3.11/site-packages/qiskit_serverless/core/job.py:190, in Job.result(self, wait, cadence, verbose, maxwait)
    187 results = self._job_service.result(self.job_id)
    189 if self.status() == "ERROR":
--> 190     raise QiskitServerlessException(results)
    192 if isinstance(results, str):
    193     try:

QiskitServerlessException: {}
```
</details>

eginez:
@frankharkins Excellent thanks for that, those changes look good.

On the stacktrace, nop that looks different. I was observing an issue with qpy serialization.
And I hear on the execution, it took me some time to get the circuit executed, but I was able to confirm that once we use the right devices, this problem gets fixed and the content is basically the same a https://github.com/Qiskit/documentation/pull/2775

PR #2767, opened 2025-03-12
Preliminary support for the C API in the MDX pipeline

arnaucasau:
This PR adds support for the Qiskit SDK C API in our MDX pipeline.

For the moment, we are using the python `Function` and `Class` components to represent the C functions and structs respectively.

I generated the docs by running:

```bash
npm run gen-api -- -p qiskit-c -v 2.0.0
```

> [!note]
> In a follow-up PR, I'll add support for the links in the signature which are currently being ignored in this PR.


scripts/js/lib/api/htmlToMd.ts
```diff
@@ -199,6 +209,44 @@ function findNodeWithProperty(nodeList: any[], propertyName: string) {
...
+  node.properties.signature = toText(node.children[0]).trim();
+  node.children = node.children.splice(1, 1);
+
+  // We use the Function component for C functions
+  node.tagName = "function";
+
+  return buildApiComponent(h, node);
+}
+
+function buildCApiStruct(h: H, node: any) {
```
arnaucasau, line 239: I divided it into two functions because I think the implementation of functions and structs will diverge when handling all the edge cases.

scripts/js/lib/api/processHtml.ts
```diff
@@ -400,8 +402,9 @@ export function maybeSetModuleMetadata(
   $: CheerioAPI,
   $main: Cheerio<any>,
   meta: Metadata,
+  isCApi: boolean,
 ): void {
-  const modulePrefix = "module-";
+  const modulePrefix = isCApi ? "group__" : "module-";
```
arnaucasau, line 407: I kept the `meta.apiType`  as  `module` because in my opinion it is strange to have the `apiType` as `group` when a single file contains several groups on it. In addition we want the same behavior in the TOC as the python modules

docs/api/qiskit-c/sparse-observable.mdx
```diff
@@ -0,0 +1,883 @@
...
+
+**<span id="group__QkSparseObservable" />*group* QkSparseObservable**
+
+This is a group of functions for interacting with an opaque (Rust-space) SparseObservable instance.
+
+<span id="breathe-section-title-functions" />
+
+**Functions**
+
+### \_CPPv411qk\_obs\_zero8uint32\_t
```
arnaucasau, line 19: The headers are generated from the function id, so maybe we should change them to be the function name instead like in the Python APIs

scripts/config/api-html-artifacts.json
```diff
@@ -30,6 +30,9 @@
     "0.24": "https://ibm.box.com/shared/static/ygdswx8s7fj970kuulkzlh8r2mjuzn3k.zip",
     "0.19": "https://ibm.box.com/shared/static/wjoea4x5tnxd0l4lgo2v3kxnx6btxvvl.zip"
   },
+  "qiskit-c": {
+    "2.0": "https://ibm.box.com/shared/static/0293fd3lwjcee97vr458wiyhebh7row9.zip"
```
Eric-Arellano, line 34: We'll need to figure out long-term if Qiskit/qiskit will give us two distinct artifacts: Python & C, or we'll have one artifact with both languages. This is sensible to start, though.

scripts/js/lib/api/htmlToMd.ts
```diff
@@ -71,7 +74,10 @@ async function generateMarkdownFile(
   return mdFile.toString().replaceAll(`<!---->`, "");
 }
 
-function prepareHandlers(meta: Metadata): Record<string, Handle> {
+function prepareHandlers(
+  meta: Metadata,
+  isCApi: boolean,
```
Eric-Arellano, line 79: The arg is unused. Can it be removed?

Eric-Arellano (commented):
Amazing how little code this takes! Great work!

scripts/js/lib/api/htmlToMd.ts
arnaucasau: Oops, good catch! Yeah, that was a leftover from the first approach

docs/api/qiskit-c/release-notes.mdx
```diff
@@ -0,0 +1,2386 @@
+---
```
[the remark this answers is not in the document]
arnaucasau: Good idea with the `ReleaseNotesConfig`! In https://github.com/Qiskit/documentation/pull/2767/commits/f1c86f87b513f1997b7929eeec08757c0ed821ec, I added it to two places to make sure we skip the release notes when is not enabled. Before, it was only used for the `_toc.json`.

Eric-Arellano (approved):

PR #2774, opened 2025-03-13
Add links to `Function`/`Class` signatures

arnaucasau:
This PR adds a mechanism to preserve the HTML links in a `Function`/`Class` signature when converting the API docs to markdown.

> [!note]
> This PR only regenerates `qiskit-addon-aqc-tensor` in order to test the new links by using `./start`. The rest of the API docs will be added in a follow-up or in this same PR after the new logic gets reviewed.


docs/api/qiskit-addon-aqc-tensor/ansatz-generation-ansatz-block.mdx
```diff
@@ -8,7 +8,7 @@ python_api_name: qiskit_addon_aqc_tensor.ansatz_generation.AnsatzBlock
 
 # AnsatzBlock
 
-<Class id="qiskit_addon_aqc_tensor.ansatz_generation.AnsatzBlock" isDedicatedPage={true} github="https://github.com/Qiskit/qiskit-addon-aqc-tensor/tree/stable/0.1/qiskit_addon_aqc_tensor/ansatz_generation.py#L32-L52" signature="AnsatzBlock(params)[source]" modifiers="class">
+<Class id="qiskit_addon_aqc_tensor.ansatz_generation.AnsatzBlock" isDedicatedPage={true} github="https://github.com/Qiskit/qiskit-addon-aqc-tensor/tree/stable/0.1/qiskit_addon_aqc_tensor/ansatz_generation.py#L32-L52" signature="AnsatzBlock(params)[[source]](https://github.com/Qiskit/qiskit-addon-aqc-tensor/tree/stable/0.1/qiskit_addon_aqc_tensor/ansatz_generation.py)" modifiers="class">
```
arnaucasau, line 11: We should remove the `[source]` links from the signature, but I think it's outside the scope of this PR

scripts/js/lib/api/generateApiComponents.ts
```diff
@@ -398,14 +411,45 @@ export function addExtraSignatures(
...
+        // by using URL-encoding form.
+        const href = determineSignatureUrl(node.properties.href)
+          .replaceAll('"', "%22")
+          .replaceAll("(", "%28")
+          .replaceAll(")", "%29");
+        const isExternal = href.startsWith("http");
+        const title = node.properties.title;
+
+        // We only show the title if it's from an external link
+        const link = title && isExternal ? `${href} "${title}"` : `${href}`;
```
arnaucasau, line 445: I decided to do this because I remember we didn't like having titles like `Qiskit v1.3` (noisy `git diff`). I kept the external references, but we can change this to have the behavior we want.

docs/api/qiskit-addon-aqc-tensor/ansatz-generation-ansatz-block.mdx
arnaucasau: FYI @Eric-Arellano. I found that the HTML contains the source link twice, so I suspect that by removing this `.first()` call will remove them all. I tested it with the current API versions and it seems to work well, but it needs more testing with the historical versions.

https://github.com/Qiskit/documentation/blob/0081fd9c159355fa4d3aa54e617d78f4f4bb4139/scripts/js/lib/api/generateApiComponents.ts#L364

If it works, we could first remove all the `[source]` links to simplify the `git diff`.

HTML of this example:

```html
...<a class="reference external" href="https://github.com/Qiskit/qiskit-addon-aqc-tensor/tree/stable/0.1/qiskit_addon_aqc_tensor/ansatz_generation.py#L32-L52"><span class="viewcode-link"><span class="pre">[source]</span></span></a><a class="reference internal" href="../_modules/qiskit_addon_aqc_tensor/ansatz_generation.html#AnsatzBlock"><span class="viewcode-link"><span class="pre">[source]</span></span></a><a class="headerlink" href="#qiskit_addon_aqc_tensor.ansatz_generation.AnsatzBlock" title="Link to this definition">¶</a>
```

Eric-Arellano: That's strange it's included twice. I'm curious how it renders in the original HTML. 

That would be great to fix this in a precursor PR! This is the only thing giving me pause to approve this PR because it makes a weird situation even weirder for our end users.

arnaucasau: This is how the original HTML renders.

Screenshot 2025-03-18 at 10 38 18


There are two links because one of them points to GitHub, and the second one shows the code under `_modules`. I created https://github.com/Qiskit/documentation/pull/2801 to remove them though.

scripts/js/lib/api/generateApiComponents.ts
```diff
@@ -15,13 +15,17 @@ import { unified } from "unified";
...
+import { visit } from "unist-util-visit";
+import { toText } from "hast-util-to-text";
 
 import { ApiType } from "./Metadata.js";
 import {
   getLastPartFromFullIdentifier,
   removeSuffix,
   APOSTROPHE_HEX_CODE,
 } from "../stringUtils.js";
+import { Root } from "mdast";
```
Eric-Arellano, line 27: Nit to put third-party imports in the above import block

scripts/js/lib/api/processHtml.ts
```diff
@@ -346,7 +347,11 @@ export async function processMembersAndSetMeta(
   $: CheerioAPI,
   $main: Cheerio<any>,
   meta: Metadata,
-  options: { isCApi: boolean },
+  options: {
+    isCApi: boolean;
+    pkgName: string;
+    kebabCaseAndShorten: boolean;
+  },
```
Eric-Arellano, line 354: This type gets used a lot. It might be worth DRYing. I'm not sure what a good name would be, maybe `ProcessHtmlOptions` or something related to the main function all these args get sent to?

scripts/js/lib/api/updateLinks.test.ts
```diff
@@ -200,6 +200,7 @@ test("normalizeUrl()", () => {
     `#qiskit_ibm_runtime.RuntimeJob.job`,
     `qiskit_ibm_runtime.RuntimeJob#qiskit_ibm_runtime.RuntimeJob.run`,
     `stubs/qiskit_ibm_runtime.RuntimeJob`,
+    `#_CPPv415qk_obs_identity8uint32_t`,
```
Eric-Arellano, line 203: Hm, this is going to fail with https://github.com/Qiskit/documentation/pull/2793. I think that suggests that `updateLinks.ts` needs to handle it? Your PR is already pretty big, so maybe create an issue in open source with this bug and ask Frank to take a look?

scripts/js/lib/api/generateApiComponents.ts
```diff
@@ -398,14 +413,58 @@ export function addExtraSignatures(
...
+        // manually creating the links as text.
+        if (!node.properties?.href || !node.children?.length) {
+          return;
+        }
+
+        const relativeHref = relativizeLink({ url: node.properties.href });
+        const normalizedHref = normalizeUrl(
+          relativeHref?.url ?? node.properties.href,
+          {},
+          new Set<string>(),
```
Eric-Arellano, line 446: How feasible would it be to pass these arguments, like how `updateLinks` does it? Otherwise, some of our URLs won't be updated properly. 

(You could refactor the code so that the creation of `resultsByName` and `itemNames` happens inside `normalizeUrl`).

scripts/js/lib/api/generateApiComponents.ts
```diff
@@ -398,14 +413,58 @@ export function addExtraSignatures(
...
+        // We encode some conflicting characters that could make the markdown link break
+        // by using URL-encoding form.
+        const href = normalizedHref
+          .replaceAll('"', "%22")
+          .replaceAll("(", "%28")
+          .replaceAll(")", "%29");
+        const isExternal = href.startsWith("http");
+        const title = node.properties.title;
+
+        // We only show the title if it's from an external link
```
Eric-Arellano, line 459:
```suggestion
        // We only show the title if it's from an external link to make the UX
        // less noisy.
```

Eric-Arellano (commented):
This new implementation looks great. Thank you for figuring out how to reuse updateLinks.ts.

FYI I opened https://github.com/Qiskit/documentation/issues/2805 to teach our link checker about these links. It is not blocking this PR.

scripts/js/lib/api/generateApiComponents.ts
arnaucasau: I'm not sure that's possible without refactoring the signature creation part. `resultsByName` and `itemsName` get computed from `initialResults` which is the result of the `convertFilesToMarkdown` function. The new code introduced in this PR is used in the `htmlSignatureToMd` function which gets called in the middle of the `convertFilesToMarkdown ` computation and therefore we can't have access to all the `initialResults`.

I've been trying to work around it, by using a unified plugin to detect them in a later step (inside `updateLinks`), but it messes up the signatures that contain markdown syntax like `_` or `*`

scripts/js/lib/api/updateLinks.test.ts
Eric-Arellano: Ah, I see. I was confusing the test with what is happening in production. Sg to keep this. Thanks for clarifying / pardon my misread!

scripts/js/lib/api/generateApiComponents.ts
arnaucasau: As we talked over Slack, I took a different approach and used placeholders to know when to update the links in `updateLinks()`. Change done in https://github.com/Qiskit/documentation/pull/2774/commits/e3b9ff0df731dd4da7f9d5f404fda32699480805

PR #2777, opened 2025-03-13
Qiskit C API: Fix code block to be C

arnaucasau:
This PR fixes the code blocks of `qiskit-c` to set the correct programming language.


Eric-Arellano (approved):
Thanks!

Nit that it's nicer to use the pattern of `options: {isCApi: boolean}` to emulate kwargs. Not blocking though

arnaucasau:
Yeah, good idea. I'll leave it as-is in this PR to minimize any possible merge conflict, but I'll probably take it into account in https://github.com/Qiskit/documentation/pull/2774 given that we need to pass more arguments as well. Thanks!

PR #2778, opened 2025-03-13
Document why we use primitives in a regular guide

beckykd:
Currently, we only document the "why" of primitives in the migration guides.

### DIFF
--- a/docs/guides/primitives.mdx
+++ b/docs/guides/primitives.mdx
@@ -6,7 +6,7 @@ description: Introduction to primitives in Qiskit and Qiskit Runtime, and an exp
 
 # Introduction to primitives
 
-## Why does Qkiskit use primitives?
+## Why does Qiskit use primitives?
 
 The `backend.run` interface was the primary access point to quantum hardware and simulators offered by IBM&reg;, as well as third-party providers that integrate with Qiskit. The evolution of user access patterns for quantum processing units (QPUs) led to the Qiskit primitives (`Sampler` and `Estimator`) superseding this interface.
 
@@ -24,7 +24,7 @@ the Qiskit [primitives interface](/primitives), which is optimized for two core 
 expectation value estimation (`Estimator`) and circuit sampling (`Sampler`). The goal is once
 again to help developers to focus more on innovation and less on data conversion.
 
-## What is a primitve?
+## What is a primitive?
 
 Computing systems are built on multiple layers of abstraction. Abstractions allow you to focus on a
 particular level of detail relevant to the task at hand. The closer you get to the hardware,
--- a/docs/guides/primitives.mdx
+++ b/docs/guides/primitives.mdx
@@ -20,7 +20,7 @@ to operate on a familiar data format and focus on the bigger picture.
 
 As access to QPUs became more widespread, and with more quantum algorithms being developed,
 again the need for a higher-level abstraction emerged. In response, IBM introduced
-the Qiskit [primitives interface](/primitives), which is optimized for two core tasks in quantum algorithm development:
+the Qiskit primitives interface, which is optimized for two core tasks in quantum algorithm development:
 expectation value estimation (`Estimator`) and circuit sampling (`Sampler`). The goal is once
 again to help developers to focus more on innovation and less on data conversion.
 
--- a/docs/guides/primitives.mdx
+++ b/docs/guides/primitives.mdx
@@ -8,21 +8,16 @@ description: Introduction to primitives in Qiskit and Qiskit Runtime, and an exp
 
 ## Why does Qiskit use primitives?
 
-The `backend.run` interface was the primary access point to quantum hardware and simulators offered by IBM&reg;, as well as third-party providers that integrate with Qiskit. The evolution of user access patterns for quantum processing units (QPUs) led to the Qiskit primitives (`Sampler` and `Estimator`) superseding this interface.
-
-<span id="qpu-access-patterns"></span>
-### User access patterns for QPUs
-
 Similar to the early days of classical computers, when developers had to manipulate CPU registers directly, the early interface to QPUs simply returned the raw data from the control electronics.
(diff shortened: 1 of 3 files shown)


docs/guides/primitives.mdx
```diff
@@ -6,6 +6,26 @@ description: Introduction to primitives in Qiskit and Qiskit Runtime, and an exp
 
 # Introduction to primitives
 
+## Why does Qiskit use primitives?
```
jyu00, line 9: It's more accurate to say Qiskit _introduced_ primitives, instead of merely using them, since the base interfaces live in Qiskit.

docs/guides/primitives.mdx
```diff
@@ -6,6 +6,26 @@ description: Introduction to primitives in Qiskit and Qiskit Runtime, and an exp
...
+Similar to the early days of classical computers, when developers had to manipulate CPU registers directly, the early interface to QPUs simply returned the raw data from the control electronics.
+This was not a big issue when QPUs lived in labs and only allowed direct access by researchers.
+Acknowledging that most developers would not and should not be familiar with distilling such raw data into 0s and 1s, IBM&reg; introduced `backend.run`, a first abstraction for accessing QPUs in the cloud. This allowed developers
+to operate on a familiar data format and focus on the bigger picture.
+
+As access to QPUs became more widespread, and with more quantum algorithms being developed,
+again the need for a higher-level abstraction emerged. In response, IBM introduced
+the Qiskit primitives interface, which is optimized for two core tasks in quantum algorithm development:
+expectation value estimation (`Estimator`) and circuit sampling (`Sampler`). The goal is once
+again to help developers to focus more on innovation and less on data conversion.
```
jyu00, line 25: We can easily consolidate these two sections into one. I also changed the wording to say primitives interface was introduced by Qiskit, since it's vendor agnostic. 

```suggestion
Similar to the early days of classical computers, when developers had to manipulate CPU registers directly, the early interface to QPUs simply returned the raw data from the control electronics.
This was not a big issue when QPUs lived in labs and only allowed direct access by researchers.
Acknowledging that most developers would not and should not be familiar with distilling such raw data into 0s and 1s, Qiskit introduced `backend.run`, a first abstraction for accessing QPUs in the cloud. This allowed developers
to operate on a familiar data format and focus on the bigger picture.

As access to QPUs became more widespread, and with more quantum algorithms being developed,
again the need for a higher-level abstraction emerged. In response, Qiskit introduced
the primitives interface, which is optimized for two core tasks in quantum algorithm development:
expectation value estimation (`Estimator`) and circuit sampling (`Sampler`). The goal is once
again to help developers to focus more on innovation and less on data conversion. The primitives interface supersedes the `backend.run` interface, since `Sampler` provides the same direct hardware access that was offered by `backend.run`. 
```

docs/migration-guides/qiskit-runtime.mdx
```diff
@@ -17,24 +17,13 @@ examples to help you migrate code from the legacy `backend.run` interface to use
...
-Acknowledging that most developers would not and should not be familiar with distilling such raw data into
-0s and 1s, IBM introduced `backend.run`, a first abstraction for accessing QPUs in the cloud. This allowed developers
-to operate on a familiar data format and focus on the bigger picture.
-
-As access to QPUs became more widespread, and with more quantum algorithms being developed,
-again the need for a higher-level abstraction emerged. In response, IBM introduced
-the Qiskit [primitives interface](/guides/primitives), which are optimized for two core tasks in quantum algorithm development:
-expectation value estimation (`Estimator`) and circuit sampling (`Sampler`). The goal is once
-again to help developers to focus more on innovation and less on data conversion.
+The `backend.run` interface was the primary access point to quantum hardware and simulators offered by IBM, as well as third-party providers that integrate with Qiskit. The [evolution of user access patterns](/guides/primitives#qpu-access-patterns) for quantum processing units (QPUs) led to the Qiskit primitives (`Sampler` and `Estimator`) superseding this interface.
```
jyu00, line 26:
```suggestion
The `backend.run` interface was the primary access point to quantum hardware and simulators offered by IBM, as well as third-party providers that integrate with Qiskit. The [evolution of user access patterns](/guides/primitives#why-did-qiskit-introduce-primitives) for quantum processing units (QPUs) led to the Qiskit primitives (`Sampler` and `Estimator`) superseding this interface.
```

jyu00 (commented):
The `user access patterns for QPUs` link on [this page](https://docs.quantum.ibm.com/migration-guides/external-providers-primitives-v2#why-should-third-party-providers-implement-primitives) should be updated to point to the main doc as well.

docs/migration-guides/qiskit-runtime.mdx
beckykd: (/guides/primitives#qpu-access-patterns) is the right link because I have an anchor there. I don't understand why the anchor doesn't show up when you hover, so I know it's confusing.  But it's better to use the anchor in case we change the title.

jyu00:
> The `user access patterns for QPUs` link on [this page](https://docs.quantum.ibm.com/migration-guides/external-providers-primitives-v2#why-should-third-party-providers-implement-primitives) should be updated to point to the main doc as well.

@beckykd can you make this ^ change? It currently points to the same verbiage on the `Migrate from backend.run to primitives` page, which then points to the actual "qpu access patterns"

beckykd:
> > The `user access patterns for QPUs` link on [this page](https://docs.quantum.ibm.com/migration-guides/external-providers-primitives-v2#why-should-third-party-providers-implement-primitives) should be updated to point to the main doc as well.
> 
> @beckykd can you make this ^ change? It currently points to the same verbiage on the `Migrate from backend.run to primitives` page, which then points to the actual "qpu access patterns"

done

jyu00 (approved):
Thanks @beckykd!

abbycross (approved):

PR #2790, opened 2025-03-17
Remove C API XML stuff

frankharkins:
We're no longer using this approach so we can remove it.


Eric-Arellano:
You can also uninstall fastxml

Issue #2792, opened 2025-03-17
`qiskit-addon-aqc-tensor` is missing some headings


arnaucasau:
Some `qiskit-addon-aqc-tensor` pages are missing the `Function` headings ([example](https://docs.quantum.ibm.com/api/qiskit-addon-aqc-tensor/simulation#-4)). We should investigate if it's a problem with our MDX pipeline or if they don't have any ID defined in the HTML.

<img width="794" alt="Image" src="https://github.com/user-attachments/assets/4de43da4-3414-4661-a599-ce4bf883bba4" />

PR #2793, opened 2025-03-17
Fix function IDs in C API

frankharkins:
Small PR to grab the ID from the function signature rather than the `id` attribute in the C API.

### DIFF
--- a/scripts/js/lib/api/processHtml.ts
+++ b/scripts/js/lib/api/processHtml.ts
@@ -78,7 +78,7 @@ export async function processHtml(options: {
   preserveMathBlockWhitespace($, $main);
 
   const meta: Metadata = {};
-  await processMembersAndSetMeta($, $main, meta);
+  await processMembersAndSetMeta($, $main, meta, isCApi);
   maybeSetModuleMetadata($, $main, meta, isCApi);
   if (meta.apiType === "module") {
     updateModuleHeadings($, $main);
@@ -339,6 +339,7 @@ export async function processMembersAndSetMeta(
   $: CheerioAPI,
   $main: Cheerio<any>,
   meta: Metadata,
+  isCApi: boolean,
 ): Promise<void> {
   let continueMapMembers = true;
   while (continueMapMembers) {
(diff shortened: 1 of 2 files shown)


scripts/js/lib/api/processHtml.ts
```diff
@@ -357,7 +358,13 @@ export async function processMembersAndSetMeta(
     }
 
     const $dl = $(dl);
-    const id = $dl.find("dt").attr("id") || "";
+    const id =
+      (isCApi
+        ? // IDs in the C API have a bunch of extra information (e.g.
+          // `_CPPv415qk_obs_free8uint32_t`) whereas we just want to display the
+          // function name (e.g. `qk_obs_free`).
+          $dl.find("span.sig-name.descname").text()
```
arnaucasau, line 366: Great! I think we tried using the function name in the python APIs and we ended up changing it for the ID, but I'm not able to remember why, and I don't see any reason for not doing it now in the C API. In the worst case, if we see any bad instance we can rework this part (this is already a good improvement). Do you remember @Eric-Arellano?

I think this line will be useful even for the Python APIs as a second option when ID is not defined. It will be relevant for https://github.com/Qiskit/documentation/issues/2792. 

Awesome job finding the alternative!

arnaucasau (approved):
Awesome!

scripts/js/lib/api/processHtml.ts
```diff
@@ -339,6 +339,7 @@ export async function processMembersAndSetMeta(
   $: CheerioAPI,
   $main: Cheerio<any>,
   meta: Metadata,
+  isCApi: boolean,
```
Eric-Arellano, line 342: Nit to consider using `kwargs: {isCApi: boolean}` to make the code more readable. I think Arnau was going to make that change somewhere else but I'm not sure he's making the change anymore.

Not blocking!

scripts/js/lib/api/processHtml.ts
Eric-Arellano: I think it was because of methods. For inline module pages, you could have >1 class that have the same method.

That isn't an issue with C:
- no classes/methods
- no function overloading

Frank, can you please add this to the comment?

```
// This is safe because C does not have function overloading so the function name
// is unique.
```

Eric-Arellano (approved):
🚀

Issue #2800, opened 2025-03-17
Improve developer experience of running local checks


Eric-Arellano:
## Summary
We have a substantial number of checks, per the README. 

However, there is a paper cut with the dev exp (DX) that most of our tools are in TypeScript and invoked with `npm`, whereas others are in Python and invoked with `tox` (Squeaky and Ruff).

### Tools audit
Require manual intervention
- images
- spelling
- metadata
- internal links
- patterns index
- orphan pages
- qiskit bot
- Ruff, some lints
- page title (proposed)
- notebook/image size (proposed)
- redirects (proposed)

Can auto-fix
- Squeaky
- Ruff, most things
- Post-process notebooks

normally not used locally
- external links
- pages render
- notebook tester
- Prettier (only relevant to devs)
- stale images (it could be used locally, but the content writers agreed it's not worth the performance)

### What we value
- easy to invoke
- speed
- cross-platform
- not noisy
- clear instructions on failure

### Not in scope
- API docs
- replacing each individual command. This is only a shortcut to invoke the tools

### Phase 1
Add two scripts:

`./fix`
`./check`

They simply run `tox -e lint/fix` and `npm run check`.

### Phase 2 - logging

Log when every tool is complete to give a sense of progress. If this ends up being too noisy, we can change to only log progress messages like (3/11 checks passed)

```
./check
✅ Success: internal link checker (8 checks remaining)
✅ Success:  orphan pages (7 checks remaining)
...
✅ Success:  Qiskit bot

🎉 All checks passed!
```

Fail eagerly once a tool fails:

./check
✅ Success:  metadata (7 checks remaining)
❌ Failure: internal link checker

---------------------------

     some tool output
    ❌ more tool output

----------------------------

### Phase 3 - better performance via Git

Moved to https://github.com/Qiskit/documentation/issues/2854

## Implementation ideas

* Rewrite our README once this is added.
* Don't parallelize `./check` yet to keep it simple. I'm not certain it would actually help because our checks are very IO-bound & risk resource contention. (It's never safe to parallelize `./fix` because it writes)
* Be on the lookout for where we can move more instructions to the tool's error messages. We don't want developers to have to learn what all our different tools are. They should simply run `./check` and it will tell them everything they need.
* It'd be neat if we could distribute this tool to private repositories like Qiskit Functions, by using a VCS requirement. We don't have to do it in the first phase, but consider how to make the design generic.

PR #2807, opened 2025-03-18
Don't run Prettier with `npm run check`

Eric-Arellano:
A quick win for https://github.com/Qiskit/documentation/issues/2800. We should optimize `npm run check` for content writers, not developers. Content writers don't work on TypeScript.

This speeds up `npm run check` by 3.8s (real) on my Macbook Pro.


Eric-Arellano:
This is a good change, but I plan to replace `npm run check` anyways this week so let's not merge to avoid noise.

PR #2815, opened 2025-03-19
Update C API artifact

frankharkins:
Regenerates the C API from a new artifact generated from https://github.com/frankharkins/qiskit-terra/tree/FH/c-api-docs. I did not check in images from the release notes or the `objects.inv` files as they'll eventually be removed during the generation script.


Eric-Arellano (approved):

docs/api/qiskit-c/_package.json
```diff
@@ -1,4 +1,4 @@
 {
   "name": "qiskit-c",
-  "version": "2.0.0"
+  "version": "2.0"
```
arnaucasau, line 3: I think our app assumes we always have a patch version. Was this change intentional?

PR #2819, opened 2025-03-19
Add a `./check` and `./fix` script

Eric-Arellano:
Minimum viable product for https://github.com/Qiskit/documentation/issues/2800. We don't have time to implement yet the follow up improvements (better logging & using Git to improve performance), but we can at least get some of the wins now.

A follow up PR will rework our README to talk about these scripts. I want to generally better differentiate between content writers versus developers.

This PR also re-orders `npm run check` so that the fastest checks run first.

### DIFF
--- a/check
+++ b/check
@@ -1,0 +1,23 @@
+#!/usr/bin/env python3
+# This code is a Qiskit project.
+#
+# (C) Copyright IBM 2025.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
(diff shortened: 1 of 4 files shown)


check
```diff
@@ -0,0 +1,23 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+from scripts.tool_runner import run_cmd
```
Eric-Arellano, line 14: I considered a few different names for the folder like `writer_tool_runner`. I wanted to emphasize this is only a runner of the tools. We should not define any tools in this folder.

I could imagine that maybe we'll have other private repositories like Functions install this new tool runner project as a VCS requirement, if it gets more complex. I'm not worried about that for now, but we should keep it in mind. Maybe, for example, we should move some `./start` code to this new folder.

PR #2824, opened 2025-03-20
Bd manage cost from UI

beckykd:


beckykd:
Looks like it was OK in main

PR #2829, opened 2025-03-20
Automate auditing new pills

### LINKED ISSUE #2728: Come up with system to audit new pill tags
how are we tracking when to add or remove the new pills?

Update:

- [ ] add a file that tracks dates for each new pill
- [ ] review list every content backlog sweep
- [ ] stretch goal: set up an automation to remove pills automatically after 1 month

kaelynj:
Closes #2728 

A new script `checkNewPills.ts` checks the toc in `/guides`, `/migration-guides`, and `/open-source` for pages with the `isNew : true` entry and returns an error if any of these entries are over two weeks old.

The proposed way to track the date in which the page was added is for the content writer to add an `isNewDate: YYYY-MM-DD` entry to track the date which the page was (approximately) published. If this date is older than two weeks (or the `isNewDate` entry is missing), then this check errors with a list of affected urls in which to remove the new pill from.

This PR also adds the `checkNewPills.ts` to our weekly cron job in order to notify us when a new pill becomes outdated.


.github/workflows/weekly-checks.yml
```diff
@@ -66,9 +66,23 @@ jobs:
...
+      steps:
+        - uses: actions/checkout@v4
+        - name: Set up Node.js
+          uses: actions/setup-node@v4
+          with:
+            node-version: 18
+        - name: Install Node.js dependencies
+          run: npm ci
+        - name: Check outdated new pills
+          run: npm run check-new-pills'
```
Eric-Arellano, line 81: There's a typo that will cause an error. The other suggestion is to use the convention we normally use for checks.

```suggestion
          run: npm run check:new-pills
```

docs/guides/_toc.json
```diff
@@ -155,7 +157,8 @@
         {
           "title": "Latest updates",
           "url": "/guides/latest-updates",
-          "isNew": true
+          "isNew": true,
+          "isNewDate": "2025-02-26"
```
Eric-Arellano, line 161: I think this is out of date?

package.json
```diff
@@ -27,6 +27,7 @@
     "check:orphan-pages": "tsx scripts/js/commands/checkOrphanPages.ts",
     "check:qiskit-bot": "tsx scripts/js/commands/checkQiskitBotFiles.ts",
     "check:stale-images": "tsx scripts/js/commands/checkStaleImages.ts",
+    "check-new-pills": "tsx scripts/js/commands/checkNewPills.ts",
```
Eric-Arellano, line 30:
```suggestion
    "check:new-pills": "tsx scripts/js/commands/checkNewPills.ts",
```

scripts/js/commands/checkNewPills.ts
```diff
@@ -0,0 +1,107 @@
+// This code is a Qiskit project.
+//
+// (C) Copyright IBM 2024.
```
Eric-Arellano, line 3:
```suggestion
// (C) Copyright IBM 2025.
```

scripts/js/commands/checkNewPills.ts
```diff
@@ -0,0 +1,107 @@
...
+    "docs/migration-guides/_toc.json",
+    "docs/open-source/_toc.json"
+  ]);
+  const allOutdatedPills = [];
+  for (const tocFile of tocFiles) {
+    const tocNewPills = await findNewPills(tocFile);
+    const outdatedPills = await getOutdatedPills(tocNewPills);
+    if (outdatedPills.length > 0) {
+      allOutdatedPills.push(...outdatedPills);
+    }
```
Eric-Arellano, line 40: FYI `push(...[])` is a no-op.

```suggestion
    allOutdatedPills.push(...outdatedPills);
```

scripts/js/commands/checkNewPills.ts
```diff
@@ -0,0 +1,107 @@
...
+      const isOld = getDateDifference(pill[1]);
+      outdatedPills.push([pill[0], pill[1]]);
+    }
+  }
+  return outdatedPills;
+}
+
+function getDateDifference(pubDate: string): boolean {
+  const [year, month, day] = pubDate.split('-').map(Number);
+  const oldDate = new Date(year, month-1, day);
```
Eric-Arellano, line 71: The Date constructor understands our strings automatically.

```suggestion
  const oldDate = new Date(pubDate);
```

scripts/js/commands/checkNewPills.ts
```diff
@@ -0,0 +1,107 @@
...
+  const dateDiff = Math.abs(TODAY.getTime() - oldDate.getTime() )
+  const diffDays = Math.ceil(dateDiff / (1000 * 3600 * 24));
+  if (diffDays > 14) {
+    return true;
+  }
+  return false;
```
Eric-Arellano, line 77: This should work. You can subtract the dates directly.

```suggestion
  const msPerDay = 1000 * 60 * 60 * 24;
  const daysDiff = (TODAY - oldDate) / msPerDay;
  return daysDiff > 14;
```

scripts/js/commands/checkNewPills.ts
```diff
@@ -0,0 +1,107 @@
...
+// Parse _toc.json to extract urls and dicts containing 'isNew' entries
+async function findNewPills(tocFilePath: string): Promise<string[]> {
+  console.log("Checking new pills in toc:", tocFilePath);
+  const raw = await fs.readFile(tocFilePath, "utf-8");
+  const rootEntries = JSON.parse(raw).children;
+  const isNewEntries = collectNewPills(rootEntries);
+  return isNewEntries;
+}
+
+function collectNewPills(entries: TocEntry[]): string[] {
```
Eric-Arellano, line 89: The return type is wrong. It is saying it returns an array of strings, but really it needs to return something like `NewPillEntry[]`. You should define `NewPillEntry`:

```
interface NewPillEntry {
  url: string;
  date: string | null;
}
```

You're already returning something similar, but using `[string, string]` to represent it, which is harder to read the code. You're also using the magic value `no-date` rather than using `null` to represent when something does not have a date.

scripts/js/commands/checkNewPills.ts
```diff
@@ -0,0 +1,107 @@
...
+      allOutdatedPills.map(item => item.join(' : ')).join("\n"),
+      "\n\n",
+    );
+  }
+  else {
+    console.log("\nNo outdated new pills found ✅\n");
+  }
+}
+
+async function getOutdatedPills(newPills: string[]): Promise<string[][]> {
```
Eric-Arellano, line 54: This function is confusing to me with the type hints. I recommend starting with my suggestion in `collectNewPills`, then it should be more clear how to fix this.

Eric-Arellano (commented):
Thanks, Kaelyn! 

When you're developing this, it will help to frequently run `npm run typecheck` to ensure the code is valid. You'll also want to run `npm run fmt` for autoformatting.

docs/guides/_toc.json
kaelynj: Yeah, I wanted to keep it there for now to make sure the check was working. I'll remove it before this PR is merged.

scripts/js/commands/checkNewPills.ts
```diff
@@ -0,0 +1,99 @@
...
+import { hideBin } from "yargs/helpers";
+import { flatten, flattenDeep } from "lodash-es";
+
+import { TocEntry } from "../lib/api/generateToc.js";
+
+const TODAY = new Date();
+
+interface NewPillEntry {
+  url: string;
+  date?: string | "null";
```
Eric-Arellano, line 27:
```suggestion
  date: string | null;
```

1. `null` is a special value, whereas you're using the string "null" right now
2. `date?` is shorthand for `date: string | "null" | undefined`. JavaScript is weird that it has both `undefined` and `null`. We only want to use `null` to model that the date is missing; it's confusing to use both `undefined` and `null`

(Applying my code suggestion in GitHub directly will break a bunch of things. You should fix this locally.)

scripts/js/commands/checkNewPills.ts
```diff
@@ -0,0 +1,99 @@
...
+  const allOutdatedPills = [];
+  for (const tocFile of tocFiles) {
+    const tocNewPills = await findNewPills(tocFile);
+    const outdatedPills = await getOutdatedPills(tocNewPills);
+    allOutdatedPills.push(...outdatedPills);
+  }
+  if (allOutdatedPills.length > 0) {
+    allOutdatedPills.forEach((outdatedPill: NewPillEntry) =>
+      console.error(
+        `\n❌ This URL needs to have their new pill removed: ${outdatedPill.url}`,
```
Eric-Arellano, line 46:
```suggestion
        `\n❌ The new pill should be removed for the url '${outdatedPill.url}' in ${tocFile}`,
```

scripts/js/commands/checkNewPills.ts
```diff
@@ -0,0 +1,99 @@
...
+      ),
+    );
+  } else {
+    console.log("\nNo outdated new pills found ✅\n");
+  }
+}
+
+async function getOutdatedPills(
```
Eric-Arellano, line 54: This doesn't need to be async because you don't `await` anything.

I'd also recommend calling it something like `detectBadNewPills` to make more clear it's a "pure function" that doesn't do any side effects like reading from the filesystem.

Finally, this function is conflating two different error paths:
1. Outdated pills
2. Missing dates

Those are distinct errors and should have different error messages. So, I recommend having this function return the type `{outdated: NewPillEntry[]; missingDate: NewPillEntry[]}`. In your `main` function, you'll have two arrays for `allOutdatedPills` and `allMissingDates`. They each get their own error message

scripts/js/commands/checkNewPills.ts
```diff
@@ -0,0 +1,99 @@
...
+  return outdatedPills;
+}
+
+// Parse _toc.json to extract urls and dicts containing 'isNew' entries
+async function findNewPills(tocFilePath: string): Promise<NewPillEntry[]> {
+  console.log("Checking new pills in toc:", tocFilePath);
+  const raw = await fs.readFile(tocFilePath, "utf-8");
+  const rootEntries = JSON.parse(raw).children;
+  const isNewEntries = collectNewPills(rootEntries);
+  return isNewEntries;
```
Eric-Arellano, line 79: The variable name introduces unnecessary noise. You can inline

```suggestion
  return collectNewPills(rootEntries);
```

scripts/js/commands/checkNewPills.ts
```diff
@@ -0,0 +1,99 @@
...
+  const isNewDates = [];
+  for (const entry of entries) {
+    if ("children" in entry) {
+      const childEntries = collectNewPills(entry.children || []);
+      isNewDates.push(...childEntries);
+    } else if (entry.isNew && entry.url) {
+      if (entry.isNewDate) {
+        isNewDates.push({ url: entry.url, date: entry.isNewDate });
+      } else {
+        isNewDates.push({ url: entry.url });
```
Eric-Arellano, line 92:
```suggestion
        isNewDates.push({ url: entry.url, date: null });
```

scripts/js/commands/checkNewPills.ts
```diff
@@ -0,0 +1,99 @@
...
+async function findNewPills(tocFilePath: string): Promise<NewPillEntry[]> {
+  console.log("Checking new pills in toc:", tocFilePath);
+  const raw = await fs.readFile(tocFilePath, "utf-8");
+  const rootEntries = JSON.parse(raw).children;
+  const isNewEntries = collectNewPills(rootEntries);
+  return isNewEntries;
+}
+
+function collectNewPills(entries: TocEntry[]): NewPillEntry[] {
+  const isNewDates = [];
```
Eric-Arellano, line 83: A better variable name is `newPillEntries`. Dates is misleading

Eric-Arellano (commented):
This is looking good. You'll want to update `mdx-guide.md` in the section "## How to add a new page" to say that you need to also set the isNewDate

kaelynj:
> This is looking good. You'll want to update `mdx-guide.md` in the section "## How to add a new page" to say that you need to also set the isNewDate

@Eric-Arellano just finished addressing these. Let me know if there's anything else that needs adjusting

Eric-Arellano (commented):
This is looking great! Should be good to go after this final round

Eric-Arellano (approved):
Great work! 🚀

PR #2832, opened 2025-03-21
Tweaks to C API parser

frankharkins:
This PR makes three changes to the API generation script:

- **Recognize C `typedef` and format as class**

- **Do not use dedicated pages in C API**
  
  In Python APIs, if a class has the same name as the page name, the page will be a "dedicated page" for that class. This commit disables that for the C API. The C API pages are a grouping of structs and associated functions, rather than a class and its methods. This change reflects that in the information heirarchy.

- **Fix TOC for C API**

  C does not have modules, so we can't group pages by module. This commit just lists all C API pages in the TOC.

### DIFF
--- a/scripts/js/lib/api/Metadata.ts
+++ b/scripts/js/lib/api/Metadata.ts
@@ -20,7 +20,7 @@ export type ApiType =
   | "exception"
   | "data"
   | "struct"
-  | "type";
+  | "typedef";
 
 export type Metadata = {
   apiName?: string;
--- a/scripts/js/lib/api/Metadata.ts
+++ b/scripts/js/lib/api/Metadata.ts
@@ -19,7 +19,8 @@ export type ApiType =
   | "function"
   | "exception"
   | "data"
-  | "struct";
+  | "struct"
+  | "type";
 
 export type Metadata = {
   apiName?: string;
--- a/scripts/js/lib/api/generateToc.ts
+++ b/scripts/js/lib/api/generateToc.ts
@@ -72,6 +72,8 @@ function getModulesAndItems(
     (result) => !isEmpty(result.meta.apiName),
   );
 
+  // We group Python packages by module for better organization.
+  // However, C does not have modules.
   const modules = options.isCApi
     ? resultsWithName
     : resultsWithName.filter((result) => result.meta.apiType === "module");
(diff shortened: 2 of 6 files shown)


scripts/js/lib/api/Metadata.ts
```diff
@@ -19,7 +19,8 @@ export type ApiType =
   | "function"
   | "exception"
   | "data"
-  | "struct";
+  | "struct"
+  | "type";
```
Eric-Arellano, line 23: Ik the code will be a little more complex for `getApiType`, but I think it'd be worth calling this `typedef` rather than `type` to make it clear this is C and not Python types. `type` is very overloaded. I think you only need to update `getApiType` to change its detection, then everything else simply s/type/typedef.

scripts/js/lib/api/generateToc.ts
```diff
@@ -31,7 +31,9 @@ type Toc = {
 };
 
 export function generateToc(pkg: Pkg, results: HtmlToMdResultWithUrl[]): Toc {
-  const [modules, items] = getModulesAndItems(results);
+  const [modules, items] = getModulesAndItems(results, {
+    isCApi: pkg.language === "C",
```
Eric-Arellano, line 35: We now repeat this line twice. A better factoring is probably to have `Pkg.isCApi()` for DRY and better encapsulation

Not blocking

scripts/js/lib/api/generateToc.ts
```diff
@@ -64,14 +66,15 @@ export function generateToc(pkg: Pkg, results: HtmlToMdResultWithUrl[]): Toc {
...
   const resultsWithName = results.filter(
     (result) => !isEmpty(result.meta.apiName),
   );
 
-  const modules = resultsWithName.filter(
-    (result) => result.meta.apiType === "module",
-  );
+  const modules = options.isCApi
+    ? resultsWithName
+    : resultsWithName.filter((result) => result.meta.apiType === "module");
```
Eric-Arellano, line 77: A comment may help, like

// In Python, we want to group by modules for better organization.
// Meanwhile, C does not have the idea of modules.

Eric-Arellano (commented):
All sensible changes. Thanks :)

Eric-Arellano (approved):

PR #2836, opened 2025-03-21
More C API tweaks

frankharkins:
- **Change min TOC level in C API docs**
- **Support C enums**


Eric-Arellano (approved):
Nice! Thoughts on naming the C types to mention C? Like CStruct and CEnum? I can't decide if I think it's overkill or not. It might make it more clear what is Python vs C

frankharkins:
I agree it could be confusing, especially if we encounter any overlap between Python and C types. I have another idea for a refactor; I'll try and make a PR today.

PR #2842, opened 2025-03-24
add noisy results

assafb:


ElePT (commented):
These changes look good to me too, provided that the formatting suggestion is applied, thanks! We might want to do a more general functions documentation content review in a follow-up, but I think it's best to merge this PR first.

PR #2843, opened 2025-03-24
Refactor API types

frankharkins:
Small refactor to keep information about API types in one place. This should make it a little cleaner to add new API types.

In addition to running the regular tests, I also ran `npm run regen-apis` up to Qiskit v0.42.0 then checked the following command had no output.

```
git diff f9acc9098af856cc11eadfa863baf4afa69b959c --name-only | grep -v '\.avif$'
```

### DIFF
--- a/scripts/js/lib/api/Metadata.ts
+++ b/scripts/js/lib/api/Metadata.ts
@@ -10,22 +10,50 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
-export type ApiType =
+export type ApiTypeName =
+  // Python API types
   | "class"
   | "method"
   | "property"
   | "attribute"
-  | "module"
   | "function"
   | "exception"
   | "data"
+  // C API types
   | "struct"
+  | "cFunction"
   | "typedef"
   | "enum"
   | "enumerator";
 
+interface ApiType {
+  cssSelector: string;
+  tagName: "class" | "function" | "attribute";
+}
+
+/**
+ * This object defines every API type we support. The key is the name we use to
+ * refer to the API type in this script, the `cssSelector` is the CSS selector
+ * in the Sphinx HTML, and the `tagName` is the tag we transform it to (and how
+ * it'll be displayed on the website).
+ */
+export const API_TYPES: { [K in ApiTypeName]: ApiType } = {
+  class: { cssSelector: "dl.py.class", tagName: "class" },
+  exception: { cssSelector: "dl.py.exception", tagName: "class" },
+  attribute: { cssSelector: "dl.py.attribute", tagName: "attribute" },
+  property: { cssSelector: "dl.py.property", tagName: "attribute" },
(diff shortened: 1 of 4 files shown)


scripts/js/lib/api/Metadata.ts
```diff
@@ -10,22 +10,51 @@
...
   | "enumerator";
 
+interface ApiObjectInfo {
+  cssSelector: string;
+  tagName: "class" | "function" | "attribute";
+}
+
+/**
+ * This object defines every API type we support. The key is the name we use to
+ * refer to the API type in this script, the `cssSelector` is the CSS selector
```
Eric-Arellano, line 37: Nit that `htmlSelector` is more accurate. The selector isn't only used for CSS, e.g. HTML uses it in forms to associate labels with inputs. JS also uses it.

scripts/js/lib/api/processHtml.ts
```diff
@@ -375,8 +377,8 @@ export async function processMembersAndSetMeta(
         : $dl.find("dt").attr("id")) || "";
     const apiType = getApiType($dl);
 
-    if (apiType && apiType === "module") {
-      throw new Error("Did not expect apiType to be 'module'");
+    if (!apiType) {
+      throw new Error(`Could not determine apiType for '${$dl}'`);
```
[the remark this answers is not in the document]
Eric-Arellano: Stricter is good. We can make it less strict if we have future errors, but this looks reasonable to me.

scripts/js/lib/api/processHtml.ts
```diff
@@ -467,33 +469,21 @@ export function updateModuleHeadings($: CheerioAPI, $main: Cheerio<any>): void {
...
-  for (const className of apiClassNames) {
-    if ($dl.hasClass(className)) {
-      if (className === "type") return "typedef";
-      return className;
-    }
+  for (const [apiTypeName, apiType] of Object.entries(API_OBJECTS)) {
+    const className = apiType.cssSelector.split(".").pop();
+    if (!className)
+      throw new Error(
+        `'cssSelector' attribute must be of form 'dl.py.class', found '${apiType}'.`,
```
Eric-Arellano, line 484: Nit that it would be better to DRY `dl.py.class` by accessing `API_OBJECTS` (even though we don't expect it to change)

Eric-Arellano (approved):
Thank you for taking the time to improve this! I was hoping you'd see ways to improve the pipeline with this project, and indeed you did 🚀

arnaucasau (approved):
Awesome work Frank! I like a lot the new `API_OBJECTS`. I run `npm run regen-apis` locally to verify we have the same behavior as before with the python APIs and everything works very well!

scripts/js/lib/api/processHtml.ts
frankharkins: Thanks, that's just supposed to be an example for the error message. I've updated it to be a bit clearer.

PR #2851, opened 2025-03-25
Remove 'Public members' rubric

frankharkins:
This removes the "Public members" rubric from structs in the C API docs. This is because the rubric appears to have the same level at the fields themselves. We decided the best solution to avoid a confusing information hierarcy is to remove the rubric.

### DIFF
--- a/scripts/js/lib/api/processHtml.ts
+++ b/scripts/js/lib/api/processHtml.ts
@@ -70,6 +70,7 @@ export async function processHtml(
   removeHtmlExtensionsInRelativeLinks($, $main);
   removePermalinks($main);
   removeDownloadSourceCode($main);
+  removePublicMembersRubric($main);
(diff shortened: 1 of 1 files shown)


scripts/js/lib/api/processHtml.ts
```diff
@@ -157,6 +158,13 @@ export function removeDownloadSourceCode($main: Cheerio<any>): void {
   $main.find("p > a.reference.download.internal").closest("p").remove();
 }
 
+function removePublicMembersRubric($main: Cheerio<any>): void {
```
Eric-Arellano, line 161: Can you please add docstring explaining why we do this? Maybe mention the better solution is inlining headers and link to that issue.

frankharkins: Good idea, thanks

Issue #2811, opened 2025-03-19
Qiskit 2.0 - updates following the circuit library modernization


ShellyGarion:
### URL to the relevant documentation

Following the circuit library modernization in https://github.com/Qiskit/qiskit/issues/13046 we should update the relevant guides and tutorials.

- [x] https://github.com/Qiskit/documentation/blob/main/docs/guides/ai-transpiler-passes.mdx
EfficientSU2 --> efficient_su2

- [x] https://github.com/Qiskit/documentation/blob/main/docs/guides/circuit-library.ipynb 
TwoLocal --> n_local or efficient_su2
ZZFeatureMap --> zz_feature_map
QuantumVolume --> quantum_volume
CDKMRippleCarryAdder --> HalfAdderGate or FullAdderGate

- [x] https://github.com/Qiskit/documentation/blob/main/docs/guides/defaults-and-configuration-options.ipynb
GroverOperator --> grover_operator
Diagonal --> DiagonalGate

- [x] https://github.com/Qiskit/documentation/blob/main/docs/guides/dynamical-decoupling-pass-manager.ipynb
EfficientSu2 --> efficient_su2

- [x] https://github.com/Qiskit/documentation/blob/main/docs/guides/get-started-with-primitives.ipynb
QAOAAnsatz --> qaoa_ansatz
EfficientSU2 --> efficient_su2

- [x] https://github.com/Qiskit/documentation/blob/main/docs/guides/local-testing-mode.ipynb
EfficientSU2 --> efficient_su2

- [x] https://github.com/Qiskit/documentation/blob/main/docs/guides/primitives-examples.ipynb
IQP --> iqp
RealAmplitudes -> real_amplitudes

- [ ] #2870
IQP --> iqp

- [x] https://github.com/Qiskit/documentation/blob/main/docs/guides/qiskit-transpiler-service.mdx
EfficientSU2 --> efficient_su2

- [x] https://github.com/Qiskit/documentation/blob/main/docs/guides/represent-quantum-computers.ipynb
EfficientSU2 --> efficient_su2

- [x] https://github.com/Qiskit/documentation/blob/main/docs/guides/simulate-stabilizer-circuits.ipynb
EfficientSu2 --> efficient_su2

- [x] https://github.com/Qiskit/documentation/blob/main/docs/guides/simulate-with-qiskit-aer.ipynb
EfficientSu2 --> efficient_su2

- [ ] #2871
ExcitationPreserving --> excitation_preserving
[...]

PR #2852, opened 2025-03-25
Update circuits in guides

ShellyGarion:
See #2811


docs/guides/qunova-chemistry.ipynb
Cryoris, line 1: Since this is documentation of a qiskit function they are probably still using the old objects under the hood. It might be better not to change this and instead tell the authors to update the function and the guide jointly

ShellyGarion: OK, I'll remove this change from here

docs/guides/q-ctrl-performance-management.ipynb
```diff
@@ -173,14 +173,14 @@
...
-    "from qiskit.circuit.library import IQP\n",
+    "from qiskit.circuit.library import iqp\n",
     "from qiskit.quantum_info import random_hermitian, SparsePauliOp\n",
     "\n",
     "n_qubits = 50\n",
     "\n",
     "# Generate a random circuit\n",
     "mat = np.real(random_hermitian(n_qubits, seed=1234))\n",
-    "circuit = IQP(mat)\n",
+    "circuit = iqp(mat)\n",
```
ElePT, line 183: this guide is also maintained externally like the qunova one. While I don't think the change would break anything, I would apply the same suggestion that Julien left in the qunova guide just to be safe.

ShellyGarion: done in [274bafa](https://github.com/Qiskit/documentation/pull/2852/commits/274bafa2839edcc320b0503b7b9425bba81dc008)

Cryoris (approved):
LGTM thanks! I didn't verify the circuits are still drawn the same, but GitHub says the notebook outputs are unchanged, so it seems correct 👍🏻

Eric-Arellano:
Thank you for leading this initiative, Shelly!

PR #2860, opened 2025-03-25
Adjust headings in API components

frankharkins:
This PR adjusts headings in API components to keep the correct information hierarchy on the page.

For example, if the title of a component is `<h3>`, then all headings inside the component should be `<h4>` or greater. This change uses the existing `getHeaderLevel` function to work out the level of the title heading, then bumps each heading inside in the component up so that the minimum heading level inside the component is one greater than the title heading level.

We need this for the C API; the docstrings are written in markdown and without this, we'd need writers to always remember to set the correct heading level. It also seems to fix a page in the dev version of Qiskit Python.

You may find it easier to review commit-wise:

1. b4775990531a270377626a427f7197f1cec517a7 – implements the logic change
2. b2c8bdc9b2aafd085679a1b7e91715f4e1b3d114 – just moving code
3. 901861386c529c525d6463b7f358875bb9a31773 – minor refactoring of the code
4. 02b01e1017ed3fe318278ef04505416eff17d5fe – the result of running `npm run regen-apis`, resetting the images, and squashing the commits.


scripts/js/lib/api/processHtml.ts
```diff
@@ -511,3 +516,83 @@ function hasPropertyModifier($dl: Cheerio<any>): boolean {
...
+function setMinimumHeadingLevel(
+  $: CheerioAPI,
+  $dl: Cheerio<any>,
+  minLevel: number,
+): void {
+  const detectedHeadingLevels = [1, 2, 3, 4, 5, 6].filter(
+    (level) => $dl.find(`h${level}`).length !== 0,
+  );
+  if (detectedHeadingLevels.length === 0) return;
+  const currentMinimumLevel = Math.min(...detectedHeadingLevels);
```
arnaucasau, line 579: Given that `.filter()` preserves the order, maybe we can just return the first one?

scripts/js/lib/api/processHtml.ts
```diff
@@ -419,18 +419,23 @@ export async function processMembersAndSetMeta(
...
-        apiType!,
+        apiType,
         id,
+        headerLevel,
         options,
       );
-      $dl.replaceWith(
-        `<div>${openTag}\n${bodyElements.join("\n")}\n${closeTag}</div>`,
-      );
+      const $body = $(`<div>${bodyElements.join("\n")}</div>`);
```
arnaucasau, line 433: Can we rename the variable to something like `$bodyElements` or `$componentBody`? I think `$body` could be mistaken by the having something to do with the HTML `<body>` tag

scripts/js/lib/api/processHtml.ts
```diff
@@ -419,18 +419,23 @@ export async function processMembersAndSetMeta(
...
+      const minHeadingLevel = openTag.includes("isDedicatedPage='true'")
```
arnaucasau, line 434: Could we have a comment saying that we are looking for a prop here? Or maybe we can generalize this in a helper function that lets us find any combination of prop/value (no need to do the latter though).

docs/api/qiskit/dev/qiskit.quantum_info.SparseObservable.mdx
```diff
@@ -93,65 +93,65 @@ $$
 
     The numeric structure of these is that they are all four-bit values of which the low two bits are the (phase-less) symplectic representation of the Pauli operator related to the object, where the low bit denotes a contribution by $Z$ and the second lowest a contribution by $X$, while the upper two bits are `00` for a Pauli operator, `01` for the negative-eigenstate projector, and `10` for the positive-eigenstate projector.
 
-    ### Values
```
arnaucasau, line 96: Interesting! This seems to be documented in the [rust code](https://github.com/Qiskit/qiskit/blob/6ce77887008c181f74306887ec494fedd2911a56/crates/accelerate/src/sparse_observable/mod.rs#L2149)

arnaucasau (approved):
Very nice! This change will make the C docs look much better. Good idea having logic here rather than imposing the devs to manually bump the headings

scripts/js/lib/api/processHtml.ts
frankharkins: I agree this isn't particularly neat, what do you think about dc813a2ac841dfd438d295bc75b96dd7b6a52161? Basically moving the assembling of the component into `createMdxComponent` where we have access to the `isDedicatedPage` variable so we don't have to re-parse the tag string.

scripts/js/lib/api/processHtml.ts
frankharkins: Good idea, thanks!

scripts/js/lib/api/processHtml.ts
arnaucasau: I like it a lot, thank you for the change! I think this will make the addition of links in the signatures much cleaner too given that everything is in the same place (we can just add the signatures inside `bodyElements` after a small post-process)

arnaucasau (approved):

scripts/js/lib/api/generateApiComponents.ts
```diff
@@ -34,25 +34,28 @@ export type ComponentProps = {
   isDedicatedPage?: boolean;
 };
 
-export async function processMdxComponent(
+export async function createMdxComponent(
```
Eric-Arellano, line 37: Nice rename!

scripts/js/lib/api/generateApiComponents.ts
```diff
@@ -465,3 +497,42 @@ function getPriorHeaderLevel(
...
+/**
+ * Sets the minimum heading level for the elements children, increasing each
+ * heading if needed to preserve the hierarchy.
+ *
+ * For example, `level=3` will convert:
+ * `<h1>Heading</h1><h2>Subheading</h2>`
+ * to
+ * `<h3>Heading</h3><h4>Subheading</h4>`
+ */
+function setMinimumHeadingLevel(
```
Eric-Arellano, line 510: What do you think of writing a unit test for this? While our generated docs give some assurance, the docs changes can be so enormous that a regression can easily slip in.

Eric-Arellano (commented):
Thanks for diving into this!! It's intricate code and I appreciate you improving it further.

It's probably worth regenerating all the docs given the latest commits? Ensure this didn't accidentally break something.

scripts/js/lib/api/generateApiComponents.ts
frankharkins: Added in 90020285d110ad103bb6fd5b614b08233d605410

PR #2861, opened 2025-03-26
Fix C API metadata handling

frankharkins:
Currently, we're looking for an ID attribute starting with `group__` to work out if a page is a module or not. This isn't very robust because the `group__` ID only appears if there's a `doxygengroup` in the page, which isn't always the case.

Instead, this PR treats all pages as modules except the index page and the release notes page, which we identify by their `<h1>`.

### DIFF
--- a/scripts/js/lib/api/processHtml.ts
+++ b/scripts/js/lib/api/processHtml.ts
@@ -47,7 +47,6 @@ export async function processHtml(
     determineGithubUrl,
     releaseNotesTitle,
     hasSeparateReleaseNotes,
-    isCApi,
   } = options;
   const $ = load(html);
   const $main = $(`[role='main']`);
@@ -55,6 +54,7 @@ export async function processHtml(
   const isReleaseNotes =
     fileName.endsWith("release_notes.html") ||
     fileName.endsWith("release-notes.html");
+  const isIndex = fileName.endsWith("index.html");
   const images = loadImages(
     $,
     $main,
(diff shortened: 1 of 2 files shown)


scripts/js/lib/api/processHtml.ts
```diff
@@ -439,9 +439,20 @@ export function maybeSetModuleMetadata(
...
   meta: Metadata,
+  isReleaseNotes: boolean,
   options: { isCApi: boolean },
 ): void {
-  const modulePrefix = options.isCApi ? "group__" : "module-";
+  if (options.isCApi) {
+    // Every page in the C API should be displayed as a module except the index
+    // and the release notes.
+    const topHeadingText = $main.find("h1").first().text();
+    const isIndex = topHeadingText === "Qiskit C API (qiskit.h)";
```
Eric-Arellano, line 449: Rather than hardcoding this, is it possible to look at the page name? It's brittle to have a magic value like this which the Qiskit SDK team could change without us realizing.

Generally, consider refactoring this to separate out C from Python, given how different they are. Like rename it to be `maybeSetPythonModuleMetadata` and `setCMetadata` etc. The two implementations are now quite different.

frankharkins: Filename is a good call, I didn't forgot that's what we did for release notes. Done both, thanks

arnaucasau (approved):
Awesome! Thanks!

Eric-Arellano (approved):
🚀

PR #2868, opened 2025-03-27
Remove empty headings for Functions without ID

arnaucasau:
Closes https://github.com/Qiskit/documentation/issues/2792

`qiskit-addon-aqc-tensor` is using the "multiple dispatch" paradigm which results in Sphinx nesting functions inside functions. In these cases, the nested functions don't have any ID defined. 

Here is an example of how Sphinx renders the `tensornetwork_from_circuit` function:

<details><summary>Screenshot</summary>
Screenshot 2025-03-27 at 11 26 45

</details>

This PR changes the MDX pipeline to stop assuming that every function will have an ID defined. If the ID is not defined the pipeline won't create a new empty heading matching that way the Sphinx output.

<details><summary>Screenshot of the current live docs</summary>
Screenshot 2025-03-27 at 11 33 29

</details>

<details><summary>Screenshot how the docs will look after this PR</summary>
Screenshot 2025-03-27 at 11 33 04

</details>

### DIFF
--- a/scripts/js/lib/api/generateApiComponents.ts
+++ b/scripts/js/lib/api/generateApiComponents.ts
@@ -329,14 +329,16 @@ function prepareFunctionProps(
     };
   }
 
-  if (id) {
-    const name = getLastPartFromFullIdentifier(id);
-    const htag = `h${headerLevel}`;
-    $(
-      `<${htag} data-header-type="method-header">${name}</${htag}>`,
-    ).insertBefore($dl);
-  }
+  // The Function ID could not exist if the function is nested inside another function.
+  // Nested functions could appear for example when using the multiple dispatch paradigm
+  // by using the python `@dispatch` decorator.
+  if (!id) return props;
 
+  const name = getLastPartFromFullIdentifier(id);
+  const htag = `h${headerLevel}`;
+  $(`<${htag} data-header-type="method-header">${name}</${htag}>`).insertBefore(
+    $dl,
+  );
   return props;
 }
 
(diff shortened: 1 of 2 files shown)


scripts/js/lib/api/generateApiComponents.ts
```diff
@@ -328,11 +328,14 @@ function prepareFunctionProps(
       isDedicatedPage: true,
     };
   }
-  const name = getLastPartFromFullIdentifier(id);
-  const htag = `h${headerLevel}`;
-  $(`<${htag} data-header-type="method-header">${name}</${htag}>`).insertBefore(
-    $dl,
-  );
+
+  if (id) {
```
arnaucasau, line 332: I decided to take this approach because:

1. We are already doing a conditional creation of the headings for methods:

https://github.com/Qiskit/documentation/blob/4bbb2ece8fc682891554676884f4d6205157ad7c/scripts/js/lib/api/generateApiComponents.ts#L212

2. Creating the headings by using the signatures name will result in a TOC with duplicated entries
<details><summary>Example of a TOC</summary>
Screenshot 2025-03-27 at 11 47 35

</details>

However, as an alternative, we could still use the signature’s name to create the heading regardless of the noisy TOC.

<details><summary>Example of how the docs will look like</summary>
Screenshot 2025-03-27 at 11 42 24

</details>

What do you think? This change only affects to one page in `qiskit-addon-aqc-tensor`

Eric-Arellano (commented):
Thank you! This is definitely the best approach. 

My only comment is if you can please add a comment explaining when we expect `id` to not be set? Also it may be clearer to invert the conditional, so your comment is focused on when `(!id)` is possible

Eric-Arellano (approved):

PR #2873, opened 2025-03-27
Attempt cross references in C

frankharkins:
This is an attempt to enable cross references in the C API docs.

## Background

The source docstrings refer to other Rust objects using square brackets. For example:

```rust
/// A [CSparseTerm] contains pointers to the indices and bit terms in the term, which
/// can be used to modify the internal data of the observable. This can leave the observable
```

[Link to source](https://github.com/Qiskit/qiskit/blob/ddb802506a499385caf436522b6fb277b51895c1/crates/cext/src/sparse_observable.rs#L299-L302).

The references link correctly in the developers' editors. The problem is that cbindgen doesn't know to convert these references to their C equivalents. Since Breathe doesn't recognise this syntax the original string ends up on the final web page. (Breathe doesn't even know Rust is involved – it just reads the header file generated by cbindgen).

The mapping of Rust names to C names is specified in Qiskit's `cbindgen.toml`:

```toml
[export.rename]
"SparseObservable" = "QkObs"
"CSparseTerm" = "QkObsTerm"
"BitTerm" = "QkBitTerm"
"Complex64" = "QkComplex64"
```

[Link to source](https://github.com/Qiskit/qiskit/blob/ddb802506a499385caf436522b6fb277b51895c1/crates/cext/cbindgen.toml#L30-L34).

## Propsed solution

I'm not convinced this is the best solution, but it does seem to work for now.

Using the `cbindgen.toml` mapping as a reference, we add the mapping of Rust name -> C name and link to a typescript file in this repo. We then search the C API markdown for possible references and insert the correct names/links ourselves.

The main problem with this is we'd need to manually update the config file. We could potentially grab the mappings automatically from `cbindgen.toml`, but we won't know which pages to link to. We might be able to get the pages from `objects.inv`, but that feels very complicated.

### Other considerations

Creating links to source code will also suffer from similar problems.

### DIFF
--- a/scripts/js/lib/api/Pkg.ts
+++ b/scripts/js/lib/api/Pkg.ts
@@ -14,6 +14,7 @@ import { join } from "path/posix";
 
 import { findSeparateReleaseNotesVersions } from "./releaseNotes.js";
 import { determineHistoricalQiskitGithubUrl } from "../qiskitMetapackage.js";
+import { insertCrossRefsVisitor } from "./handleCrossRefs";
 import {
   TocGrouping,
   QISKIT_TOC_GROUPING,
@@ -267,6 +268,11 @@ export class Pkg {
 
   isCApi(): boolean {
     return this.language === "C";
+  }
+
+  crossRefsVisitorFn(): any {
+    if (!this.isCApi()) return null;
+    return insertCrossRefsVisitor(this.name, this.version);
   }
 
   isProblematicLegacyQiskit(): boolean {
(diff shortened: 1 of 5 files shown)


Eric-Arellano:
> The main problem with this is we'd need to manually update the config file. We could potentially grab the mappings automatically from cbindgen.toml, but we won't know which pages to link to. We might be able to get the pages from objects.inv, but that feels very complicated.

Yeah, I would only want to do this if we could automate the process like pulling `cbindgen.toml` from the Box artifact.

Regarding not knowing which pages to link to, is it viable to intervene between cbindgen -> Breathe? Have some script that processes the cbindgen output to use RST syntax for cross-references, which Breathe will then understand?

I know you said it wasn't viable to directly put the RST in the Rust code.

frankharkins:
Supposedly it is possible https://breathe.readthedocs.io/en/latest/markups.html#inline-rst, but I can't get it to work – the markup always appears in the final HTML (not parsed into a cross-reference).

Issue #2879, opened 2025-03-28
API generation should fail when patch version not specified


Eric-Arellano:
I tried using `-v 0.2` recently with `gen-api`, when it needs to be `-v 0.2.0`. @arnaucasau caught this, which is great, but our tooling should error instead.

PR #2888, opened 2025-03-31
Error when API versions are incomplete

arnaucasau:
Closes https://github.com/Qiskit/documentation/issues/2879

This PR adds a mechanism to catch when a version is incomplete when running `npm run gen-api`

### DIFF
--- a/scripts/js/commands/api/updateApiDocs.ts
+++ b/scripts/js/commands/api/updateApiDocs.ts
@@ -15,7 +15,7 @@ import { hideBin } from "yargs/helpers";
 
 import { Pkg } from "../../lib/api/Pkg.js";
 import { zxMain } from "../../lib/zx.js";
-import { parseMinorVersion } from "../../lib/apiVersions.js";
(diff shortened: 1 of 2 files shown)


scripts/js/commands/api/updateApiDocs.ts
```diff
@@ -46,6 +46,14 @@ const readArgs = (): Arguments => {
       type: "string",
       demandOption: true,
       description: "The full version string of the --package, e.g. 0.44.0",
+      coerce: (version) => {
```
arnaucasau, line 49: By using `coerce`, we can transform the version in the future if we want. I noticed that we sometimes tend to run `npm run gen-api -- -p <package> -v <major>.<minor>` without the patch when the patch version is 0 (e.g. qiskit 2.0.0). With this set up, we could make assumptions and transform the version if needed.

Also, the errors inside `coerce` are treated as a validation failure (Documentation [here](https://yargs.js.org/docs/)).

frankharkins (approved):
Thanks!

PR #2891, opened 2025-03-31
Test code snippets from mdx files in CI

kaelynj:
We currently have no testing infrastructure for the code snippets in our markdown files. This PR will work on generating executable scripts from these code snippets and incorporate them into our CI pipeline.

A few things will be needed:

- [ ] Script to generate `.py` files
- [ ] A means to monkey patch specific files (likely using nb-tester)
- [ ] Add the related scripts to our fornightly cron job


Eric-Arellano:
We decided to instead convert more MDX files to Jupyter notebooks. That avoids a lot of issues with testing MDX, like how we would need to invent a syntax to say that a code cell should not be tested. https://github.com/Qiskit/documentation/issues/2928

Issue #2903, opened 2025-03-31
Clean up link ignores post-Qiskit 2.0


Eric-Arellano:
https://github.com/Qiskit/documentation/blob/494283300b58d54340da9ccf6daa90e22fc7b170/scripts/js/lib/links/ignores.ts#L20-L22

https://github.com/Qiskit/documentation/blob/494283300b58d54340da9ccf6daa90e22fc7b170/scripts/js/lib/links/ignores.ts#L125-L129

Eric-Arellano:
All that remains is improving the ignores for C's object.inv @frankharkins. Some options:

1. Fix the bad entries entirely, such as removing them or pointing them to where we expect
2. Make our ignore regex more robust for future C versions, like using a more generic regex

PR #2907, opened 2025-04-01
Clean up link ignores

arnaucasau:
Part of https://github.com/Qiskit/documentation/issues/2903#event-17063149873

This PR removes the removed Qiskit 2.0 pages from the link checker ignores and adds them to files to load for our guides and runtime. We originally added to `QISKIT_REMOVED_PAGES` all the removed pages, but I ended up removing the pages that we don't make any reference in this repo.


scripts/js/lib/links/ignores.ts
```diff
@@ -119,14 +113,22 @@ function _runtimeObjectsInvRegexes(): string[] {
...
+  return [
+    "qk-obs-term#structqkobsterm",
+    "qk-obs-term#structqkobsterm_1a14ff1665641903565439ad9877fd2c8e",
+    "qk-obs-term#structqkobsterm_1a3e532a189705a89456f2d2b96d2ff759",
+    "qk-obs-term#structqkobsterm_1aaab7bf8363689e37e5c9a8131c54f674",
+    "qk-obs-term#structqkobsterm_1af1019513a1a00f578d33b38d10b36670",
+    "qk-obs-term#structqkobsterm_1af131526e747cdcac7ecc1dffc66b439c",
+    "qk-obs-term#structqkobsterm_1autotoc_md2",
+  ].map((path) => `\/api\/qiskit-c\/${path}$`);
+}
```
arnaucasau, line 126: @frankharkins FYI - These links were caught by the link checker. I found that the last one exists, but in camel case:
https://github.com/Qiskit/documentation/blob/0d83af1e0bccf3322c5cca94eefb90eb5557d09a/docs/api/qiskit-c/qk-obs-term.mdx?plain=1#L22

frankharkins: Good spot, that's fine; the links should still take users to the correct page. We can fix this after merging.

PR #2908, opened 2025-04-01
Pages render: Add instructions for platform specific pages

arnaucasau:
This PR adds instructions about how to fix `checkPagesRender` errors on platform specific files.

### DIFF
--- a/scripts/js/commands/checkPagesRender.ts
+++ b/scripts/js/commands/checkPagesRender.ts
@@ -134,7 +134,9 @@ zxMain(async () => {
       `💔 ${failures.length} pages crash when rendering. This is usually due to invalid syntax, such as forgetting ` +
         "the closing component tag, like `</Admonition>`. You can sometimes get a helpful error message " +
         "by previewing the docs locally or in CI. See the README for instructions.\n\n" +
-        failures.join("\n"),
+        failures.join("\n") +
+        "\n\nIf your files are platform specific, you should add them to the `LEGACY_ONLY_PAGES` or `CLOUD_ONLY_PAGES` list " +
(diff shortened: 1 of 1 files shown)


scripts/js/commands/checkPagesRender.ts
```diff
@@ -134,7 +134,9 @@ zxMain(async () => {
       `💔 ${failures.length} pages crash when rendering. This is usually due to invalid syntax, such as forgetting ` +
         "the closing component tag, like `</Admonition>`. You can sometimes get a helpful error message " +
         "by previewing the docs locally or in CI. See the README for instructions.\n\n" +
-        failures.join("\n"),
+        failures.join("\n") +
+        "\n\nIf your files are platform specific, you should add them to the `LEGACY_ONLY_PAGES` or `CLOUD_ONLY_PAGES` list " +
+        "at the beginning of `scripts/js/commands/checkPagesRender.ts`",
```
arnaucasau, line 139: Thoughts on moving this two arrays to a more easy location for everyone to modify? E.g. `scripts/config`

Eric-Arellano: That's a good idea. Generally I like the idea of moving things to `scripts/config`. Maybe use TS files rather than JSON files so that we can still have comments and variables to DRY things? You could timebox that to ~20 minutes and switch to JSON if TS is not working.

Eric-Arellano: Actually, let's not spend time on this, please. The config is going away in 2 months.

Eric-Arellano (approved):
Thanks for improving this :)

PR #2921, opened 2025-04-02
Add missing `version-info` cells to notebooks

frankharkins:
This PR adds empty markdown cells with `version-info` tags to any notebooks that don't have them. These are currently invisible to the user, but they'll be populated with the version information when the notebooks are next executed.

Also adds a check to make sure we add them in the future.


Eric-Arellano (commented):
Thanks for fixing! 

Can you please remove the new check in this PR and open a dedicated PR with it? That way we can review it more closely without all the content owners getting notifications. Other than that, this looks great.

PR #2922, opened 2025-04-02
Fix CI by pinning osqp

frankharkins:
Workaround for https://github.com/osqp/osqp/issues/728

### DIFF
--- a/scripts/nb-tester/requirements.txt
+++ b/scripts/nb-tester/requirements.txt
@@ -18,4 +18,4 @@ pyscf~=2.8.0; sys.platform != 'win32'
 networkx~=3.4.2
 
 # We pin to v1.0.1 due to https://github.com/osqp/osqp/issues/728
-osqd==1.0.1
+osqp==1.0.1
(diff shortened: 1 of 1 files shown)


scripts/nb-tester/requirements.txt
```diff
@@ -16,3 +16,6 @@ scipy~=1.15.2
 scikit-learn~=1.6.1
 pyscf~=2.8.0; sys.platform != 'win32'
 networkx~=3.4.2
+
+# We pin to v1.0.1 due to https://github.com/osqp/osqp/issues/728
+osqp==1.0.1
```
Eric-Arellano, line 21: Instead, more idiomatic is to add a constraints.txt file: https://github.com/Qiskit/qiskit/blob/99d0fa99cd81258946bf3973ac046a1442609eb6/tox.ini#L9

But maybe this is a good time to switch us to `uv` (probably instead of `tox`)? I'd love for us to use an actual lockfile. If you're interested in working on that, I'd timebox it to <3 hours

frankharkins: I'd be interested in looking at that, however I think we should merge this for now to unblock CI.

jabbera: @Eric-Arellano, just as an FYI uv actually has a worse experience here: https://github.com/astral-sh/uv/issues/12618. That said we use tox-uv with great success!

Eric-Arellano (approved):

Issue #2854, opened 2025-03-25
Speed up `./check` by only running on changed files


Eric-Arellano:
Follow up to https://github.com/Qiskit/documentation/issues/2800.

Add a mode that uses `git status` to only run based on changed files. This allows us to optimize:

- Only run on changed files:
    - spellcheck (4s)
    - image check (1.8s)
    - metadata (0.4s)
    - Squeaky
    - Ruff
    - notebook post-process
- Only run if new files added
    - qiskit-bot (0.3s)
    - orphan (0.5s)
    - patterns index (0.3s)
- Won't speed up because not possible
    - link checker (6s)

TBD what we want the default behavior to be:

* default to changed files; use `--all` (`-a`) for all
* default to all files; use `--changed` (`-c`) for changed files

Note that CI will always run on all files. So long as we can trust that CI is working and we don't merge on red, then it's safe to use `--changed`.

PR #2929, opened 2025-04-02
Check all notebooks contain `version-info` cells

frankharkins:
Checks that all notebooks have a markdown cell with tag `version-info`. We add these cells so the notebook tester knows where to insert package version information so users can see which package versions were used to run the notebook.

### DIFF
--- a/scripts/ci/check-for-version-info-cells.py
+++ b/scripts/ci/check-for-version-info-cells.py
@@ -22,25 +22,20 @@ guides_path = Path("./docs/guides")
 for notebook in guides_path.glob("*.ipynb"):
     nb = nbformat.read(notebook, as_version=4)
     num_version_info_cells = len(
-        list(
-            filter(
-                lambda cell: "version-info" in cell.metadata.get("tags", []),
-                nb.cells,
-            )
-        )
+        [cell for cell in nb.cells if "version-info" in cell.metadata.get("tags", [])]
     )
     if num_version_info_cells == 1:
         continue
     if num_version_info_cells == 0:
         print(
-            f"\n\n❌ No version info cell found in {notebook}; add an empty markdown "
-            "cell immediately after the top-level heading and add a "
-            "`version-info` tag. This is so the package versions will be added "
-            "to the notebook next time it's executed.\n\n"
+            f"\n\n❌ No version info cell found in {notebook}. (This allows users to "
+            "see what versions of requirements were used when developing the notebook.)\n\n"
+            "To fix, add a new markdown cell immediately after the top-level heading. Keep its content "
+            "empty. In the `tags` metadata, add 'version-info'.\n\n"
         )
     else:
         print(
-            f"\n\n❌ Found more than one cell with tag `version-info` in {notebook}; "
-            "remove all but one.\n\n"
+            f"\n\n❌ Found more than one cell with tag `version-info` in {notebook}. "
+            "To fix, remove all but one.\n\n"
         )
     sys.exit(1)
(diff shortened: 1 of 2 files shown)


scripts/ci/check-for-version-info-cells.py
```diff
@@ -0,0 +1,46 @@
...
+
+"""
+Check all notebooks have exactly one `version-info` cell.
+"""
+
+from pathlib import Path
+import nbformat
+import sys
+
+guides_path = Path("./docs/guides")
```
Eric-Arellano, line 21: We should look for all notebooks, such as tutorials. I recommend `Path("./docs").rglob("*.ipynb")`

scripts/ci/check-for-version-info-cells.py
```diff
@@ -0,0 +1,46 @@
...
+for notebook in guides_path.glob("*.ipynb"):
+    nb = nbformat.read(notebook, as_version=4)
+    num_version_info_cells = len(
+        list(
+            filter(
+                lambda cell: "version-info" in cell.metadata.get("tags", []),
+                nb.cells,
+            )
+        )
```
Eric-Arellano, line 30: Nit/FYI a list comprehension is more idiomatic in Python because it's seen as more readable

```suggestion
        [cell for cell in nb.cells if "version-info" in cell.metadata.get("tags", [])]
```

scripts/ci/check-for-version-info-cells.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+    if num_version_info_cells == 1:
+        continue
+    if num_version_info_cells == 0:
+        print(
+            f"\n\n❌ No version info cell found in {notebook}; add an empty markdown "
+            "cell immediately after the top-level heading and add a "
+            "`version-info` tag. This is so the package versions will be added "
+            "to the notebook next time it's executed.\n\n"
```
Eric-Arellano, line 39:
```suggestion
            f"\n\n❌ No version info cell found in {notebook}. (This allows users to "
            "see what versions of requirements were used when developing the notebook.)\n\n"
            "To fix, add a new markdown cell immediately after the top-level heading. Keep its content "
            "empty. In the `tags` metadata, add 'version-info'.\n\n"
```

scripts/ci/check-for-version-info-cells.py
```diff
@@ -0,0 +1,46 @@
...
+        )
+    else:
+        print(
+            f"\n\n❌ Found more than one cell with tag `version-info` in {notebook}; "
+            "remove all but one.\n\n"
```
Eric-Arellano, line 44:
```suggestion
            f"\n\n❌ Found more than one cell with tag `version-info` in {notebook}. "
            "To fix, remove all but one.\n\n"
```

scripts/ci/check-for-version-info-cells.py
Eric-Arellano, line 1: Nit that the convention in Python is to use a main guard and `main()` function. https://docs.python.org/3/library/__main__.html#idiomatic-usage tl;dr is that it allows other modules and e.g. test runners to import your file without executing code. 

While this script is pretty simple and we don't plan to test it, it's a good convention for us to be consistent with. (Also the script could maybe get more complex once we add learning)

Eric-Arellano (commented):
Thanks! Some high level things that seem sensible to me but it's worth making explicit:

1. This eagerly errors on the first notebook failure rather than collecting all notebooks. Given that this check is only relevant to _new_ files, I think that makes sense. Usually you only add one new file at a time.
2. This will be part of `./check`. I think that's worth it, unless this is unusually slow. But it'd be great to eventually have https://github.com/Qiskit/documentation/issues/2854 so that we only run this check when we detect a new notebook was added

scripts/ci/check-for-version-info-cells.py
frankharkins: Are you sure? I think we should check with the tutorials team before we enforce this on them

Eric-Arellano: That's fair

Eric-Arellano (approved):

PR #2930, opened 2025-04-02
Rename `public/` folder to `public/docs`

Eric-Arellano:
This is prep for Learning migrating to this repository and Next.js app. For that to happen, our links are going to need to start having a `/docs` prefix, like `/docs/images` rather than `/images`.

This PR renames the `public/` folder to `public/docs`, and adjusts our tooling to handle the change. We do _not_ yet start using the new links of `/docs/images` rather than `/images`, but that will be done in a follow up. This PR is only meant to split up the change and allow content writers to start organizing files in the appropriate place.

The closed source app will copy the `public/docs` folder to live at `public/`, like how this PR updates our local & PR previews to strip the `/docs` prefix.


Eric-Arellano:
The commits are distinct

scripts/js/commands/checkInternalLinks.ts
```diff
@@ -77,7 +77,7 @@ async function main() {
 
   const fileBatches = await determineFileBatches(args);
   const otherFiles = [
-    ...(await globby("public/{images,videos,open-source}/**/*")).map(
+    ...(await globby("public/docs/{images,videos,open-source}/**/*")).map(
```
Eric-Arellano: This file will need to be updated when we add `public/learning`.

scripts/js/commands/checkStaleImages.ts
```diff
@@ -15,8 +15,11 @@ import { globby } from "globby";
 
 import { zxMain, isInstalled } from "../lib/zx.js";
 
+// When users link to this folder, they leave off this prefix.
+const PUBLIC_FOLDER_PREFIX = "public/docs/";
```
Eric-Arellano: This will need to be rethought when we have `public/learning`

scripts/js/lib/links/InternalLink.ts
```diff
@@ -66,7 +66,7 @@ export class InternalLink {
       this.value.startsWith("/videos") ||
       this.value.endsWith(".pdf")
     ) {
-      return [path.join("public/", this.value)];
+      return [path.join("public/docs/", this.value)];
```
Eric-Arellano: Once the link is `/docs/images`, we'll go back to the old code

scripts/js/lib/markdownImages.ts
```diff
@@ -43,7 +43,7 @@ export async function collectInvalidImageErrors(
             ".avif",
           );
           imagesErrors.add(
-            `Convert '${imageName}' to AVIF. You can use the command \`magick public${node.url} public${urlWithAvifExtension}\`. ` +
+            `Convert '${imageName}' to AVIF. You can use the command \`magick public/docs${node.url} public/docs${urlWithAvifExtension}\`. ` +
```
Eric-Arellano: I don't think we'll be able to give this precise command in the future when we add Learning

scripts/pr-previews/builder.py
```diff
@@ -120,13 +120,13 @@ def _copy_local_content(root_dir: Path) -> None:
...
-        "public/images/optimize",
-        "public/images/qiskit-patterns",
+        "public/docs/videos",
+        "public/docs/images/guides",
+        "public/docs/images/optimize",
+        "public/docs/images/qiskit-patterns",
     ]:
         dest = (
-            root_dir / "packages/preview" / dir
+            root_dir / "packages/preview" / dir.replace("/docs", "")
```
Eric-Arellano: This is temporary

start
```diff
@@ -47,7 +47,7 @@ def main() -> None:
         f"{PWD}/docs:/home/node/app/docs/en",
         *skip_apis(),
         "-v",
-        f"{PWD}/public:/home/node/app/packages/preview/public",
+        f"{PWD}/public/docs:/home/node/app/packages/preview/public",
```
Eric-Arellano: This is temporary

arnaucasau (commented):
Awesome! To fix CI you'll need to update this line as well to be `public/docs`:

https://github.com/Qiskit/documentation/blob/7d5a193ef504c8ddfa023483a792e7e20bbc8bd0/scripts/js/lib/links/extractLinks.ts#L104

Eric-Arellano:
Thank you, @arnaucasau! 

> Also, I'm not super familiar with checkInternalLinksFromOtherRepo.ts

I skipped updating it because I'm not sure yet how we are going to set up private repositories. We don't have any in use currently for Functions. We will soon with Tutorials, but I figure we can update this as part of the issue to get Tutorials working. I want us to first switch over how links work to have the `/docs` prefix.

PR #2933, opened 2025-04-03
Unpin osqp

frankharkins:
We pinned this temporarily to unblock CI, but it shouldn't be needed any more.


arnaucasau (approved):
Thanks!

Eric-Arellano (commented):
Thanks for following up on this! You saved me from creating a tech debt issue today 😅

PR #2943, opened 2025-04-04
Fix spellcheck CI

frankharkins:
Turns out the spellcheck won't block CI if it finds an error in an MDX file, only notebooks.


Eric-Arellano (approved):
Oof. Thanks!!

Issue #2914, opened 2025-04-01
Spellcheck should run on table of contents


Eric-Arellano:
A high-profile release was delayed due to a bad typo in _toc.json. Our spellchecker isn't configured to run on _toc.json, but it could be.

Some thoughts:

* This becomes less necessary if we do https://github.com/Qiskit/documentation/issues/2489. It's still useful for divergences, which do happen. But 2489 is probably more important
* This will slightly slow down the spellcheck performance. https://github.com/Qiskit/documentation/issues/2854 would mitigate that

PR #2944, opened 2025-04-04
Refactor checkSpelling.ts

Eric-Arellano:
Prework for https://github.com/Qiskit/documentation/issues/2914. This makes the code more generic so that we can add more content types than MDX and Jupyter.

### DIFF
--- a/scripts/js/commands/checkSpelling.ts
+++ b/scripts/js/commands/checkSpelling.ts
@@ -44,53 +44,68 @@ function readArgs(): Arguments {
 
 zxMain(async () => {
   const args = readArgs();
-  const cspellCmd = ["npx", "cspell", "--no-progress", "--config", args.config];
-  let allGood = true;
-
-  try {
-    await $`${cspellCmd} docs/**/*.mdx !docs/api/**/*.mdx`.pipe(process.stdout);
-    if (args.apis) {
-      await $`${cspellCmd} docs/api/**/*.mdx`.pipe(process.stdout);
-    }
-  } catch (p) {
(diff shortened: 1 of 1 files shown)


frankharkins (changes requested):
This doesn't seem to work; I tried introducing a spelling mistake but `npm run check:spelling` still passes 🤔 

Screenshot 2025-04-07 at 12 33 33


EDIT: The same thing happens on main too 😬

arnaucasau:
I also reproduce what Frank said in the main branch. I found that if you run the `npx cspell --no-progress --config scripts/config/cspell/cSpell.json` directly over a file it doesn't work either. It found my spell error when I removed the `--config` flag (and a bunch more which were expected because of the absence of our dictionaries).

My change was the following in `/docs/guides/install-qiskit-source.mdx`:

```diff
-Installing the Qiskit SDK from source allows you to access the current development version, instead of using the version in the Python Package Index (PyPI) repository. This lets you inspect and extend the latest version of the Qiskit code more efficiently.
+Instaling the Qiskit SDK from source allows you to access the current development version, instead of using the version in the Python Package Index (PyPI) repository. This lets you inspect and extend the latest version of the Qiskit code more efficiently.
``

Eric-Arellano:
Thank you for investigating! You found a major bug! Becky is going to fix this in https://github.com/Qiskit/documentation/pull/2954.

I believe this PR should be good to go to unblock https://github.com/Qiskit/documentation/issues/2914.

frankharkins (approved):
Thanks!

PR #2952, opened 2025-04-07
Unify Python and C API generation

### LINKED ISSUE #2863: Create formalized pipeline with Qiskit SDK repo for C API docs
We added C API docs in Qiskit 2.0, but have not yet set up the automated system in qiskit/qiskit for them to upload the C docs as an artifact (zip file) to GitHub Actions. We want them to upload their built docs, so that our team can then download them and upload them to Box.

We decided that we prefer for their CI to have a dedicated artifact for C docs, separate from Python docs. It is much simpler for our Sphinx->MDX pipeline if we treat the C docs as entirely independent of the Python docs.

### LINKED ISSUE #2900: Nightly dev docs sync should update C API
https://github.com/Qiskit/documentation/blob/main/.github/workflows/dev-docs-sync.yml should also work on the Qiskit C package.

Per https://github.com/Qiskit/documentation/issues/2863, we should reuse the same GitHub Artifact for both Python and C. api-html-artifacts.json should be updated for both packages.

frankharkins:
This PR combines the generation of Qiskit's Python and C APIs into a single pipeline. Both these APIs use the same artifact and should be generated at the same time.

***

Closes #2863 and closes #2900.

### DIFF
--- a/scripts/js/commands/api/updateApiDocs.ts
+++ b/scripts/js/commands/api/updateApiDocs.ts
@@ -92,21 +92,6 @@ export async function generateVersion(
   console.log(`Run pipeline for ${pkg.name}:${pkg.versionWithoutPatch}`);
   await runConversionPipeline(sphinxArtifactFolder, "docs", "public/docs", pkg);
   await generateHistoricalRedirects();
-
-  if (pkg.name === "qiskit") {
-    const majorVersion = Number(pkg.version.split(".").pop());
-    if (majorVersion >= 2) {
-      await generateVersion(
-        await Pkg.fromArgs(
-          "qiskit-c",
-          pkg.version,
-          pkg.versionWithoutPatch,
-          pkg.type,
-        ),
-        args,
-      );
-    }
-  }
 }
 
 export function determineMinorVersion(args: Arguments): string {
--- a/scripts/js/commands/api/updateApiDocs.ts
+++ b/scripts/js/commands/api/updateApiDocs.ts
@@ -92,6 +92,21 @@ export async function generateVersion(
   console.log(`Run pipeline for ${pkg.name}:${pkg.versionWithoutPatch}`);
   await runConversionPipeline(sphinxArtifactFolder, "docs", "public/docs", pkg);
   await generateHistoricalRedirects();
+
+  if (pkg.name === "qiskit") {
+    const majorVersion = Number(pkg.version.split(".").pop());
+    if (majorVersion >= 2) {
+      await generateVersion(
+        await Pkg.fromArgs(
+          "qiskit-c",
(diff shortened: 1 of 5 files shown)


frankharkins:
See https://github.com/Qiskit/documentation/pull/2957 for an example of this working

scripts/js/commands/api/updateApiDocs.ts
```diff
@@ -92,6 +92,21 @@ export async function generateVersion(
...
+        await Pkg.fromArgs(
+          "qiskit-c",
+          pkg.version,
+          pkg.versionWithoutPatch,
+          pkg.type,
+        ),
+        args,
+      );
+    }
+  }
```
Eric-Arellano, line 109: This might be annoying when generating exclusively Python when you're iterating on Python. But I'm unsure, I can see both sides. I'm curious what @arnaucasau thinks is the best workflow?

Eric-Arellano (commented):
Thanks!

Regardless of my comment, can you please add a brief section to the README at https://github.com/Qiskit/documentation?tab=readme-ov-file#generate-new-api-docs about "Qiskit SDK" where you explain that you should reuse the same artifact for both qiskit and qiskit-c? Then whatever the instruction will be to generate both, whether that's only running on `qiskit` or running both `qiskit` and `qiskit-c`

scripts/js/commands/api/updateApiDocs.ts
frankharkins: If it makes a difference, the C docs are very quick to generate so it shouldn't be too annoying

Eric-Arellano: Sgtm. This PR looks good then other than the README

frankharkins: While writing that up, I realised I'm pushing another thing to remember on to future us, and it shouldn't be too hard to have the C API use the qiskit entry in the json file. Gonna convert to draft while I test it out.

frankharkins:
Sorry for the last minute change, this is ready for review again

scripts/js/commands/api/updateApiDocs.ts
Eric-Arellano: Yeah, I agree with Arnau. My ideal would be:

1. Keep your unification of the HTML artifacts file. That's great!
2. Manually trigger creation of both packages

frankharkins: Thanks for the feedback, done in 588dd806efc725e4137351fa0a45b2e2246a7e8b

Eric-Arellano (approved):
Great work! I really like this final factoring.

PR #2960, opened 2025-04-08
Fix some broken external links

Eric-Arellano:
The changes in `ignores.ts` are valid links, but the link checker fails to ping them.


arnaucasau (approved):
Thanks!

Issue #2896, opened 2025-03-31
Fix Q-Ctrl and Qunasys notebooks


kaelynj:
### URL to the relevant documentation

https://quantum.cloud.ibm.com/docs/en/guides/qunasys-quri-chemistry

https://quantum.cloud.ibm.com/docs/en/guides/q-ctrl-optimization-solver

### Select all that apply

- [ ] typo
- [x] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

### Describe the fix.

While auditing our notebooks for 2.0 I noticed in [the last run](https://github.com/Qiskit/documentation/actions/runs/14135537468/job/39606259215) two of the functions notebooks error either b/c the problem is build incorrectly or parsing the results is misconfigured.

javabster:
cc @miamico @christopher-porter1

might also affect some tutorials and/or courses?

christopherporter1:
Does it affect this tutorial: https://learning.quantum.ibm.com/tutorial/compute-dissociation-curves-for-strong-coupling-systems-with-quna-sys-qsci

kaelynj:
Here are the relevant logs for the notebook failures: https://github.com/Qiskit/documentation/actions/runs/14135537468/job/39606259215#step:4:221

kaelynj:
> Does it affect this tutorial: https://learning.quantum.ibm.com/tutorial/compute-dissociation-curves-for-strong-coupling-systems-with-quna-sys-qsci

Yes @christopherporter1 I think it will. I think what ended up happening is that the `parameters` argument when calling `function.run()` has been split out into `molecule` and `circuit_options` which need to be specified.

kaelynj:
This issue will also be closed by #2970 and #2971

PR #2970, opened 2025-04-09
Fix arguments in job submission for Qunasys notebook

kaelynj:
Closes (partly) #2896 

An incorrect argument name when executing `function.run()` was causing two of the jobs in the examples to fail. This PR corrects the function call.


frankharkins (approved):
Thank you!!

PR #2971, opened 2025-04-09
Fix problem size for maxcut graph

kaelynj:
Closes (partly) #2896 

The first example in the Q-Ctrl otpimization solver created a graph with a problem size too large to run on the IBM fleet of QPUs. This PR fixes this issue.


frankharkins (approved):
Sorry I missed this. Looks great, thanks!

PR #2985, opened 2025-04-14
[Suggestion]: Reduce frequency of dev docs sync

frankharkins:
The dev docs sync runs every day and we often have a few dev docs PRs open at a time which means many are usually not merged. This PR reduces the frequency to run only three times a week, avoiding weekends.


arnaucasau (approved):
Good idea, thanks!

PR #2978, opened 2025-04-11
Add tutorials overview and TOC

frankharkins:
- **Add page and TOC**
- **Add page check**

PR #2965, opened 2025-04-09
Spellcheck table of contents

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/2914.

PR #2993, opened 2025-04-15
Do not spellcheck URLs

frankharkins:
#2965 added support to spellcheck TOCs, but this has the side-effect of also checking URLs.

Unfortunately, there are some words that are valid in a URL but shouldn't be added to the dictionary as they're not valid elsewhere. One example of this is `grovers`; usually this should be corrected to `Grover's`, but it _is_ valid in a URL. 

Since we can't add file-level ignores in the `_toc.json`, I think the best approach is to ignore these URLs using a regex.

This is not a problem on `main`, but it's blocking CI in #2978.


arnaucasau (approved):
Thanks!

PR #2998, opened 2025-04-15
fix import formatting

kevinsung:
it was bothering me


miamico (approved):
Haha right, we should have black and all the various tools to prettify before pushing!

Eric-Arellano:
> we should have black and all the various tools to prettify before pushing!

We do. We run Ruff with `./fix`, and CI also checks for formatting. Ruff is based on Black.

The issue is that Ruff/Black pay attention to when you add a comma at the end and think it was intentional. So, it will keep the import on a dedicated line.

PR #3011, opened 2025-04-17
Document docker system prune

Eric-Arellano:


arnaucasau (approved):
Thanks!

README.md
```diff
@@ -100,6 +100,16 @@ The preview application does not include the top nav bar. Instead, navigate to t
 
 Maintainers: when you release a new version of the image, you need to update the image digest in `./start` by following the instructions at the top of the file and opening a pull request.
 
+### Tip: periodically prune Docker
+
+Occasionally, Docker may fail due to running out of disk space. For example, you may encounter an error like this:
```
abbycross, line 105:
```suggestion
### Tip: Periodically prune Docker

Occasionally, Docker might fail when disk space runs low. For example, you might encounter an error like this:
```

abbycross: I think I see that it's Docker's space specifically, based on the last line

Eric-Arellano: Docker runs out of space. I still had 600GB on my machine but Docker was giving the error about not enough space. Does that change your recommendation?

abbycross (approved):

PR #3034, opened 2025-04-22
Update fractional gates with note on how angle constraints are considered

### LINKED ISSUE #2479: Consider updating the fractional gates tutorial
Regarding this tutorial: https://github.com/Qiskit/documentation/blob/main/docs/guides/fractional-gates.ipynb
We recently added a new feature: transpilation of circuits with a backend that supports fractional gates will automatically transform the circuits, such that all `rzz` numeric angles are in the valid range of `0 - pi/2`.

I think we should add a comment about it in the tutorial, because:
1. Users may be confused that their circuit ran successfully, in spite of invalid angles therein.
2. To inform users about a modification of their circuit during transpilation.

But I also hesitate, having the concern that this will make the tutorial too detailed and complex. In particular note that the transformation applies to numeric angles but not to parametric ones.

Discussion a bit of discussion in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2043.

cc: @nkanazawa1989 @wshanks @kaelynj

kaelynj:
Closes #2479 

I've added in some information about the validation Runtime does for the angles in `rzz` gates when transpiling/executing circuits against a backend with fractional gates enabled. 

@yaelbh, a quick question for you.
Just to make sure I have the feature(s)/user story clear in my head. Can you tell me if the following is correct?:
1. The `FoldRzzAngle` pass -- which appropriately transforms circuits such that all `rzz` gates will have numeric angles in the valid range `0 - pi/2` --  is installed into the Qiskit SDK as a plugin, and will run inside of the standard preset pass manager (assuming the user specified a backend with fractional gates enabled as the target).
2. If any `rzz` gates are defined using `Parameter`s, the pass will ignore these gates and assume the user will specify valid angles at runtime, and if a PUB contains invalid parameter values, the job will fail.


docs/guides/fractional-gates.ipynb
```diff
@@ -220,6 +221,15 @@
     ")"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Angle constraints\n",
+    "\n",
+    "For the two-qubit $R_{ZZ}(\\theta)$ gate, only angles between $0$ and $\\pi/2$ can be executed on IBM Quantum hardware. If a circuit contains any $R_{ZZ}(\\theta)$ gates with an angle outside of this range, then the standard transpilation pipeline will generally correct this with an appropriate circuit transformation (through the [`FoldRzzAngle`](../api/qiskit-ibm-runtime/transpiler-passes-fold-rzz-angle) pass).  However, any $R_{ZZ}(\\theta)$ gates containing one or more [`Parameter`](../api/qiskit/qiskit.circuit.Parameter)s, the transpiler will assume these parameters will be assigned angles within this range at runtime. However the job will fail if any of the parameter values specified in the PUB submitted to Qiskit Runtime are outside of this range."
```
yaelbh, line 230:
```suggestion
    "For the two-qubit $R_{ZZ}(\\theta)$ gate, only angles between $0$ and $\\pi/2$ can be executed on IBM Quantum hardware. If a circuit contains any $R_{ZZ}(\\theta)$ gates with an angle outside of this range, then the standard transpilation pipeline will generally correct this with an appropriate circuit transformation (through the [`FoldRzzAngle`](../api/qiskit-ibm-runtime/transpiler-passes-fold-rzz-angle) pass).  However, any $R_{ZZ}(\\theta)$ gates containing one or more [`Parameter`](../api/qiskit/qiskit.circuit.Parameter)s, the transpiler will assume these parameters will be assigned angles within this range at runtime. The job will fail if any of the parameter values specified in the PUB submitted to Qiskit Runtime are outside of this range."
```

yaelbh:
Yes, what you wrote is correct. @nkanazawa1989 and @wshanks can you please also verify?

docs/guides/fractional-gates.ipynb
abbycross: I think this sentence is missing something, maybe a "for" before "any"? (note that I removed some formatting/linking text here to make it more readable, so don't copy/paste it back like this 😄 )

> However, any R_ZZ gates containing one or more Parameters, the transpiler will assume these parameters will be assigned angles within this range at runtime.

wshanks:
It looks good to me besides the wording issues that Yael and Abby pointed out.

abbycross (approved):

PR #3037, opened 2025-04-23
Squash `gh-pages` commit history every night

### LINKED ISSUE #2799: Improve GitHub Pages cleanup job to force push
`git-sizer` shows that our gh-pages is enormous. This makes sense: it has the entire built site (but no API docs) for every PR preview we've built. 

We don't actually care about this Git history. We only use the `gh-pages` branch because it allows us to have multiple PRs concurrently push previews to the same site, each with their own folder. It works around the problem of otherwise needing to _atomically_ deploy the entire site with all preview folders at the same time.

So, we should force push to clean up the branch. Normally, force pushing is dangerous, but it should be fine in this case because no one actually uses the gh-pages branch other than CI. Still, we should do this force push at night when fewer people are contributing to reduce the risk of interfering with active CI jobs.

Idea:

1. Improve https://github.com/Qiskit/documentation/blob/main/.github/workflows/pr-cleanup.yml to force push to delete the old history
2. Run the job every night, rather than on merges to `main`.

frankharkins:
Part of #2799.

This PR changes the PR preview cleanup to squash all commits on the branch, and to run at midnight rather than every push to `main`.

### DIFF
--- a/scripts/pr-previews/cleanup.py
+++ b/scripts/pr-previews/cleanup.py
@@ -25,8 +25,9 @@ from utils import (
     setup_git_account,
     switch_branch,
     changed_files,
-    commit_all_and_push,
 )
+
+INITIAL_COMMIT = "499a5040585d02593cdd8237e19c9ee4a84ae126"
 
 
 def main() -> None:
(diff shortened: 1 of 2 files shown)


scripts/pr-previews/cleanup.py
```diff
@@ -39,8 +40,20 @@ def main() -> None:
...
 
-        commit_all_and_push("Clean up PR previews")
-        logger.info("Pushed updates to gh-pages branch")
+        # Be extra sure about the branch before we force-push
+        current_branch = run_subprocess(
+            ["git", "branch", "--show-current"]
+        ).stdout.strip()
+        assert (
+            current_branch == "gh-pages"
+        ), f"Did not expect to be on branch '{current_branch}'; exiting"
```
frankharkins, line 49: This is maybe a bit paranoid, but it's possible a bug introduced to the shared utilities might leave us on a branch we're not expecting.

scripts/pr-previews/cleanup.py
```diff
@@ -39,8 +40,20 @@ def main() -> None:
...
+
+        # Squash all commits on this branch to save space
+        run_subprocess(["git", "reset", "--soft", INITIAL_COMMIT])
+        run_subprocess(["git", "add", "."])
+        run_subprocess(["git", "commit", "-m", "Clean up PR previews"])
+        run_subprocess(["git", "push", "-f"])
```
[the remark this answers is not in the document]
frankharkins: Yep, just checked again to be sure: Committed changes become staged, and uncommitted changes remain unstaged. Since we do `git add .`, the full working state of the repo should be committed.

scripts/pr-previews/cleanup.py
frankharkins: Good idea, done :)

Eric-Arellano (approved):

Eric-Arellano:
FYI I updated this to not automatically close https://github.com/Qiskit/documentation/issues/2799 - let's close it after verifying it works tonight.

PR #3047, opened 2025-04-25
Relativize cloud base links

### LINKED ISSUE #3043: Relativize cloud links in API generation script
When we switch packages' intersphinx config to point to https://quantum.cloud.ibm.com/, it'll include links to that website rather than the legacy website. We should relativize those links in the same way we do the legacy links.

Here's an example of generating the runtime dev docs with the new `objects.inv` location:
Image

frankharkins:
Closes #3043

### DIFF
--- a/scripts/js/lib/api/updateLinks.ts
+++ b/scripts/js/lib/api/updateLinks.ts
@@ -10,6 +10,7 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
+import { join } from "path";
 import { initial, keyBy, keys, last } from "lodash-es";
 import { Root } from "mdast";
 import { visit } from "unist-util-visit";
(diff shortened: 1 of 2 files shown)


scripts/js/lib/api/updateLinks.ts
```diff
@@ -166,7 +172,8 @@ export function relativizeLink(link: Link): Link | undefined {
 
   const newText = link.url === link.text ? url : undefined;
   const newPrefix = priorPrefixToNewPrefix.get(priorPrefix)!;
-  return { url: `${newPrefix}/${url}`, text: newText };
+  const relativeUrl = removePrefix(join(newPrefix, url), "/");
+  return { url: `/${relativeUrl}`, text: newText };
```
frankharkins, line 173: I had to change the logic here to fix some cases of double-slashes (e.g. `/docs//guides/page`).

Eric-Arellano: Sg. Do your tests capture that edge case?

frankharkins: Yep, the tests actually uncovered it

PR #3048, opened 2025-04-25
Run API and notebook tests on workflow change

### LINKED ISSUE #2885: Run notebook tester & API checks when changing their GitHub workflows
We shouldn't skip these jobs if their files in `.github` change. With the notebook tester, it's important that our nb-tester and install-linux-deps logic recognizes that _every_ notebook should run.

frankharkins:
Fixes #2885


Eric-Arellano (approved):
Awesome, thanks for tackling this!

arnaucasau (approved):
Thanks

PR #3054, opened 2025-04-25
Remove path from image conversion advice

frankharkins:
It was a nice idea for writer experience to provide the full command they need to run, but it keeps getting outdated (currently says `public/docs/docs/...`), and won't be accurate if a user runs this check in inner-source. Since we rarely add images to this repo, I think it's better to just give an example command instead.


Eric-Arellano (approved):
Good call

PR #3057, opened 2025-04-25
No backends use fixed rate

### LINKED ISSUE #3005: We no longer have backends that use fixed rep rate

https://docs.quantum.ibm.com/guides/repetition-rate-execution

### Select all that apply

- [ ] typo
- [ ] code bug
- [x] out-of-date content
- [ ] broken link
- [ ] other

### Describe the fix.

In the linked page, the first sentence says `Most IBM® quantum processing units (QPUs) execute circuits at a fixed rate.`, which is no longer true. All of our backends use dynamic rep rate now, and we should just remove the mention of fixed rep rate. 

Having said that, I think there's value in keep the page to explain what rep delay is and how to set it. Although it should explain while we do circuit0_shot0 -> circuit1_shot0 etc, there is no guarantee on the order the circuits from PUBs are executed. e.g. if you submit `pubs=[pub1, pub2]`, there is no guarantee the circuits from pub1 will run before pub2.

beckykd:
closes #3005


docs/guides/repetition-rate-execution.mdx
```diff
@@ -1,32 +1,14 @@
...
-
-Therefore, each circuit is equal in duration to the longest circuit in the batch, and there is a common idle time that is placed in front of all circuits to match the repetition rate.
-
-## Dynamic repetition rate execution
-
-Some IBM QPUs allow for dynamic repetition rate execution. These QPUs are identified in Qiskit using `backend.configuration().dynamic_reprate_enabled`, and return a value of `True`. On these QPUs, it is possible to manually set the above idle time by setting the `rep_delay` of the submitted job. One can see from the above figures that by reducing the idle time one can potentially see a greater throughput of circuits on the QPUs that support dynamic repetition rates.
+Note that while circuits are processed in order; that is, circuit0_shot0 -> circuit1_shot0, and so on, there is no guarantee on the order the circuits from PUBs are executed. Thus, even though you submit pubs=[pub1, pub2], there is no guarantee the circuits from pub1 will run before those from pub2.
```
jyu00, line 11: I think we need more wording on what rep delay is and the effect of changing it. Something like

> When a circuit is executed on an IBM quantum processing unit (QPU), an implicit reset is typically inserted at the beginning to ensure the qubits are initialized to zero. However, errors in the reset process mean that this initialization is not perfect, leading to state preparation errors. To alleviate the error, the system also inserts repetition delay time (or `rep_delay`) between circuits. Each backend has a different default `rep_delay`, but it's usually longer than T1 to allow the environment help reset the qubits. The default rep_delay can be queried via `backend.default_rep_delay`. 
>
> All IBM quantum processing units (QPUs) use dynamic repetition rate execution, which allows you to change the `rep_delay` for each job. Circuits you submit in a primitive job are batched together for execution on the QPU. These circuits are executed by iterating over the circuits for each shot requested; the execution is column-wise over a matrix of circuits and shots (see below).
> 
> \<Column-wise execution matrix>
> 
> Since `rep_delay` is inserted between circuits, each shot of the execution would encounter this delay. Lowering the `rep_delay` would decrease the total QPU execution time, but at the expense of increased state preparation error rate. 
>
> Note that while circuits in a primitive job are batched together for QPU execution, there is no guarantee on the order the circuits from PUBs are executed. Thus, even though you submit `pubs=[pub1, pub2]`, there is no guarantee the circuits from `pub1` will run before those from `pub2`. There is also no guarantee that circuits from the same job would run as a single batch on the QPU. 

Paul wrote an excellent [blog](https://nonhermitian.org/posts/2021/2021-11-07-rep_delay.html) on this topic few years ago, if you want more reference.

PR #3064, opened 2025-04-28
add tutorials authors

miamico:
modify qiskit-bot yaml to update github handles of various tutorial authors

### DIFF
--- a/qiskit_bot.yaml
+++ b/qiskit_bot.yaml
@@ -570,6 +570,15 @@ notifications:
     - "@RitajitMajumdar"
   "docs/tutorials/qunova-hivqe":
     - "@miamico"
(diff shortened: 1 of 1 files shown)


qiskit_bot.yaml
```diff
@@ -445,6 +445,7 @@ notifications:
     - "@jenglick"
     - "@garrison"
     - "@johannesgreiner"
+    - "@ElePT"
```
frankharkins, line 448: This one is not a tutorial, is that intentional?

frankharkins (approved):

qiskit_bot.yaml
miamico: Yes, I updated based on feedback from @pandasa123

PR #3069, opened 2025-04-29
Minor fixes to the timing pages

jyu00:
Some minor fixes to the pages on timing, including

- Use the same wording to describe quantum time
- Change `QPU-calculated max execution time` to `service-calculated`, since QPU is just hardware
- Fix hyperlink text to match the new topic names
- Add wording on minimizing run time for estimator jobs

### DIFF
--- a/docs/guides/minimize-time.mdx
+++ b/docs/guides/minimize-time.mdx
@@ -14,7 +14,7 @@ There are several ways you can limit the amount of time spent processing and run
 
 - Use only the necessary settings for error suppression and error mitigation, because higher values can cause your jobs to run longer. See [Introduction to options](./runtime-options-overview), [Configure error suppression](./configure-error-suppression), and [Configure error mitigation](configure-error-mitigation) for details.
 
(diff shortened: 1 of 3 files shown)


docs/guides/minimize-time.mdx
```diff
@@ -6,13 +6,16 @@ description: How to minimize the amount of quantum time spent processing and run
...
 
-- Use only the necessary settings for error suppression, error mitigation, and optimization, because higher values can cause your jobs to run longer. See [Algorithm tuning options](./runtime-options-overview), [Configure runtime compilation](./configure-error-suppression), and [Configure error mitigation](configure-error-mitigation) for details.
+- Use only the necessary settings for error suppression and error mitigation, because higher values can cause your jobs to run longer. See [Introduction to options](./runtime-options-overview), [Configure error suppression](./configure-error-suppression), and [Configure error mitigation](configure-error-mitigation) for details.
+
+- If you are running multiple jobs that contain the same (likely parameterized) circuits and using an error mitigation method that requires noise models, such as PEA and PEC, consider using `NoiseLearner`. This helper program allows you to learn the noise model of a circuit once and reuse the model in subsequent Estimator queries. Note that a noise model becomes stale after certain time, so this is only practical if there is no long delay between jobs (e.g. within a Session). See [Noise learning helper
```
abbycross, line 17:
```suggestion
- If you are running multiple jobs that contain the same (likely parameterized) circuits and are using an error mitigation method that requires noise models, such as PEA and PEC, consider using `NoiseLearner`. With this helper program, you can learn the noise model of a circuit once and reuse the model in subsequent Estimator queries. Note that a noise model becomes stale after a certain time, so this is only practical if there is no long delay between jobs (for example, within a session). See [Noise learning helper
```

PR #3074, opened 2025-04-30
update license file to include creative commons terms

javabster:
After team discussions the decision has been made to update our license to protect content under a creative commons license that is more appropriate than apache 2.0 (which is designed for code). all code remains licensed under apache 2.0


Eric-Arellano (approved):
It looks like there is a formatting fix. You can run `npm run fmt`

PR #3080, opened 2025-05-02
Document how to ignore gh-pages branch

Eric-Arellano:
Running `git fetch` is very slow in this repository because of the `gh-pages` branch, even though you never use that locally. https://stackoverflow.com/a/64439191 says how to ignore the branch.


frankharkins (approved):
Worked nicely for me, thanks!

PR #3089, opened 2025-05-06
Update max ttl defaults

### LINKED ISSUE #2996: Investigate and verify that the max ttl for premium users is correctly documented

_No response_

### Select all that apply

- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [x] other

### Describe the fix.

Investigate and verify that the max ttl for premium users is correctly documented. Is it  8 hours by default, instead of 10 min like for open users?

run-jobs-batch#batch-length

beckykd:
closes #2996

### DIFF
--- a/docs/guides/run-jobs-batch.mdx
+++ b/docs/guides/run-jobs-batch.mdx
@@ -61,15 +61,16 @@ with Batch(backend=backend, max_time="25m"):
   ...
 ```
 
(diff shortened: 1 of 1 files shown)


abbycross (approved):
👍

jyu00:
@beckykd doesn't the session page needs to be updated as well?

beckykd:
> @beckykd doesn't the session page need to be updated as well?

This information was pulled from the batch TTL table.  I'll take a look and see if there are similar updates for sessions.

PR #3093, opened 2025-05-06
Adds additional tutorials requirements for nb-tester

### LINKED ISSUE #3071: Update `requirements.txt` for tutorials
We've noticed at least one tutorial ([Generating long-range entanglement with measurements](https://github.com/Qiskit/documentation/blob/main/docs/tutorials/generating-long-range-entanglement-with-measurements.ipynb)) that has a requirement not listed in `requirements.txt`. We should update the requirements file to include it.

emilkovacev:
Closes #3071

### DIFF
--- a/scripts/nb-tester/requirements.txt
+++ b/scripts/nb-tester/requirements.txt
@@ -19,5 +19,5 @@ scikit-learn~=1.6.1
 pyscf~=2.8.0; sys.platform != 'win32'
 python-sat~=1.8.dev16
 gem-suite~=0.1.6
-ffsim~=0.0.54
+ffsim~=0.0.54; sys.platform != 'win32'
 sympy~=1.14.0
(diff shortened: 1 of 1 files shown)


Eric-Arellano (commented):
Thanks!

@emilkovacev do you happen to know if any of these won't work on Windows? A good way to check is going in PyPI to the "Download files" tab like https://pypi.org/project/python-sat/#files. Then see if it's a universal wheel (only written in Python) vs a platform-specific wheel. If it's platform-specific, they should have a Windows wheel distributed.

emilkovacev:
Thanks for the pointer! After a brief look, ffsim doesn't distribute a Windows-specific wheel, the others all seem to. Also if I'm not mistaken, gem-suite seems to only have releases up to python3.10.

frankharkins (approved):
Thanks!

PR #3095, opened 2025-05-07
Add script to extract notebook image outputs

frankharkins:
This PR adds a script to extract image outputs from notebooks and convert them to AVIF if necessary. We're already doing this in inner-source as part of the sync but it was taking too long with the new tutorial notebooks. This PR just introduces the script, a follow-up PR will extract the images and add this to `./fix`.


scripts/image-output-extractor/qiskit_docs_notebook_image_extractor/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def main():
+    """
+    Search for notebooks and extract image outputs if necessary.
+    """
+    for nb_path in NOTEBOOK_PATHS:
+        images_folder = prepare_image_folder(nb_path)
+        nb = nbformat.read(nb_path, 4)
+        print(f"{nb_path}")
```
Eric-Arellano, line 55: Leftover from debugging?

scripts/image-output-extractor/qiskit_docs_notebook_image_extractor/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+    return RasterImage(filepath=new_path, data=new_data)
+
+
+def prepare_image_folder(nb_path: Path) -> Path:
+    """
+    Determine the appropriate output folder for the extracted images, and ensure it exists and is empty.
+    """
+    images_folder = Path(
+        "public",
+        nb_path.parts[0],
```
Eric-Arellano, line 127: It'd be helpful to have a comment explaining what the parts correspond to, e.g. `/my_path/foo.ipynb` -> `my_path`

scripts/image-output-extractor/qiskit_docs_notebook_image_extractor/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+        "images",
+        *nb_path.with_suffix("").parts[1:],
```
Eric-Arellano, line 129: Ditto comment to give an example what this entails

Eric-Arellano (commented):
Nice work! I'm glad you wrote this in Python rather than JS, and good call to use imagemagick.

It's fine for this to be in a follow up, but some things that may impact how this is designed:

1. We also need to move the handling of `CIRCUIT_DRAW_HTML_PREFIX` from closed source to open source. Presumably this file would do that. Rather than just doing image extraction, this script is really more about normalizing notebooks
2. We'll want to hook this script up to `./fix` (fine as a follow up)
3. We will need a lint mode so that CI can validate there are no missing conversions. You may want to use `argparse` to switch between fix vs lint mode.

frankharkins:
> We'll want to hook this script up to `./fix` (fine as a follow up)

Yep: I'll make a follow-up PR that actually runs this on the notebooks and adds the check to CI, `./check`, and `./fix`

scripts/image-output-extractor/qiskit_docs_notebook_image_extractor/__init__.py
frankharkins: I thought it'd be good to see which files have been processed, however happy to remove. It will be hidden from `./fix` users anyway.

scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+
+def normalize_notebook(
+    nb: nbformat.NotebookNode, image_folder: Path, check_only: bool = False
+) -> NormalizationResult:
+    """
+    Extracts images (converting if necessary) and returns an updated notebook.
+    """
+    images = []
+    change_made = False
+    for cell_index, cell in enumerate(nb.cells):
```
Eric-Arellano, line 128: This function has gotten pretty big. Maybe extract a helper function `normalize_cell`

scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+@dataclass
+class NormalizationNeeded:
+    changes = True
+    nb: nbformat.NotebookNode
+    images: list[Image]
+
+
+@dataclass
+class AlreadyNormalized:
+    changes = False
```
Eric-Arellano, line 70: More idiomatic is to make clear this is a classvar by using `ClassVar[bool]` here and line 63

Eric-Arellano (approved):
🚀

scripts/image-output-extractor/qiskit_docs_notebook_image_extractor/__init__.py
Eric-Arellano: I lean towards deleting it because we run on every file, even if the file was already processed. I think that's noisy given how many notebooks we have.

If you do keep it, I'd probably recommend adding a prefix like `Processed {notebook}`

scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/__init__.py
frankharkins: Good point, I tried splitting the specific normalization "passes" into their own functions – do you think that's enough? Or do you think it's better to have a per-cell function too?

scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/cell_output_data.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def extract_image_output(
+    output_data: dict, filestem: Path, skip_conversion: bool
+) -> Image | None:
+    """Extract image output if one exists and mutate the cell to point to the destination image."""
+
+    def _get_image(
+        output_data: dict, filestem: Path, skip_conversion: bool
+    ) -> Image | None:
```
Eric-Arellano, line 52: Nit that it's better to either move this to an entirely separate function, or to keep this as an inner function but remove the arguments. Right now, the arguments for this nested function are "shadowing" the outer arguments. Nested functions have access to the outer scope (aka the "closure")

I'd probably move it to a dedicated function. It's easier to read and easier to test in the future.

scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/__init__.py
```diff
@@ -60,7 +60,6 @@ def main():
         if is_hidden(nb_path):
             continue
 
-        print(f"{nb_path}")
```
Eric-Arellano: The print on 78 doesn't make sense anymore. It needs to put the notebook name and probably don't indent things. Maybe also use a ✍️ emoji to make it fun hah

scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+            # 2. Extract image outputs
+            filestem = Path(image_folder, f"{cell.id}-{index}")
+            if image := extract_image_output(
+                data, filestem, skip_conversion=check_only
+            ):
+                change_made = True
+                images.append(image)
+
+            if change_made and check_only:
+                # We know the notebook needs linting so we can stop here
```
Eric-Arellano, line 120:
```suggestion
                # We now know the notebook needs linting, so we don't need to keep looking at
                # more cells.
```

PR #3110, opened 2025-05-09
Update session max TTL 

beckykd:
@jyu00 noticed that when I updated the batch max TTL, I needed to also update the session max TTL.

### DIFF
--- a/docs/guides/run-jobs-session.mdx
+++ b/docs/guides/run-jobs-session.mdx
@@ -82,7 +82,7 @@ with Session(backend=backend):
 <span id="specify-length"></span>
 ## Session length
 
-The maximum session time to live (TTL) determines how long a session can run.  You can set this value with the `max_time` parameter. This should exceed the longest job's execution time.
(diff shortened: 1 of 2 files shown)


abbycross (approved):
👍

docs/guides/run-jobs-session.mdx
```diff
@@ -97,30 +97,32 @@ with Session(backend=backend, max_time="25m"):
...
+| --- | --- |
+| Premium Plan | 8 h* |
+| Open Plan | sessions run in job mode |
 </LegacyContent>
 <CloudContent>
-| Instance type (Open or Premium Plan) |  Interactive TTL | Maximum TTL |
-| --- | --- | --- |
-| Premium Plan | 60 sec* | 8 h* |
+
+| Instance type  | Default maximum TTL |
```
jyu00, line 111:
```suggestion
| Plan type  | Default maximum TTL |
```

Isn't it clearer to just call it `Plan type`?

jyu00 (approved):
One small comment but otherwise looks good.

docs/guides/run-jobs-session.mdx
beckykd: sure

PR #3114, opened 2025-05-12
Update testing env to include `PendingDeprecation` warnings

### LINKED ISSUE #2716: Consider testing for PendingDepreactionWarnings
There are many DeprecationWanings coming in 2.1 (see https://github.com/Qiskit/qiskit/pull/13604) Maybe you want to check PendingDeprecationWarnings to be ahead of them?

kaelynj:
Closes #2716 

According to the official python docs, `PendingDeprecation` warnings are [ignored by default](https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning) unless [Python Development Mode](https://docs.python.org/3/library/devmode.html) is enabled (done either through the `-Xdev` or `-Wd` flags when invoking `python` or by setting the environment variable `PYTHONDEVMODE=1`). 

This PR sets the `tox` environment to run using dev mode with the `PYTHONDEVMODE` env var.

### DIFF
--- a/docs/guides/circuit-library.ipynb
+++ b/docs/guides/circuit-library.ipynb
@@ -56,7 +56,11 @@
    "cell_type": "code",
    "execution_count": 1,
    "id": "a846a845-7ac5-4c92-b124-d2b90a773ba2",
-   "metadata": {},
+   "metadata": {
+    "tags": [
+        "ignore-warnings"
+    ]
+   },
    "outputs": [
     {
      "data": {
@@ -97,6 +101,17 @@
     "    \"# Add a controlled-phase gate to the circuit 'qc'\"\n",
     "  ]}\n",
     "/>"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "ff50da63",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import warnings\n",
+    "warnings.simplefilter('default', category=PendingDeprecationWarning)\n"
    ]
   },
   {
@@ -482,7 +497,7 @@
  "metadata": {
   "description": "Read more about out-of-the-box circuits provided by the Qiskit circuit library, including N-local, time-evolution and data-encoding circuits",
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "test-pendingDeprecation",
    "language": "python",
    "name": "python3"
   },
@@ -496,7 +511,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3"
+   "version": "3.11.11"
   },
   "title": "Circuit library"
  },
--- a/docs/guides/circuit-library.ipynb
+++ b/docs/guides/circuit-library.ipynb
@@ -58,7 +58,7 @@
    "id": "a846a845-7ac5-4c92-b124-d2b90a773ba2",
    "metadata": {
     "tags": [
-        "ignore-warnings"
+     "ignore-warnings"
     ]
    },
    "outputs": [
@@ -111,7 +111,8 @@
    "outputs": [],
    "source": [
     "import warnings\n",
-    "warnings.simplefilter('default', category=PendingDeprecationWarning)\n"
+    "\n",
+    "warnings.simplefilter(\"default\", category=PendingDeprecationWarning)"
    ]
   },
   {
@@ -497,7 +498,7 @@
  "metadata": {
   "description": "Read more about out-of-the-box circuits provided by the Qiskit circuit library, including N-local, time-evolution and data-encoding circuits",
   "kernelspec": {
-   "display_name": "test-pendingDeprecation",
+   "display_name": "Python 3",
    "language": "python",
    "name": "python3"
   },
@@ -511,7 +512,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.11"
+   "version": "3"
   },
   "title": "Circuit library"
  },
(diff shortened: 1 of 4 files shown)


Eric-Arellano:
Thanks, Kaelyn. A better approach is to configure our `warnings` code to look for `PendingDeprecationWarning`. We want to avoid all the other implications of dev mode when possible, such as slower performance.

See https://docs.python.org/3/library/warnings.html

> The determination whether to issue a warning message is controlled by the warning filter, which is a sequence of matching rules and actions. Rules can be added to the filter by calling filterwarnings() and reset to its default state by calling resetwarnings().

In particular, you can use https://docs.python.org/3/library/warnings.html#warnings.simplefilter with `category=PendingDeprecationWarning`. Look at https://docs.python.org/3/library/warnings.html#the-warnings-filter for the actions

--

I think you'll add this code here, but @frankharkins feel free to add other suggestions:

https://github.com/Qiskit/documentation/blob/7cd57d7591e0214eea97dbba2b0f0bfe29545162/scripts/nb-tester/qiskit_docs_notebook_tester/config.py#L26-L33

--

In case you haven't yet, be sure to test this is working correctly. You can do that by adding `warnings.warn()` inside a notebook and executing it with nb-tester to confirm that the script fails due to the warning.

frankharkins:
Thanks Eric, I agree changing the warnings filter would be better (@kaelynj sorry I didn't think of this when we discussed this earlier).

It's a little bit more work, but it'd be best if we enable it through a CLI argument. I think `get_notebook_job` should append to the `PRE_EXECUTE_CODE` if the arg is present.

kaelynj:
Thanks for the feedback all! There is now a new CLI arg `--check-pending-deprecations` that appends a modified warning filter to the `PRE_EXECUTE_CODE` in a hopefully satisfying way. (Though the variable names are feeling a bit cumbersome :sweat_smile: )

frankharkins (approved):
Thanks! Thoughts on adding the arg to `tox.ini` so it's enabled by default? I only found a couple of warnings when I ran it: One in `docs/guides/circuit-library.ipynb` which we should fix in a separate PR, and one in `docs/guides/qiskit-addons-mpf-get-started.ipynb` which we can't control but can ignore. You should be able to ignore both warnings for now by adding an `ignore-warnings` tag to the cells.

<details><summary>Warnings</summary>

```
⚠️ Warnings in docs/guides/circuit-library.ipynb:
Cell 4:
 │ /Users/frankharkins/Documents/documentation/.tox/py311/lib/python3.11/site
 │ -packages/qiskit/circuit/library/standard_gates/x.py:1118:
 │ PendingDeprecationWarning: The method ``qiskit.circuit.library.standard_ga
 │ tes.x.MCXGate.get_num_ancilla_qubits()`` is pending deprecation as of
 │ Qiskit 1.3. It will be marked deprecated in a future release, and then
 │ removed no earlier than 3 months after the release date. For an MCXGate it
 │ is no longer possible to know the number of ancilla qubits that would be
 │ eventually used by the transpiler when the gate is created. Instead, it is
 │ recommended to use MCXGate and let HighLevelSynthesis choose the best
 │ synthesis method depending on the number of ancilla qubits available.
 │ However, if a specific synthesis method using a specific number of ancilla
 │ qubits is require, one can create a custom gate by calling the
 │ corresponding synthesis function directly.
 │   num_ancilla_qubits =
 │ self.__class__.get_num_ancilla_qubits(num_ctrl_qubits)

⚠️ Warnings in docs/guides/qiskit-addons-mpf-get-started.ipynb:
Cell 7:
 │ /Users/frankharkins/Documents/documentation/.tox/py311/lib/python3.11/site
 │ -packages/osqp/interface.py:405: PendingDeprecationWarning: The default
 │ value of raise_error will change to True in the future.
 │   warnings.warn(
```

</details>

kaelynj:
> Thanks! Thoughts on adding the arg to `tox.ini` so it's enabled by default? I only found a couple of warnings when I ran it: One in `docs/guides/circuit-library.ipynb` which we should fix in a separate PR, and one in `docs/guides/qiskit-addons-mpf-get-started.ipynb` which we can't control but can ignore. You should be able to ignore both warnings for now by adding an `ignore-warnings` tag to the cells.
> Warnings
> 
> ```
> ⚠️ Warnings in docs/guides/circuit-library.ipynb:
> Cell 4:
>  │ /Users/frankharkins/Documents/documentation/.tox/py311/lib/python3.11/site
>  │ -packages/qiskit/circuit/library/standard_gates/x.py:1118:
>  │ PendingDeprecationWarning: The method ``qiskit.circuit.library.standard_ga
>  │ tes.x.MCXGate.get_num_ancilla_qubits()`` is pending deprecation as of
>  │ Qiskit 1.3. It will be marked deprecated in a future release, and then
>  │ removed no earlier than 3 months after the release date. For an MCXGate it
>  │ is no longer possible to know the number of ancilla qubits that would be
>  │ eventually used by the transpiler when the gate is created. Instead, it is
>  │ recommended to use MCXGate and let HighLevelSynthesis choose the best
>  │ synthesis method depending on the number of ancilla qubits available.
>  │ However, if a specific synthesis method using a specific number of ancilla
>  │ qubits is require, one can create a custom gate by calling the
>  │ corresponding synthesis function directly.
>  │   num_ancilla_qubits =
>  │ self.__class__.get_num_ancilla_qubits(num_ctrl_qubits)
> 
> ⚠️ Warnings in docs/guides/qiskit-addons-mpf-get-started.ipynb:
> Cell 7:
>  │ /Users/frankharkins/Documents/documentation/.tox/py311/lib/python3.11/site
>  │ -packages/osqp/interface.py:405: PendingDeprecationWarning: The default
>  │ value of raise_error will change to True in the future.
>  │   warnings.warn(
> ```

This makes sense to me. I added it to `tox.ini` and included the appropriate `ignore-warnings` tags. (Also opened an issue to fix `circuit-library.ipynb` which is being tracked by #3118 )

frankharkins (approved):
Thank you!

Eric-Arellano (approved):

PR #3115, opened 2025-05-12
Fix small error

asorini:
The Hello World guide seems to have a error (or potentially misleading statement) in its description of an expectation value. While it is true that the referenced expectation value is 1 for the specific entangled state created by the example circuit, that expectation value does not have to be 1 for all entangled states.


kaelynj (approved):
This is correct for the maximally entangled state, thanks for pointing it out!

abbycross:
Thank you @asorini ! We need you to [sign the CLA](https://github.com/Qiskit/documentation/pull/3115#issuecomment-2873944769) before we can merge this PR, if you don't mind doing so?

asorini:
Hi. Thanks for letting me know. I think I did it, but it doesn't seem to be showing up yet.

Eric-Arellano (approved):
Thanks for fixing! This is our most read page in the docs, so you're helping a lot of people with this change 🙏

PR #3143, opened 2025-05-15
Use lock to avoid kernel address conflicts in nb-tester

frankharkins:
This PR uses a lock to try and avoid this error when running in a GitHub action:

```
zmq.error.ZMQError: Address already in use (addr='tcp://127.0.0.1:45769')
```

This is a bit of a shot in the dark because I can't reproduce it locally.

This is related to #2387, but is not necessarily a full fix as I believe there is more than one thing going on there.

***

Inspired by https://github.com/jupyter/nbclient/pull/327

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ poetry.lock
 /dist/
 /test-results/
 Untitled*.ipynb
+*.lock
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,3 @@ poetry.lock
 /dist/
 /test-results/
 Untitled*.ipynb
-*.lock
--- a/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
+++ b/scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
@@ -32,9 +32,13 @@ async def _main() -> None:
 
     start_time = datetime.now()
     print("Executing notebooks:")
-    setup_kernel_lock = asyncio.Lock()
+
+    # New kernels choose a port on creation. They usually detect if the port is
+    # in use and will choose another if so, but there can be race conditions
+    # when creating many kernels at once. We use a lock to avoid this.
+    kernel_setup_lock = asyncio.Lock()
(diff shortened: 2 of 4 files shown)


frankharkins:
I re-ran the notebook test ~10 times and didn't experience the issue, so hopefully this fixes it.

.gitignore
```diff
@@ -19,3 +19,4 @@ poetry.lock
 /dist/
 /test-results/
 Untitled*.ipynb
+*.lock
```
Eric-Arellano, line 22: Not necessary anymore, right?

scripts/nb-tester/qiskit_docs_notebook_tester/__init__.py
```diff
@@ -18,6 +18,8 @@
 import platform
 from datetime import datetime
 
+import nbclient
```
Eric-Arellano, line 21: Unused?

scripts/nb-tester/qiskit_docs_notebook_tester/execute.py
```diff
@@ -130,11 +131,12 @@ async def _execute_notebook(
     """
     nb = nbformat.read(job.path, as_version=4)
 
-    kernel_manager, kernel = await start_new_async_kernel(
-        kernel_name="python3",
-        extra_arguments=["--InlineBackend.figure_format='svg'"],
-        cwd=working_directory,
-    )
+    async with kernel_setup_lock:
```
Eric-Arellano, line 134: It'd be helpful to add a comment, e.g.

> We've noticed non-deterministic failures when running multiple notebooks in parallel. We believe the issue is when starting up new kernels, whereas it's okay to have multiple already created kernels continue their execution in parallel.

Probably look to the issue you got inspiration from

Eric-Arellano (approved):
Awesome! Good job using asyncio.lock rather than the filelock library

Eric-Arellano:
@Qiskit/docs-content-team test

Issue #3090, opened 2025-05-06
Turn off inheritance for Qiskit SDK QuantumCircuit subclasses


Eric-Arellano:
We recently improved the docs for QuantumCircuit, but that resulted in all the subclasses in the circuit library inheriting the same docs. `@jakelishman` and I think that is somewhat confusing. See https://github.com/Qiskit/documentation/pull/3082/files

Jake explained that they don't actually want the circuit library classes like `And` to subclass `QuantumCircuit`, and that's more of an implementation. 

> in the nearer term, I expect that what would be most useful to users would be if we don’t inherit all that documentation, and instead set the documentation template of those classes to simply be the class+__init__ docstrings, and any non-derived methods they have (which shouldn’t be any), and put a top-level description of them on the page that lists them explaining that they’re only meant to be used as if they were gates

This would probably be achieved by configuring a dedicated `autosummary` template, such as using https://github.com/Qiskit/qiskit/blob/main/docs/_templates/autosummary/class_no_inherited_members.rst or maybe using a variant of it.

PR #3150, opened 2025-05-16
Generate Qiskit 2.0.1

Eric-Arellano:
This adds the new docs for QuantumCircuit. While we don't love that this means https://github.com/Qiskit/documentation/issues/3090 is a problem, we should not block the update on that.


arnaucasau (approved):
Thanks!

PR #3178, opened 2025-05-23
Use tempfiles in image extraction

frankharkins:
Originally, I was trying to do everything in-memory so we could push side-effects to the end of the program, but that didn't seem to work on Fedora for some reason. This works for me on OSX, and @kaelynj tested this out and it works on her Fedora machine.


Eric-Arellano (approved):
🚀

PR #3190, opened 2025-05-27
remove unused version info box

### LINKED ISSUE #3188: Broken (and unneeded) package version box in Hardware considerations file
There are no packages listed in the version-info cell of the Hardware considerations page - but it looks like it's not necessary since there are no code blocks anyway. I recommend removing the cell altogether.

Image

abbycross:
Closes #3188

### DIFF
--- a/docs/guides/dynamic-circuits-considerations.ipynb
+++ b/docs/guides/dynamic-circuits-considerations.ipynb
@@ -10,27 +10,6 @@
   },
   {
    "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


abbycross:
> Hm, this should probably be converted to an MDX file then? MDX is much simpler to maintain and we should only use Jupyter notebooks if there is actual code.

I would be fine with that! But I wonder if there's a reason why we converted it from MDX in the first place? I'll investigate that, but the page will be revamped in the next few weeks since dynamic circuits is being updated.  As part of that work we'll consider changing the page back to MDX.

PR #3209, opened 2025-05-29
Document Qiskit C API compiler requirements

mtreinish:
Following discussions around Qiskit/qiskit#14138 we decided to document explicitly that Qiskit's C API requires the C11 standard. So you need a C11 compatible C compiler to use the C API. This commit updates the C installation guide to document this requirement.

### DIFF
--- a/docs/guides/install-c-api.mdx
+++ b/docs/guides/install-c-api.mdx
@@ -49,7 +49,7 @@ This section provides build instructions for UNIX-like systems.
 
 Compilation requires the following tools:
 * A Rust compiler: see for example the [guide on installing Qiskit from source](/docs/guides/install-qiskit-source)
-* A C compiler: for example, GCC on Linux and Clang on MacOS. Qiskit's C API is compatible with a [C11](https://en.wikipedia.org/wiki/C11_(C_standard_revision) compatible compiler.
+* A C compiler: for example, GCC on Linux and Clang on MacOS. Qiskit's C API is compatible with a [C11](https://en.wikipedia.org/wiki/C11_(C_standard_revision)) compatible compiler.
 * [`cbindgen`](https://github.com/mozilla/cbindgen): a tool to create the C header, which you can install with `cargo install cbindgen`
     Note that running the tool from the command line should be enabled, which might require exporting your `PATH` variable to include `/path/to/.cargo/bin`
 * (GNU) Make: this is optional but is recommended to use automated install processes
--- a/docs/guides/install-c-api.mdx
+++ b/docs/guides/install-c-api.mdx
@@ -49,7 +49,7 @@ This section provides build instructions for UNIX-like systems.
 
 Compilation requires the following tools:
 * A Rust compiler: see for example the [guide on installing Qiskit from source](/docs/guides/install-qiskit-source)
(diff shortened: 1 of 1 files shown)


docs/guides/install-c-api.mdx
```diff
@@ -49,7 +49,7 @@ This section provides build instructions for UNIX-like systems.
 
 Compilation requires the following tools:
 * A Rust compiler: see for example the [guide on installing Qiskit from source](/docs/guides/install-qiskit-source)
-* A C compiler: for example, GCC on Linux and Clang on MacOS
+* A C compiler: for example, GCC on Linux and Clang on MacOS. Qiskit's C API is compatible with a `C11 <https://en.wikipedia.org/wiki/C11_(C_standard_revision)>`__ compatible compiler.
```
mtreinish, line 52: I wasn't sure of the best link for this. I decided on wikipedia but I don't think it's the best choice. The official ISO spec is not useful: https://www.iso.org/standard/57853.html and also costs a lot of money. There is: https://www.open-std.org/jtc1/sc22/wg14/www/standards.html which isn't super useful either but it does link to the standard: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf Maybe it doesn't need a link at all. I'm fine to defer to whatever others think is best.

mtreinish: Oops, I've been writing too many sphinx docs lately forgot to context switch to markdown.

docs/guides/install-c-api.mdx
```diff
@@ -49,7 +49,7 @@ This section provides build instructions for UNIX-like systems.
 
 Compilation requires the following tools:
 * A Rust compiler: see for example the [guide on installing Qiskit from source](/docs/guides/install-qiskit-source)
-* A C compiler: for example, GCC on Linux and Clang on MacOS
+* A C compiler: for example, GCC on Linux and Clang on MacOS. Qiskit's C API is compatible with a [C11](https://en.wikipedia.org/wiki/C11_(C_standard_revision) compatible compiler.
```
Eric-Arellano, line 52:
```suggestion
* A C compiler: for example, GCC on Linux and Clang on MacOS. Qiskit's C API is compatible with a [C11](https://en.wikipedia.org/wiki/C11_(C_standard_revision)) compatible compiler.
```

docs/guides/install-c-api.mdx
```diff
@@ -49,7 +49,7 @@ This section provides build instructions for UNIX-like systems.
 
 Compilation requires the following tools:
 * A Rust compiler: see for example the [guide on installing Qiskit from source](/docs/guides/install-qiskit-source)
-* A C compiler: for example, GCC on Linux and Clang on MacOS
+* A C compiler: for example, GCC on Linux and Clang on MacOS. Qiskit's C API is compatible with a [C11](https://en.wikipedia.org/wiki/C11_(C_standard_revision)) compatible compiler.
```
Cryoris: Just a suggestion to avoid word repetition, but also fine as is if you prefer 🙂 
```suggestion
* A C compiler: for example, GCC on Linux and Clang on MacOS. Qiskit's C API is compatible with a compiler conforming to the [C11](https://en.wikipedia.org/wiki/C11_(C_standard_revision)) standard.
```

PR #3222, opened 2025-05-30
Add Runtime .40 to latest updates

abbycross:


docs/guides/latest-updates.mdx
```diff
@@ -8,7 +8,7 @@ in_page_toc_max_heading_level: 2
 
 # Latest updates
 
-*Last updated 16 May 2025*
+*Last updated 2 June 2025*
```
abbycross:
```suggestion
*Last updated 4 June 2025*
```

jyu00:
Is there a PR coming to update the IQP migration guide with the new syntax? If so, perhaps we should wait for that. Otherwise it might be confusing when this page says to use, for example, `channel="ibm_quantum_platform"` but the migration guide says `channel="ibm_cloud"`.

beckykd:
> Is there a PR coming to update the IQP migration guide with the new syntax? If so, perhaps we should wait for that. Otherwise it might be confusing when this page says to use, for example, `channel="ibm_quantum_platform"` but the migration guide says `channel="ibm_cloud"`.

Yes - @kt474 updated it already.

PR #3229, opened 2025-06-02
Print first ten results rather than counts in "Get started with primitives"

frankharkins:
Printing the full counts dictionary is really noisy, and it also doesn't really make sense since the dictionary is huge and every value is just `1`. I think it makes more sense to just print a handful of results instead.

Draft while I re-run the notebook, I'll have to do this overnight probably due to queue times.


Eric-Arellano:
This seems like a good change to land. Is the notebook running now?

frankharkins:
Yep this is now good to go

Eric-Arellano (approved):

PR #3243, opened 2025-06-03
Switch testing to `ibm_cloud`

frankharkins:
Hopefully this is all we need to do


scripts/nb-tester/qiskit_docs_notebook_tester/patches/qiskit-ibm-runtime-open
```diff
@@ -5,7 +5,7 @@ from qiskit_ibm_runtime import QiskitRuntimeService
 QiskitRuntimeService._original_least_busy = QiskitRuntimeService.least_busy
 
 def patched_least_busy(self, *args, **kwargs):
-    open_backends = [b.name for b in self.backends(instance="ibm-q/open/main")]
+    open_backends = [ "alt_brisbane", "alt_sherbrooke", "alt_torino" ]
```
Eric-Arellano, line 8: Given what Emil said about queues, it sounds like this is a bad idea to use

PR #3262, opened 2025-06-05
Stop using images from Learning API

Eric-Arellano:


docs/tutorials/depth-reduction-with-circuit-cutting.ipynb
```diff
@@ -136,7 +136,7 @@
    "id": "68c00476",
    "metadata": {},
    "source": [
-    "![swaps-2.png](https://learning-api.quantum.ibm.com/assets/ce74853a-8e5b-460b-9dc6-69eeadca2d05)"
+    "![swaps.png](/docs/images/tutorials/depth-reduction-with-circuit-cutting/swaps.avif)"
```
Eric-Arellano: I opened https://github.com/Qiskit/documentation/issues/3263 to fix this alt text

frankharkins: Fixing in a follow-up is fine, but would be great if we can fix it now

```suggestion
    "![Coupling map showing the qubits that will need to be swapped](/docs/images/tutorials/depth-reduction-with-circuit-cutting/swaps.avif)"
```

PR #3273, opened 2025-06-06
[nb-tester]: Handle our cloud instance

frankharkins:


scripts/config/notebook-testing.toml
```diff
@@ -138,6 +137,9 @@ notebooks = [
 # Don't ever test the following notebooks
 [groups.exclude]
 notebooks = [
+    # TODO: See if this works again after migrating to cloud
+    "docs/guides/save-jobs.ipynb",
```
Eric-Arellano, line 141: This was already broken on main? What's the issue? Should we open a GH issue to track it?

Eric-Arellano (commented):
Thanks! This looks good.

1. It'd be safer to require `''` or `""` surrounding the string, since they will be string literals. 
2. How hard do you think this would be to unit test? While this code is temporary, there's a real risk it lasts much longer due to competing priorities. Maybe spend <45 minutes testing this -- we don't need perfect tests.

scripts/config/notebook-testing.toml
frankharkins: It's timing out, I don't know why 😕 

I hadn't noticed it break before, but it does rely on IQP and the previous history of jobs, so I think it's most likely related to that.

scripts/nb-tester/test/test_backend_replacements.py
```diff
@@ -0,0 +1,67 @@
...
+from qiskit_docs_notebook_tester.execute import (
+    open_backends_to_internal,
+    internal_backends_to_open,
+)
+
+def in_notebook(cell_source: str, outputs: list[dict] | None = None) -> NotebookNode:
+    return new_notebook(cells=[new_code_cell(
+        source=cell_source,
+        outputs=[new_output(**o) for o in outputs] if outputs else []
+    )])
```
Eric-Arellano: Naming nit - feel free to disagree. `in_notebook` is a little confusing to me. I'd recommend `create_notebook` or `new_notebook`. Of course, the latter clashes with nbformat; so, you can switch `nbformat` to `import nbformat.v4`, then the call sites would be `nbformat.v4.new_notebook`. While that's more verbose, it's arguably more clear what is our code vs nbformat's code

scripts/nb-tester/test/test_backend_replacements.py
```diff
@@ -0,0 +1,67 @@
...
+    received = internal_backends_to_open(nb)
+    assert len(received.cells) == 1
+    assert received.cells[0].source == "alt_brisbane_coupling_map = CouplingMap(...)"
+
+def test_internal_to_open_stream_outputs():
+    nb = in_notebook(
+        "print(\"Using backend: {backend.name}\")",
+        outputs=[{
+            "output_type": "stream",
+            "text": "Using backend: alt_brisbane"
```
Eric-Arellano, line 45: Consider adding more complex output, like newlines and some strings with quotes and some without.

PR #3279, opened 2025-06-06
Remove channel specification

abbycross:
Remove channel specification and mention of Sherbrooke from get-qpu-information file


Eric-Arellano (approved):
Thanks

PR #3291, opened 2025-06-10
Fix magick improper header error

arnaucasau:
This PR aims to fix the following error when running the notebook normalizer:

```
magick: improper image header `/var/folders/d1/z9l0cjmx5jn7nthr53czl1qm0000gn/T/tmpp5co82hx.png' @ error/png.c/ReadPNGImage/3947.
```

Given that the error is with the header, I suspect that magick wasn't able to read it when opening the images. This could be caused because after the `write`, the File Handle is not located at the beginning of the file.

This is an example of some positions printed after the `write` command for some images:

```
  6560
  41510
  55567
  54349
  86080
  104641
  2564
```

This PR changes the File Handle position to the beginning of the file before running the magick command to prevent that


Eric-Arellano (approved):
Thanks! I'm surprised we haven't been seeing failures before this.

Issue #3315, opened 2025-06-13
Code for Grover's oracle fails when one of the marked states is all 1s


iyanmv:
### URL to the relevant tutorial

https://learning.quantum.ibm.com/tutorial/grovers-algorithm

[docs/tutorials/grovers-algorithm.ipynb](https://github.com/Qiskit/documentation/blob/main/docs/tutorials/grovers-algorithm.ipynb)

### Select all that apply

- [ ] new content request
- [ ] typo
- [x] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

### Describe the fix or the content request.

The function `grover_oracle()` used in the tutorial fails if one of the marked states contains all 1s. This is because the `zero_inds` in that case is an empty list. This edge case is not considered and the methods `qc.x()` fail.

### For new content requests - if the request is accepted, do you want to write the content?

I will write (or already have written) a draft of the proposed content

PR #3316, opened 2025-06-13
Fix `grover_oracle()`

iyanmv:
This PR is just a small change in one of the cells of the [Grover's tutorial notebook](https://learning.quantum.ibm.com/tutorial/grovers-algorithm) to consider the edge case where one of the marked states is all 1s.

Before this PR, doing something like this would fail:
```python
grover_oracle("111")
```
After this PR, it works as expected and the circuit contains no X gates in that edge case.

I have limited the change to that cell in the notebook, and I have not rerun the whole notebook to avoid modifying cells metadata and other irrelevant things.

Fixes: #3315


henryzou50 (approved):
LGTM! I reviewed the changes and testing, and the edge case now works correctly. Thanks!

Eric-Arellano (approved):
Thank you for taking the time to fix this!

PR #3320, opened 2025-06-13
fix bug in real-time qubit selection tutorial

### LINKED ISSUE #3197: Improve "real-time benchmarking for qubit selection"

https://github.com/Qiskit/documentation/blob/main/docs/tutorials/real-time-benchmarking-for-qubit-selection.ipynb

### Select all that apply

- [ ] new content request
- [ ] typo
- [x] code bug
- [ ] out-of-date content
- [ ] broken link
- [x] other

Few team members have helped flag some benefits for updating notebook:

- Should fix transpilation to ensure that trivial qubit selection is not done
- T2 characterization should instead be done without neighboring qubits due to always on ZZ  coupling

I can help the team by providing enough information or by referring them to a specific Subject Matter Expert to write the material

miamico:
Closes  #3197 where it was noted that the optimization setting chosen for the transpilation is not noise aware, thus the new calibrations are not used.

### DIFF
--- a/docs/tutorials/real-time-benchmarking-for-qubit-selection.ipynb
+++ b/docs/tutorials/real-time-benchmarking-for-qubit-selection.ipynb
@@ -461,7 +461,7 @@
    "metadata": {},
    "source": [
     "###  Execute a quantum circuit with default qubit selection\n",
-    "As a reference result of performance, we'll take a look at an example of executing a quantum circuit on a QPU by using the default qubits, that is the qubits selected with the reported backend properties. `optimization_level = 3` setting for the transpiler includes the most advanced transpilation optimizations and the use of target properties (like operation errors) to select best performing qubits for execution.  "
+    "As a reference result of performance, we'll execute a quantum circuit on a QPU by using the default qubits, which are the qubits selected with the requested backend properties. We will use `optimization_level = 3`.  This setting includes the most advanced transpilation optimization, and uses target properties (like operation errors) to select the best performing qubits for execution.  "
    ]
   },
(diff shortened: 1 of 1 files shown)


Eric-Arellano (commented):
Thanks! What do you think about adding description to the paragraph above explaining why you're using optimization_level=3? Or as a comment. Otherwise I'm worried users won't understand why that 'magic value' is being used

docs/tutorials/real-time-benchmarking-for-qubit-selection.ipynb
```diff
@@ -461,12 +461,12 @@
    "metadata": {},
    "source": [
     "###  Execute a quantum circuit with default qubit selection\n",
-    "As a reference result of performance, we'll take a look at an example of executing a quantum circuit on a QPU by using the default qubits, that is the qubits selected with the reported backend properties."
+    "As a reference result of performance, we'll take a look at an example of executing a quantum circuit on a QPU by using the default qubits, that is the qubits selected with the reported backend properties. `optimization_level = 3` setting for the transpiler includes the most advanced transpilation optimizations and the use of target properties (like operation errors) to select best performing qubits for execution.  "
```
beckykd:
```suggestion
    "As a reference result of performance, we'll execute a quantum circuit on a QPU by using the default qubits, which are the qubits selected with the requested backend properties. We will use `optimization_level = 3`.  This setting includes the most advanced transpilation optimization, and uses target properties (like operation errors) to select the best performing qubits for execution.  "
```

PR #3330, opened 2025-06-16
Remove unnecessary tar file

arnaucasau:
This PR removes the `ImageMagick.tar.gz` file which was introduced accidentally in https://github.com/Qiskit/documentation/pull/3220


Eric-Arellano (commented):
Hm, I guess https://github.com/Qiskit/documentation/pull/3215 wasn't pulled in yet.

PR #3348, opened 2025-06-17
Make announcements links relative to the platform

arnaucasau:
Until now, announcements were legacy only and we needed to link to them externally. Now, announcements has been incorporated into the cloud app, so we don't need to do so anymore. This PR updates the links to make them relative again.


Eric-Arellano (approved):
Thanks!

PR #3362, opened 2025-06-18
Cp/last

christopherporter1:
Reviewed whatever was live on staging. Found two references to surveys that were hiding somewhere other than the start of the course. Removed. That's it!


arnaucasau (approved):
Thanks!

Issue #3202, opened 2025-05-29
KQD tutorial - KeyError while transpiling to QPUs


nirajvenkat:
### URL to the relevant tutorial

Hi, I am attempting to run [the KQD tutorial](https://quantum.cloud.ibm.com/docs/en/tutorials/krylov-quantum-diagonalization) and I get the following error:
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[19], [line 19](vscode-notebook-cell:?execution_count=19&line=19)
     [16](vscode-notebook-cell:?execution_count=19&line=16)                     continue
     [18](vscode-notebook-cell:?execution_count=19&line=18) for q in cmap_list:
---> [19](vscode-notebook-cell:?execution_count=19&line=19)     twoq_gate_err = target["cz"][q].error
     [20](vscode-notebook-cell:?execution_count=19&line=20)     if twoq_gate_err > 0.015:
     [21](vscode-notebook-cell:?execution_count=19&line=21)         for q_pair in cmap_list:

File ~/anaconda3/envs/qi/lib/python3.11/site-packages/qiskit/transpiler/target.py:601, in Target.__getitem__(self, key)
    [600](https://file+.vscode-resource.vscode-cdn.net/Users/nvenkat/Desktop/Code/mitre_qchem/~/anaconda3/envs/qi/lib/python3.11/site-packages/qiskit/transpiler/target.py:600) def __getitem__(self, key):
--> [601](https://file+.vscode-resource.vscode-cdn.net/Users/nvenkat/Desktop/Code/mitre_qchem/~/anaconda3/envs/qi/lib/python3.11/site-packages/qiskit/transpiler/target.py:601)     return self._gate_map[key]

KeyError: 'cz'
```

When I output `target.keys()`:
```
dict_keys(['if_else', 'rz', 'id', 'ecr', 'x', 'reset', 'measure', 'for_loop', 'delay', 'switch_case', 'sx'])
```

I tried both the backends available to me on the Open Plan, `service.backends()` shows:
```
[<IBMBackend('ibm_brisbane')>, <IBMBackend('ibm_sherbrooke')>]
```

### Select all that apply

- [ ] new content request
- [ ] typo
- [x] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

### Describe the fix or the content request.

Please test on the Open Plan QPUs if possible!

### For new content requests - if the request is accepted, do you want to write the content?

[...]

PR #3217, opened 2025-05-30
add fix for KQD tutorial

miamico:
Fix error in qubit selection snippet in the KQD tutorial. Closes #3202

PR #3389, opened 2025-06-19
Fix Krylov tutorial being Heron-specific

Eric-Arellano:
Closes https://github.com/Qiskit/documentation/issues/3202. See https://github.com/Qiskit/documentation/pull/3217 for the original PR.


Eric-Arellano:
> maybe we should just leave out the min_num_qubits argument in the backend?

Do you remember why it was changed in the first place? I was only copying your change. I'm happy to keep it as is with 127. 

I think @beckykd normally has us set the argument so that things like simulators aren't used, iirc? Or maybe it's a holdover from when we had smaller devices and isn't necessary anymore?

Issue #3342, opened 2025-06-16
Research if there is any fix to race condition in PR preview deployment with GitHub Pages


Eric-Arellano:
## Problem

We periodically have gotten PR deployment failures in the `poll_deployment.py` step. They've always happened, but have been much worse in the past month, which I now realize is because of how much activity this repo has had.

It turns out that the issue is a race condition with PRs concurrently requesting a deployment with GitHub Pages to the same environment. This results in GitHub Actions cancelling the deployment 

![Image](https://github.com/user-attachments/assets/accb1044-fe76-4cf8-b162-66bf763afc17)

That happens in the `deploy` step with actions/deploy-pages@v4, e.g. in https://github.com/Qiskit/documentation/actions/runs/15691700791/job/44208343380

![Image](https://github.com/user-attachments/assets/c1ee7049-ecc7-4981-bfce-cdcdcec55b36)

```
Run actions/deploy-pages@v4
  with:
    token: ***
    timeout: 600000
    error_count: 10
    reporting_interval: 5000
    artifact_name: github-pages
    preview: false
Fetching artifact metadata for "github-pages" in this workflow run
Found 1 artifact(s)
Creating Pages deployment with payload:
{
	"artifact_id": 3339897245,
	"pages_build_version": "0226b2635e7e28044e7fc62819dd1059f27edf1f",
	"oidc_token": "***"
}
Created deployment for 0226b2635e7e28044e7fc62819dd1059f27edf1f, ID: 0226b2635e7e28044e7fc62819dd1059f27edf1f
Getting Pages deployment status...
Current status: deployment_in_progress
Getting Pages deployment status...
Current status: deployment_in_progress
Getting Pages deployment status...
Current status: deployment_in_progress
Getting Pages deployment status...
Current status: deployment_in_progress
Canceling Pages deployment...
Canceled deployment with ID 0226b2635e7e28044e7fc62819dd1059f27edf1f
Error: The operation was canceled.
```

Those logs correspond to https://github.com/actions/deploy-pages/blob/854d7aa1b99e4509c4d1b53d69b7ba4eaf39215a/src/index.js#L12 and https://github.com/actions/deploy-pages/blob/854d7aa1b99e4509c4d1b53d69b7ba4eaf39215a/src/internal/deployment.js#L219, which are triggered specifically when GH Actions cancels the workflow.

## Any solution?

Is there any mechanism to address this race condition, such as locking? Note that the issue is happening at the GH Actions level.

[...]

PR #3394, opened 2025-06-20
Bump PR preview polling timeout

Eric-Arellano:
Per https://github.com/Qiskit/documentation/issues/3342#issuecomment-2988072602, we are getting timeouts in our polling even when there is no race condition happening. That suggests that our deployment to GH pages is for some reason taking a long time, presumably because we have so many subsites being deployed, each with a lot of content.


arnaucasau (approved):
Thanks!

PR #3363, opened 2025-06-18
Migrate testing to cloud

frankharkins:
This PR migrates our notebook testing setup to use the IBM Cloud provider rather than the legacy provider.

PR #3404, opened 2025-06-23
[Bug] Fix GDQ function loader

pandasa123:
- **Original**: `gdq/quantum_portfolio_optimizer`
- **Expected**: `global-data-quantum/quantum-portfolio-optimizer`


abbycross (approved):

Eric-Arellano:
The CI issues are a regression from https://github.com/Qiskit/documentation/pull/3363. We'll fix that promptly and it's a simple fix. If we cannot fix it quickly enough, we'll merge this on red

PR #3408, opened 2025-06-23
[WIP] Add check to detect overly long inline katex expressions

arnaucasau:
Inline katex expressions are problematic because they will soon start to overflow the prose of the docs. This is because we won't be controlling the overflow on any paragraph anymore to allow the use of tooltips on them.

For this, we decided to remove the inline katex (`$`) references and instead use `katex-display` (`$$`) when the expressions do not fit in our smallest screen supported. Given that this is not easy to foresee, this PR adds a new check that will verify that all the inline katex expressions introduced in a file are always going to render without overflowing.

When a bad expressions is detected, we will see the conflicting expression and file with a message indicating what we can do to fix it:

- Run `./fix` for the output of the notebooks, or
- Manually replace the inline math (`$`) by equations (`$$`)

In addition, this PR updates all the bad expressions found in our non-API docs.


scripts/js/commands/checkKatexRender.ts
```diff
@@ -0,0 +1,269 @@
...
+      // We skip math expressions that are not inline
+      if (isKatexDisplay) {
+        continue;
+      }
+
+      const box = await mathExpr.boundingBox();
+      if (box && box.width >= MIN_WIDTH) {
+        const content = await mathExpr.textContent();
+        errors.push(
+          `  -  Error found. Approximately preview of the expression:\n\t $${content}}$`,
```
frankharkins:
```suggestion
          `  -  Error found. Approximate preview of the expression:\n\t $${content}}$`,
```

scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/cell_output_data.py
```diff
@@ -52,6 +53,16 @@ def remove_circuit_drawing_html(output_data: dict) -> bool:
             return True
     return False
 
+def remove_inline_katex_expression(output_data: dict) -> bool:
+    """Convertes inline katex expressions into display mode if needed and returns True if any changes were made."""
+    if latex := output_data.get("text/latex"):
+        if latex.startswith('$\\displaystyle') and latex.endswith('$'):
+            latex = latex.replace('$\\displaystyle', '$$\n') # Replace the first '$'
+            latex = re.sub(r'\$$', '\n$$', latex) # Replace the last '$'
+            output_data["text/latex"] = latex
```
frankharkins: Is starting with `\\displaystyle` guaranteed? If not, I think it's safe enough that we're in the `text/latex` output and have starting and ending `$`.

Also might be even safer to strip the output too:

```suggestion
    if latex := output_data.get("text/latex").strip():
        if latex.startswith('$\\displaystyle') and latex.endswith('$'):
            latex = latex.replace('$\\displaystyle', '$$\n') # Replace the first '$'
            latex = re.sub(r'\$$', '\n$$', latex) # Replace the last '$'
            output_data["text/latex"] = latex
```

frankharkins (approved):
Awesome, thank you!

scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/cell_output_data.py
arnaucasau: > Is starting with \\displaystyle guaranteed?

It is for all the conflicting instances we have in our docs, but not necessarily for all of them in the future. I found that having the `\\displaystyle` does not affect how we render the expression, so I'm now replacing just the opening `$` to be more generic.

I also added the `strip()`. Thanks for the review!

docs/api/qiskit-addon-cutting/0.9/qiskit-addon-cutting.mdx
Eric-Arellano, line 1: To double check, you updated the Box artifact?

scripts/js/commands/checkKatexRender.ts
```diff
@@ -0,0 +1,269 @@
...
+    })
+    .option("qiskit-release-notes", {
+      type: "boolean",
+      description:
+        "Check the pages in the `docs/api/qiskit/release-notes` folder.",
+    })
+    .option("ignored-files", {
+      type: "boolean",
+      description:
+        "Check the pages the IGNORED_FILES list. Useful to fix the pages",
```
Eric-Arellano:
```suggestion
        "Check the pages in the IGNORED_FILES list. This is useful to fix the pages",
```

scripts/js/commands/checkKatexRender.ts
```diff
@@ -0,0 +1,269 @@
...
+
+    numFilesChecked++;
+  }
+
+  await browser.close();
+
+  if (!allGood) {
+    console.log(
+      "\nThe previous inline math expressions are too long. Please run `./fix` to fix the output of the notebooks' code cells and/or manually convert the expressions into a 'Katex Display' by using `$$` instead of `$` in a new line.\n",
+      "Example of the convertion:\n",
```
Eric-Arellano:
```suggestion
      "Example of the conversion:\n",
```

scripts/js/commands/checkKatexRender.ts
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    numFilesChecked++;
+  }
+
+  await browser.close();
+
+  if (!allGood) {
+    console.log(
+      "\nThe previous inline math expressions are too long. Please run `./fix` to fix the output of the notebooks' code cells and/or manually convert the expressions into a 'Katex Display' by using `$$` instead of `$` in a new line.\n",
```
Eric-Arellano: 'Katex display' is a wonky internal name. Better is to probably call it something like "manually convert the inline expressions into math blocks"

scripts/js/commands/checkKatexRender.ts
```diff
@@ -0,0 +1,269 @@
...
+      "Example of the convertion:\n",
+      "--------------------------\n",
+      "$[sample_{0}observable_{0}, \ldots, sample_{0}observable_{N-1}, sample_{1}observable_{0}, \ldots, sample_{M-1}observable_{N-1}]$\n",
+      "->\n",
```
Eric-Arellano: Nit to maybe use a down arrow `↓` and new lines between each section

scripts/js/commands/checkKatexRender.ts
```diff
@@ -0,0 +1,269 @@
...
+async function checkPage(page: Page, pathname: string): Promise<string[]> {
+  const errors: string[] = [];
+  try {
+    await page.goto(`http://localhost:3000/${pathname}`);
+
+    const allKatex = page.locator("span.katex");
+    const count = await allKatex.count();
+
+    for (let i = 0; i < count; i++) {
+      const mathExpr = allKatex.nth(i);
```
Eric-Arellano: You can simplify by using Locator.all() https://playwright.dev/docs/api/class-locator#locator-all

scripts/js/commands/checkKatexRender.ts
```diff
@@ -0,0 +1,269 @@
...
+      // We skip math expressions that are not inline
+      if (isKatexDisplay) {
+        continue;
+      }
+
+      const box = await mathExpr.boundingBox();
+      if (box && box.width >= MIN_WIDTH) {
+        const content = await mathExpr.textContent();
+        errors.push(
+          `  -  Error found. Approximate preview of the expression:\n\t $${content}}$`,
```
Eric-Arellano: Consider escaping the outer $ with `\` to make it more clear what's happening.

scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/cell_output_data.py
```diff
@@ -52,6 +53,24 @@ def remove_circuit_drawing_html(output_data: dict) -> bool:
             return True
     return False
 
+def remove_inline_katex_expression(output_data: dict) -> bool:
+    """Convertes inline katex expressions into display mode if needed and returns True if any changes were made."""
+    if "text/latex" in output_data:
```
Eric-Arellano: Nit to invert the conditional to early return

scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/cell_output_data.py
```diff
@@ -52,6 +53,24 @@ def remove_circuit_drawing_html(output_data: dict) -> bool:
...
+        latex = output_data.get("text/latex").strip()
+
+        # We skip displayed equiations because they already have the desired style
```
Eric-Arellano:
```suggestion
        # We skip displayed equations because they already have the desired style
```

scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/cell_output_data.py
```diff
@@ -52,6 +53,24 @@ def remove_circuit_drawing_html(output_data: dict) -> bool:
             return True
     return False
 
+def remove_inline_katex_expression(output_data: dict) -> bool:
+    """Convertes inline katex expressions into display mode if needed and returns True if any changes were made."""
```
Eric-Arellano:
```suggestion
    """Converts inline katex expressions into display mode if needed and returns True if any changes were made."""
```

scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/cell_output_data.py
```diff
@@ -52,6 +53,24 @@ def remove_circuit_drawing_html(output_data: dict) -> bool:
...
+    if "text/latex" in output_data:
+        latex = output_data.get("text/latex").strip()
+
+        # We skip displayed equiations because they already have the desired style
+        if latex.startswith('$$'):
+            return False
+
+        # We transforms inline katex expressions
```
Eric-Arellano:
```suggestion
        # We transform inline katex expressions
```

scripts/js/commands/checkKatexRender.ts
```diff
@@ -0,0 +1,269 @@
...
+    }
+  }
+
+  return filesWithMath.map((f) =>
+    // The pages must be accesses without the file extensions
+    removeFileExtension(f),
+  );
+}
+
+main().then(() => process.exit());
```
Eric-Arellano, line 251: I would expect this to mess up the tests because importing the file runs this code. This is why Python has `if __name__ == "__main__"`. You can see similar code we have in the closed source infra repository for redirects. Or, the approach we normally take in this repo is to move the helper functions to `lib/`

scripts/js/commands/checkKatexRender.test.ts
```diff
@@ -0,0 +1,62 @@
...
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+import { expect, test } from "@playwright/test";
+
+import { hasInlineMath, removeFileExtension } from "./checkKatexRender";
+
+test("hasInlineMath()", () => {
+  const md1 = `
```
Eric-Arellano: Nit that it'd help to name these more clearly. Maybe `baselineMd`, `mathBlock`, and `inlineMath`

docs/api/qiskit-addon-cutting/0.9/qiskit-addon-cutting.mdx
arnaucasau: Yes, I created a PR on the original repo and modified both artifacts

Eric-Arellano (approved):

PR #3409, opened 2025-06-23
[nb-tester]: Handle forks better

frankharkins:
Previously, `save_account` would happily accept an empty string for the `token`, but this seems to have changed. Since some notebooks can be tested without a valid account, this PR adds a try/except to continue even if account setup fails. The existing script will give an appropriate error message if any notebooks fail on a fork.

### DIFF
--- a/.github/actions/set-up-notebook-testing/action.yml
+++ b/.github/actions/set-up-notebook-testing/action.yml
@@ -37,12 +37,22 @@ runs:
       shell: python
       run: |
         from qiskit_ibm_runtime import QiskitRuntimeService
-        QiskitRuntimeService.save_account(
-            channel="ibm_cloud",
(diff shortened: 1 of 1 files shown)


Eric-Arellano (commented):
Thoughts on this alternative approach? 

In Python, check if the token is the empty string and don't attempt to run `save_account`. You could log a warning with the message you currently have.

Otherwise, if the token is set, still eagerly error. We want to error if there is an unexpected problem like maybe the token has expired.

My concern with the current approach is it could mask legitimate problems. While we can catch it through later failures, it's better to fail eagerly when possible.

frankharkins:
Yep I agree that's better, makes the code simper too.

Eric-Arellano (approved):
Thanks!

Issue #3413, opened 2025-06-23
Problem in `docs/guides/transpiler-stages.ipynb`


frankharkins:
The "Transpiler stages" notebook raises an error with the latest dependencies.

Here's the offending cell:

```python
ghz = QuantumCircuit(5)
ghz.h(0)
ghz.cx(0, range(1, 5))


# Use fake pulse-enabled backend
backend = FakeWashingtonV2()

# Run with optimization level 3 and 'asap' scheduling pass
pass_manager = generate_preset_pass_manager(
    optimization_level=3,
    backend=backend,
    scheduling_method="asap",
    seed_transpiler=1234,
)


circ = pass_manager.run(ghz, backend)
circ.draw(output="mpl", idle_wires=False)
```

And the traceback:

```pytb
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[7], line 19
     10 pass_manager = generate_preset_pass_manager(
     11     optimization_level=3,
     12     backend=backend,
     13     scheduling_method="asap",
     14     seed_transpiler=1234,
     15 )
     18 circ = pass_manager.run(ghz, backend)
---> 19 circ.draw(output="mpl", idle_wires=False)

File ~/work/documentation/documentation/.tox/py311/lib/python3.11/site-packages/qiskit/circuit/quantumcircuit.py:3808, in QuantumCircuit.draw(self, output, scale, filename, style, interactive, plot_barriers, reverse_bits, justify, vertical_compression, idle_wires, with_layout, fold, ax, initial_state, cregbundle, wire_order, expr_len)
   3805 # pylint: disable=cyclic-import
   3806 from qiskit.visualization import circuit_drawer
-> 3808 return circuit_drawer(
   3809     self,
   3810     scale=scale,
   3811     filename=filename,
   3812     style=style,
   3813     output=output,
   3814     interactive=interactive,
   3815     plot_barriers=plot_barriers,
   3816     reverse_bits=reverse_bits,
   3817     justify=justify,
   3818     vertical_compression=vertical_compression,
   3819     idle_wires=idle_wires,
   3820     with_layout=with_layout,
   3821     fold=fold,
   3822     ax=ax,
   3823     initial_state=initial_state,
   3824     cregbundle=cregbundle,
   3825     wire_order=wire_order,
   3826     expr_len=expr_len,
   3827 )

[...]

PR #3416, opened 2025-06-24
Fix "Transpiler stages" notebook

frankharkins:
From my https://github.com/Qiskit/documentation/issues/3413#issuecomment-2998012384:

> Looking at [`run`](https://docs.quantum.ibm.com/api/qiskit/qiskit.transpiler.PassManager#run), I think we're passing the backend object to the `output_name` parameter, which expects a string. I guess this was always a mistake – we already pass the `backend` object in `generate_preset_pass_manager` – Qiskit must just have been ignoring the `output_name` parameter until recently.


Eric-Arellano (approved):
🚀

PR #3432, opened 2025-06-26
Delete stale PR previews

### LINKED ISSUE #3412: Expire PR previews after a week
Relates to https://github.com/Qiskit/documentation/issues/3342. 

One of the issues with our PR previews is that it takes a long time for GitHub Pages to deploy our site. We suspect this is due to how many subsites we have, each of which are fairly large. So, we want to reduce the number of sites we have, which we currently only can do by closing out PRs.

One idea is to expire old previews after 7 days. The content team signed off on this.

We would update https://github.com/Qiskit/documentation/blob/main/.github/workflows/pr-cleanup.yml to start also expiring old builds. This should be based on the deployment's metadata, rather than the PR's metadata; that is important because it's possible to restart a GH Actions job that results in a new deployment despite no new commits. TBD how to get the deployment's metadata.

frankharkins:
Attempt to improve GitHub pages deployment time by expiring PR previews that have not been updated in a week.

This approach uses Git's last modification date to return the timestamp of the most recent commit that affects each PR folder. This timestamp will correspond to the last time the PR was updated. If a PR preview is deleted, restarting the PR preview job will commit a new folder, starting a new seven day active period. The only downside I can see is that restarting a job that doesn't modify the preview before the preview is cleaned up will have no effect, so writers can't pre-emptively re-trigger a job to keep a preview alive; they can only do this after it's deleted. If that's a problem, we could maybe commit a timestamp file to each preview folder when we create them.

I propose starting by just logging, rather than deleting so we can check it's behaving correctly. I'll then follow up with a PR to start deleting things.

---

Closes #3412

### DIFF
--- a/scripts/pr-previews/cleanup.py
+++ b/scripts/pr-previews/cleanup.py
@@ -15,6 +15,7 @@
 import json
 import logging
 import shutil
+import time
 from pathlib import Path
 
 logger = logging.getLogger(__name__)
@@ -29,6 +30,8 @@ from utils import (
 
 INITIAL_COMMIT = "499a5040585d02593cdd8237e19c9ee4a84ae126"
 
+THREE_DAYS_IN_SECONDS = 60 * 60 * 24 *3
+PR_EXPIRATION_TIME = THREE_DAYS_IN_SECONDS
 
 def main() -> None:
     setup_git_account()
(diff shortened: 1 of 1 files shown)


scripts/pr-previews/cleanup.py
```diff
@@ -64,14 +67,25 @@ def get_active_pr_folders() -> set[str]:
...
-        if folder.name not in active_pr_folders:
-            logger.info(f"Deleting {folder}")
+        if is_closed(folder.name):
+            logger.info(f"Deleting {folder} as PR is closed")
             shutil.rmtree(folder)
-
+            continue
+        if is_stale(folder.name):
+            logger.info(f"Would delete {folder} as it is stale")
+            # shutil.rmtree(folder)
```
arnaucasau, line 88: I like the idea of starting by just logging. Could you add a comment on top of this line for us to remember why it's commented?

arnaucasau (approved):
Awesome, thanks for working on this!

scripts/pr-previews/cleanup.py
frankharkins:
```suggestion
            # TODO (#3433) Change the log message and uncomment the following line
            # shutil.rmtree(folder)
```

scripts/pr-previews/cleanup.py
```diff
@@ -64,14 +67,26 @@ def get_active_pr_folders() -> set[str]:
...
 
+def is_stale(folder_name: str) -> bool:
+    # All time measured in seconds, from the unix epoch
+    current_timestamp = time.time()
+    last_modified_timestamp = int(run_subprocess(["git", "log", "-n", "1", "--format=%at", folder_name]).stdout)
+    return (current_timestamp - last_modified_timestamp) > PR_EXPIRATION_TIME_SECONDS
 
 def delete_closed_pr_folders() -> None:
     active_pr_folders = get_active_pr_folders()
+    is_closed = lambda folder: folder not in active_pr_folders
```
Eric-Arellano, line 78: Style nit: more idiomatic is to do this:

```python
for folder in Path(".".).glob("pr-*"):
    is_closed = folder.name not in active_pr_folders
    if is_closed:
        ...
```

Lambdas are generally discouraged in Python due to PEP 8's style guide, particularly assigning them to a variable. Better is to use `def`, which is allowed as a nested function. Or, refactor to not need a function. https://stackoverflow.com/questions/25010167/e731-do-not-assign-a-lambda-expression-use-a-def.

PR #3468, opened 2025-07-01
Stop using the legacy app in previews

arnaucasau:
This PR deletes the `platform` option from the `start` script. This is because we are removing the legacy app in inner source and from the preview app. It also simplifies the render checker to only verify cloud pages.

This change will be accompanied by a follow-up that finishes removing missing parts once we don't use any platform specific tag in the content.

### DIFF
--- a/scripts/js/commands/checkPagesRender.ts
+++ b/scripts/js/commands/checkPagesRender.ts
@@ -25,25 +25,6 @@ const LEGACY_ONLY_PAGES: Set<string> = new Set([
   "docs/guides/estimate-job-run-time.mdx",
 ]);
 
-// These are expected to 404 in the legacy app due to being cloud only.
-const CLOUD_ONLY_PAGES: Set<string> = new Set([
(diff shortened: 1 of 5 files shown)


scripts/js/commands/checkPagesRender.ts
```diff
@@ -220,20 +188,9 @@ async function determineFilePaths(args: Arguments): Promise<string[]> {
...
-    return (
-      !CLOUD_ONLY_PAGES.has(page) &&
-      !page.startsWith("docs/api") &&
-      !page.startsWith("docs/tutorials") &&
-      !page.startsWith("learning")
-    );
-  }
-
+function filterPlatformSpecificPage(page: string) {
   return !LEGACY_ONLY_PAGES.has(page);
```
arnaucasau, line 195: This is necessary until we remove the legacy only pages. Otherwise, they could 404 and make CI to fail

Eric-Arellano (approved):
👋

PR #3487, opened 2025-07-02
In SQD tutorial, scale up from 6-31g (32 qubits) to cc-pvdz (52 qubits)

kevinsung:
This was requested by @nathanearnestnoble .

The configuration recovery here took 14 minutes on my computer and the energy error is now 69 mH.


nathanearnestnoble (changes requested):
I think we just need to make minor change in first paragraph highlighting the new number of qubits.   Otherwise, I think it is good!

kevinsung:
> I think we just need to make minor change in first paragraph highlighting the new number of qubits. Otherwise, I think it is good!

Done

abbycross (approved):

PR #3479, opened 2025-07-01
Add cut bell pair benchmarking for dynamic circuits

miamico:
Notebook adding a demonstration of the new dynamic circuit capabilities. In particular, dynamic circuit performance is benchmarked device-wide by preparing 4-qubit chains where Bell pairs on the middle qubits are consumed through mid-circuit measurement and feedforward to create a Bell pair on the edge qubits.

PR #3493, opened 2025-07-03
Update long range entanglement tutorial

miamico:
This notebook brings substantial updates to the long-range entanglement tutorial to support the novel capabilities. In particular:

- [ ] removed post-processing technique and text
- [ ] removed randomized certification of fidelity and corresponding text
- [ ] limit number of qubits measured to amount currently supported (16)
- [ ] use hellinger fidelity of the bell pair created by the CNOT between the terminal qubits of the chain as measure of performance


docs/tutorials/long-range-entanglement.ipynb
```diff
@@ -192,65 +180,22 @@
...
-  },
   {
    "cell_type": "markdown",
    "id": "11e35681-d7d0-4d5d-8764-5517181d91d0",
    "metadata": {},
    "source": [
     "#### Set primary parameters\n",
     "\n",
-    "In this section are definitions for some common parameters that you will use later. You'll need to specify these parameters for a particular backend. In order to do so, you will need an account on [IBM Quantum&trade; Platform](https://quantum-computing.ibm.com/). More details on how to initialize your account can be found in the [documentation](/docs/guides/setup-channel#set-up-to-use-ibm-quantum-platform)."
+    "In this section are definitions for some common parameters that you will use later. You'll need to specify these parameters for a particular backend. In order to do so, you will need an account on [IBM Quantum&trade; Platform](https://quantum.cloud.ibm.com/). More details on how to initialize your account can be found in the [documentation](/docs/guides/setup-channel#set-up-to-use-ibm-quantum-platform)."
```
abbycross:
```suggestion
    "In this section are definitions for some common parameters that you will use later. You'll need to specify these parameters for a particular backend. In order to do so, you will need an account on [IBM Quantum&reg; Platform](https://quantum.cloud.ibm.com/). More details on how to initialize your account can be found in the [documentation](/docs/guides/setup-channel#set-up-to-use-ibm-quantum-platform)."
```

docs/tutorials/long-range-entanglement.ipynb
```diff
@@ -1096,597 +665,88 @@
...
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "210a251f-a4b0-43f1-b081-67da215719bb",
-   "metadata": {},
-   "source": [
-    "## Step 3: Execute using Qiskit primitives\n",
-    "In this step you execute the experiment on the specified backend. A lightweight transpilation is done before submission to ensure that the circuits have all the physical parameters needed for execution on the device. This is done in the `submit_circuits` function, which also splits the circuits into smaller batches to be submitted to the device."
+    "Because you have already specified the physical qubit layout and built the circuits with a line topology in mind, there is no need to further optimize the circuits and we just make sure that the circuits are fully compatible with the chosen backend."
```
abbycross:
```suggestion
    "Because you have already specified the physical qubit layout and built the circuits with a line topology in mind, there is no need to further optimize the circuits; we only need to make sure that the circuits are fully compatible with the chosen backend."
```

abbycross (approved):
Thanks!

jyu00:
I guess I'm a bit too late, but this should also be updated with the same comment I left for #3479 for early access - i.e. don't use `dynamic_circuits=True` (because it'll always return an empty list), and explicitly call out that adding `IfElseOp` & `gen3-experimental` is only necessary during early access.

Issue #3474, opened 2025-07-01
Instances page out-of-date


abbycross:
https://quantum.cloud.ibm.com/docs/en/guides/instances has some outdated links to the old platform, and should also be checked for up-to-date authentication info.

abbycross:
Fixed by https://github.com/Qiskit/documentation/pull/3572

PR #3503, opened 2025-07-07
Remove legacy content and both legacy/cloud tags

abbycross:


docs/guides/_toc.json
```diff
@@ -18,13 +18,11 @@
             },
             {
               "title": "Set up an IBM Quantum channel",
-              "url": "/docs/guides/setup-channel",
-              "platform": "legacy"
```
Eric-Arellano, line 22: Legacy only pages should be deleted entirely

abbycross: oops didn't delete the title!! good catch

docs/guides/_toc.json
```diff
@@ -597,8 +572,7 @@
             },
             {
               "title": "Instances",
-              "url": "/docs/guides/instances",
-              "platform": "legacy"
+              "url": "/docs/guides/instances"
```
Eric-Arellano: Should be deleted entirely

abbycross: This gives me pause - maybe this was an error to be a legacy-only page?  I think we want a page about instances relevant to the new platform, and the /docs/guides/instances file does have content relevant to Cloud in it. But we also have a [PR open](https://github.com/Qiskit/documentation/issues/3474) to review some of the language in it, since it refers to things like the Standard Plan and the ibm-q/open/main instance (even in the Cloud content). I think we should err on the side of keeping the file unless @beckykd @javabster you have feedback?

arnaucasau: I agree with Abby about keeping the page, but notice that the entry is currently being duplicated. We have two entries for the same page because it was placed in a different section depending on the platform you were in.

Cloud entry:
```
{
  "title": "Create and manage instances",
  "url": "/docs/guides/instances",
  "platform": "cloud"
},
```

Legacy entry:
```
{
  "title": "Instances",
  "url": "/docs/guides/instances",
  "platform": "legacy"
},
```

Given that we want to keep the cloud content, I think Eric is right about deleting this entry from the TOC (not the page) entirely.

arnaucasau (commented):
Thanks for working on this! It will help unblocking more cleanup of our code :)

FYI the `docs/guides/manage-cost.mdx` page is still having `CloudContent` and `LegacyContent` tags that should be removed.

learning/courses/quantum-chem-with-vqe/ground-state.ipynb
```diff
@@ -414,14 +414,6 @@
...
    "metadata": {},
    "outputs": [
-    {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
-      "/home/porter284/.pyenv/versions/3.11.12/lib/python3.11/site-packages/qiskit_ibm_runtime/qiskit_runtime_service.py:1129: UserWarning: IBM Quantum Platform is moving, and this version will be sunset on July 1. To get started on the new platform, read the migration guide at https://docs.quantum.ibm.com/migration-guides/classic-iqp-to-cloud-iqp\n",
-      "  warnings.warn(warning_message)\n"
-     ]
-    },
```
frankharkins, line 424: I think we should leave this in? Would a user still see this warning when running the code examples?

docs/guides/_toc.json
abbycross: Aha! That makes sense. I did not notice the duplication. Thanks @arnaucasau !

learning/courses/quantum-chem-with-vqe/ground-state.ipynb
abbycross: We had recently scrubbed the courses of similar errors, but if it's instructional at all to leave it in, that makes sense - let's check with @christopherporter1.

abbycross:
Many thanks to @Eric-Arellano @arnaucasau @frankharkins for your careful look at this PR!

learning/courses/quantum-chem-with-vqe/ground-state.ipynb
christopherporter1: @abbycross Thanks for checking. I don't even know why that warning was thrown. I have not seen it in other files. I looked at the file and it seems like someone already removed it. That's fine.

abbycross: Chris is good to remove it.

PR #3509, opened 2025-07-09
Correctly normalize notebooks with partially-extracted outputs

### LINKED ISSUE #3440: Notebook normalizer can't handle notebooks with partially extracted outputs
We currently don't handle the current situation well:
1. Notebook has multiple image outputs and is normalized (with `./fix`)
2. Writer runs only _some_ cells of the notebook, regenerating only _some_ image outputs
3. Writer runs `./fix` on the same notebook

Since the normalizer detects images that need extracting, it'll delete the existing image folder and extract the new images. This is a problem as the notebook will contain `<Image />` tags that reference images that don't exist.

We should generally discourage only running some cells of a notebook, but our normalization script should be able to handle it.

Until we fix this, the workaround is to either:
1. Re-run the notebook from start to finish (preferably [using `tox`](https://github.com/Qiskit/documentation?tab=readme-ov-file#execute-notebooks))
2. If that's not an option, restore the deleted images using Git

frankharkins:
Fixes #3440

Rather than deleting the whole image folder and re-write, this PR finds images that don't appear in cell outputs and removes them.

### DIFF
--- a/scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/__init__.py
+++ b/scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/__init__.py
@@ -11,11 +11,10 @@
 # that they have been altered from the originals.
 
 import argparse
-import shutil
 from dataclasses import dataclass
 from itertools import chain
 from pathlib import Path
-from typing import TypeGuard, ClassVar
+from typing import TypeGuard, ClassVar, Iterator
 
 import nbformat
 
@@ -71,9 +70,11 @@ def main():
         if args.check:
             continue
 
-        ensure_exists_and_empty(images_folder)
+        images_folder.mkdir(exist_ok=True, parents=True)
         for image in result.images:
             image.write()
+        for image in get_unused_images(images_folder, nb):
+            image.unlink()
         nbformat.write(result.nb, nb_path)
         print(f"✍️ Written '{nb_path}' and {len(result.images)} image(s)")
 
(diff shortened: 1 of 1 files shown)


scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/__init__.py
```diff
@@ -125,14 +126,16 @@ def normalize_notebook(
                 # keep looking at other cells
                 return NormalizationNeeded(nb=nb, images=[])
 
-    if change_made:
+    has_unused_images = any(image for image in get_unused_images(image_folder, nb))
```
Eric-Arellano, line 129: You can speed up this algorithm by using a set difference. Right now, it's O(n^2) because you have a nested for loop: for every image, loop through every output cell. Instead, pre-compute:

- all images in the folder
- all used image names in the notebook

Then, take the set difference of `all_image_files - used_images`. O(n)

frankharkins: That's cool, I wasn't aware of that approach.

Implementing it here does have a trade off: We'd need to use a regex (or similar) to extract the filenames from the outputs, which means we need to make assumptions about the form of the mdx component (`<Image src="...`). This should be fine for now, but it could cause problems if we ever change the MDX component (e.g. changing arrangement of properties).

Do you think the better scaling is worth it?

Eric-Arellano: Ah, touché. I didn't think about that being more brittle with extracting the image names. Your approach sounds good!

Eric-Arellano (approved):
Thanks for fixing!

scripts/notebook-normalizer/qiskit_docs_notebook_normalizer/__init__.py
frankharkins: Sounds good, I'll definitely use that tip in the future though

Issue #3525, opened 2025-07-14
[EPIC] New guide pages for GA


kaelynj:
As part of our efforts in upgrading the platform, we need to add several new pages documenting GA features. Many of these are based on templates that can already be found on the IBM Cloud platform and we'll keep track of each of these in this epic. A lot of this work has been started already so content authors refer to the epic that exists internally for the associated links and templates to work from.

## Pages to add

- [x] Virtual Private Endpoints (@abbycross) PR: https://github.com/Qiskit/documentation/pull/3527
- [x] Data retention (@kaelynj) PR: https://github.com/Qiskit/documentation/pull/3529
- [x] Responsibilities matrix (@abbycross) #3571 
- [x] High availability / disaster recovery (@beckykd) https://github.com/Qiskit/documentation/pull/3545
- [x] Securing your data (@abbycross ) PR: https://github.com/Qiskit/documentation/pull/3530
  - Examples:
    - https://cloud.ibm.com/docs/secrets-manager?topic=secrets-manager-secrets-manager-ha-dr
    - https://cloud.ibm.com/docs/vpc?topic=vpc-ha-dr-vpc
    - https://cloud.ibm.com/docs/codeengine?topic=codeengine-ha-dr
    - https://test.cloud.ibm.com/docs-draft/resiliency?topic=resiliency-postgresql-ha-dr
- [x] New logging topic (already started in #3294 ) (@beckykd)
- [x] Accessibility page (@beckykd) PR #3555 
- [x] Documenting observability (@abbycross) PR https://github.com/Qiskit/documentation/pull/3724
- [ ] External metrics PR #3757 

### Pages structure
We determined that these pages should have the following structure

#### Footer > Accessibility (`/docs/accessibility` page)
- Accessibility

#### Footer > Security and compliance (new section) (`/docs/security` folder)

- Private connectivity with VPE and VPC
- Securing your data
- Platform roles and responsibilities (responsibility matrix)
- High availability / Disaster recovery
- Data retention

#### Guides left nav > IBM Quantum Platform

- Logging

PR #3554, opened 2025-07-16
Merge GA feature branch 

kaelynj:
Related to #3525

This pr merges the feature branch we were initially using for the GA pages. We've decided to just merge directly to main instead.

This PR also makes some of our check code more generic to handle new pages/folders automatically.


scripts/js/commands/checkQiskitBotFiles.ts
```diff
@@ -43,15 +43,22 @@ const ALLOWED_OWNERLESS_FILES = new Set([
   "docs/migration-guides/qiskit-1.0",
   "docs/migration-guides/qiskit-1.0-features",
   "docs/migration-guides/qiskit-1.0-installation",
+  // TODO: everything needs an owner
```
Eric-Arellano, line 46: Can you please either create an issue or open a PR today to fix this?

Eric-Arellano: https://github.com/Qiskit/documentation/issues/3557

PR #3512, opened 2025-07-09
Updates instances guide

kt474:
Not sure the best way to handle this but the [instances guide](https://quantum.cloud.ibm.com/docs/en/guides/instances) still refers to the legacy platform. This is confusing because the rest of the guide is only for the new platform. 

If bigger changes to this guide are required then feel free to close this PR

related to #3474

PR #3572, opened 2025-07-18
Update instances guide

kt474:
Reopening https://github.com/Qiskit/documentation/pull/3512 from a branch off of the main repo instead of my fork.


docs/guides/instances.ipynb
```diff
@@ -35,9 +35,8 @@
...
-    "\n",
-    "Those using the Open Plan instance have up to 10 minutes total of quantum time per 28-day rolling window. Open Plan users can track their usage on the [Platform dashboard,](https://quantum.ibm.com/) [Workloads,](https://quantum.ibm.com/workloads) and [Account](https://quantum.ibm.com/account) pages.\n",
+    "By default, users who sign up for an IBM Quantum account are assigned to the Open Plan.\n",
+    "Open Plan users can run quantum circuits on the world's best QPUs for free (up to 10 minutes quantum time per 28-day rolling window). The QPU usage can be tracked on the [Platform dashboard](https://quantum.cloud.ibm.com/) and the [Workloads](https://quantum.cloud.ibm.com/workloads) pages.\n",
```
kt474, line 38: I don't know if "By default, users who sign up for an IBM Quantum account are assigned to the Open Plan" is actually true. Is there any other content we want to have in this "Open Plan" section? @beckykd @abbycross 

We have the same thing for the standard plan in the section below too: 
"By default, users who sign up for an IBM Quantum account are assigned to the Standard Plan"

abbycross: Hm, I think you're right - it's unclear and confusing, given the necessary setup steps. And I definitely don't think it's true for Standard (which now needs to be called Pay-As-You-Go)

abbycross: If I think of being assigned by default translating to, I am able to make use of this plan once I have an account, it makes sense - but I don't think it communicates that clearly.I think we could strike the line from both sections

PR #3583, opened 2025-07-22
Use command to view available QPUs

### LINKED ISSUE #3537: Issue from support: Describe how to see available QPUs via code

https://quantum.cloud.ibm.com/docs/en/guides/cloud-setup#set-up-to-use-ibm-cloud

### Select all that apply

- [ ] new content request
- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

What is  the string to view the same instances I see here:
https://eu-de.quantum.cloud.ibm.com/instances

service = QiskitRuntimeService(token=token,channel=channel,instance=instance) (if you didn't save credentials)

Or to save the credentials

```
# save account
service = QiskitRuntimeService.save_account(
  # No channel needs to be specified, ibm_quantum_platform is the default
  token=token,
  region="eu-de", # Optional
  plans_preference=['Open',...], #Optional
  set_as_default=True  #Optional
  )
```

We need to post about the region switcher and the instances specific to regions. Should be useful to share some screenshots

**This may have been fixed in the latest updates to the getting started topics/ FAQ. **

-----------------------------

So at the moment, we have access to multiple instances, which provide different QPUs. Is there a way to query (via code) the backends corresponding to a specific instance. For example, ibm_aachen is accessed through an EU region instance but not a US instance. 

beckykd:
closes #3537

### DIFF
--- a/docs/guides/cloud-setup.mdx
+++ b/docs/guides/cloud-setup.mdx
@@ -11,7 +11,7 @@ If you have an invitation to join an account, follow the steps in [Set up your I
 
 You can run workloads on IBM&reg; quantum processing units (QPUs) by setting up an account on IBM Cloud&reg;. You will also need to [install and set up Qiskit and Qiskit Runtime](/docs/guides/install-qiskit#local).
 
-Your user account is associated with one or more [instances](./instances), identified by _Cloud Resource Names_ (CRNs) that give access to IBM Quantum services. Additionally, a unique API token (also called a key) is assigned to each account, allowing for IBM Quantum access from Qiskit.  For instructions to choose a specific instance, see [Connect to an instance](./instances#connect-instance).
+Your user account is associated with one or more [instances,](./instances) identified by _Cloud Resource Names_ (CRNs) that give access to IBM Quantum services. Additionally, a unique API token (also called a key) is assigned to each account, allowing for IBM Quantum access from Qiskit.  For instructions to choose a specific instance, see [Connect to an instance](./instances#connect-instance).
 
     <Admonition type="note">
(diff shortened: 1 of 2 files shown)


docs/guides/get-qpu-information.ipynb
```diff
@@ -44,7 +44,11 @@
    "source": [
     "## List backends\n",
     "\n",
-    "To view the backends you have access to, you can either view a list on the [Compute resources page,](https://quantum.cloud.ibm.com/computers) or you can use the [`QiskitRuntimeService.backends()`](../api/qiskit-ibm-runtime/qiskit-runtime-service#backends) method. This method returns a list of [`IBMBackend`](../api/qiskit-ibm-runtime/ibm-backend) instances:"
+    "To view the backends you have access to, you can either view a list on the [Compute resources page,](https://quantum.cloud.ibm.com/computers) or you can use the [`QiskitRuntimeService.backends()`](../api/qiskit-ibm-runtime/qiskit-runtime-service#backends) method. This method returns a list of [`IBMBackend`](../api/qiskit-ibm-runtime/ibm-backend) instances:\n",
+    "\n",
+    "<Admonition type=\"note\">\n",
+    "If you are logged in to a specific instance, only the backends available to you on that instance are returned.\n",
```
jyu00: This applies to the how you initialize your `QiskitRuntimeService` as well. For example, if you do `QiskitRuntimeService(instance=...)`, then `QiskitRuntimeService.backends()` would return only backends that can be accessed via that instance (with or without filters).

jyu00 (approved):

PR #3593, opened 2025-07-24
Migrate the link checker to use the remark parser

arnaucasau:
This PR migrates the link checker to use the `remark` parser instead of the `rehype` parser. Rehype is meant to take html as input whereas Remark expects markdown.

Turns out that the only visitor that was extracting links before was the one looking for `text` nodes. After this change, we can remove the visitor and rely on the node types `link` and `mdxJsxTextElement` for `<a>` tags. It also allows us to remove the `markdown-link-extractor` dependency.

This new approach also finds links that we weren't checking before (it even found some broken ones). The previous approach was checking 1638 links on our non-api docs, and after this PR we will check 1651 links.


learning/courses/variational-algorithm-design/instances-and-extensions.ipynb
```diff
@@ -806,7 +806,7 @@
...
+    "One of the main issues with VQE is the multiple calls to a quantum computer that are required to obtain the parameters for each step, including $k$, $k-1$, etc. This is especially problematic when access to quantum devices is queued. While a [`Session`](/docs/api/qiskit-ibm-runtime/session) can be used to group multiple iterative calls, an alternative approach is to use sampling. By utilizing more classical resources, we can complete the full optimization process in a single call. This is where [Quantum Sampling Regression](https://arxiv.org/pdf/2012.02338) comes into play. Since access to quantum computers is still a low-offer/high-demand commodity, we find this trade-off to be both possible and convenient for many current studies. This approach harnesses all available classical capabilities while still capturing many of the inner workings and intrinsic properties of quantum computations that do not appear in simulation.\n",
```
arnaucasau: @abbycross @beckykd @christopherporter1 I found this broken link that our link checker wasn't verifying. Do you think the API docs is the best place to link here? We also have https://quantum.test.cloud.ibm.com/docs/en/guides/execution-modes#session-mode

scripts/js/lib/links/extractLinks.ts
```diff
@@ -116,6 +121,13 @@ export async function parseFile(filePath: string): Promise<ParsedFile> {
     };
   }
 
+  if (filePath.endsWith(".json"))
```
arnaucasau, line 124: We need to early return with JSON files because otherwise the unified plugin will break. We weren't checking any JSON file before either, so this should preserve the status quo.

We could extend the link checker to validate the `_toc.json` files in the future

learning/courses/variational-algorithm-design/instances-and-extensions.ipynb
abbycross: How about https://quantum.cloud.ibm.com/docs/en/guides/run-jobs-session?

Eric-Arellano (approved):
Thank you!

PR #3597, opened 2025-07-24
fix new pill checker to handle folder

### LINKED ISSUE #3539: New pill expiration check doesn't detect stale folders
We have this folder marked `isNew`:

https://github.com/Qiskit/documentation/blob/582c84da03b10059442fc2d477ee82803acadcba/docs/guides/_toc.json#L95-L97

It's been over 3 months since it was set up, so it should be marked as expired by scripts/js/commands/checkNewPills.ts. However, it looks we only detect when individual pages are expired.

The issue is our logic here:

https://github.com/Qiskit/documentation/blob/582c84da03b10059442fc2d477ee82803acadcba/scripts/js/commands/checkNewPills.ts#L87-L106

We are only collecting pages with the new pill, when we should be collecting entries even if they are a folder. To understand this code, it's helpful to understand how `_toc.json` is set up:

```typescript
export type TocEntry = {
  title: string;
  url?: string;
  children?: TocEntry[];
  isNew?: boolean;
  isNewDate?: string;
};
```

## Precursor change: add some tests

This script is getting more complex and we should add a test to it so that we make sure it always works. Throughout this, you may want to run `npm run check:new-pills` to invoke the script and see how it's behaving.

There will be a few steps to this:

1. Move the function `collectNewPills` and the type definition `NewPillEntry` to a new file called `scripts/js/lib/newPills.ts`. You will need to add `export` to both values, then update `checkNewPills.ts` to import the two values.
   - `npm run typecheck` is useful to run TypeCheck to make sure the imports are set up properly
   - Once this is working properly, consider committing it.
3. Add `scripts/js/lib/newPills.test.ts`. Look at scripts/js/lib/stringUtils.test.ts for an example of how tests are set up. Then, write a test for `collectNewPills` to check that the existing behavior is working correctly.
    - You should make the input as simple as possible, but also sufficiently complex to handle some edge cases. For the `title`, use descriptive names like "Root-level page 1", "Root-level page 2", "Root-level folder", etc. I recommend you have one root-level page be marked `isNew` and one leave it off. Then have two root-level folders. In one folder, set a child page to have `isNew: true` 
    - Why do we have the pages/folders without `isNew`? These make sure that we don't have _false positives_, where our script complains about entries that are actually okay
    - Make sure this test works with `npm run test`. Then consider committing it
4. Improve your new test to have the bug in this GitHub issue. Mark one of the folders as `isNew` (rather than one of the pages). Make sure `npm run test` fails. This is called test-driven development
5. Now, fix the production code & make sure your test fixes things

GantaRoja:
closes #3539
1. Fixed new pill check to handle folder
2. Added test cases to new pill check


scripts/js/lib/newPills.ts
```diff
@@ -0,0 +1,26 @@
+import { TocEntry } from "./api/generateToc.js";
+export interface NewPillEntry {
+  toc: string;
+  url: string;
```
frankharkins: Now that this isn't always a URL (sometimes it's a title), we should change the name to something else (such as `identifier`).

After you've done that, you'll need to change some other parts of the codebase – you can use `npm run typecheck` to find places that need updating.

scripts/js/lib/newPills.test.ts
```diff
@@ -0,0 +1,109 @@
...
+            {
+              title: "child-1.1",
+              url: "/child-1.1",
+              isNew: "true",
+            },
+          ],
+        },
+      ],
+    },
+  ];
```
frankharkins, line 70: Great idea to test two levels of nesting

scripts/js/lib/newPills.ts
Eric-Arellano: Agreed. You'll also want to change these logs in our `main` function:

```
`\n❌ The new pill should be removed for the url: '${outdatedPill.url}' in '${outdatedPill.toc}'`,
```

```
`\n❌ There is a missing 'isNewDate' entry for the url: '${missingDatePill.url}' in '${missingDatePill.toc}'`,
````

They should say this

```
`\n❌ The new pill should be removed for this entry: '${outdatedPill.identifier}' in '${outdatedPill.toc}'`,
```

```
`\n❌ There is a missing 'isNewDate' value for this entry: '${missingDatePill.identifier}' in '${missingDatePill.toc}'`,
````

PR #3608, opened 2025-07-28
Add conditional meas to DC limits list

jyu00:
I was asked about this a few times, so it should go into the docs (that conditional measurements are also not supported).


abbycross (approved):
👍

PR #3618, opened 2025-07-29
Sphinx's versionadded directive

### LINKED ISSUE #3249: Ensure we support Sphinx's `versionadded` directive
## Background

Sphinx is a Python tool to generate documentation from source code. Our API partners use Sphinx to generate their docs.

We need to convert their Sphinx docs from Sphinx into our format, MDX. So, we first have Sphinx generate HTML, and then we convert the HTML into MDX using a conversion pipeline we wrote.

## Goal

Sphinx has a "directive" called `versionadded`, which you can add to your API docs to signal in what version a change was introduced. The Qiskit SDK team wants to start using it. We need to see if our conversion script can handle it.

### Part 1: improve qiskit_sphinx_theme

We have an integration test that checks how our conversion pipeline works from start-to-finish. The test is based off of sample docs that we have in this repository: https://github.com/Qiskit/qiskit_sphinx_theme.

We want to change our example docs in https://github.com/Qiskit/qiskit_sphinx_theme to test out `versionadded`. To do this, we will add a line to the docstring (documentation) of this function:

https://github.com/Qiskit/qiskit_sphinx_theme/blob/f00f5bdc2e36061b7b647cc18ddfe49fb3ee71f8/example_docs/api_example/electron.py#L51-L61

Add this line to the bottom of the docstring:

```

        .. versionadded:: 2.5
            The *velocity* parameter.
```

Pay attention to whitespace. The `..` should be the same column as `Raises:`. There needs to be a blank line between `.. versionadded` and the `ValueError` line above it.

### Part 2: improve our integration test

Follow these instructions. First, regenerate the test input by following the second paragraph. Then, regenerate our code by running `npm test -- --updateSnapshot`:

https://github.com/Qiskit/documentation/blob/960c592dfa1c3c6bf3e762eb4717722cda94d43b/scripts/js/lib/api/conversionPipeline.test.ts#L23-L28

We want to see how our pipeline handles the input. Open a draft PR with the change so that we can see the behavior.

### Part 3: maybe fix?

Depending on part 2, we may need to change our conversion pipeline code to support `versionadded`. It depends on what happens in part 2.

GantaRoja:
closes #3249 Added "versionadded" directive, checked and tested and it's working fine.


scripts/js/lib/api/htmlToMd.ts
```diff
@@ -266,6 +270,13 @@ function buildDeprecatedAdmonition(
     "versionmodified",
   );
   const title = toText(titleNode).trim().replace(/:$/, "");
+
+  //  Determine type based on title content
+  let type = "info"; // default for versionadded
+  if (title.toLowerCase().includes("deprecated")) {
+    type = "danger";
+  }
```
Eric-Arellano: Nice job! This works well.

However, I think we'll want a different approach to make this "less magical". I'd rather that the caller specifies what behavior it should do because it will make the code more clear that the function's behavior changes based on the type of input. Right now, you have to read the actual implementation of `buildApiVersionAdmonition` to know that there is this special conditional.

I propose that you add an argument called `admonitionType: 'info' | 'danger'`. This is using TypeScript's "union" feature with `|`, https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#union-types. You'd need to update lines 128 and 131 to set the right value. You'd also need to update our tests. I like this design because it keeps things more flexible for if we ever add support for `versionchanged`: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionchanged

Alternatively, I considered having an argument like `type: 'deprecation' | 'added'`, but I think that is less clear.

scripts/js/lib/api/htmlToMd.ts
```diff
@@ -257,7 +260,8 @@ function buildAdmonition(
   };
 }
 
-function buildDeprecatedAdmonition(
+
+function buildApiVersionAdmonition(
```
Eric-Arellano, line 263: Good name!

scripts/js/lib/api/htmlToMd.test.ts
```diff
@@ -482,6 +482,26 @@ test("parse deprecations warnings", async () => {
 </Admonition>
 `);
 });
+// ------------------------------------------------------------------
+// Handle version added derivative admonition
+// ------------------------------------------------------------------
```
Eric-Arellano: Often, this type of comment is really helpful. Here, it's a bit noisy and I'd remove it because the test for versionadded is part of the section we already have for "admonitions"

Eric-Arellano (approved):
🚀

PR #3627, opened 2025-07-30
Update multiverse singularity machine learning documentation

azamat-bakytzhan:
Closes Qiskit/documentation#3626

### DIFF
--- a/docs/guides/multiverse-computing-singularity.ipynb
+++ b/docs/guides/multiverse-computing-singularity.ipynb
@@ -114,8 +114,9 @@
       "Job status:  QUEUED\n",
       "Action result status:  ok\n",
       "Action result message:  Classifier created, fitted, and predicted.\n",
-      "Predictions (first five results):  [0, 1, 0, 0, 0]\n",
-      "Probabilities (first five results):  [[0.7975136940667589, 0.20248630593324138], [0.08459026150938889, 0.9154097384906114], [0.8043402543461567, 0.19565974565384334], [0.7975136940667589, 0.20248630593324138], [0.7975136940667589, 0.20248630593324138]]\n"
+      "Predictions (first five results):  [1, 1, 1, 1, 0]\n",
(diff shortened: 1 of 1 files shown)


docs/guides/multiverse-computing-singularity.ipynb
```diff
@@ -853,6 +856,7 @@
     "    weight_update_method=\"logarithmic\",\n",
     "    sample_scaling=True,\n",
     "    optimizer_options={\"simulator\": True},\n",
+    "    backend_name=\"test_eagle_us-east\",\n",
```
Eric-Arellano: Was this change intentional to use a test device?

azamat-bakytzhan: After migration to the new IQP, I have found that `backend_name` is important to use if users are using a test instance and it has only a single machine with `test_*` prefix. It works fine if users use quantum machines - without `test_*` prefix.

This is because we have an internal filter that removes test instances from the list if `backend_name` is None, which satisfies the `least_busy` logic.

pandasa123: @azamat-bakytzhan, I'd recommend a true device and `simulator: False`, since only a very limited number of people have access to the test devices

pandasa123:
Beyond the change @Eric-Arellano pointed out, this is good

docs/guides/multiverse-computing-singularity.ipynb
azamat-bakytzhan: I have removed the `backend_name` parameter since the simulator runs fine it the instance has at least one machine without `test_*` prefix.

pandasa123: @azamat-bakytzhan, is `backend_name` a required parameter?

azamat-bakytzhan: No, it is not required if the instance contains at least one quantum machine. It is required only if there are no quantum machines in the instance. By quantum machines, I mean machines without the `test_*` prefix.

Issue #2489, opened 2024-12-16
Add check that page title is in sync with ToC, h1, and metadata


Eric-Arellano:
You have to set the page title in three places:

* `_toc.json`
* H1, i.e. starting with `#`
* Metadata at the top of the page

It's actually quite useful that these three can be different values. However, ~90% of the time, we want them to be in sync. We should require allowlisting if they are not in sync.

Background for this idea: someone was confused why searching for the _toc.json title was not resulting in the page showing up.

> allowlisting

TBD the best approach to do this. The allowlist will need to be the file name. But do you only list the file to say it's okay for them to diverge, or do you have to get more granular to say which of the 3 are divergent? (Imo the former is less confusing and better)

PR #3669, opened 2025-08-06
Add check that page title is in sync with ToC, h1, and metadata 

GantaRoja:
#2489 Added a new function and test case for mismatched title and heading

### DIFF
--- a/scripts/js/lib/markdownTitles.test.ts
+++ b/scripts/js/lib/markdownTitles.test.ts
@@ -1,0 +1,39 @@
+// This code is a Qiskit project.
+//
+// (C) Copyright IBM 2023.
+//
+// This code is licensed under the Apache License, Version 2.0. You may
+// obtain a copy of this license in the LICENSE file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+import { expect, test } from "@playwright/test";
+
+import { collectHeadingTitleMismatch } from "./markdownTitles";
+
+test("Test to match mismatched titles and headings", async () => {
+  const markdown1 = `
+---
+title: My Awesome Guide
+---
+
+# My Awesome Guide
+    `;
+const markdown2 = `---
+title: Qiskit Doc
+author: John
+---
+
+# Introduction
+
+This guide will walk you through everything.`;
+  const mismatched = await collectHeadingTitleMismatch(markdown1);
+  const mismatched2 = await collectHeadingTitleMismatch(markdown2)
+  const result: string[] = [];
+  const result2: string[] = [`Mismatch: frontmatter title "Qiskit Doc" does not match heading "Introduction"`];
+expect(mismatched).toEqual(result);
+expect(mismatched2).toEqual(result2);
+});
--- a/scripts/js/lib/markdownTitles.ts
+++ b/scripts/js/lib/markdownTitles.ts
@@ -1,0 +1,47 @@
+import { unified } from "unified";
+import remarkParse from "remark-parse";
+import remarkGfm from "remark-gfm";
+import remarkFrontmatter from "remark-frontmatter";
+import { visit } from "unist-util-visit";
+import { Root } from "mdast";
+import yaml from "js-yaml";
+
+export async function collectHeadingTitleMismatch(markdown: string): Promise<string[]> {
+  const mismatches: string[] = [];
+  let frontmatterTitle: string | undefined;
+  let headingText: string | undefined;
+
+  const processor = unified()
+    .use(remarkParse)
+    .use(remarkGfm)
+    .use(remarkFrontmatter, ["yaml"]);
+
+  const tree = processor.parse(markdown);
+
+  // Run the transformer manually
+  visit(tree, "yaml", (node: any) => {
+    const data = yaml.load(node.value);
+    if (typeof data === "object" && data !== null && "title" in data) {
+      frontmatterTitle = (data as any).title;
+      console.log("Frontmatter title:", frontmatterTitle);
+    }
+  });
+
+  visit(tree, "heading", (node: any) => {
+    if (node.depth === 1 && !headingText) {
+      headingText = node.children
+        .filter((child: any) => child.type === "text")
+        .map((child: any) => child.value)
+        .join(" ");
+      console.log("Heading text:", headingText);
+    }
+  });
+
+  if (frontmatterTitle && headingText && frontmatterTitle !== headingText) {
+    mismatches.push(
+      `Mismatch: frontmatter title "${frontmatterTitle}" does not match heading "${headingText}"`
+    );
+  }
+
+  return mismatches;
+}


scripts/js/lib/markdownTitles.test.ts
```diff
@@ -0,0 +1,39 @@
...
+# Introduction
+
+This guide will walk you through everything.`;
+  const mismatched = await collectHeadingTitleMismatch(markdown1);
+  const mismatched2 = await collectHeadingTitleMismatch(markdown2)
+  const result: string[] = [];
+  const result2: string[] = [`Mismatch: frontmatter title "Qiskit Doc" does not match heading "Introduction"`];
+expect(mismatched).toEqual(result);
+expect(mismatched2).toEqual(result2);
+});
```
frankharkins: I'd split this into two tests: One where the titles match, and another where they don't. This means one test failing won't block the other, which will make it easier for a future contributor to work out how they've broken something.

Might also be good to test we handle heading with nested markdown, e.g.

```markdown
# This is a *Heading*
```

I don't think we'll ever want that in a heading, but we should check the script doesn't crash

scripts/js/lib/markdownTitles.ts
```diff
@@ -0,0 +1,47 @@
...
+    .use(remarkFrontmatter, ["yaml"]);
+
+  const tree = processor.parse(markdown);
+
+  // Run the transformer manually
+  visit(tree, "yaml", (node: any) => {
+    const data = yaml.load(node.value);
+    if (typeof data === "object" && data !== null && "title" in data) {
+      frontmatterTitle = (data as any).title;
+      console.log("Frontmatter title:", frontmatterTitle);
```
frankharkins: We can remove these now, I just recommended them while debugging

scripts/js/lib/markdownTitles.ts
```diff
@@ -0,0 +1,47 @@
+import { unified } from "unified";
+import remarkParse from "remark-parse";
+import remarkGfm from "remark-gfm";
+import remarkFrontmatter from "remark-frontmatter";
+import { visit } from "unist-util-visit";
+import { Root } from "mdast";
+import yaml from "js-yaml";
+
+export async function collectHeadingTitleMismatch(markdown: string): Promise<string[]> {
+  const mismatches: string[] = [];
```
frankharkins: Can you use a `Set` rather than an array to match the images check? Then you'd just do `.add()` rather than `.push()` on line 41

scripts/js/lib/markdownTitles.ts
```diff
@@ -0,0 +1,64 @@
...
+    .use(remarkGfm)
+    .use(remarkFrontmatter, ["yaml"]);
+
+  const tree = processor.parse(markdown);
+
+  // Extract frontmatter title
+  visit(tree, "yaml", (node: any) => {
+    const data = yaml.load(node.value);
+    if (typeof data === "object" && data !== null && "title" in data) {
+      frontmatterTitle = (data as any).title;
```
frankharkins: I couldn't find documentation for this, but I believe if we import `EXIT` from `unist-util-visit` then we can stop traversing the tree once we've got the information we're looking for (same for line 52):

```suggestion
      frontmatterTitle = (data as any).title;
      return EXIT;
```

Maybe see if that works and if it speeds up the check at all. Sorry I didn't think of this earlier.

scripts/js/lib/markdownTitles.ts
```diff
@@ -0,0 +1,64 @@
...
+  if (node.children && Array.isArray(node.children)) {
+    return node.children.map(extractText).join(" ");
+  }
+
+  return "";
+}
+
+export async function collectHeadingTitleMismatch(
+  markdown: string,
+): Promise<string[]> {
```
frankharkins: Thanks for switching to `Set`, however, I meant we should return a `Set` to match the interface of the images check.

```suggestion
): Promise<Set<string>> {
```

Then you can remove `Array.from` on line 63 and change the test results to be

```ts
const result: Set<string> = new Set(['...'])
```

frankharkins (commented):
Thanks!

scripts/js/commands/checkMarkdown.ts
```diff
@@ -35,29 +37,53 @@ const readArgs = (): Arguments => {
...
+    }
+
+    //Handle title/heading mismatch errors
+    if (mismatchedTitleHeadingErrors.length) {
+      errorsInFile.push(
+        ...mismatchedTitleHeadingErrors.map(
+          (err) => `Title/Heading mismatch: ${err}`,
+        ),
+      );
+    }
```
frankharkins: I think the error messages themselves are clear enough that we don't need to prefix them, which also means we can simplify the code to:

```suggestion
    const errorsInFile: string[] = [
      ...imageErrors,
      ...mismatchedTitleHeadingErrors,
    ];
```

scripts/js/commands/checkMarkdown.ts
```diff
@@ -35,29 +37,53 @@ const readArgs = (): Arguments => {
...
   }
 
+  // Final error report
   if (fileErrors.length) {
     fileErrors.forEach((error) => console.log(error));
     console.error(
-      "💔 Some images have problems. See https://github.com/Qiskit/documentation#images for instructions.\n",
+      "Some issues were found in your Markdown files. Please fix them before proceeding.\n" +
+        "Image help: https://github.com/Qiskit/documentation#images\n" +
+        "Title/Heading help: https://github.com/Qiskit/documentation#titles-and-headings\n",
```
frankharkins, line 198: Thanks for adding this!

package.json
```diff
@@ -16,7 +16,7 @@
     "test": "playwright test",
     "typecheck": "tsc",
     "check": "npm run check:patterns-index && npm run check:qiskit-bot && npm run check:metadata && npm run check:images && npm run check:orphan-pages && npm run check:spelling && npm run check:internal-links",
-    "check:images": "tsx scripts/js/commands/checkImages.ts",
+    "check:images": "tsx scripts/js/commands/checkMarkdown.ts",
```
frankharkins: We should also change the script name:

```suggestion
    "check:markdown": "tsx scripts/js/commands/checkMarkdown.ts",
```

You'll need to update these files too:

```
check
.github/workflows/main.yml
.github/workflows/api-checks.yml
```

package.json
```diff
@@ -57,7 +57,7 @@
     "remark-mdx": "^2.3.0",
     "remark-parse": "^10.0.1",
     "remark-stringify": "^10.0.3",
-    "transform-markdown-links": "^2.1.0",
+    "transform-images-links": "^2.1.0",
```
frankharkins: I think this was an accident

```suggestion
    "transform-markdown-links": "^2.1.0",
```

scripts/js/lib/markdownTitles.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  let frontmatterTitle: string | undefined;
+  let headingText: string | undefined;
+
+  const processor = unified()
+    .use(remarkParse)
+    .use(remarkGfm)
+    .use(remarkFrontmatter, ["yaml"]);
+
+  const tree = processor.parse(markdown);
```
frankharkins: This is the parsing step we want avoid doing twice; move this out of this function and into `checkMarkdown.ts`

scripts/js/lib/markdownTitles.test.ts
```diff
@@ -0,0 +1,64 @@
+// This code is a Qiskit project.
+//
+// (C) Copyright IBM 2023.
```
frankharkins:
```suggestion
// (C) Copyright IBM 2025.
```

scripts/js/lib/markdownTitles.ts
```diff
@@ -0,0 +1,66 @@
+import { unified } from "unified";
```
frankharkins: You need to add the copyright comment to the top of this file (just copy from another script and update the year)

scripts/js/commands/checkMarkdown.ts
```diff
@@ -0,0 +1,146 @@
...
+        `Error in file '${file}':\n\t- ${errorsInFile.join("\n\t- ")}\n`,
+      );
+    }
+  }
+
+  // Final error report
+  if (allErrors.length) {
+    allErrors.forEach((error) => console.log(error));
+    console.error(
+      "Some issues were found in your Markdown files. Please fix them before proceeding.\n" +
```
frankharkins: Kinda silly, but how do you feel about adding back in the broken heart emoji (💔)? Aside from being fun, I think it also made it a bit easier to scan the logs. Same for the tick emoji (✅) on line 124.

frankharkins (commented):
Thank you! This is looking really good 🎉 

Aside from a few small things, there's one change I think we should make before merging this:

At the moment, both `collectInvalidImageErrors` and `collectHeadingTitleMismatch` accept a markdown string, then parse that string into an abstract syntax tree. The parsing step is relatively slow, so I think we should pull that out of the functions so we only parse the tree once.

To do this, you'd move the parsing step to L92, then change the two "collect..." functions to accept a `tree: Root` rather than `markdown: string`. Then tweak `collectInvalidImageErrors` so it re-uses the same tree.

scripts/js/lib/markdownImages.test.ts
```diff
@@ -11,6 +11,19 @@
...
+import remarkFrontmatter from "remark-frontmatter";
+import { Root } from "mdast";
+
+function parseMarkdown(markdown: string): Root {
+  return unified()
+    .use(remarkParse)
+    .use(remarkGfm)
+    .use(remarkFrontmatter, ["yaml"])
+    .parse(markdown);
+}
```
frankharkins: This is a nice helper function, what do you think about moving it to a new file (e.g. `scripts/js/lib/markdownUtils.ts`) and re-using it in `checkMarkdown.ts`?

frankharkins (approved):
Awesome work, thanks again

scripts/js/commands/checkMarkdown.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+
+const readArgs = (): Arguments => {
+  return yargs(hideBin(process.argv))
+    .version(false)
+    .option("apis", {
+      type: "boolean",
+      default: false,
+      description:
+        "Check the images in the current and dev versions of the API docs have alt text.",
```
frankharkins: Sorry, I missed this in my earlier review

```suggestion
        "Check files in the current and dev versions of the API docs have matching titles and metadata.",
```

scripts/js/commands/checkMarkdown.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    //Collect all errors for this file
+    const errorsInFile: string[] = [
+      ...imageErrors,
+      ...mismatchedTitleHeadingErrors,
+    ];
+
+    if (errorsInFile.length) {
+      allErrors.push(
+        // `Error in file '${file}':\n\t- ${[...imageErrors].join("\n\t- ")}\n`,
```
frankharkins: This comment also needs removing

scripts/js/commands/checkMarkdown.ts
```diff
@@ -0,0 +1,147 @@
+// This code is a Qiskit project.
+//
+// (C) Copyright IBM 2024.
```
Eric-Arellano:
```suggestion
// (C) Copyright IBM 2025.
```

scripts/js/commands/checkMarkdown.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+const readArgs = (): Arguments => {
+  return yargs(hideBin(process.argv))
+    .version(false)
+    .option("apis", {
+      type: "boolean",
+      default: false,
+      description:
+        "Check files in the current and dev versions of the API docs have matching titles and metadata.",
```
Eric-Arellano: Now that we check both titles and images, we shouldn't be overly specific

```suggestion
        "Check files in the current and dev versions of the API docs.",
```

scripts/js/lib/markdownImages.test.ts
Eric-Arellano: This comment still seems relevant

Eric-Arellano (commented):
Awesome! Great work. I really like the approach of combining checks for better performance.

check
```diff
@@ -25,7 +25,7 @@ CHECKS = {
     "metadata": ["npm", "run", "check:metadata"],
     "patterns index pages": ["npm", "run", "check:patterns-index"],
     "tutorials index page": ["python3", "scripts/ci/check-tutorials-index.py"],
-    "images": ["npm", "run", "check:images"],
+    "images": ["npm", "run", "check:markdown"],
```
Eric-Arellano:
```suggestion
    "markdown": ["npm", "run", "check:markdown"],
```

scripts/js/lib/markdownTitles.ts
```diff
@@ -0,0 +1,67 @@
...
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+import { visit, EXIT } from "unist-util-visit";
+import { Root } from "mdast";
+import yaml from "js-yaml";
+
+// Helper to recursively extract visible text from heading node
+function extractText(node: any): string {
```
Eric-Arellano: Can you please call this `extractHeadingText`? And the parameter `headingNode`. Then you can remove the comment. We prefer "self-documenting code"

It'd be helpful to move this to `markdownUtils.ts` because we may want to use it in the future in other places. This is pretty generic code

scripts/js/lib/markdownTitles.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  if (node.type === "link" && node.children) {
+    return node.children.map(extractText).join(" ");
+  }
+
+  if (node.children && Array.isArray(node.children)) {
+    return node.children.map(extractText).join(" ");
+  }
+
+  return "";
```
Eric-Arellano: Rather than returning an empty string, it's better to throw an error: ``new Error(`Could not parse heading node: ${node}`)``. This is "defensive programming" - our assumptions will have been invalidated that there is an edge case we haven't thought about. We want to know that right away. 

It is safe to error here because these are developer productivity scripts; we are not breaking users of IBM Quantum, only developers of these docs.

scripts/js/lib/markdownTitles.ts
```diff
@@ -0,0 +1,67 @@
...
+  // Extract first level-1 heading with full formatting
+  visit(tree, "heading", (node: any) => {
+    if (node.depth === 1 && !headingText) {
+      headingText = extractText(node).trim();
+      return EXIT;
+    }
+  });
+
+  // Compare and collect mismatch
+  if (frontmatterTitle && headingText && frontmatterTitle !== headingText) {
```
Eric-Arellano, line 36: FYI, in the future, we will want to error if the frontmatterTitle or headingText are missing. Currently, `check:metadata` already does that. We'll consolidate the checks in a follow up.

scripts/js/lib/markdownTitles.test.ts
```diff
@@ -0,0 +1,67 @@
...
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+import { expect, test } from "@playwright/test";
+
+import { collectHeadingTitleMismatch } from "./markdownTitles";
+import { parseMarkdown } from "./markdownUtils";
+
+test("Test for matching titles and headings", async () => {
```
Eric-Arellano: Because all these tests are for the same function, it's more conventional to group the tests all together by using `test.describe()`, like this

https://github.com/Qiskit/documentation/blob/eec1dae8fcb9d2ecc2c06922b34f96ab72d8154e/scripts/js/lib/api/TocGrouping.test.ts#L145-L167

Here, you could put in the description for `test.describe()` the string `"collectHeadingTitleMismatch")`. Then, name the `test()` cases as:

* `"valid"`
* `"mismatched - simple h1"`
* `"mismatched - complex h1"`

Furthermore, there is a lot of duplication between the tests. It is conventional to DRY (Don't Repeat Yourself) this, i.e. to deduplicate it. You can define a helper function:

```typescript
const assert = (markdown: string, expected: Set<string>): void => {
    const tree = parseMarkdown(markdown);
    const result = await collectHeadingTitleMismatch(tree);
    expect(result).toEqual(expected)
}
```

scripts/js/lib/markdownTitles.ts
```diff
@@ -0,0 +1,67 @@
...
+  if (node.children && Array.isArray(node.children)) {
+    return node.children.map(extractText).join(" ");
+  }
+
+  return "";
+}
+
+export async function collectHeadingTitleMismatch(
+  tree: Root,
+): Promise<Set<string>> {
```
Eric-Arellano: Do you remember why Frank had you return a `Set<string>` rather than `string | null`?

Eric-Arellano: Ah, I imagine it comes from copying the image code, right? That makes sense with images because there can be >1 error per file. 

It'd be more accurate for this to return `string | null`. However, after looking at the `checkMarkdown.ts` script, that would make the code much more annoying to deal with. So, let's stick with this current implementation.

scripts/js/lib/markdownTitles.ts
```diff
@@ -0,0 +1,67 @@
...
+  const mismatches = new Set<string>();
+
+  let frontmatterTitle: string | undefined;
+  let headingText: string | undefined;
+
+  // Extract frontmatter title
+  visit(tree, "yaml", (node: any) => {
```
Eric-Arellano: I realized this isn't going to work with Jupyter notebooks. They set their markdown differently. We'll need to figure out a new approach. We can pair program on this

Eric-Arellano:
Hey Swati, I thought of how to get this working with Jupyter notebooks. In `markdownReader.ts`, define a new asynchronous function `readMarkdownAndMetadata` with the same arguments as `readMarkdown`. Return `Promise<{content: string; metadata: Record<string, string>}`. The implementation should look like `readMarkdown`, but you need to enhance it to also parse metadata by using the techniques in this function with `grayMatter()` and `json.metadata`:

https://github.com/Qiskit/documentation/blob/9dc9194e2b11ca1e09e28f1dcbae4237980e8fa3/scripts/js/commands/checkMetadata.ts#L40-L51

(Why do we want to combine markdown reading and metadata reading into one function? We want to reuse the `readFile()` call so that we are not reading the raw file twice. This should help with performance. Filesystem operations are expensive, along with buffering the whole file in memory.)

You're then going to want to have `checkMarkdown.ts` call `readMarkdownAndMetadata`. `collectInvalidImageErrors` should still stay the same, but `collectHeadingTitleMismatch` is going to change that it needs both the parsed markdown tree & the metadata as an argument. Update `collectHeadingTitleMismatch` to simply read `metadata.title`, and remove the `yaml` parsing. Update the unit tests.

Finally, refactor `checkMetadata.ts` to use your new `readMarkdownAndMetadata` function.

scripts/js/commands/checkMetadata.ts
```diff
@@ -37,18 +39,18 @@ const readArgs = (): Arguments => {
...
+//   if (ext === "md" || ext === "mdx") {
+//     const content = await fs.readFile(filePath, "utf-8");
+//     return grayMatter(content).data;
+//   } else if (ext === "ipynb") {
+//     const json = await readJsonFile(filePath);
+//     return json.metadata;
+//   } else {
+//     throw new Error(`Unknown extension for ${filePath}: ${ext}`);
+//   }
+// };
```
Eric-Arellano: This should be deleted. We don't like commented out code because it's confusing why it's still there

scripts/js/commands/checkMetadata.ts
```diff
@@ -19,6 +19,8 @@ import { globby } from "globby";
 
 import { readJsonFile } from "../lib/fs";
 
+import {readMarkdownAndMetadata} from "../lib/markdownReader";
```
Eric-Arellano: It looks like the formatting is messed up. Reminder to run `npm run fmt` and `npm run lint`

scripts/js/lib/markdownReader.ts
```diff
@@ -10,22 +10,21 @@
...
-  filePath: string,
-  options: { includeCodeCellSourceCode?: boolean } = {},
-): Promise<string> {
-  if (path.extname(filePath) === ".ipynb") {
-    const notebook = await readJsonFile(filePath);
-    return markdownFromNotebook(notebook, options);
-  }
-  return await readFile(filePath, { encoding: "utf8" });
-}
+// export async function readMarkdown(
```
Eric-Arellano: Why is this commented out? This is still useful code.

scripts/js/lib/markdownTitles.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+import { collectHeadingTitleMismatch } from "./markdownTitles";
+import { readMarkdownAndMetadata } from "./markdownReader"
+
+
+const assert = async (
+  markdown: string,
+  expected: Set<string>,
+): Promise<void> => {
+  const { tree, metadata } = await readMarkdownAndMetadata(markdown);
```
Eric-Arellano: This code needs to be updated. Please make sure `npm test` passes

PR #3692, opened 2025-08-10
Fix code example in the C API installation guide

### LINKED ISSUE #3689: Example code bug in C API install guide

Install the Qiskit C API [https://quantum.cloud.ibm.com/docs/en/guides/install-c-api](https://quantum.cloud.ibm.com/docs/en/guides/install-c-api)

### Select all that apply

- [ ] new content request
- [ ] typo
- [x] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

The C code example at the top of the page does not compile. The source of the problem was traced to line 16.

```QkObsTerm term = {coeff, 3, bit_terms, indices, num_qubits};```

C compiler returns type errors.

PROPOSED FIX: The ```coeff``` object needs to be converted to a ```QkComplex64``` type before assignment to the ```QkObsTerm``` struct.

### Environment
- OS: Linux
- Compiler: GCC v15.1.1
- Qiskit source code: commit eb8f791 (main branch, August 8, 2025)

### PR DESCRIPTION
**Pull request by eliarbel**
This PR fixes a compilation error in the code example of the C API installation guide. 

Fixes #3689

### DIFF
--- a/docs/guides/install-c-api.mdx
+++ b/docs/guides/install-c-api.mdx
@@ -13,10 +13,10 @@ The following example builds an observable with C:
 ```c
 // file: example.c
 #include <stdio.h>
-#include <complex.h>
+#include <stdint.h>
 #include <qiskit.h>
 
-int main(int argc, char* argv[]) {
+int main(int argc, char *argv[]) {
     // build a 100-qubit empty observable
     uint32_t num_qubits = 100;
     QkObs *obs = qk_obs_zero(num_qubits);
--- a/docs/guides/install-c-api.mdx
+++ b/docs/guides/install-c-api.mdx
@@ -13,17 +13,16 @@ The following example builds an observable with C:
 ```c
 // file: example.c
 #include <stdio.h>
-#include <stdint.h>
 #include <complex.h>
 #include <qiskit.h>
 
 int main(int argc, char* argv[]) {
     // build a 100-qubit empty observable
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: docs/guides/install-c-api.mdx [hunk 1]
```diff
@@ -13,17 +13,16 @@ The following example builds an observable with C:
 ```c
 // file: example.c
 #include <stdio.h>
-#include <stdint.h>
```
Eric-Arellano, line 16: Why remove this line?

eliarbel: Oh, maybe I was too eager to keep the example as small as possible... and just relied on it being included by `qiksit.h`. Should we add it back?

Eric-Arellano: I'm not enough of a C programmer to know what is the best practice. My naive take would be that depending on a transitive dependency of another dependency seems risky and less explicit. Semi-related, why include `stdio.h` and `complex.h` too?

But I'm not a C programmer. Maybe this is best practice

eliarbel: `stdio.h` is required for the `printf` calls. 

I'm rusty in C programming as well 😄 It might actually be better to avoid indirect include dependencies. I pushed another commit to align with this. Also we don't need to include `complex.h`. And to better align with other examples and Qiskit's convention, `*` should be placed next to identifiers.

PR #3698, opened 2025-08-11
Clarify service init behavior

kt474:
A user saved their account with a default instance but then initialized the `QiskitRuntimeService` with a token so the saved account was not used. This confusion caused them to mistakenly use a premium instance, thinking they were using the open instance. 

```python
QiskitRuntimeService.save_account(
    token=token,
    channel="ibm_quantum_platform",
    instance=CRN
)

# This initializes a new account with the token and channel, it does not use the saved account. 
service = QiskitRuntimeService(channel="ibm_quantum_platform", token=token)
```


abbycross (approved):
Thanks!

PR #3715, opened 2025-08-13
Fixed figure caption (changed 'red vector' to 'yellow vector')

### LINKED ISSUE #3711: Mismatch between figure and caption in course: A practical introduction to quantum-safe cryptography

https://quantum.cloud.ibm.com/learning/en/courses/quantum-safe-cryptography/quantum-safe-cryptography#the-shortest-vector-problem

### Select all that apply

- [ ] new content request
- [x] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

The caption on Figure 2 states that the vector is red, whereas the figure displays a yellow vector.

christopherporter1:
Fixes #3711

### DIFF
--- a/learning/courses/basics-of-quantum-information/entanglement-in-action/chsh-game.ipynb
+++ b/learning/courses/basics-of-quantum-information/entanglement-in-action/chsh-game.ipynb
@@ -261,7 +261,7 @@
     "\n",
     "Now we can describe the quantum strategy.\n",
     "\n",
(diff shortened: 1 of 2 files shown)


abbycross (approved):
👍

Eric-Arellano (commented):
To double check for accessibility, do we have any other way to refer to the value other than color? With color blindness, we ideally want a second mechanism like dotting the line or making it thicker. (I'm on my phone so can't check)

If we don't, it'd be helpful to either fix now or track it in an issue for the future.

PR #3749, opened 2025-08-18
Add imagemagick to notebook testing environment

kaelynj:
Our cron job was not running `tox -e fix` correctly when generating new notebook outputs. This was because imagemagick wasn't included in the CI environment. This PR fixes that bug.


Eric-Arellano (approved):
Thanks!

Issue #3740, opened 2025-08-18
RequestsApiError: "HTTPSConnectionPool(host='quantum.cloud.ibm.com', port=443): Max retries exceeded with url:


kevinsung:
### URL to the relevant tutorial

https://quantum.cloud.ibm.com/docs/en/tutorials/real-time-benchmarking-for-qubit-selection

### Select all that apply

- [ ] new content request
- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

### Describe the fix or the content request.

```
RequestsApiError: "HTTPSConnectionPool(host='quantum.cloud.ibm.com', port=443): Max retries exceeded with url: /api/v1/backends/ibm_torino/properties?updated_before=2024-10-10T04%3A00%3A00%2B00%3A00 (Caused by ResponseError('too many 500 error responses'))"
```

With the `ibm_torino` backend, I get this error in the code cell
```python
errors_list = []
for day_idx in range(10, 17):
    calibrations_time = datetime(
        year=2024, month=10, day=day_idx, hour=0, minute=0, second=0
    )
    targer_hist = backend.target_history(datetime=calibrations_time)

    t1_dict, t2_dict = {}, {}
    for qubit in range(targer_hist.num_qubits):
        t1_dict[qubit] = targer_hist.qubit_properties[qubit].t1
        t2_dict[qubit] = targer_hist.qubit_properties[qubit].t2

    errors_dict = {
        "1q": targer_hist["sx"],
        "2q": targer_hist["ecr"],
        "spam": targer_hist["measure"],
        "t1": t1_dict,
        "t2": t2_dict,
    }

    errors_list.append(errors_dict)
```

Issue #3739, opened 2025-08-18
BackendPropertyError: "Couldn't find the property 'T1' for qubit 146."


kevinsung:
### URL to the relevant tutorial

https://quantum.cloud.ibm.com/docs/en/tutorials/real-time-benchmarking-for-qubit-selection

### Select all that apply

- [ ] new content request
- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

### Describe the fix or the content request.

With backend `ibm_kingston`, this error is thrown with the code cell beginning with
```python
# Create T1 experiments on all qubit in parallel
t1_exp = ParallelExperiment(
    [
        T1(
            physical_qubits=[qubit],
            delays=np.linspace(
                1e-6, 2 * backend.properties().t1(qubit), 5, endpoint=True
            ),
        )
        for qubit in qubits
    ],
    backend,
    analysis=None,
)
```

### For new content requests - if the request is accepted, do you want to write the content?

I only want to review the material when it's finished

miamico:
This can happen if the calibration data is missing the information about some qubit. Maybe a different default for the t1 experiment will fix it. Let's think about possibilities

Issue #3736, opened 2025-08-16
"(flattened) layered coupling map" needs explanation


kevinsung:
### URL to the relevant tutorial

https://quantum.cloud.ibm.com/docs/en/tutorials/real-time-benchmarking-for-qubit-selection#setting-up-backend-and-coupling-map

### Select all that apply

- [x] new content request
- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

### Describe the fix or the content request.

The code block creates a "flattened layered coupling map" which is just the coupling map but with edges grouped by layer. The prose should explain what is happening and why.

### For new content requests - if the request is accepted, do you want to write the content?

I can help the team by providing enough information or by referring them to a specific Subject Matter Expert to write the material

PR #3811, opened 2025-08-27
Fix issues in real-time becnhmarking for qubit selection tutorial

miamico:
Small changes to fix issues raised in:

- #3736 (removed flattened layered coupling map as not useful)
- #3739 (added exception if properties are not found)
- #3740 (updated date for backend property retrieval, still kind of brittle solution)

Also changed to Heron device, updated requirements and re-run


abbycross:
Seeing if I can resolve the merge conflicts...

docs/tutorials/real-time-benchmarking-for-qubit-selection.ipynb
```diff
@@ -734,7 +836,7 @@
    "id": "3705d85c-e11d-4a8f-bfae-20bd501124d8",
    "metadata": {},
    "source": [
-    "With increasing chain length, and thus less freedom to choose physical qubits, the fact that we have updated information about qubit properties becomes less substantial, with there being no difference when all of the qubits are used."
+    "Not all runs will show improvement in performance due to the real-time characterization and with increasing chain length, and thus less freedom to choose physical qubits, the importance of updated device information becomes less substantial. However, it is good practice to collect fresh data on the device properties to understand its performance. Occasionally, transient two-level systems may affect the performance of some of the qubits. Real-time data can inform us when such events are happening and help us avoid experimental failures in such instances."
```
abbycross:
```suggestion
    "Not all runs will show improvement in performance due to the real-time characterization - and with increasing chain length, and thus less freedom to choose physical qubits, the importance of updated device information becomes less substantial. However, it is good practice to collect fresh data on the device properties to understand its performance. Occasionally, transient two-level systems may affect the performance of some of the qubits. Real-time data can inform us when such events are happening and help us avoid experimental failures in such instances."
```

abbycross:
@miamico There are errors outputting from the code in [this section](https://qiskit.github.io/documentation/pr-3811/docs/tutorials/real-time-benchmarking-for-qubit-selection#execute-a-quantum-circuit-with-real-time-qubit-selection).

kevinsung (approved):

miamico:
> @miamico There are errors outputting from the code in [this section](https://qiskit.github.io/documentation/pr-3811/docs/tutorials/real-time-benchmarking-for-qubit-selection#execute-a-quantum-circuit-with-real-time-qubit-selection).

I thought warnings would be fine, I'm not sure we can properly fix them but we could ignore them with the warnings package

miamico:
@abbycross is this good to be merged?

docs/tutorials/real-time-benchmarking-for-qubit-selection.ipynb
```diff
@@ -161,18 +162,42 @@
...
-    "\n",
+    "one_dir_coupling_map = coupling_graph.edge_list()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d48b5ba2-74ec-4484-9ba6-b0681ae32919",
+   "metadata": {},
+   "source": [
+    "In order to benchmark as many two-qubit gates as possible simultaneously we separate the coupling map into a `layered_coupling_map`. This object contains a list of layers where each layer is a list of edges on which two-qubit gates can be executed at the same time. This is also called a coloring of the coupling map."
```
kevinsung:
```suggestion
    "In order to benchmark as many two-qubit gates as possible simultaneously we separate the coupling map into a `layered_coupling_map`. This object contains a list of layers where each layer is a list of edges on which two-qubit gates can be executed at the same time. This is also called an edge coloring of the coupling map."
```

abbycross:
@miamico @kevinsung Sorry I didn't respond right away - I was waiting for a response from our team on how to handle publishing errors. Bottom line - we do not want code to error for the user, and we definitely don't want to include output that contains errors and deprecation warnings. Will need to solve this a different way in a followup PR. Proposal from the team:

- Remove erroring output
- Add note in the text before the code, explaining the errors that a user might get, depending on their Python version

PR #3849, opened 2025-09-03
docs: add v1->v2 migration guide to the index

kandolfp:
In the index for the migration guides the new guide vor v2 was missing. I added the context as stated in the menu file [`_toc.json`](https://github.com/Qiskit/documentation/blob/main/docs/migration-guides/_toc.json)


Eric-Arellano:
We ended up removing the index page. Thanks for this PR, though!

PR #3855, opened 2025-09-04
fix: Dirac notation for Pauli-Y

tongkiat:
In "Utility-scale quantum computing", chapter "Quantum bits, gates, and circuits" (https://github.com/Qiskit/documentation/blob/main/learning/courses/utility-scale-quantum-computing/bits-gates-and-circuits.ipynb), section "3.2 Single-qubit quantum state and unitary evolution", paragraph 3, the Pauli-Y gate (using Dirac notation) is given as `-1 |0><1| + i |1><0|`.

Shouldn't it be `-i |0><1| + i |1><0|`?


christopherporter1 (approved):
Nice catch. Thanks!

PR #3860, opened 2025-09-04
Fix notebook failures in CI

### LINKED ISSUE #3643: Weekly notebook test failed
Today's scheduled test of notebooks that submit jobs failed.
Please [check the logs](https://github.com/Qiskit/documentation/actions/runs/16668709460/).
> [!NOTE]
> This issue was created by a GitHub action.

kaelynj:
Closes #3643 

Many of the issues with the notebook testing were credential based, but a few valid bugs were caught. This PR should fix all of them. But I'll run the notebook tester again just to be safe.


.github/actions/set-up-notebook-testing/action.yml
```diff
@@ -51,6 +51,11 @@ runs:
               token="${{ inputs.ibm-cloud-token }}",
               set_as_default=True
           )
+          QiskitRuntimeService.save_account(
```
Eric-Arellano, line 56: Would it make sense to delete lines 48-53?

Eric-Arellano (commented):
Thank you for working on this!

.github/actions/set-up-notebook-testing/action.yml
kaelynj: We coud, I'd just need to change all the catalog instantiations to use `"ibm_cloud"` instead. If a channel isn't specified when instantiating a `QiskitFunctionsCatalog` it assumes you're using the old quantum platform and throws an error.

In general, both the `"ibm_cloud"` and `"ibm_quantum_platform"` are valid channel options and point to the same thing in qiskit runtime under the hood. I'm just not sure if we should have a preference in our docs pages either way.

Eric-Arellano: Could you please add a comment explaining why you are calling this same function twice?

docs/guides/q-ctrl-performance-management.ipynb
```diff
@@ -105,23 +105,20 @@
    "source": [
     "## Get started\n",
     "\n",
-    "Fire Opal Performance Management uses Qiskit v`2.0.0`, which is the recommended version. Supported versions are Qiskit v`2.0.0` - `2.0.2`.\n",
-    "\n",
-    "Authenticate using your [IBM Quantum Platform API key](http://quantum.cloud.ibm.com/), and select the Qiskit Function as follows:"
+     "Fire Opal Performance Management uses Qiskit v`2.0.0`, which is the recommended version. Supported versions are Qiskit v`2.0.0` - `2.0.2`.\n",
```
Eric-Arellano: This line about Qiskit versions is out-of-date. Could you please ask the Functions team if this can be updated or removed?

scripts/config/notebook-testing.toml
```diff
@@ -100,6 +99,7 @@ notebooks = [
 test-strategies.hardware = { patch="qiskit-ibm-runtime-open" }
 notebooks = [
     "docs/tutorials/hello-world.ipynb",
+    "docs/guides/monitor-job.ipynb",
```
Eric-Arellano: Why this change? (For my understanding)

kaelynj: Ah actually putting it here was a mistake, I should have moved it so that it isn't tested until we can get runtime `0.41.1` working in our environment.

docs/guides/q-ctrl-performance-management.ipynb
kaelynj: I changed the line to read 

> Supported versions are Qiskit >=v`2.0.0`.

This should be a safe change since if the function supports 2.0, it should work with any 2.X version

PR #3870, opened 2025-09-05
Merge metadata check into markdown check

### LINKED ISSUE #3750: Merge metadata check into new markdown check
This builds on top of https://github.com/Qiskit/documentation/pull/3669, which combines our image check and new title check into a single script. We do that for performance reasons because reading and parsing the same files multiple times is expensive.

In this issue, we should remove `checkMetadata.ts` and instead have `checkMarkdown.ts` do everything.

## Step 1: new `lib` file

Move the logic from `checkMarkdown.ts` into `lib/metadataChecker.ts`. Have a `checkMetadata` function that takes as an argument the `metadata: Record<string, string>` and returns a `Set<string>` with the errors for that file.

You're going to use the logic from `isValidMetadata`, but should improve it to return a Set of error messages rather than a single boolean: 

* If title is missing, return "Missing `title` in the metadata"
* If description is missing, return "Missing `description` in the metadata"
* If the description is too short or too long, return "The `description` in the metadata must be between 50 and 160 characters, but was ${the actual length}"

Add unit tests for `checkMetdata()`.

## Step 2: hook up to `checkMarkdown.ts`

Have `checkMarkdown.ts` call your new function from `metadataChecker.ts` and ensure we error on issues. Update the error message to say "Metadata issues: https://github.com/Qiskit/documentation/blob/main/mdx-guide.md#page-metadata"

## Step 3: set up the allowlist

In our original file `checkMetadata.ts`, you'll notice that we have a lot of files we skip. We need to keep skipping those.

However, our config for what to skip is now getting pretty long. We should centralize all of our allowlists into a central file in `config/` called `allowLists.ts`. Move the existing allowlists from `checkMarkdown.ts` into that file, and also move the allowlist from `checkMetadata.ts` there. You may need to rename variables to make sure things are not ambiguous between our different checks.

GantaRoja:
merged metadata into markdown and moved ignore files inlo new file config\allowlist. Closes #3750


scripts/js/lib/metadataChecker.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      title: "My Awesome Article",
+      description:
+        "This is a well-written description that meets the length requirements perfectly.",
+    };
+    expect(checkmetadata(metadata)).toEqual(new Set());
+  });
+
+  test("should return error for missing title", () => {
+    const metadata = {
+      title: "",
```
Eric-Arellano: Even better is to delete the `title` entirely from the object. Usually our content writers will forget to set it at all

scripts/js/lib/metadataChecker.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    };
+    expect(checkmetadata(metadata)).toEqual(
+      new Set(["Missing title in the metadata"]),
+    );
+  });
+
+  test("should return error for missing description", () => {
+    const metadata = {
+      title: "Valid Title",
+      description: "",
```
Eric-Arellano: Ditto that this is best to leave off entirely

scripts/js/commands/checkMarkdown.ts
```diff
@@ -195,7 +78,7 @@ async function main() {
     console.error(
       "💔 Some issues were found in your Markdown files. Please fix them before proceeding.\n" +
         "Image help: https://github.com/Qiskit/documentation#images\n" +
-        "Title/Heading help: https://github.com/Qiskit/documentation#titles-and-headings\n",
```
Eric-Arellano, line 198: It's important to keep this line. You'll want to be adding the new line for metadata without changing the title/heading link

scripts/js/commands/checkMarkdown.ts
```diff
@@ -176,10 +53,16 @@ async function main() {
         ? new Set<string>()
         : collectHeadingTitleMismatch(tree, metadata);
 
+    const metadataErrors =
+      IGNORE_TITLE_MISMATCHES.includes(file) || file.startsWith("docs/api")
```
Eric-Arellano: This part is wrong with the allowlist. You need a dedicated allowlist for metadata, something like `METADATA_ALLOWLIST`. Each checker gets its own allowlist inside `allowLists.ts`. 

In `allowLists.ts`, this means that you'll need to rename `ALLOWED_VIOLATIONS` to something like `METADATA_ALLOWLIST` and make it public. While you're at it, you should also rename the helper function `qiskitLegacyIgnores` to a less general name -- rename it to something like `metadataIgnoresForLegacyQiskit()`

Finally, you can remove `|| file.startsWith("docs/api")` from this specific line. Our metadata checker actually works on API docs.

scripts/js/commands/checkMarkdown.ts
```diff
@@ -18,131 +18,8 @@ import { collectInvalidImageErrors } from "../lib/markdownImages.js";
...
-  "learning/courses/foundations-of-quantum-error-correction/quantum-code-constructions/css-codes.ipynb",
-  "learning/courses/fundamentals-of-quantum-algorithms/grover-algorithm/grover-algorithm-description.ipynb",
-  "learning/courses/fundamentals-of-quantum-algorithms/grover-algorithm/number-of-iterations.ipynb",
-  "learning/courses/fundamentals-of-quantum-algorithms/phase-estimation-and-factoring/phase-estimation-problem.ipynb",
-  "learning/courses/fundamentals-of-quantum-algorithms/phase-estimation-and-factoring/phase-estimation-procedure.ipynb",
-  "learning/courses/fundamentals-of-quantum-algorithms/quantum-algorithmic-foundations/factoring-and-gcd.ipynb",
-  "learning/courses/general-formulation-of-quantum-information/general-measurements/discrimination-and-tomography.ipynb",
-  "learning/modules/computer-science/vqe.ipynb",
-];
+import { checkmetadata } from "../lib/metadataChecker.js";
```
Eric-Arellano: Should be called `checkMetadata` (capitalization)

scripts/js/commands/checkMarkdown.ts
```diff
@@ -165,9 +42,9 @@ const readArgs = (): Arguments => {
 
 async function main() {
   const args = readArgs();
-  const files = await determineContentFiles(args);
+  const [mdFiles] = await determineContentFiles(args);
```
Eric-Arellano: This is a bad change. Ditto line 47

Eric-Arellano (commented):
Looks great! You will need to remove references to `check: metadata `, like updating our `.github` folder. With GitHub, you need to delete the check: metadata section entirely because now `check:markdown` does that check for us. Use ripgrep, `rg --hidden 'check:metadata'`

.github/workflows/api-checks.yml
```diff
@@ -43,6 +43,5 @@ jobs:
       - name: Check for orphan pages
         run: npm run check:orphan-pages -- --apis
       - name: Check metadata
```
Eric-Arellano:
```suggestion
```

scripts/js/commands/checkMarkdown.ts
```diff
@@ -56,9 +56,10 @@ async function main() {
         ? new Set<string>()
         : collectHeadingTitleMismatch(tree, metadata);
 
-    const metadataErrors = METADATA_ALLOWLIST.has(file) || file.startsWith("docs/api")
-      ? new Set<string>()
-      : checkMetadata(metadata);
+    const metadataErrors =
+      METADATA_ALLOWLIST.has(file) || file.startsWith("docs/api")
```
Eric-Arellano: @GantaRoja this is the wrong fix. We do want to check metadata for the API docs. We only don't want to check specifically the line length. Look at the original code in `checkMetadata.ts`: you need to pass the filepath to the function `checkMetadata`

scripts/js/lib/metadataChecker.ts
```diff
@@ -24,10 +27,12 @@ export function checkMetadata(metadata: Record<string, string>): Set<string> {
     errors.add("Missing description in the metadata");
   } else {
     const length = description.trim().length;
-    if (length < 50 || length > 160) {
-      errors.add(
-        `The description in the metadata must be between 50 and 160 characters, but was ${length}`,
-      );
+    if (!filePath.startsWith("docs/api/")) {
+      if (length < 50 || length > 160) {
```
Eric-Arellano: This should be inlined for better style. `if (!filePath.startsWith("docs/api/") || ((length < 50 || length > 160))`

scripts/js/lib/metadataChecker.test.ts
```diff
@@ -21,15 +21,17 @@ test.describe("checkMarkdown", () => {
       description:
         "This is a well-written description that meets the length requirements perfectly.",
     };
-    expect(checkMetadata(metadata)).toEqual(new Set());
+    const filePath = "docs/tutorials/intro.md";
```
Eric-Arellano: Imo these should all be inlined, i.e. don't create a dedicated variable. The value is simple and short enough that it's clearer to put it directly in `checkMetadata`. 

(This is subjective code style)

PR #3918, opened 2025-09-15
Followup fixes to the qpu-information page

### LINKED ISSUE #3904: Fix to the RX error definition on QPU information page and other follow-on details
From @dcmckayibm: 

I was wrong about this

Image

it's reported as the same as SX/X/Id so add it to this section

Image

At the end of that section "The RX gate is also the same duration of the sqrt(x) and X gate with variable amplitude, and so is reported as having the same error as these gates" 

Image
This should read "The 2Q error per edge from the same batch of measurements used to calculate the 2Q median [link] and 2Q best [link]"

Image
we should note that the per edge data is provided in the calibration data section of the view

abbycross:
Closes #3904 

cc: @dcmckayibm @marcolincs

### DIFF
--- a/docs/guides/qpu-information.mdx
+++ b/docs/guides/qpu-information.mdx
@@ -154,7 +154,7 @@ The readout_length specifies the duration of the readout operation for a qubit. 
 <span id="finite-duration-discrete"></span>
 ### ID error / √x (sx) error / Pauli-X error / RX error
 
-Error in the finite-duration discrete one-qubit gates, measured from randomized benchmarking. The randomized benchmarking sequence includes SX, ID, X, and RX gates, and it is assumed their errors are the same. The ID gate is a delay of duration equal to the duration of the √X and X gates. The RX gate is also the same duration of the √X and X gates with variable amplitude, and so it is reported as having the same error as these gates.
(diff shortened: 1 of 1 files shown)


docs/guides/qpu-information.mdx
```diff
@@ -152,18 +152,14 @@ Similarly, this parameter represents the probability of measuring a qubit in the
...
 
 <span id="finite-duration-discrete"></span>
-### ID error / √x (sx) error / Pauli-X error
+### ID error / √x (sx) error / Pauli-X error / RX error
 
-Error in the finite-duration discrete one-qubit gates, measured from randomized benchmarking. The randomized benchmarking sequence includes SX, ID, and X gates, and it is assumed their errors are the same. The ID gate is a delay of duration equal to the duration of the √X and X gates.
+Error in the finite-duration discrete one-qubit gates, measured from randomized benchmarking. The randomized benchmarking sequence includes SX, ID, X, and RX gates, and it is assumed their errors are the same. The ID gate is a delay of duration equal to the duration of the √X and X gates. The RX gate is also the same duration of the √X and X gates with variable amplitude, and so it is reported as having the same error as these gates.
```
marcolincs:
```suggestion
Error in the finite-duration discrete one-qubit gates, measured from randomized benchmarking. The randomized benchmarking sequence includes SX, ID, and X gates, and it is assumed their errors are the same. The ID gate is a delay of duration equal to the duration of the √X and X gates. The RX gate is also the same duration as the √X and X gates with variable amplitude, and so it is reported as having the same error as these gates.
```
the additional phrase is because RX is not part of the sequence.

abbycross: Thanks for catching, @marcolincs!

Also, I just now changed The RX gate is also the same duration _of_ the √X and X gates with variable amplitude
to The RX gate is also the same duration _as_ the √X and X gates with variable amplitude. Does that make sense?

Issue #785, opened 2024-02-08
Remove some historical Qiskit API docs post Qiskit 1.0


Eric-Arellano:
Currently, we host in docs.quantum.ibm.com all the API docs for Qiskit that used to be in qiskit.org. Some of these docs are quite old, like Qiskit 0.19 from June 2020 (3.5 years old). It seems unlikely folks are still using this.

The main benefit of deleting old versions is to use fewer resources, such as speeding up CI in this repository because we don't need to run quality checks on those docs. Another benefit to users is less noise in the version selector because fewer options to choose.

But, we lose some history.

--

If we delete, we need to adjust our redirect strategy to avoid introducing 404s.

--

Update February 20, 2024: We discussed this in the dev meeting. Devs are not strongly against this, but also not super convinced it's worth removing. They challenged that we can simply choose to not run quality checks on old API docs.

They were more receptive to keeping the URLs up, but hiding them in the version selector. So have to manually change the URL or arrive at it from Google to access the version.

Either way, they're curious metrics on how many people visit the old docs.

--

Update Dec 18, 2024: we should revisit this as part of adding globalization and related changes. It might be a natural time. People wanting very old versions like Qiskit 0.19 can still access GitHub for the original source of truth; although, we should recognize users do often need to migrate their programs from ~a year before to newer versions.

Eric-Arellano:
This has been completed. We have a custom 410 page like this:

<img width="707" height="379" alt="Image" src="https://github.com/user-attachments/assets/79b2d1c5-4e98-41d8-b15e-93c39f6c0379" />

PR #3919, opened 2025-09-15
Remove old Qiskit SDK and qiskit-ibm-runtime versions

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/785.

Two of the Runtime release notes will get fixed in https://github.com/Qiskit/qiskit-ibm-runtime/pull/2402. In the meantime, we ignore them; this isn't great, but is fine because we'll have a custom 410 error page when users click these two links.


arnaucasau (approved):
🙌

PR #3942, opened 2025-09-22
How to use job tags

abbycross:
Add an example of adding, updating, and retrieving job tags.

PR #3947, opened 2025-09-22
Add job tags page

abbycross:
A followup to #3942 : in discussion with the original requestors, we decided the save/retrieve jobs topic is too different from the add job tags topic, and so we separate them in this PR.

### DIFF
--- a/docs/guides/_toc.json
+++ b/docs/guides/_toc.json
@@ -572,7 +572,7 @@
             },
             {
               "title": "Add and search by job tags",
-              "url": "/docs/guides/job-limits"
+              "url": "/docs/guides/add-job-tags"
             },
             {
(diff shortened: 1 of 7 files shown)


docs/guides/_toc.json
```diff
@@ -571,7 +571,11 @@
               "url": "/docs/guides/job-limits"
             },
             {
-              "title": "Tag, retrieve, and save results",
+              "title": "Add and search by job tags",
```
drew-distefano: A broader, outcome focused title like "Organize and search jobs using tags" or "Manage and search job tags" (if it needs to be shorter) might connect more with users and also encompassees the 'update' part of the guide.

beckykd (approved):

PR #3966, opened 2025-09-25
Update MDX guide and fix some GP issues

arnaucasau:
This PR fixes some files that were causing problems when they were sent to translate & updates the MDX guide to reflect some constrains when using definition tooltips and setting the markdown metadata.


mdx-guide.md
```diff
@@ -56,6 +56,8 @@ In Jupyter notebooks, set `title` and `description` in the `metadata` section fo
 }
 ```
 
+Warning: In any case, try to avoid using hyphens, colons, semicolons, backticks, and pipes when writing the `title` and `description`. These characters might be used differently in other languages, and their use might complicate the translation process. The Globalization Pipeline could replace them with other symbols that could change their meaning or even break the page.
```
Eric-Arellano:
```suggestion
Avoid using hyphens, colons, semicolons, backticks, and pipes when writing the `title` and `description`. These characters might be used differently in other languages, and their use might complicate the translation process. Our translation tooling often replaces them with other symbols that could change their meaning or even break the page.
```

mdx-guide.md
```diff
@@ -219,7 +221,10 @@ To use a `DefinitionTooltip`, use the following syntax:
 
 For full list of props, please check [here](https://react.carbondesignsystem.com/?path=/docs/components-definitiontooltip--playground#component-api).
 
-Warning: do not use LaTeX/math expressions in the same paragraph as a definition tooltip because it will break the styling. Use a new line to separate out the two into separate paragraphs.
+Caveats:
+
+1. Do not use any interactive content (e.g. links) as a definition tooltip child nor into the `definition` prop. If you use such content, it will not be available to users of screen reader software.
```
Eric-Arellano: Content writers don't know what children means

```suggestion
1. Do not use any interactive content (e.g. links) with the `definition` prop or inside the tags. If you use such content, it will not be available to users of screen reader software.
```

mdx-guide.md
```diff
@@ -219,7 +221,10 @@ To use a `DefinitionTooltip`, use the following syntax:
...
+2. Avoid using single and double quotes in the `definition` prop when possible. Even if they are escaped, they could be a problem when translating the files.
```
Eric-Arellano:
```suggestion
2. Do not use single and double quotes in the `definition` prop. Even if they are escaped, they cause problems when translating the files.
```

learning/courses/quantum-safe-cryptography/quantum-safe-cryptography.ipynb
```diff
@@ -56,7 +56,7 @@
    "source": [
     "## Quantum algorithms and impacts to cryptography\n",
     "\n",
-    "| Quantum Algorithm | Functionality      | <DefinitionTooltip definition=\"As defined by the [NIST](https://csrc.nist.gov/glossary/term/security_strength): *A number characterizing the amount of work that is expected to suffice to “defeat” an implemented cryptographic mechanism (for example, by compromising its functionality and/or circumventing the protection that its use was intended to facilitate). Security strength is often expressed in bits. If the security strength of a particular implementation of a cryptographic mechanism is s bits, it is expected that the equivalent of (roughly) 2<sup>s</sup> basic operations of some sort will be sufficient to defeat it in some way.*\">Security Strength</DefinitionTooltip> ($n$ = number of bits) | Impacted Cryptographic Protocols                 | Mitigation             |\n",
```
abbycross, line 59: Did these quotation marks cause a problem because they are not the usual "straight" quotation marks? “” vs. ""

arnaucasau: Hmm.. probably that contributed, but the GP is not super fond of having quotation marks in their translations. This is because in other language they can make no sense, so they get replaced by different characters. Usually, it's not a big deal, but given that these ones are inside the definition prop, it was breaking our page in Japanese and Korean.

I think we should try to avoid using quotations inside props when possible, so I thought it was best to remove them entirely here. If we can't remove them in a particular case, we will need to manually fix the translation later on.

Issue #2904, opened 2025-03-31
Release notes folder for Qiskit SDK is very large in table of contents


Eric-Arellano:
Would it make any sense to stop having `release-notes` be a collapsible folder and instead only have it be the index page `/api/qiskit/release-notes`? 

<img width="254" alt="Image" src="https://github.com/user-attachments/assets/596edb43-1f40-461f-aaea-84bd7fdfeb06" />

Eric-Arellano:
- major version folders in left ToC
- create dedicated design issue for the index page itself

gracelindsell:
@Eric-Arellano designs are completed and can be found [here](https://www.figma.com/design/sNXucK1uMaZ8XXyLTZrdrs/IBM-Quantum-Documentation?node-id=11179-116175&t=B6NgAfSHwgAefkfe-1)

Eric-Arellano:
This is the design. Very similar to https://github.com/Qiskit/documentation/issues/2941, such as same order of versions and grouping by major version:

<img width="447" height="441" alt="Image" src="https://github.com/user-attachments/assets/b9b94efa-1696-462a-9212-ed7424625813" />

The main change is that this needs to update the `_toc.json` file for every folder in `api/qiskit`.

## Step 1

We have this code to generate the `_toc.json`:

https://github.com/Qiskit/documentation/blob/c5780a96bdb96401f9bbfea8fc34791c6ebd9a91/scripts/js/lib/api/generateToc.ts#L232-L247

Lines 240-246 are what we need to change.

This code should use the new function from https://github.com/Qiskit/documentation/pull/3902 to group by major versions. Then, have TOC entries like:

```json
{"title": "v2", "children": [
  {"title": "v2.1", "url": "/docs/api/qiskit/release-notes/2.1"}, 
  {"title": "v2.0", "url": "/docs/api/qiskit/release-notes/2.0"}
]}
```

You should make generateReleaseNotesEntry public so that you can write unit tests for it. Reminder to use `Pkg.mock()` to help with writing the unit tests.

Then test that things are actually working in the real world by running `npm run gen-api -- -p qiskit -v 2.1.2` and checking that the `_toc.json` file changes are expected. If that is all working, commit the change. Then run `npm run regen-apis -- -p qiskit` - you should see the `_toc.json` updated for every version, i.e. they should all be using the same new format.

## Step 2

TODO. I'm still thinking of the best approach. We need to rework `updateHistoricalTocFiles` in `releaseNotes.ts`.

[...]

PR #3979, opened 2025-09-26
Release notes folder for Qiskit SDK is very large in table of contents

GantaRoja:
Added groupBymajorversion function in generateReleaseNotesEntry  and wrote test case for this function #2904

### DIFF
--- a/docs/api/qiskit/0.46/_toc.json
+++ b/docs/api/qiskit/0.46/_toc.json
@@ -3493,204 +3493,219 @@
       "title": "Release notes",
       "children": [
         {
-          "title": "2.2",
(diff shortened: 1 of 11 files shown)


scripts/js/lib/api/generateToc.test.ts
```diff
@@ -314,3 +318,43 @@ test("generate a toc without modules and releaes notes", () => {
     title: "My Quantum Project",
   });
 });
+
+test("returns nested entries grouped by major version", () => {
```
Eric-Arellano: Can you actually please get rid of this test and combine its logic into `TOC with separate release note files`? Like update the arguments to `Pkg.mock` so that you end up testing this behavior of grouping by major versions.

PR #3984, opened 2025-09-26
Refactor how we determine release note page versions

Eric-Arellano:
Prework for part 2 of https://github.com/Qiskit/documentation/issues/2904. 

This PR removes a confusing part of our release note code: `ReleaseNotesConfig.separatePagesVersions` was not guaranteed to have every version it needed! Now, the expectation is that when creating `ReleaseNotesConfig`, you already have all the versions populated.

This doesn't fully unblock https://github.com/Qiskit/documentation/issues/2904, but it makes the code less confusing.

PR #3987, opened 2025-09-29
Refactor release notes handling for historical API docs when new version

Eric-Arellano:
This implements part 2 of https://github.com/Qiskit/documentation/issues/2904. 

Before, our code was very brittle for when we have a brand new version of Qiskit and need to add its release note to the _toc.json for all historical versions. It was going to break with https://github.com/Qiskit/documentation/pull/3979.

Now, the new code should be more resilient. The new code generates in-memory the canonical "Release notes" section one-time, as it should be the same for all versions of Qiskit. Then, we simply overwrite the `_toc.json` for every version with this canonical `TocEntry`. When combined with https://github.com/Qiskit/documentation/pull/3984, we should now always handle release notes correctly.

### DIFF
--- a/scripts/js/lib/api/releaseNotes.ts
+++ b/scripts/js/lib/api/releaseNotes.ts
@@ -22,6 +22,7 @@ import type { HtmlToMdResultWithUrl } from "./HtmlToMdResult.js";
 import { C_API_BASE_PATH, DOCS_BASE_PATH } from "./conversionPipeline.js";
 import { kebabCaseAndShortenPage } from "./normalizeResultUrls.js";
 import { removePrefix } from "../stringUtils.js";
+import { generateReleaseNotesEntry, TocEntry } from "./generateToc.js";
 
 // ---------------------------------------------------------------------------
 // Generic release notes handling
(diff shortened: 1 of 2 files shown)


scripts/js/lib/api/releaseNotes.ts
```diff
@@ -182,44 +183,27 @@ New features, bug fixes, and other changes in previous versions of ${pkg.title}.
 async function updateHistoricalTocFiles(pkg: Pkg): Promise<void> {
   console.log("Updating _toc.json files for the historical versions");
 
+  const releaseNotesEntry = generateReleaseNotesEntry(pkg);
```
arnaucasau: Thoughts on renaming it to `maybeReleaseNotes`?

scripts/js/lib/api/releaseNotes.ts
```diff
@@ -182,44 +183,27 @@ New features, bug fixes, and other changes in previous versions of ${pkg.title}.
...
-  // Add the current API version in the second position of the list
-  if (releaseNotesNode.children[0].title != newVersion) {
-    releaseNotesNode.children.unshift({
-      title: newVersion,
-      url: `${DOCS_BASE_PATH}/api/qiskit/release-notes/${newVersion}`,
+    const tocPath = `docs/api/${pkg.name}/${folder.name}/_toc.json`;
+    const rawToc = await readFile(tocPath, {
+      encoding: "utf8",
     });
+    let tocJson = JSON.parse(rawToc)
```
arnaucasau: This could be `const`:

```suggestion
    const tocJson = JSON.parse(rawToc)
```

arnaucasau (approved):
Nice!

Issue #3980, opened 2025-09-26
Qiskit release notes: the latest major version should be open by default


Eric-Arellano:
In docs/api/qiskit/release-notes, @1ucian0 requested that the newest major version is open by default:

<img width="438" height="322" alt="Image" src="https://github.com/user-attachments/assets/55c84d15-c766-445a-bbd4-907f62a25e90" />

We can do that by simply adding `open` to the `<details>`, like this:

```
<details open>
<summary>v2</summary>
- [v2.2](./2.2)
- [v2.1](./2.1)
- [v2.0](./2.0)
</details>
```
 
You'll need to modify this function to take an additional argument like `kwargs: {isLatestVersion: boolean}`. Change how `<details>` is set up based on it. 

https://github.com/Qiskit/documentation/blob/513f344b6dd68eeb6dc6c5f6a66d5728760d5a7d/scripts/js/lib/api/releaseNotes.ts#L289

Update the call site to set `isLatestVersion` if it's the first entry in the for loop. You could switch the for loop to instead be a call to `.forEach((version, idx) => ...)`

https://github.com/Qiskit/documentation/blob/513f344b6dd68eeb6dc6c5f6a66d5728760d5a7d/scripts/js/lib/api/releaseNotes.ts#L174

PR #3999, opened 2025-09-30
latest version should open by default

GantaRoja:
Modified releasenotes.ts files so that latest version should open by default and updated the testcases #3980


scripts/js/lib/api/releaseNotes.ts
```diff
@@ -170,9 +170,16 @@ New features, bug fixes, and other changes in previous versions of ${pkg.title}.
...
-  }
+
+  let isFirst = true;
+  grouped.forEach((versionList, majorVersion) => {
+    const isLatestVersion = isFirst;
+    markdown +=
+      renderVersionGroup(majorVersion, versionList, { isLatestVersion }) +
+      "\n\n";
+    isFirst = false;
+  });
```
Eric-Arellano: I believe this should work. It's more idiomatic

```suggestion
  grouped.forEach(([versionList, majorVersion], idx) => {
    markdown +=
      renderVersionGroup(majorVersion, versionList, { isLatestVersion: idx === 0 }) +
      "\n\n";
  });
```

You might need to `npm run fmt`, and you should verify this code works

PR #4003, opened 2025-09-30
Qedma: QESEM qiskit function + OBP addon tutorial

oria-qedma:
This is a draft of a Qesem QF tutorial showing simulation of 2d transverse field Ising model with Qesem error mitigation combined with qiskit operator backpropagation module. 
Feedback from IBM team is welcome. 

FYI @assafb

### DIFF
--- a/docs/tutorials/qedma-2d-ising-with-qesem.ipynb
+++ b/docs/tutorials/qedma-2d-ising-with-qesem.ipynb
@@ -1354,7 +1354,7 @@
     "<div class=\"alert alert-block alert-info\">\n",
     "<b>Note:</b>\n",
     "\n",
-    "The reason that the 9th step has a large statistical error bar is because we limited the QPU time to 5 minutes. If you run this step for 15 minutes (as the empirical time estimation suggests), you will get a smaller error bar. Hence, the mitigated value will become closer to the ideal value\n"
+    "The reason that the 9th step has a large statistical error bar is because we limited the QPU time to 5 minutes. If you run this step for 15 minutes (as the empirical time estimation suggests), you will get a smaller error bar. Hence, the mitigated value will become closer to the ideal value."
    ],
    "id": "e14cb01633a8451a"
   },


valeriechiang:
Thank you for the draft @oria-qedma! Is it possible to scale this up to 30+ qubits? 

cc: @miamico @pandasa123

oria-qedma:
Yes it is possible to run this using a larger number of qubits , we just have to play a bit with the parameters to keep the QPU runtime short enough on the one hand but still make sure the example shows some interesting behavior on the other hand.

valeriechiang:
More feedback from the team: It would be interesting to include a result+depth comparison when using OBP+ QESEM versus just QESEM. Thoughts?

cc: @pandasa123 @jenglick

talishnaider-qedma:
> More feedback from the team: It would be interesting to include a result+depth comparison when using OBP+ QESEM versus just QESEM. Thoughts?
> 
> cc: @pandasa123 @jenglick

Hey Valerie, Thanks fo the feedback! Yeah it could be a nice comparison. Currently the notebook takes around 20 minutes of QPU time (for all the trotter steps up to 9), I'm afraid it might be intimidating for users if it were to take double the time (approximately, haven't tried it yet) let's say 50 minutes.. What do you think?

abbycross:
Authors: If there is an option along the right-hand side of this PR to "Allow edits from maintainers" - can you please enable it, if you don't mind? That way we can commit to your branch.

valeriechiang:
> > More feedback from the team: It would be interesting to include a result+depth comparison when using OBP+ QESEM versus just QESEM. Thoughts?
> > cc: @pandasa123 @jenglick
> 
> Hey @valeriechiang , Thanks fo the feedback! Yeah it could be a nice comparison. Currently the notebook takes around 20 minutes of QPU time (for all the trotter steps up to 9), I'm afraid it might be intimidating for users if it were to take double the time (approximately, haven't tried it yet) let's say 50 minutes.. What do you think?

@talishnaider-qedma This is a good point. Maybe we can make the comparison to just QESEM optional (and add a disclaimer note that it will take ~50 minutes)? Then we can keep the overall tutorial time estimate as 20 minutes. 

cc: @miamico @jenglick @pandasa123

oria-qedma:
> Authors: If there is an option along the right-hand side of this PR to "Allow edits from maintainers" - can you please enable it, if you don't mind? That way we can commit to your branch.

@abbycross  I don't see such an option, maybe because I don't have admin rights to the fork? (I think maybe @assafb created the fork originally for other documentation work)

abbycross:
> > Authors: If there is an option along the right-hand side of this PR to "Allow edits from maintainers" - can you please enable it, if you don't mind? That way we can commit to your branch.
> 
> @abbycross I don't see such an option, maybe because I don't have admin rights to the fork? (I think maybe @assafb created the fork originally for other documentation work)

Thanks for checking! - it's certainly possible that it's an admin-only option.

docs/tutorials/qedma-2d-ising-with-qesem.ipynb
miamico, line 1: Awesome, this is looking much better! A few more comments:
- Rename the H2 titles for the Qiskit Patterns steps to follow our standards (e.g. "Step 1: Define circuit and observables" should be "Step 1: Map classical inputs to a quantum problem", others steps need correction too)
- Do I understand correctly that the experiment ran is a 10q example? If that's the case, we really need to bump that number as high as reasonably possible. Something on the 30+ qubits scale would be great! If it becomes problematic for the QPU or classical runtime than it's something that we may want to discuss and highlight resource usage of the technique in the discussion
- nevermind the comment about the layers, I think the visualization of the coupling map is good

miamico: - I would skip the noisy backend example and go straight to hardware unless it is adding any particular insight ( I can imagine that you could show how the learnt noise model is close to the defined noise model but I don't think this is happening at the moment)

valeriechiang: @talishnaider-qedma can you take a look? Thank you!

talishnaider-qedma: Hey guys, thanks! I'm working on it will let you know when there's a new version

abbycross:
@talishnaider-qedma, two questions:

1. We need a usage estimate on a specific processor family (Heron, which revision?).

2. Also, which versions of the package installs are needed - the latest for all?

- qiskit-ibm-catalog
- qiskit-addon-obp and qiskit-addon-utils
- qiskit-aer
- matplotlib

talishnaider-qedma:
> @talishnaider-qedma it might be that you haven't [signed the CLA](https://github.com/Qiskit/documentation/pull/4003#issuecomment-3353009902) yet - can you please check?
> 
> I am reviewing the tutorial now - will hope to get it in today's deploy if possible.

Hey @abbycross, you're right I signed it now

abbycross:
Hm, the branches should have been the same?...  I'll sort out the differences so that we can update it!

PR #4021, opened 2025-10-03
Updated state with proper qubit order

christopherporter1:
Very complicated state vector correction.

### DIFF
--- a/learning/courses/quantum-machine-learning/data-encoding.ipynb
+++ b/learning/courses/quantum-machine-learning/data-encoding.ipynb
@@ -945,42 +945,64 @@
    "id": "0f573f85-c2a2-449e-bcd4-74d42072e0db",
    "metadata": {},
    "source": [
-    "Up to the first barrier (a point we label $b1$), our states are:\n",
+    "As we write our state, we will use the Qiskit convention that least-significant qubits are ordered to the far right, as in $|q_2,q_1,q_0\\rangle$ or $|q_2\\rangle\\otimes|q_1\\rangle\\otimes|q_0\\rangle.$ These states can become very complicated very quickly, and this rare example may help explain why such states are seldom written out explicitly.\n",
+    "\n",
+    "Our system starts in the state $|00\\rangle.$ Up to the first barrier (a point we label $b1$), our states are:\n",
     "\n",
     "$$\n",
-    "|\\psi\\rangle_{b1} = \\left(\\cos(\\theta_0/2)|0\\rangle+\\sin(\\theta_0/2)e^{i\\theta_2}|1\\rangle\\right)\\otimes\\left(\\cos(\\theta_1/2)|0\\rangle+\\sin(\\theta_1/2)e^{i\\theta_3}|1\\rangle\\right)\n",
+    "|\\psi\\rangle_{b1} = \\left(\\cos\\left(\\frac{\\theta_1}{2}\\right)|0\\rangle+\\sin\\left(\\frac{\\theta_1}{2}\\right)e^{i\\theta_3}|1\\rangle\\right)\\otimes\\left(\\cos\\left(\\frac{\\theta_0}{2}\\right)|0\\rangle+\\sin\\left(\\frac{\\theta_0}{2}\\right)e^{i\\theta_2}|1\\rangle\\right)\n",
     "$$\n",
     "\n",
-    "That's just dense encoding, which we've seen before. Now after the CNOT gate, at the second barrier ($b2$), our state is\n",
-    "\n",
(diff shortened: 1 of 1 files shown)


learning/courses/quantum-machine-learning/data-encoding.ipynb
```diff
@@ -989,11 +1014,11 @@
...
    "metadata": {},
    "source": [
-    "At first glance, it may appear that we have loaded so more parameters onto just a few states than makes sense, since the final state can be written as $\\psi_\\text{final} = c_0|00\\rangle+c_1|01\\rangle+c_2|10\\rangle+c_3|11\\rangle$. But note that each prefactor is complex! Written like this:\n",
+    "This is likely too complicated to parse. But step back and think about how many parameters we loaded onto the state: 8. But we have with just four computational basis states. At first glance, it may appear that we have loaded more parameters than makes sense, since the final state can be written as $\\psi_\\text{final} = c_0|00\\rangle+c_1|01\\rangle+c_2|10\\rangle+c_3|11\\rangle$. But note that each prefactor is complex! Written like this:\n",
     "$$\n",
     "\\psi_\\text{final} = (a_0+ib_0)|00\\rangle+(a_1+ib_1)|01\\rangle+(a_2+ib_2)|10\\rangle+(a_3+ib_3)|11\\rangle\n",
     "$$\n",
-    "One can see that we do, indeed, have 8 parameters on the state on which to encode our 8 features.\n",
+    "ne can see that we do, indeed, have 8 parameters on the state on which to encode our 8 features.\n",
```
```suggestion
    "One can see that we do, indeed, have eight parameters on the state on which to encode our eight features.\n",
```

learning/courses/quantum-machine-learning/data-encoding.ipynb
```diff
@@ -989,11 +1014,11 @@
    "id": "93e9e6ae-85e8-4077-9c25-f947daebe610",
    "metadata": {},
    "source": [
-    "At first glance, it may appear that we have loaded so more parameters onto just a few states than makes sense, since the final state can be written as $\\psi_\\text{final} = c_0|00\\rangle+c_1|01\\rangle+c_2|10\\rangle+c_3|11\\rangle$. But note that each prefactor is complex! Written like this:\n",
+    "This is likely too complicated to parse. But step back and think about how many parameters we loaded onto the state: 8. But we have with just four computational basis states. At first glance, it may appear that we have loaded more parameters than makes sense, since the final state can be written as $\\psi_\\text{final} = c_0|00\\rangle+c_1|01\\rangle+c_2|10\\rangle+c_3|11\\rangle$. But note that each prefactor is complex! Written like this:\n",
```
```suggestion
    "This is likely too complicated to parse. But step back and think about how many parameters we loaded onto the state: eight. But we have with just four computational basis states. At first glance, it may appear that we have loaded more parameters than makes sense, since the final state can be written as $\\psi_\\text{final} = c_0|00\\rangle+c_1|01\\rangle+c_2|10\\rangle+c_3|11\\rangle$. But note that each prefactor is complex! Written like this:\n",
```

abbycross: I note also that three sentences in this paragraph start with the word "But" 🤣  Probably best to only start one of them that way.

learning/courses/quantum-machine-learning/data-encoding.ipynb
```diff
@@ -989,11 +1014,11 @@
    "id": "93e9e6ae-85e8-4077-9c25-f947daebe610",
    "metadata": {},
    "source": [
-    "At first glance, it may appear that we have loaded so more parameters onto just a few states than makes sense, since the final state can be written as $\\psi_\\text{final} = c_0|00\\rangle+c_1|01\\rangle+c_2|10\\rangle+c_3|11\\rangle$. But note that each prefactor is complex! Written like this:\n",
+    "This is likely too complicated to parse. But step back and think about how many parameters we loaded onto the state: eight. But we have with just four computational basis states. At first glance, it may appear that we have loaded more parameters than makes sense, since the final state can be written as $\\psi_\\text{final} = c_0|00\\rangle+c_1|01\\rangle+c_2|10\\rangle+c_3|11\\rangle$. But note that each prefactor is complex! Written like this:\n",
```
christopherporter1:
```suggestion
    "This is likely too complicated to parse. Instead, just step back and think about how many parameters we loaded onto the state: eight. But we have with just four computational basis states. At first glance, it may appear that we have loaded more parameters than makes sense, since the final state can be written as $\\psi_\\text{final} = c_0|00\\rangle+c_1|01\\rangle+c_2|10\\rangle+c_3|11\\rangle$. Note, however, that each prefactor is complex! Written like this:\n",
```

PR #4039, opened 2025-10-07
Add tutorial notebook for M3 readout mitigation

jlapeyre:
This tutorial presents an application of [qiskit-addon-mthree](https://github.com/Qiskit/qiskit-addon-mthree), the "Matrix-free Measurement Mitigation (M3)" technique. (It is actually readout mitigation). The technique is applied to a simple [hidden shift problem](https://en.wikipedia.org/wiki/Hidden_shift_problem)

In particular, we follow [citation 5](https://en.wikipedia.org/wiki/Hidden_shift_problem#cite_note-5) in the article linked above ([arXiv](https://arxiv.org/abs/0811.3208)).

This PR was co-authored by @kevinsung

PR #4054, opened 2025-10-10
Mthree tutorial

abbycross:
@jlapeyre Here is your branch with my changes. Sorry it split off into a new PR.

[You can see a preview here](https://qiskit.github.io/documentation/pr-4054/docs/tutorials/readout-error-mitigation-sampler)

Continuation of #4039 

Authors: @kevinsung @jlapeyre


miamico (approved):
Looks good to me! @jlapeyre please confirm whether this is good to be merged

miamico:
> Ok @jlapeyre I've finished copyediting - ready for you to review. 
> 
> @miamico how do the tutorial surveys work? Do we put one at the end of all tutorials, and if so, do we generate a custom link for each? I see different URLs on different tutorials (but I also don't see a survey on every one.)

The surveys are from before the migration. I think we need a new solution. I have been removing whenever we edit an old tutorial and not including them in new ones.

docs/tutorials/readout-error-mitigation-sampler.ipynb
```diff
@@ -0,0 +1,1414 @@
...
+    "or by a density matrix.\n",
+    "Measurement of the qubit register into a classical bit register then proceeds in two steps.\n",
+    "First the quantum measurement proper is performed.\n",
+    "This means that the state of the qubit register\n",
+    "is projected onto a single basis state that is characterized\n",
+    "by a string of $1$s and $0$s.\n",
+    "The second step consists of reading the bitstring characterizing this basis state\n",
+    "and writing it into classical computer memory.\n",
+    "We call this step *readout*.\n",
+    "It turns out that readout error is more important than error because it can incorrectly collapse the state.\n",
```
jlapeyre:
```suggestion
    "It turns out that the second step (readout) incurs more error than the first step (projection onto basis states).\n",
```
My original text here was poorly worded and unclear. Your edit changed my intended meaning. I have tried to make it more clear in this suggestion. But I'm still not 100% happy with the syntax.

docs/tutorials/readout-error-mitigation-sampler.ipynb
```diff
@@ -0,0 +1,1414 @@
...
+    "If the sampled bitstring (stored in classical memory) differs from the bitstring characterizing\n",
+    "the projected quantum state, we say that a readout error has occurred.\n",
+    "These errors are observed to be random and uncorrelated from sample to sample.\n",
+    "It has proven useful to model readout error as a _noisy classical channel_.\n",
+    "That is, for every pair of\n",
+    "bitstrings $i$ and $j$, there is a fixed probability that a true value of $j$ will\n",
+    "be incorrectly read as $i$.\n",
+    "\n",
+    "More precisely, for every pair of bitstrings $(i, j)$, there is a (conditional) probability ${M}_{i,j}$\n",
+    "that $i$ is read, given that the true value is $j$.\n",
```
jlapeyre: In my view of the rendered document, the final period is on a separate line. I think putting the `.` inside, like this: `$j.$` might fix this.

docs/tutorials/readout-error-mitigation-sampler.ipynb
abbycross: Glad you read closely to find that I changed the meaning - apologies!! I think it makes sense now, if you're trying to convey that the two steps are projection onto basis states, and readout, and that the second step is more prone to error.

docs/tutorials/readout-error-mitigation-sampler.ipynb
abbycross: You're correct, that should fix it. I'll move the period.

abbycross:
```suggestion
    "that $i$ is read, given that the true value is $j.$\n",
```

abbycross:
> Looking at all the tutorials, it seems that the doc-local spelling word list is always visible at the top. I followed other tutorials as an example. Just checking that this is indeed the case.

@jlapeyre you mean the cspell statement? It does render in notebooks but will not display in the html guide. ([See preview here](https://qiskit.github.io/documentation/pr-4054/docs/tutorials/readout-error-mitigation-sampler).)

docs/tutorials/readout-error-mitigation-sampler.ipynb
```diff
@@ -0,0 +1,1414 @@
...
+   "id": "c98d8409-66ad-452f-b87a-d8df886d77d4",
+   "metadata": {},
+   "source": [
+    "### Qiskit M3 addon: Scaling to longer bitstrings\n",
+    "\n",
+    "The M3 method is an alternative to the direct method of equation (2)\n",
+    "for mitigating readout error in the experimental distribution over bitstrings.\n",
+    "While the direct method is limited to bitstrings no longer than about $10$ bits,\n",
+    "M3 can handle much longer bitstrings.\n",
+    "The two important features of M3 are\n",
```
jlapeyre: My original text was ambiguous and lacked precision. Because of this, your edit changed the intended (but unclear) meaning. The following is better, I think:

Solving equation (2) using standard numeric linear algebra routines is limited to bitstrings no longer than about $10$ bits. M3, however, can handle much longer bitstrings. Two key properties of M3 that make this possible are:

docs/tutorials/readout-error-mitigation-sampler.ipynb
```diff
@@ -0,0 +1,1414 @@
...
+    "$$\n",
+    "  g(x, y) = {f}(x+a, y+b) = (-1)^{(x+a) \\cdot (y+b)},\n",
+    "  $$\n",
+    "  where $a$ and $b$ are (hidden) parameters.\n",
+    "  We are given two black boxes, one implementing $f$, and the other $g$.\n",
+    "  We suppose that we know that they compute the functions defined above, except that we know\n",
+    "  neither $a$ nor $b$. The game is to determine the hidden bitstrings (shifts)\n",
+    "  $a$ and $b$ by making queries to $f$ and $g$. It's clear that if we play the game classically,\n",
+    "  we need $O(2m)$ queries to determine $a$ and $b$. For example, we can query $g$\n",
+    "  with all pairs of strings that, when combined, have exactly one element set to $1$.\n",
```
jlapeyre: Once again, the original is ambiguous.
Better:
For example, we can query $g$ with all pairs of strings such that one element of the pair is all zeros, and the other element has exactly one element set to $1$.

jlapeyre (changes requested):
Thanks, Abby. I went through the diff of all of your changes. They are all clear improvements to the document. The few comments I made are at places where my original text was unclear.

Apart from these points, this looks good to me.

docs/tutorials/readout-error-mitigation-sampler.ipynb
```diff
@@ -0,0 +1,1414 @@
...
+   "id": "c98d8409-66ad-452f-b87a-d8df886d77d4",
+   "metadata": {},
+   "source": [
+    "### Qiskit M3 addon: Scaling to longer bitstrings\n",
+    "\n",
+    "The M3 method is an alternative to the direct method of equation (2)\n",
+    "for mitigating readout error in the experimental distribution over bitstrings.\n",
+    "While the direct method is limited to bitstrings no longer than about $10$ bits,\n",
+    "M3 can handle much longer bitstrings.\n",
+    "The two important features of M3 are\n",
```
abbycross:
```suggestion
    "Solving equation (2) using standard numeric linear algebra routines is limited to bitstrings no longer than about 10 bits. M3, however, can handle much longer bitstrings. Two key properties of M3 that make this possible are:\n",
```

docs/tutorials/readout-error-mitigation-sampler.ipynb
```diff
@@ -0,0 +1,1410 @@
...
+    "$$\n",
+    "  g(x, y) = {f}(x+a, y+b) = (-1)^{(x+a) \\cdot (y+b)},\n",
+    "  $$\n",
+    "  where $a$ and $b$ are (hidden) parameters.\n",
+    "  We are given two black boxes, one implementing $f$, and the other $g$.\n",
+    "  We suppose that we know that they compute the functions defined above, except that we know\n",
+    "  neither $a$ nor $b$. The game is to determine the hidden bitstrings (shifts)\n",
+    "  $a$ and $b$ by making queries to $f$ and $g$. It's clear that if we play the game classically,\n",
+    "  we need $O(2m)$ queries to determine $a$ and $b$. For example, we can query $g$\n",
+    "  with all pairs of strings that, when combined, have exactly one element set to $1$.\n",
```
abbycross:
```suggestion
    "  we need $O(2m)$ queries to determine $a$ and $b$. For example, we can query $g$ with all pairs of strings such that one element of the pair is all zeros, and the other element has exactly one element set to $1$.\n",
```

PR #4085, opened 2025-10-15
Added a lot of explanation on ordering of data vs qubits

### LINKED ISSUE #4076: IBM Qiskit Learning platform Quantum Machine Learning Course

https://quantum.cloud.ibm.com/learning/en/courses/quantum-machine-learning/data-encoding

### Select all that apply

- [ ] new content request
- [x] typo
- [x] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

1) The efficient SU(2) encoding of the output of the code does not align with the calculation of the state after the CNOT gate is applied the order is different; it should be q0 to q1  as per IBM standard Qiskit syntax notation order. Might be wrong may need to correct the calculations after that.
2)Similarly, in the Z feature map, the equation of encoding order does not follow the explanation given, which correctly follows the IBM standard Qiskit syntax notation order. 

Please verify with the expert and correct it if necessary.
thank you
Regards,
Pradhyumna Parthasarathy.

Image
Image

christopherporter1:
Fixes #4076

### DIFF
--- a/learning/courses/quantum-machine-learning/data-encoding.ipynb
+++ b/learning/courses/quantum-machine-learning/data-encoding.ipynb
@@ -1034,7 +1034,7 @@
     "\\vec{x} = (0.1,0.2,0.3,0.4,0.5,0.6,0.7,0.8,0.9,1.0,1.1,1.2)\n",
     "$$\n",
     "\n",
-    "In this data vector, the features are shown in a particular order. It is not obvious if it matters that they be encoded in this order, as opposed to the reverse order. What is important is keeping track of it. Note in the circuit diagram that `efficient_su2` assumes a certain ordering of encoding, specifically filling the first layer of parameterized gates from qubit 0 to qubit 2, and then moving to the next layer. This is neither consistent nor inconsistent with little-endian notation, since here the data features cannot be ordered by qubit *a priori*, before an encoding circuit has been specified."
+    "In this data vector, the features are shown in a particular order. In isolation, it doesn't matter if they are encoded in this order or in the reverse. What is important is keeping track of it and being consistent. Note in the circuit diagram that `efficient_su2` assumes a certain ordering of encoding, specifically filling the first layer of parameterized gates from qubit 0 to qubit 2, and then moving to the next layer. This is neither consistent nor inconsistent with little-endian notation, since here the data features cannot be ordered by qubit *a priori*, before an encoding circuit has been specified."
    ]
   },
   {


learning/courses/quantum-machine-learning/data-encoding.ipynb
```diff
@@ -1105,10 +1107,12 @@
...
+    "To start with a small example, a two-qubit ZFM circuit is coded using Qiskit and drawn to display the simple circuit structure. In the example, a single repetition, $r=1$, is implemented with the data vector $\\vec{x} = \\left(\\textstyle\\frac{1}{2}\\pi, \\textstyle\\frac{1}{3}\\pi\\right)$. Note that this is written in the standard order of a vector in Python, meaning the $0^\\text{th}$ element is $\\textstyle\\frac{1}{2}\\pi.$ We are free to encode this $0^\\text{th}$ feature onto our $0^\\text{th}$ qubit, or onto our $N^\\text{th}.$ Again, there cannot always be a single 1:1 mapping from feature order to qubit order, since different feature maps encode different numbers of features to each qubit. Again what is important is that we are aware of where each feature is being encoded. When providing a parameter list to the $Z$ feature map, it will encode feature 0 from the list to the least-significant qubit with a parameterized gate, as in qubit 0. So we will follow that convention when doing this by hand. We will encode $\\textstyle\\frac{1}{2}\\pi$ on the $0^\\text{th}$ qubit, and $\\textstyle\\frac{1}{3}\\pi$ on the $1^\\text{st}$ qubit,.\n",
```
```suggestion
    "To start with a small example, a two-qubit ZFM circuit is coded using Qiskit and drawn to display the simple circuit structure. In the example, a single repetition, $r=1$, is implemented with the data vector $\\vec{x} = \\left(\\textstyle\\frac{1}{2}\\pi, \\textstyle\\frac{1}{3}\\pi\\right)$. Note that this is written in the standard order of a vector in Python, meaning the $0^\\text{th}$ element is $\\textstyle\\frac{1}{2}\\pi.$ We are free to encode this $0^\\text{th}$ feature onto our $0^\\text{th}$ qubit, or onto our $N^\\text{th}.$ Again, there cannot always be a single 1:1 mapping from feature order to qubit order, since different feature maps encode different numbers of features to each qubit. Again what is important is that we are aware of where each feature is being encoded. When providing a parameter list to the $Z$ feature map, it will encode feature 0 from the list to the least-significant qubit with a parameterized gate, as in qubit 0. So we will follow that convention when doing this by hand. We will encode $\\textstyle\\frac{1}{2}\\pi$ on the $0^\\text{th}$ qubit, and $\\textstyle\\frac{1}{3}\\pi$ on the $1^\\text{st}$ qubit.\n",
```

learning/courses/quantum-machine-learning/data-encoding.ipynb
```diff
@@ -1032,7 +1032,9 @@
     "\n",
     "$$\n",
     "\\vec{x} = (0.1,0.2,0.3,0.4,0.5,0.6,0.7,0.8,0.9,1.0,1.1,1.2)\n",
-    "$$"
+    "$$\n",
+    "\n",
+    "In this data vector, the features are shown in a particular order. It is not obvious if it matters that they be encoded in this order, as opposed to the reverse order. What is important is keeping track of it. Note in the circuit diagram that `efficient_su2` assumes a certain ordering of encoding, specifically filling the first layer of parameterized gates from qubit 0 to qubit 2, and then moving to the next layer. This is neither consistent nor inconsistent with little-endian notation, since here the data features cannot be ordered by qubit *a priori*, before an encoding circuit has been specified."
```
beckykd: If it's accurate, I think it would be more clear to say "It doesn't matter if they are encoded in this order or the reverse.  What is...it.  However, in the circuit diagram, `efficient_su2` assumes...

christopherporter1:
```suggestion
    "In this data vector, the features are shown in a particular order. In isolation, it doesn't matter if they are encoded in this order or in the reverse. What is important is keeping track of it and being consistent. Note in the circuit diagram that `efficient_su2` assumes a certain ordering of encoding, specifically filling the first layer of parameterized gates from qubit 0 to qubit 2, and then moving to the next layer. This is neither consistent nor inconsistent with little-endian notation, since here the data features cannot be ordered by qubit *a priori*, before an encoding circuit has been specified."
```

christopherporter1: I added a minor qualifier "In isolation..." but otherwise used your phrasing. Thanks!

PR #4122, opened 2025-10-23
Fix typo in PEA tutorial =+ should be +

jlapeyre:
There is an erroneous extra `=` in an equation.

This PR removes it.

For the correct expression, see, for example Eq (1) in [this paper](https://arxiv.org/pdf/1805.04492)  ([abstract for same paper](https://arxiv.org/abs/1805.04492))


abbycross (approved):
Thanks!

PR #4182, opened 2025-11-03
move eagle to retired page

beckykd:


abbycross (approved):
Thanks! Just delete the image and lint will be happy

PR #4183, opened 2025-11-03
Document how to use the Card component

Eric-Arellano:


mdx-guide.md
```diff
@@ -287,6 +287,31 @@ There is a specific use case where you want to show instructions for different o
 </OperatingSystemTabs>
 ```
 
+### Card
+
+The `Card` component creates a clickable card that links to another page, such as our [API index page](https://quantum.cloud.ibm.com/docs/api). 
+
+To use, first set up a `CardGroup` component. Then, insert one or more `Card` components inside the `CardGroup`. For each `Card`, you must set the `title`, `description`, `href`, and and `analyticsName`, such as the following example:
```
abbycross: Is there a standard format for the `analyticsName` field entry?

Also, I thought I remembered @arnaucasau said something is optional - is that still the case?

arnaucasau: @abbycross I would say that `analyticsName` should be a brief "title" that can be used to identify the card among the rest. When sending the analytics, we will automatically add the prefix `MDX Card:`, so something like the card title and its location could be valid (see Eric's examples down below)

The optional argument is `linkText`, which is missing in this line:

```suggestion
To use, first set up a `CardGroup` component. Then, insert one or more `Card` components inside the `CardGroup`. For each `Card`, you must set the `title`, `description`, `href`, and `analyticsName`. Moreover you can change the default link's text by setting the optional property `linkText`.
```

mdx-guide.md
```diff
@@ -287,6 +287,31 @@ There is a specific use case where you want to show instructions for different o
...
+      description="Supports the construction of high-fidelity circuits with reduced depth"
+      href="/docs/api/qiskit-addon-aqc-tensor"
+      analyticsName="Overview page: card: addon AQC Tensor"
+    />
+
+    <Card
+      title="Qiskit SDK"
+      description="A radical software development kit"
+      href="/docs/api/qiskit"
+      analyticsName="Overview page: card: Qiskit SDK"
```
arnaucasau:
```suggestion
      analyticsName="Overview page: card: Qiskit SDK"
      linkText="Learn more"
```

mdx-guide.md
arnaucasau: FYI - https://github.com/Qiskit/documentation/pull/4183#discussion_r2489387460

abbycross (approved):

PR #4203, opened 2025-11-07
Ajc/qedma qesem

abbycross:
Supersedes #4003

### DIFF
--- a/docs/guides/_toc.json
+++ b/docs/guides/_toc.json
@@ -11,7 +11,9 @@
         },
         {
           "title": "Quickstart",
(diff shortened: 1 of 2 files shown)


talishnaider-qedma:
@abbycross @miamico @pandasa123 @oria-qedma I think there was some kind of a mix up here. The version in this PR is definitely not the last version I uploaded to the previous PR https://github.com/Qiskit/documentation/pull/4003/. The version here is the first version we sent you.
The latest version has 4 steps and they are named according to your convention and a bunch of other changes... Could you address this? and update the notebook? :)

PR #4218, opened 2025-11-10
Change symbol characters to code versions

abbycross:
Found several instances where ® ™ and † are used instead of writing out the html or latex code. Potential for globalization issues, so fixing here proactively.


abbycross:
Nooooooooooo broken

PR #4219, opened 2025-11-10
replace tm, dagger, r symbols with code

abbycross:
Proactively replacing these symbols with code in case they run afoul of the globalization process.

### DIFF
--- a/learning/modules/computer-science/qft.ipynb
+++ b/learning/modules/computer-science/qft.ipynb
@@ -963,7 +963,7 @@
     "2. T/F QFT can be implemented using only Hadamard and CNOT gates.\n",
     "3. T/F QFT is a key component of Shor’s algorithm.\n",
     "4. T/F The output of Quantum Phase Estimation is a quantum state representing the eigenvector of the operator.\n",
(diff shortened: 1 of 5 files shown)


Eric-Arellano (approved):
Thanks!

learning/modules/computer-science/qft.ipynb
```diff
@@ -963,7 +963,7 @@
     "2. T/F QFT can be implemented using only Hadamard and CNOT gates.\n",
     "3. T/F QFT is a key component of Shor’s algorithm.\n",
     "4. T/F The output of Quantum Phase Estimation is a quantum state representing the eigenvector of the operator.\n",
-    "5. T/F QPE requires the use of the inverse Quantum Fourier Transform (QFT†).\n",
+    "5. T/F QPE requires the use of the inverse Quantum Fourier Transform (QFT$^\\dag$).\n",
```
arnaucasau, line 966: For future cases:

Would it be possible to include `QFT` into the math expression? Otherwise the pipeline can (and probably will) add a space between the dagger symbol and the text. Ideally this should be `${QFT}^\\dag$` to be treated as a whole.

PR #4258, opened 2025-11-14
Fixes `KeyError` when transpling to a QPU in Krylov tutorial

henryzou50:
Resolves the issue of `KeyError: (0, 14)` when doing `twoq_gate_err = target[op_name][q].error` as we transpile to a QPU. The fix here is to add the `if_else` instruction to the backend, as the `op_name`can be `if_else`. With this fix, the notebook will run successfully with the latest versions of the requirements.


miamico:
BTW there are some further comment on this tutorial as part of this issue #4265

PR #4277, opened 2025-11-18
Add contributor tooling setup script

### LINKED ISSUE #3233: Add `./setup` script to check for valid installation
One of our writers was confused when `./fix` failed:

```
/Users/opt/miniconda3/envs/qaoa/lib/python3.11/subprocess.py", line 1026, in __init__
      self._execute_child(args, executable, preexec_fn, close_fds,
    File "/Users/opt/miniconda3/envs/qaoa/lib/python3.11/subprocess.py", line 1953, in _execute_child
      raise child_exception_type(errno_num, err_msg, err_filename)
  FileNotFoundError: [Errno 2] No such file or directory: 'magick'
```

It'd be nicer to proactively check that `magick` is installed, probably with https://docs.python.org/3/library/shutil.html#shutil.which. If not, fail with a good error message. 

I'm thinking we should create a script called `./setup` that will try to automatically run any necessary commands like `npm install`, and will also check that all necessary tools are installed like tox and npm. If not, it will fail with a nice error message. It should not install any system dependencies.

If `./setup` is fast enough, we should have `./check` and `./fix` automatically invoke it. (They could do that by calling the `main()` function; this is why we have the main guard `if __name__ == "__main__")`.

Double check this works on Windows.

## npm install

When invoking `./setup`, we should always call `npm install`.

However, if the setup logic is being called from `./fix` and `./check`, we should avoid automatically calling `npm install` because of the performance hit. Only run `npm install` if `node_modules` is non-existent or empty.

frankharkins:
This PR adds a new script (`./setup`) that will check the system has the appropriate packages installed.

Closes #3233.

### DIFF
--- a/scripts/tool_runner/qiskit_docs_contributor_tools/__init__.py
+++ b/scripts/tool_runner/qiskit_docs_contributor_tools/__init__.py
@@ -56,7 +56,7 @@ def install_node_dependencies() -> None:
     run_cmd("Installed Node dependencies", ["npm", "install"], progress=None)
 
 
-def check_packages(verbose=False) -> None:
+def check_packages(print_on_success=False) -> None:
     """Check the current system has the required packages"""
     checks = [
         (
@@ -90,5 +90,5 @@ def check_packages(verbose=False) -> None:
         if not installed:
             print(message)
             raise SystemExit(1)
-        if verbose:
+        if print_on_success:
             print(message)
--- a/setup
+++ b/setup
@@ -20,7 +20,7 @@ from scripts.tool_runner.qiskit_docs_contributor_tools import (
 
 def main() -> None:
     print("Checking for installed packages...")
-    check_packages(verbose=True)
+    check_packages(print_on_success=True)
     install_node_dependencies()
 
 


scripts/tool_runner/qiskit_docs_contributor_tools/__init__.py
```diff
@@ -47,3 +49,41 @@ def run_cmd(name: str, cmd: list[str], *, progress: str | None) -> None:
...
+def check_packages(verbose=False) -> None:
+    """Check the current system has the required packages"""
+    checks = [
+        (
+            "node",
+            (
+                "If you expect to use JavaScript in other projects, consider "
+                "using NVM (https://github.com/nvm-sh/nvm). Otherwise, consider "
+                "using Homebrew (https://formulae.brew.sh/formula/node) or "
+                "installing Node.js directly (https://nodejs.org/en)."
```
arnaucasau: The nodejs website automatically sets the locale based on the previous user choice (like in our app)
```suggestion
                "installing Node.js directly (https://nodejs.org)."
```

setup
```diff
@@ -0,0 +1,32 @@
...
+    run_cmd,
+)
+from scripts.tool_runner.qiskit_docs_contributor_tools import check_packages
+
+
+def main() -> None:
+    print("Checking for installed packages...")
+    check_packages(verbose=True)
+    print("Installing Node dependencies (this may take a while)...")
+    run_cmd("Installed Node dependencies", ["npm", "install"], progress=None)
```
arnaucasau: Thoughts on DRYing this line with the `./check` script?

scripts/tool_runner/qiskit_docs_contributor_tools/__init__.py
```diff
@@ -47,3 +49,41 @@ def run_cmd(name: str, cmd: list[str], *, progress: str | None) -> None:
...
+                "downloading the binary."
+            ),
+        ),
+    ]
+    for command, install_instructions in checks:
+        installed, message = check_for_package(command, install_instructions)
+        if not installed:
+            print(message)
+            raise SystemExit(1)
+        if verbose:
```
arnaucasau: What do you think about renaming the argument to something like `print_on_success` or `success_output`. At first I was confused about it because I was expecting the function to silently fail when `verbose=False`

scripts/tool_runner/qiskit_docs_contributor_tools/__init__.py
```diff
@@ -47,3 +49,41 @@ def run_cmd(name: str, cmd: list[str], *, progress: str | None) -> None:
 
     # This ends the program without adding a noisy stacktrace.
     raise SystemExit(1)
+
+
+def check_packages(verbose=False) -> None:
+    """Check the current system has the required packages"""
+    checks = [
```
arnaucasau, line 61: This array is very nice! It will be super easy to add new checks in the future!

arnaucasau (commented):
Awesome! I tested it locally and everything worked well! 🚀

scripts/tool_runner/qiskit_docs_contributor_tools/__init__.py
frankharkins: Good spot, thanks

Eric-Arellano:
This is great! I created two follow up issues:

* https://github.com/Qiskit/documentation/issues/4384
* https://github.com/Qiskit/documentation/issues/4385

PR #4282, opened 2025-11-19
Add more information to external link checker logs

frankharkins:
It's helpful to be able to glance at the logs and determine why a link failed.

For example: A 502 status code is most likely a temporary flake, which we can ignore. Whereas a 410 status code means the link is definitely broken.

This PR adds the status code to the logs and reports if the fetch failed for some other reason.

### DIFF
--- a/scripts/js/lib/links/ExternalLink.ts
+++ b/scripts/js/lib/links/ExternalLink.ts
@@ -73,9 +73,6 @@ function responseToErrorMessage(
   const isOk = httpCode >= 100 && httpCode < 300;
   if (isOk) return undefined;
 
-  const isRedirectLike = httpCode >= 300 && httpCode < 399;
-  if (isRedirectLike) return undefined;
-
   if (httpCode === 404) return `Could not find link '${link}' (${httpCode})`;
   if (httpCode === 410) return `Link '${link}' has been removed (${httpCode})`;
   if (httpCode === 418) return `Link '${link}' is a teapot (${httpCode})`;
--- a/scripts/js/lib/links/ExternalLink.test.ts
+++ b/scripts/js/lib/links/ExternalLink.test.ts
@@ -55,26 +55,6 @@ test.describe("ExternalLink.check()", () => {
     );
   });
 
-  test("temporary redirect", async () => {
-    global.fetch = () => Promise.resolve(new Response("", { status: 302 }));
-    let link = new ExternalLink("https://bad-link.com", ["/testorigin.mdx"]);
-    const result = await link.check();
-    expect(result).toBeUndefined();
-  });
-
-  test("permanent redirect", async () => {
-    global.fetch = () =>
-      Promise.resolve(
-        new Response("", {
-          status: 308,
-          headers: { location: "https://good-link.com" },
-        }),
-      );
-    let link = new ExternalLink("https://bad-link.com", ["/testorigin.mdx"]);
-    const result = await link.check();
-    expect(result).toBeUndefined();
-  });
-
   test("other problem", async () => {
     global.fetch = () => Promise.resolve(new Response("", { status: 502 }));
     let link = new ExternalLink("https://bad-link.com", ["/testorigin.mdx"]);


scripts/js/lib/links/ExternalLink.ts
```diff
@@ -58,6 +51,47 @@ export class ExternalLink {
...
 }
 
+async function safeFetch(
+  link: string,
+): Promise<{ response: Response } | { error: Error }> {
+  try {
+    const response = await fetch(link, {
+      headers: getHeaders(link),
+      method: "HEAD",
+      redirect: "manual",
```
frankharkins: This prevents `fetch` from throwing an error when redirected

arnaucasau: I'm curious, have you run the script with all our docs (including APIs) to see how many new issues we find? Some of them are very subtle like:

```
arnaucasau@Arnaus-MacBook-Pro documentation % npm run check:external-links -- 'docs/guides/composer.mdx'

> qiskit-documentation@0.1.0 check:external-links
> tsx scripts/js/commands/checkExternalLinks.ts docs/guides/composer.mdx

❌ Link 'https://arxiv.org/pdf/1707.03429.pdf' has permanently moved to '/pdf/1707.03429' (301). Appears in:
    docs/guides/composer.mdx
❌ Link 'https://arxiv.org/pdf/1707.03429.pdf' has permanently moved to '/pdf/1707.03429' (301). Appears in:
    docs/guides/composer.mdx


💔 Some external links appear broken, although it's possible they are flakes. Checked 7 links.
```

Thoughts on using a different icon (other than ❌) for these type of errors? I feel like it'd be helpful to visually see what are the links that are broken vs the links that still work, in case we have a lot of them

frankharkins: Embarrassingly, no. I mistakenly assumed we'd been reporting 3xx error codes due to the following line, plus the fact that some redirected links were reported in the logs.

```ts
      if (response.status >= 300) {
```

However, it seems fetch _was_ following redirects most of the time, and this PR flags many more issues rather than reducing the number as I expected. I don't know if that's desirable or not. I think it makes sense to fix some, but it's not _that_ important.

frankharkins: I've reflected on it, and I've pushed a commit to ignore all 3xx codes. This means this PR should flag the same number of issues as before, the only difference being the extra information provided. Maybe in the future we can look at updating links.

arnaucasau: > assumed we'd been reporting 3xx error codes due to the following line

Yeah, it's a bit confusing when looking at the original code. I think the key is that without `redirect: "manual"`, `fetch` will follow every redirect until it gets to the final destination (or it reaches a maximum hop limit). That means that it will never return a 3xx code (if I'm not mistaken).

What do you think about removing the `redirect: "manual"` again? Otherwise, I think we can hit an edge case in the following scenario:

1. Fetch URL X
2. X redirects to Y
3. Y redirects to Z
4. Z returns a 404

With `redirect: "manual"` we will stop at step 2, get a 3xx code and our code will say that everything is okay. Without `redirect: "manual"`, we will get to step 4 and detect the 404

> I don't know if that's desirable or not. I think it makes sense to fix some, but it's not that important.

Agreed! To me it's also good to fix, but definitely lower priority than fixing the broken links. So, I think it's okay to punt it for now as you said 👍

scripts/js/lib/links/ExternalLink.ts
```diff
@@ -58,6 +51,39 @@ export class ExternalLink {
...
+function responseToErrorMessage(
+  link: string,
+  response: Response,
+): string | undefined {
+  const httpCode = response.status;
+  const isOk = httpCode >= 100 && httpCode < 300;
+  if (isOk) return undefined;
+
+  const isRedirectLike = httpCode >= 300 && httpCode < 399;
+  if (isRedirectLike) return undefined;
```
arnaucasau: Could we add a small comment explaining why we skip them?

scripts/js/lib/links/ExternalLink.ts
frankharkins: Yes good point, done :)

scripts/js/lib/links/ExternalLink.ts
frankharkins: Maybe we should remove that block altogether if we're not setting `redirect: 'manual'`?

arnaucasau: Yeah, I think that works too :)

scripts/js/lib/links/ExternalLink.ts
Eric-Arellano: Yeah, agreed it's good to fix redirects because there's a risk they take away the redirect in the future. Although, it's indeed lower priority. For example, imagine people linking to the old Qiskit textbook, if we ever took away those redirects because it's years later.

> I mistakenly assumed we'd been reporting 3xx error codes due to the following line

Do you remember how many issues we're talking about?

Issue #4200, opened 2025-11-07
Fix the Qiskit C API `objects.inv` broken links


arnaucasau:
The Qiskit C API has always had some broken links in the `objects.inv` file that we never fixed:

https://github.com/Qiskit/documentation/blob/3798d09d86eff2d66c4d928d756c3811ba594378/scripts/js/lib/links/ignores.ts#L423-L445

After the release of Qiskit v2.2.3, our link checker found roughly 320 new broken links that could be caused by our script not fully understanding the changes done in https://github.com/Qiskit/qiskit/pull/15250

To unblock the release of the docs, we ignored the `objects.inv` entirely from our link checker in https://github.com/Qiskit/documentation/pull/4159, so this issue is used to track fixing them.

PR #4304, opened 2025-11-24
Fix some C API `objects.inv` links

frankharkins:
Part of https://github.com/Qiskit/documentation/issues/4200

Looking at the broken links, the vast majority have a period (`.`) in their anchors because they're links to members of an object.

* One example is `/docs/api/qiskit-c/qk-bit-term#qk_bitterm_label.bit_term`. The part after the period is a parameter of the function [`qk_bitterm_label`](https://quantum.cloud.ibm.com/docs/en/api/qiskit-c/qk-bit-term#qk_bitterm_label). Parameters don't have HTML anchors, hence the broken link.
* Another example is `/docs/api/qiskit-c/qk-circuit#qkdelayunit.qkdelayunit_ms`, a value of [`QkDelayUnit`](https://quantum.cloud.ibm.com/docs/en/api/qiskit-c/qk-circuit#qkdelayunit). This enum value _does_ have an anchor, but due to our conversion script, the anchor is `#enumerator-qkdelayunit_ms`.

This PR fixes this by mapping such links to their parent objects. This fixes 220/236 links. While it would be nice to be able to link directly to the members, it's going to be quite a lot of work. Mapping to the parent object is very simple (see 02758e8a4c224cb74d151dfc7d2a8b3344dc5c19), and will take the user very close to the memeber.

I regenerated all APIs to make sure this didn't break any other links (it fixed two links in `qiskit-c v2.1`). I will continue investigating the other 16 broken links.

### DIFF
--- a/scripts/js/lib/api/objectsInv.ts
+++ b/scripts/js/lib/api/objectsInv.ts
@@ -169,12 +169,8 @@ export class ObjectsInv {
     // `QkObsTerm::num_qubits` or `qk_bitterm_label.bit_term`), so we instead
     // just point to the parent object. This is a best-effort attempt that
     // should get users close to the right place.
-    if (name.includes("::")) {
-      const objectName = name.split("::")[0].toLowerCase();
-      return `${path}#${objectName}`;
-    }
-
-    const objectName = name.split(".")[0].toLowerCase();
+    const delimiter = name.includes("::") ? "::" : ".";
+    const objectName = name.split(delimiter)[0].toLowerCase();
     return `${path}#${objectName}`;
   }
 


scripts/js/lib/api/objectsInv.ts
```diff
@@ -166,10 +166,15 @@ export class ObjectsInv {
...
-    const objectName = name.split("::")[0].toLowerCase();
+    // `QkObsTerm::num_qubits` or `qk_bitterm_label.bit_term`), so we instead
+    // just point to the parent object. This is a best-effort attempt that
+    // should get users close to the right place.
+    if (name.includes("::")) {
+      const objectName = name.split("::")[0].toLowerCase();
+      return `${path}#${objectName}`;
+    }
+
+    const objectName = name.split(".")[0].toLowerCase();
```
arnaucasau: Thoughts on adding tests for this case?

arnaucasau: Nit: You could DRY the code by creating a variable for the delimiter. Something like:

```
const delimiter = name.includes("::") ? "::" : ".";
const objectName = name.split(delimiter)[0].toLowerCase();
return `${path}#${objectName}`;
```

But feel free to ignore this if you think it's clearer as-is 👍

frankharkins: Yep good idea

frankharkins: nice, thanks!

Issue #1970, opened 2024-09-17
Add branch protection for "Execute Notebooks" and "API Checks"


Eric-Arellano:
It's currrently possible to merge on red. This is particularly dangerous because the button to "merge when ready" will merge even if "API Checks" and "Execute Notebooks" are red. This is because we didn't set up those CI checks in branch protection due to issues with how we skip those jobs - when skipped, it would block the PR from merging since it never runs.

To fix, implement https://www.pantsbuild.org/blog/2022/10/10/skipping-github-actions-jobs-without-breaking-branch-protection.

Eric-Arellano:
Looks like GitHub might have fixed this quirk, per https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks

![Image](https://github.com/user-attachments/assets/035381da-f0d1-4cb4-be71-37e55284d5c8)

I'm trying to test this out and will close if true.

Eric-Arellano:
The test for https://github.com/Qiskit/documentation/pull/2323 shows that this should be okay now 🚀 At least, I confirmed that the status checks don't block being added to a merge queue. 

I didn't confirm that the merge queue works properly because I don't want to merge that PR - I'll reopen this issue if it fails.

Eric-Arellano:
For some reason, the api-checks and Execute notebook jobs weren't actually being enforced. Perhaps we didn't save the config?

PR #4321, opened 2025-11-26
Use Python script for status check path filtering in CI

frankharkins:
Work towards #1970

This PR has changed since I first opened it and I have force-pushed to clean up the commits.

We can't use GitHub's path filtering to have status checks be required for some files but not for others. If a check is required but does not run due to path filtering, then CI can never pass (see https://github.com/orgs/community/discussions/44490#discussion-4761945).

This PR adds a workaround for the notebook test job: We use a separate job to decide if the check should run and use conditionals to skip the main job if the check isn't required.

### DIFF
--- a/.github/workflows/notebook-test.yml
+++ b/.github/workflows/notebook-test.yml
@@ -13,17 +13,31 @@
 name: Test notebooks
 on:
   pull_request:
-    paths:
-      - ".github/workflows/notebook-test.yml"
-      - "docs/**/*.ipynb"
-      - "learning/**/*.ipynb"
-      - "!docs/api/**/*"
-      - "scripts/nb-tester/**/*"
   workflow_dispatch:
 jobs:
+  check_paths:
+    name: Decide whether to run
+    if: github.repository_owner == 'Qiskit'
+    runs-on: ubuntu-latest
+    outputs:
+      SHOULD_RUN_NOTEBOOK_TESTER: ${{ steps.filter.outputs.SHOULD_RUN_NOTEBOOK_TESTER }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: scripts/ci/pr-path-filter.py
+      - name: Check paths
+        id: filter
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          CHANGED_FILES=$(gh pr diff ${{ github.event.number }} --name-only)
+          echo $CHANGED_FILES | python scripts/ci/pr-path-filter.py
+          cat $GITHUB_OUTPUT
   execute:
     name: Execute notebooks
-    if: github.repository_owner == 'Qiskit'
+    needs: [check_paths]
+    if: needs.check_paths.outputs.SHOULD_RUN_NOTEBOOK_TESTER == 'true'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
--- a/scripts/ci/pr-path-filter.py
+++ b/scripts/ci/pr-path-filter.py
@@ -1,0 +1,101 @@
+# This code is a Qiskit project.
+#
+# (C) Copyright IBM 2025.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+import os
+import sys
+from typing import TypedDict
+
+
+class Config(TypedDict):
+    SHOULD_RUN_NOTEBOOK_TESTER: str
+
+
+def filter_paths(changed_files: list[str]) -> Config:
+    contains_script_file = any(
+        path == ".github/workflows/notebook-test.yml"
+        or path.startswith("scripts/nb-tester")
+        for path in changed_files
+    )
+    if contains_script_file:
+        return {"SHOULD_RUN_NOTEBOOK_TESTER": "true"}
+
+    def is_content_notebook(path: str) -> bool:
+        return path.endswith(".ipynb") and (
+            path.startswith("learning/")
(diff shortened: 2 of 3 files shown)


.github/workflows/notebook-test.yml
```diff
@@ -28,15 +22,19 @@ jobs:
...
-          files: "{{docs,learning}/**/*.ipynb,scripts/nb-tester/**/*,.github/workflows/notebook-test.yml}"
           separator: "\n"
           write_output_files: true
 
+      - name: Check if this action should run
+        id: path-filter
+        run: python scripts/ci/should-run-notebook-tester.py
+
       - name: Check if extra linux deps needed
+        if: ${{ steps.path-filter.outputs.SHOULD_RUN }} == "true"
```
arnaucasau: Rather than having the `if` in each step, what do you think about creating a precursor job called something like `evaluate`? That way we can have the `execute` job wait for it and use the conditional on like 20 to skip the entire process. Maybe that could simplify other parts too 🤔

frankharkins: Thanks! Yes that would definitely work. Only downside is I think it would mean two "Test notebooks" jobs would appear under each PR, but maybe that doesn't really matter. It would certainly make the code simpler.

arnaucasau: Yeah, I think that's fine as long as the job names are descriptive enough. We can test it out in this PR to see how it looks like.

If we do that, what do you think about not using python at all for the path filtering? Even though I agree with you that it has some benefits, I think it adds some complexity + it's more code to maintain. We could maybe use the same action we were using until now to get the changed paths and use grep with a regex to decide if we have to skip or not

frankharkins: That's fair, this ended up being more complex than I planned. What do you think of https://github.com/Qiskit/documentation/pull/4324 as a compromise?

Bear in mind that the grep+regex is still code to maintain. It's less lines for sure, but we can't test it easily. For example, I discovered while adding tests for this PR that https://github.com/Qiskit/documentation/pull/3048 didn't work because I didn't update `scripts/ci/pr-execute-notebook.py`.

.github/workflows/notebook-test.yml
```diff
@@ -13,17 +13,31 @@
...
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: scripts/ci/pr-path-filter.py
+      - name: Check paths
+        id: filter
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          CHANGED_FILES=$(gh pr diff ${{ github.event.number }} --name-only)
```
frankharkins, line 34: I don't know why I didn't think of this before: We can use the `gh` cli to get the names of the changed files, no need for the `tj-actions/changed-files` action. We could replace that action everywhere in a follow-up.

.github/workflows/notebook-test.yml
```diff
@@ -13,17 +13,31 @@
...
+  check_paths:
+    name: Decide whether to run
+    if: github.repository_owner == 'Qiskit'
+    runs-on: ubuntu-latest
+    outputs:
+      SHOULD_RUN_NOTEBOOK_TESTER: ${{ steps.filter.outputs.SHOULD_RUN_NOTEBOOK_TESTER }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: scripts/ci/pr-path-filter.py
```
frankharkins, line 27: This means we only checkout the path filter script, which saves ~10s

.github/workflows/notebook-test.yml
```diff
@@ -13,17 +13,31 @@
...
+      - name: Check paths
+        id: filter
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          CHANGED_FILES=$(gh pr diff ${{ github.event.number }} --name-only)
+          echo $CHANGED_FILES | python scripts/ci/pr-path-filter.py
```
frankharkins, line 35: I should also mention that we could replace this with a grep call and set `GITHUB_OUTPUT` conditioned on that result (as per @arnaucasau suggestion). I think that would be fine, especially if we include a list of paths to test against on regex101

```
# If changing this regex, test it on https://regex101.com/ with the
# following paths (you can paste the block in and uncomment it):
#
#  -- Should match:
#  .github/workflows/notebook-test.yml
#  scripts/nb-tester/requirements.txt
#  docs/guides/thing.ipynb
#  learning/courses/my-course/notebook.ipynb
#
#  -- Should not match:
#  docs/guides/thing.mdx
#  docs/api/thing.mdx
#  docs/api/thing.ipynb
#  scripts/js/lib/thing.ts
#
# Add new tests if necessary
```

I found the following pattern worked correctly on regex101, but for some reason couldn't make it work in `grep`.

```
(\.github\/workflows\/notebook-test\.yml|(docs|learning)(?!\/api\/)\/.*\.ipynb|scripts\/nb-tester\/.*)
```

arnaucasau: Oh, good idea with adding the comment with some tests to run in regex101. I think it'd be nice if we can follow this approach and have something similar to the current `paths`.

For example, I have been testing with this script and I think it works well:

```bash
INCLUDE=(
  '^\.github/workflows/notebook-test\.yml$'
  '^docs/.*\.ipynb$'
  '^learning/.*\.ipynb$'
  '^scripts/nb-tester/.*$'
)
EXCLUDE='^docs/api/.*$'

INCLUDE_REGEX=$(IFS='|'; echo "${INCLUDE[*]}")
EXCLUDE_REGEX=$(IFS='|'; echo "${EXCLUDE[*]}")

CHANGED_FILES=(docs/api/qiskit/index.ipynb scripts/nb-tester/test.py learning/courses/test/test.ipynb)

echo "${CHANGED_FILES[@]}" | tr ' ' '\n' | grep -E "$INCLUDE_REGEX" | grep -Eqv "$EXCLUDE_REGEX" && echo "True" || echo "False"
```

I suggested using bash because I think it's easier to understand, to generalize to other jobs like API checks, and to update when needed (you just go to the workflow file and add/remove a regex). However, if you think python will be more maintainable, I'm also okay with it.

I'm also curious to know @Eric-Arellano's opinion?

PR #4326, opened 2025-11-28
Speed up internal link checker

frankharkins:
Improving check time is important to our team (see #2854). A large portion of the time is spent in the internal link checker.

I profiled the link checker to see if there were any opportunities to improve things. It looks like most of the time is spent parsing markdown, which we can't really avoid. However, we can do a few things to mitigate it without introducing any more complexity:

* c09865fd512d24ffc828d200ac985583136c7e1d does two things
  * Remove the stringification step, we don't need to convert the AST back to markdown, so we can avoid that work.
  * Traverse the tree one time, rather than three separate times
* 2b92c9ecac8727ea2209fde2354c23fab1e4b721 avoids blocking on file read. This means we're parsing markdown while we're waiting for the filesystem.

A very crude test shows somewhere between a 10%–30% speedup on my machine.

Before: (30.53, 29.10, 29.08, 29.41) Mean 29.53 (+-~1) 
After:  (25.19, 23.84, 23.95, 25.56) Mean 24.64 (+-~1)

EDIT: More evidence this is a real speedup is the link checker in the API checks takes ~6m for [this PR](https://github.com/Qiskit/documentation/actions/runs/19760027099/job/56619546894), whereas it took between 8–9m in the last three PRs that it ran on, which roughly matches the speedup estimate on my machine.

### DIFF
--- a/scripts/js/lib/links/FileBatch.ts
+++ b/scripts/js/lib/links/FileBatch.ts
@@ -67,22 +67,26 @@ export class FileBatch {
    *   2. A list of InternalLink objects to validate.
    */
   async load(): Promise<[File[], InternalLink[]]> {
-    const files: File[] = [];
-    for (let filePath of this.toLoad) {
+    const parsedFilesToLoad = this.toLoad.map(async (filePath) => {
       const parsed = await parseFile(filePath);
-      files.push(new File(filePath, parsed.anchors));
-    }
+      return new File(filePath, parsed.anchors);
+    });
 
     const linksToOriginFiles = new Map<string, string[]>();
-    for (const filePath of this.toCheck) {
+    const parsedFilesToCheck = this.toCheck.map(async (filePath) => {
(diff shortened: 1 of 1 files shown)


scripts/js/lib/links/extractLinks.ts
```diff
@@ -77,26 +77,24 @@ export async function parseLinks(
...
-        }
-      });
-      visit(tree, "link", (TreeNode) => addLink(TreeNode.url));
-      visit(tree, "image", (TreeNode) => addLink(TreeNode.url));
-    })
-    .use(remarkStringify)
-    .process(markdown);
+    .parse(markdown);
+
+  visit(tree, (node: any) => {
```
arnaucasau, line 88: Interesting! I didn't know you could have the visitors outside like this. TIL.

arnaucasau (approved):
Great investigation! I tested out the code locally and it was working well for me 🚀

PR #4337, opened 2025-12-01
Document how to use lists with tables

Eric-Arellano:
Turns out, we already supported lists, admonitions, and links. You only need a special syntax for lists. A closed-source PR will improve the styling of lists.

screenshot_2025-12-01_at_4 11 32___pm_480


beckykd (approved):
Thanks!

PR #4340, opened 2025-12-02
Restore image for Qiskit 0.45 release notes

Eric-Arellano:
I manually restored the file and updated the release note link. We don't regenerate the release notes for old Qiskit SDK, so this is safe.


arnaucasau (approved):
Thanks!

PR #4345, opened 2025-12-02
Move content out of `version-info` cells

frankharkins:
These admonitions will be removed when the notebook is next run. I've also added a warning to the version info cells to prevent this happening again in the future.


Eric-Arellano (approved):
Thanks. Good catch

PR #4349, opened 2025-12-03
Combine PR checks into single workflow

frankharkins:
This PR enables fixing #1970 by combining all PR checks into a single workflow. We'll then need to edit the settings to require all jobs in this workflow.

There are some things I'd like to neaten up in follow-up PRs, but this is already a large PR to review, so I've just done the minimum to keep the same behaviour as before while satisfying #1970.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
     name: Decide which jobs to run
     runs-on: ubuntu-latest
     # We don't run these tests in the merge queue
-    if: false && github.repository_owner == 'Qiskit' && github.event_name == 'pull_request'
+    if: github.repository_owner == 'Qiskit' && github.event_name == 'pull_request'
     outputs:
       run-notebook-tester: ${{ steps.decide.outputs.RUN_NOTEBOOK_TESTER }}
       run-api-checks: ${{ steps.decide.outputs.RUN_API_CHECKS }}


.github/workflows/main.yml
```diff
@@ -10,9 +10,70 @@
...
 # that they have been altered from the originals.
 
-name: Lint
+name: Pull request
 on: [pull_request, merge_group]
 jobs:
+  decide-jobs:
+    name: Decide which jobs to run
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'Qiskit' && github.event_name == 'pull_request'
```
frankharkins: `github.event_name == 'pull_request'` prevents the notebook tester and api checks from running in the merge queue. This is the same behaviour as before.

.github/workflows/main.yml
```diff
@@ -104,3 +165,93 @@ jobs:
...
+      - uses: actions/checkout@v4
+
+      - name: Get changed files
+        id: all-changed
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          echo "CHANGED_FILES<<EOF" >> $GITHUB_OUTPUT
+          gh pr diff -R Qiskit/documentation ${{ github.event.number }} --name-only >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
```
frankharkins: I migrated this from the "changed files" action because it wasn't working here and I couldn't work out why 🤔

.github/workflows/main.yml
```diff
@@ -104,3 +164,93 @@ jobs:
               "To fix, install tox and run `tox -e fix`."
             )
             sys.exit(1)
+  notebook-test:
```
Eric-Arellano, line 170: Nit to add some whitespace to make this more readable

scripts/ci/pr-execute-notebooks.py
```diff
@@ -29,12 +29,13 @@
...
 
 import os
+import sys
 import subprocess
 from pathlib import Path
 
 CONTENT_DIRECTORIES = ["docs", "learning"]
 
-all_changed_files = Path(".github/outputs/all_changed_files.txt").read_text().split("\n")
+all_changed_files = sys.argv[1].strip().split("\n")
```
Eric-Arellano: This is somewhat brittle to pass filenames through the CLI because it can fail if the arg list is too long. Maybe we have so few notebooks it is fine, but it would be better to still write to a file. If you don't like how we hardcode the path here, you could have the filename be sys.argv[1].

.github/workflows/main.yml
```diff
@@ -104,3 +164,93 @@ jobs:
...
+    name: API checks
+    needs: [decide-jobs]
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'Qiskit' && needs.decide-jobs.outputs.run-api-checks == 'true'
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
```
Eric-Arellano: I didn't realize this is so old. We should be using 22

.github/workflows/main.yml
Eric-Arellano: That would be helpful to comment

.github/workflows/main.yml
```diff
@@ -104,3 +166,94 @@ jobs:
...
+          EXTRA_DEPS_NOTEBOOKS = """\
+          docs/guides/visualize-circuits.ipynb
+          docs/guides/custom-backend.ipynb
+          docs/guides/transpiler-stages.ipynb
+          docs/guides/represent-quantum-computers.ipynb
+          docs/guides/common-parameters.ipynb
+          docs/guides/DAG-representation.ipynb
+          """.strip().split("\n")
+          import os
+          import path
```
Eric-Arellano:
```suggestion
          import pathlib
```

scripts/ci/pr-execute-notebooks.py
frankharkins: Sounds good, I forgot we had that problem with args being too long

.github/workflows/main.yml
```diff
@@ -10,9 +10,71 @@
...
 
-name: Lint
+name: Pull request
 on: [pull_request, merge_group]
 jobs:
+  decide-jobs:
+    name: Decide which jobs to run
+    runs-on: ubuntu-latest
+    # We don't run these tests in the merge queue
+    if: false && github.repository_owner == 'Qiskit' && github.event_name == 'pull_request'
```
Eric-Arellano:
```suggestion
    if: github.repository_owner == 'Qiskit' && github.event_name == 'pull_request'
```

arnaucasau (approved):

PR #4352, opened 2025-12-04
Update learning back button

### LINKED ISSUE #3883: Update the learning back button to use our new landing pages
Every `_toc.json` in Learning sets `parentUrl`, which sets up the back-left arrow in the left ToC. 

https://github.com/Qiskit/documentation/blob/90bdf5ebfe1ac97725f909e8b7ad4e5201f97e8a/learning/courses/quantum-chem-with-vqe/_toc.json#L2

Now that we're launching the new pages `/learning/courses` and `/learning/modules`, we should change these `parentUrl` values to be more specific rather than just `/learning`.

Be careful with merging this into `main` because we need to ensure that the new pages are actually live.

christopherporter1:
Fixes #3883


learning/courses/quantum-machine-learning/classical-ml-review.ipynb
Eric-Arellano: Is this intentional? The image file name is `unsup` but this talks about supervised learning.

learning/courses/quantum-machine-learning/introduction.ipynb
Eric-Arellano, line 53: Ditto if this is intentional

learning/courses/quantum-machine-learning/classical-ml-review.ipynb
christopherporter1: @Eric-Arellano Do you mean "qml-cr-background-sup-unsup.avif"? It is a comparison of supervised and unsupervised. So it has "sup" and "unsup" in the title.

learning/courses/quantum-machine-learning/introduction.ipynb
christopherporter1: @Eric-Arellano Ditto.

learning/courses/quantum-machine-learning/classical-ml-review.ipynb
```diff
@@ -20,15 +20,15 @@
...
     "\n",
     "![A diagrammatic representation of supervised and unsupervised learning.](/learning/images/courses/quantum-machine-learning/classical-ml-review/qml-cr-background-sup-unsup.avif)\n",
     "\n",
+    "The left image shows two categories of labeled data as in supervised learning. In this case, the categories are linearly separable. The right image shows clusters of data. In an unsupervised learning task, these data would not initially be labeled and the algorithm would study the distribution, perhaps looking for clusters. For the purposes of visualizing example clusters the algorithm might identify, the data points have now been labeled. A key difference between the two is that the supervised learning process starts with the data already labeled and the unsupervised process starts with unlabeled data, even if the data are labeled at the end.\n",
```
kaelynj:
```suggestion
    "The left image shows two categories of labeled data as in supervised learning. In this case, the categories are linearly separable. The right image shows clusters of data which, in an unsupervised learning task, would not initially be labeled and the algorithm would study the distribution. (For the purposes of visualizing example clusters the algorithm might identify, the data points have now been labeled.) A key difference between the two is that the supervised learning process starts with the data already labeled and the unsupervised process starts with unlabeled data, even if the data are labeled at the end.\n",
```

Small nit and doesn't need to happen to merge this. But I added this edit in to help a bit with readability.

Issue #2883, opened 2025-03-28
Identify why changed-files GitHub Actions failed


Eric-Arellano:
See https://github.com/Qiskit/documentation/pull/2881. We want to be using the latest commit from tj-actions/changed-files if possible. 

However, we had to revert ttps://github.com/Qiskit/documentation/pull/2881 because `separator` was not working and there was no separator between the changed files, at least here:

https://github.com/Qiskit/documentation/blob/b3df62bf43a481b73fba924772292cc53d8d2963/.github/workflows/notebook-test.yml#L50

We should open a bug report in https://github.com/tj-actions/changed-files to 

1. reduce our tech debt to allow us to use more modern versions, and
2. be a good open-source citizen.

It'd be helpful to find a minimum reproduction: create a dedicated GitHub repository that only calls the action in GH Actions & prints the output, such as using `print("""${{ steps.all-changed.outputs.all_changed_files }}""")` in a Python step.

Even better is to bisect where this broke in their repository. We know that `af2816c65436325c50621100d67f6e853cd1b0f1` works, but `2f7c5bfce28377bc069a65ba478de0a74aa0ca32` doesn't. So the issue must have happened somewhere between those two commits. You can use `git bisect` to do a binary search on their commit history to figure out where this broke. You'd clone their repository, then manually use `git bisect` to annotate the good and bad starting commits; try the new commits Git tells you with the minimum reproduction, and annotate if it was good or bad. (Time box this to ~20 minutes; it's still helpful to narrow the range even if we don't identify the exact commit.)

PR #4357, opened 2025-12-05
Clean up workflow files

frankharkins:
Most notably closes https://github.com/Qiskit/documentation/issues/2883 by replacing all uses of `@tj-actions/changed-file` with `gh pr diff` (we decided we didn't need to investigate if we replaced the action).

Also makes a couple of other tweaks:

* Moves some environment variables to be defined in YML: I think this makes the workflow easier to read because the comments get syntax highlighting, and it's clearer which variables to edit.
* Fix some of the regexes (😬)
* Simplifies the extended notebook test: We rarely use this, so when we do I think it's acceptable to run on all files rather than those chaned between `main`.


Eric-Arellano (dismissed):
Thanks!

PR #4368, opened 2025-12-08
Add dynamic circuit tutorial

haimeng-zhang:


docs/tutorials/dc-hex-ising.ipynb
miamico, line 1: Awesome! Here's some comments on this version:

- [ ] edit the "usage estimate" line. I think we don't say the specific processor anymore but something like "_Usage estimate: 7.5 minutes on a Heron r3 processor._"
- [ ] I would emphasize the fact that we are using dynamic circuits to implement interactions that go beyond the physical connectivity of the device. It took me a couple reading of the introduction to get it. Not sure whether also edit the title but for sure highlighting it in the introduction a bit more and earlier on
- [ ] Review the standard naming of the various Step .. sections, I think there's some discrepancies with existing tutorials
- [ ] The final plot is a little crowded. Let's use the following conventions: simulations (solid lines) experimental data (scatter plots with different markers and error bars). That hopefully cleans it up a bit

haimeng-zhang: Thank you for the review and comments, @miamico! I fixed the points raised above. Could you take another look?

haimeng-zhang:
> Awesome! something else that jumped out in my latest read:
> 
> * [ ]  Just clarifying question. Why does the unitary circuit implement the time evolution as Rx Rzz Rx?
>   Shouldn't there be only one Rx? Or are they each half angle?
> * [ ]  Let's add some more comments throughout dense parts of the code. For example, I got a bit lost reading the `gen_hex_dynamic` function.
> * [ ]  When you plot the DC circuit layout with `plot_circuit_layout` I'm confused at what I'm looking at. Are the black numbered qubits the ones that are being used in the circuit? maybe just add a sentence to clarify. If so, it could also be nice to distinguish between data and ancilla qubits in this plot by giving slightly different colors
> * [ ]  I'm a bit conflicted about the depth and duration plots. We show the depth plot to claim the dynamic circuits are less deep but then the total duration is longer. If we think that most errors comes from total duration of the circuit only then showing the depth savings is not useful (since the dc are anyway longer). If we want to argue that errors come from number of gates applied then we could say that the DC are longer but use much fewer gates and the errors arising during idling (or long measurement process) can be mitigated with DD more effectively than errors that arise due to many gate applications. Let's think about this point a little more and reflect our conclusions in the tutorial

Thank you for the comments and suggestions, @miamico!
Re. 1: There is one layer of Rx; see the equation for the Trotter layer at the beginning of the tutorial. The final `Rx` layer is not necessary but is added to be consistent with the implementation in [this](https://arxiv.org/pdf/2405.08810) paper. I added this reference to the introduction paragraph.
Re. 2: I added more comments to the code.
Re. 3: Yes, the black circles are the physical qubits being used. I revised the description to make it clearer.
Re. 4: I think it is informative to present both the depth and the duration plots to show the dynamic circuit implementation drastically reduces the SWAP gate overhead (shown in the depth plot), at the cost of introducing lengthy mid-circuit measurements and classical feedforward operations (shown in the duration plot). I reworded the text around the two plots to present this trade-off.

PR #4373, opened 2025-12-09
Delete stale PR previews

frankharkins:
Fixes #3432 by using a timestamp written to file rather than the last git commit time.

You can see the timestamp written to this PR's preview:

https://github.com/Qiskit/documentation/blob/6d0c3ff7ac6226d83093413f032e6d599f36a339/pr-4373/last_modified.txt#L1

### DIFF
--- a/scripts/pr-previews/cleanup.py
+++ b/scripts/pr-previews/cleanup.py
@@ -71,7 +71,7 @@ def get_active_pr_folders() -> set[str]:
 def is_stale(folder_name: str) -> bool:
     # All time measured in seconds, from the unix epoch
     current_timestamp = time.time()
-    last_modified_timestamp = get_timestamp(folder_name)
+    last_modified_timestamp = get_timestamp(Path(folder_name))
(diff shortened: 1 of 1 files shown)


scripts/pr-previews/cleanup.py
```diff
@@ -64,14 +68,26 @@ def get_active_pr_folders() -> set[str]:
...
 
+def is_stale(folder_name: str) -> bool:
+    # All time measured in seconds, from the unix epoch
+    current_timestamp = time.time()
+    last_modified_timestamp = get_timestamp(folder_name)
+    return (current_timestamp - last_modified_timestamp) > PR_EXPIRATION_TIME_SECONDS
 
 def delete_closed_pr_folders() -> None:
     active_pr_folders = get_active_pr_folders()
+    is_closed = lambda folder: folder not in active_pr_folders
```
Eric-Arellano: Lambdas are discouraged in Python, including due to bad typing support. Instead, I'd either:

1. Inline it again, or
2. In the for loop, save a variable:
```python
is_closed = folder.name not in active_pr_folders
if is_closed:
```

Issue #4366, opened 2025-12-08
PR preview only has changed files


Eric-Arellano:
Our PR previews take a very long time to deploy and the deployment often times out, as explained in https://github.com/Qiskit/documentation/issues/3412 and https://github.com/Qiskit/documentation/issues/3342.

One idea we had to improve this is to reduce the size of our deployments by only building changed files, rather than the entire site. The content team agreed that they are okay with this tradeoff.

You would achieve this by modifying this function:

https://github.com/Qiskit/documentation/blob/9f1a68f28d4ac1d9ead0a6f43675188e71a6f81c/scripts/pr-previews/builder.py#L117

Either only copy over what we need, or delete what we don't want after the copy. 

Note that you must always copy the `_toc.json` for the corresponding folder, when relevant. We also need to copy over images. Because you can in theory use images from anywhere, maybe do a simple check like "if any learning file, copy all of /learning". 

As before, we should never copy API files.

Finally, we need to update the homepage in iqp-channel-docs for the preview app to give more clear instructions that PR previews won't have every page loaded. Audit if the README needs updates.

Eric-Arellano:
Let's wait to close this until the documentation is updated: https://github.com/Qiskit/documentation/pull/4405#pullrequestreview-3627127672

PR #4405, opened 2025-12-12
Only include changed files in PR preview

frankharkins:
Closes https://github.com/Qiskit/documentation/issues/4366

### DIFF
--- a/docs/guides/install-qiskit.mdx
+++ b/docs/guides/install-qiskit.mdx
@@ -7,7 +7,7 @@ description: Install the Qiskit SDK and Qiskit Runtime on various operating syst
 {/* cspell:ignore manylinux miniconda */}
 
 <span id="qiskit-install"></span>
-# Install Qiskit?
+# Install Qiskit
 
 Whether you will work locally or in a cloud environment, the first step for all users is to install Qiskit.
 


scripts/pr-previews/builder.py
```diff
@@ -103,35 +109,74 @@ def save_output(root_dir: Path, dest: Path) -> None:
...
+        """For input to shutil.copytree. This function takes the directory path
+        (such as `docs/guides`) and a list of file and folder names (entries) in
+        that directory. It should output a list of entries to ignore.
+        """
+        ignores = []
+
+        # Ignore all paths with "/api/" in them
+        # We intentionally don't copy over API docs to speed up the build.
+        if "api" in contents:
+            ignores.append("api")
```
Eric-Arellano: We need to no matter what ignore `/api/`, specifically with the `main.yml` workflow. Also, I would check for `/api/` and not `api` because we have legit guides like `primitives-rest-api`

scripts/pr-previews/builder.py
```diff
@@ -54,7 +54,13 @@ def main() -> None:
         write_proof_of_concept(args.dest)
         return
 
-    with setup_dir() as dir:
+    try:
+        lines = Path(".github/outputs/changed-content-files.txt").read_text().split("\n")
+        changed_content_files = set(line for line in lines if line != "")
+    except FileNotFoundError:
+        changed_content_files: set[str] = set()
```
Eric-Arellano, line 65: Maybe log (with logger)?

scripts/pr-previews/builder.py
frankharkins: This does do that: `contents` is a list of directory entries (e.g. `["api", "index.mdx", ... ]`), so this has the effect of ignoring all paths matching `/api/` and would not match `primitives-rest-api`. I think the confusion is because the interface of `copytree`'s `ignore` is a bit unusual, I'll try to clarify in the comments.

scripts/pr-previews/builder.py
```diff
@@ -103,35 +110,76 @@ def save_output(root_dir: Path, dest: Path) -> None:
...
+            if entry == "_toc.json":
+                continue
+
+            # We also need to copy over the index files because the app doesn't
+            # build without them
+            if entry.startswith("index."):
+                continue
+
+            # Finally, continue traversing if the directory contains one of our
+            # changed files.
```
Eric-Arellano: This comment makes me think that we will include all of the files in a directory if at least one is changed, i.e. sibling files. I don't think that's what the code is doing though. Maybe this?


```suggestion
            # Finally, include files that were directly changed.
```

scripts/pr-previews/builder.py
```diff
@@ -103,35 +110,76 @@ def save_output(root_dir: Path, dest: Path) -> None:
...
         logger.info(f"Using tmpdir {root_dir}")
 
-        _copy_local_content(root_dir)
+        _copy_local_content(root_dir, changed_content_files)
         _extract_docker_files(root_dir)
         yield root_dir
 
+def _copy_local_content(root_dir: Path, changed_files: set[str]) -> None:
+
+    def ignore_contents(dir: str, contents: list[str]) -> list[str]:
```
Eric-Arellano: Often I'm a fan of inlined functions, but here I think it makes the code more confusing than necessary. You aren't using anything from the outer closure, and the helper function is quite verbose. I recommend moving it to a standalone function.

Eric-Arellano (commented):
This needs to use the new approach for changed files

scripts/pr-previews/builder.py
frankharkins: It does actually use the `changed_files` variable from the outer closure. Is there something else I can do to make it clearer?

Eric-Arellano: Ah. You could fix that though by adding a new `changed_files` arg to the function`. I'd still recommend splitting it out because both functions are now quite big

Eric-Arellano (approved):
Nice use of `partial`!

This looks good. Thanks for iterating on it. Please make sure to communicate this to content writers. Some places to consider doing that:

* Audit if our README needs tweaking
* Probably mention this restriction in the preview app's homepage (changed in closed source)
* Directly tell the core content writers in Slack

Issue #3836, opened 2025-09-02
Job limits section improvements


acastellane:
### URL to the relevant course

[https://quantum.cloud.ibm.com/docs/en/guides/job-limits#maximum-executions](https://quantum.cloud.ibm.com/docs/en/guides/job-limits#maximum-executions)

### Select all that apply

- [x] new content request
- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

### Describe the fix or the content request.

Based on Support Team experience, we do suggest the following mods.

1) An example of calculation could be provided. Using for example [https://quantum.cloud.ibm.com/docs/en/guides/custom-backend#transpile-against-custom-backends-1](https://quantum.cloud.ibm.com/docs/en/guides/custom-backend#transpile-against-custom-backends-1) to get the number of ops.

2) As per the documentation, number of total executions calculation is given as follows:
_For example, if you have a PUB with one circuit, observables with shape (1, 6), and parameters with shape (4, 1), this would render 4×6=24 circuits (or fewer, if some observables commute). If you requested 2,000 shots, then the total number of executions is 24×2,000=48,000_

We suggest to extend the explanation.
When the observables do commute the number of executions can be further reduced. Generally, from observables we calculate number of observables using qiskit.
`number_of_commuting_groups = observables.group_commuting(qubit_wise=True)`
    OR
`number_of_commuting_groups = observables.group_qubit_wise_commuting()`
Once number of commuting groups has been found, the total execution is given by:
Total executions = number_of_commuting_groups * number_of_parameter_set * number_of_shots
If Total Executions > 10 Million, that workload cannot be handled by the quantum software stack currently and usually will fail due to timeout or 1500 errors.

Further, If Error Mitigation Strategies/Twirling are employed, the number of actual executions during runtime would further increase from the Total Executions calculated.

### For new content requests - if the request is accepted, do you want to write the content?

I can help the team by providing enough information or by referring them to a specific Subject Matter Expert to write the material

christopherporter1:
@abbycross @beckykd Are you the best people to assign to this since it is from a guide? Thanks.

PR #4420, opened 2025-12-15
Update cost topics

### LINKED ISSUE #4229: Investigate and combine cost topics
Verify that all the cost topics are accurate and put them near each other (parking lot item)

Manage cost, 
view cost
workload usage
minimize job run time

beckykd:
Closes #4229

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -135,7 +135,7 @@ jobs:
           fi
 
           CHANGED_FILES=$(gh pr diff -R Qiskit/documentation $PR_NUMBER --name-only)
-          CHANGED_CONTENT_FILES=$(echo $CHANGED_FILES | grep -P $CONTENT_FILE_REGEX || true)
+          CHANGED_CONTENT_FILES=$(echo "${CHANGED_FILES[@]}" | grep -P $CONTENT_FILE_REGEX || true)
           if [ "$CHANGED_CONTENT_FILES" != '' ]
           then
             echo "ANY_CHANGED=true" >> "$GITHUB_OUTPUT"


jyu00:
Some comments about [this page](https://quantum.cloud.ibm.com/docs/en/guides/estimate-job-run-time): 

1. > Session usage is the time from when the first job starts until the session goes inactive, is closed, or when its last job completes, whichever happens last...

    I think we can simplify this to just say "Session usage is the wall-clock time the session is active. See [Session length](https://quantum.cloud.ibm.com/docs/en/guides/run-jobs-session#session-length) for more information about session status transition." 
    The reason for this suggestion is because the original sentence might make people think it includes time a session is deactivated  - e.g. if a session is deactivated at T1, then re-activated at T2, and the last job completes at T3, one can easily think its usage includes T2-T1 based on that sentence.  
    Note that if we make this change, we'll need to shift `(time spent by the QPU complex to process your job)` to the next setence.

2. Same for this sentence
    > Session mode: The reported usage is the wall-clock time from when the first job started executing in the session until the session terminates,

3. > Determine a workload's actual usage

    I think "Query a workload's actual usage" is a better title? "Determine" feels to me as something one needs to calculate. 

4. > <num executions> is the total number of circuits times the number of shots, where the circuits are those generated after PUB elements are broadcasted. 

    As suggested in [this issue](https://github.com/Qiskit/documentation/issues/3836), we could add "Estimator groups qubit-wise commuting observables, which reduces the number of executions".

docs/guides/estimate-job-run-time.ipynb
```diff
@@ -26,8 +26,8 @@
     "<span id=\"usage\"></span>\n",
     "Usage is a measurement of the amount of time the QPU is locked for your workload, and it is calculated differently, depending on which execution mode you're using.\n",
     "\n",
-    "* Session usage is the time from when the first job starts until the session goes inactive, is closed, or when its last job completes, whichever happens last. It includes both classical and quantum time (time spent by the QPU complex to process your job).\n",
-    "* Batch usage is the sum of quantum time of all jobs in the batch.\n",
+    "* Session usage is the wall-clock time the session is active. See [Session length](/docs/guides/run-jobs-session#session-length) for more information about session status transition.\n",
+    "* Batch usage is the sum of quantum time (time spent by the QPU complex to process your job) of all jobs in the batch.\n",
     "* Single job usage is the quantum time the job uses in processing.\n",
     "\n",
     "The reported usage is the time a QPU is locked for your workload. Failed or canceled jobs count toward your usage in certain circumstances - see the [Failed and canceled jobs](#failed-job) section for details.\n",
```
jyu00:
```suggestion
    "Note that failed or canceled jobs count toward your usage in certain circumstances - see the [Failed and canceled jobs](#failed-job) section for details.\n",
```

nit - we already said usage is the time QPU is locked just few sentences above, so I don't think it needs to be repeated again.

docs/guides/view-cost.mdx
```diff
@@ -7,29 +7,21 @@ description: View how much cost has been incurred by an instance
...
 
-## View job cost
-
-To determine how much has been billed to each job associated with an instance, from the [Instances page](https://quantum.cloud.ibm.com/instances), click the instance to open its details page. Next, on the left side, click Jobs.
-
-These are the columns relevant to cost:
-
-- **Usage**: Qiskit Runtime used by this job. This usage is the time counted by Qiskit Runtime to process a job, and is determined by the use of internal resources.
-- **Cost**: The total cost of running this job.
+To determine how much usage a specific workload incurred, open the [Workloads page](https://quantum.cloud.ibm.com/workloads) and view view the "Usage" column for the workload.
```
jyu00:
```suggestion
To determine how much usage a specific workload incurred, open the [Workloads page](https://quantum.cloud.ibm.com/workloads) and view the "Usage" column for the workload. 
```

jyu00 (dismissed):
Note that I didn't review the admin pages since I don't get that view

PR #4461, opened 2025-12-24
Fix CI for large PRs

Eric-Arellano:
https://github.com/cli/cli/issues/10712 resulted in our Runtime PR failing.

I debated using Git to get the information, but didn't because our `fetch-depth` is 1 and we would need to fetch more, which isn't great with our repository's size. I also don't want to use `tj-actions/changed-files` due to supply chain security and because it also under-the-hood Git fetches more commits.

So, we now use the API directly. The REST API has overly verbose information with the actual patch for each file, so we use GraphQL.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,7 +54,7 @@ jobs:
           # If changing these regexes, test them on https://regex101.com/ with the
           # following paths (you can paste the block in and uncomment it):
           # Add new tests if necessary
-          NOTEBOOK_TEST_REGEX: ((docs|learning)(?!\/api\/)\/.*\.ipynb|scripts\/nb-tester\/.*)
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -16,16 +16,41 @@ jobs:
   decide-jobs:
     name: Decide which jobs to run
     runs-on: ubuntu-latest
-    # We don't run these tests in the merge queue
-    if: github.repository_owner == 'Qiskit' && github.event_name == 'pull_request'
+    if: github.repository_owner == 'Qiskit'
     outputs:
+      all-changed-files: ${{ steps.changed-files.outputs.ALL_CHANGED_FILES }}
```
frankharkins, line 21: Good idea to share this across jobs

PR #4481, opened 2026-01-07
Expire PR previews based on PR activity

frankharkins:
The current PR expiration logic looks at the time PR files were last modified. This PR switches the check to look at the time the PR was last updated. This could be a better metric in cases where the PR is undergoing a long review.

### DIFF
--- a/scripts/pr-previews/cleanup.py
+++ b/scripts/pr-previews/cleanup.py
@@ -28,7 +28,6 @@ from utils import (
     setup_git_account,
     switch_branch,
     changed_files,
-    get_timestamp,
 )
 
 INITIAL_COMMIT = "499a5040585d02593cdd8237e19c9ee4a84ae126"
--- a/scripts/pr-previews/builder.py
+++ b/scripts/pr-previews/builder.py
@@ -24,7 +24,7 @@ from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Iterator
 
-from utils import configure_logging, run_subprocess, write_timestamp
+from utils import configure_logging, run_subprocess
 
 # You can change this to `iqp-channel-docs-preview-builder` when running locally, if
 # you're able to create a local copy of the builder image through the closed source repo.
@@ -67,7 +67,6 @@ def main() -> None:
     with setup_dir(changed_content_files) as dir:
         yarn_build(dir, args.basepath)
         save_output(dir, args.dest)
-    write_timestamp(args.dest)
 
 
 def write_proof_of_concept(dest: Path) -> None:
--- a/scripts/pr-previews/utils.py
+++ b/scripts/pr-previews/utils.py
@@ -14,26 +14,12 @@ from __future__ import annotations
 
 import logging
 import subprocess
-import time
 from contextlib import contextmanager
 from pathlib import Path
 from typing import Iterator
 
 
 logger = logging.getLogger(__name__)
-
-
-def get_timestamp(dest: Path) -> int | None:
-    try:
-        return int(Path(dest / "last_modified.txt").read_text())
-    except (FileNotFoundError, ValueError):
-        return None
-
-
-def write_timestamp(dest: Path) -> None:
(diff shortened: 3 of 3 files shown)


scripts/pr-previews/cleanup.py
```diff
@@ -17,6 +17,8 @@
 import shutil
 import time
 from pathlib import Path
+from datetime import datetime
+from typing_extensions import TypedDict
```
Eric-Arellano: It's in the modern std lib.

```suggestion
from typing import TypedDict
```

scripts/pr-previews/cleanup.py
```diff
@@ -61,36 +64,77 @@ def main() -> None:
         logger.info("Cleaned up closed PR previews")
 
 
-def get_active_pr_folders() -> set[str]:
+class PrFolders(TypedDict):
+    open: set[str]
```
Eric-Arellano: Maybe call this `all_open` to make clear that it is a superset of `stale`. Or, add docstring. I was at first confused reading delete_closed_pr_folders.

scripts/pr-previews/cleanup.py
```diff
@@ -82,7 +82,7 @@ def parse_updated_at(data):
         return data
 
     parsed = json.loads(api_response, object_hook=parse_updated_at)
-    all_pr_numbers = (number for number in parsed)
+    all_pr_numbers = (pr["number"] for pr in parsed)
```
frankharkins, line 84: Good thing you suggested tests – this wasn't caught by the type checker because `parsed` is `Unknown`

scripts/pr-previews/cleanup.py
```diff
@@ -61,36 +64,136 @@ def main() -> None:
...
+        if "updatedAt" in data:
+            iso_format = data["updatedAt"]
+            timestamp = datetime.fromisoformat(iso_format).timestamp()
+            data["updatedAt"] = timestamp
+        return data
+
+    parsed = json.loads(api_response, object_hook=parse_updated_at)
+    all_pr_numbers = (pr["number"] for pr in parsed)
+
+    # === Extract stale PRs ===
```
Eric-Arellano: Nit that I'm a big fan of using 3-line section headers because they are more visible, especially in the right sidebar in VSCode. Feel free to disagree


```suggestion
    # ========================================================
    # Extract stale PRs
    # ========================================================
```

scripts/pr-previews/cleanup.py
```diff
@@ -61,36 +64,136 @@ def main() -> None:
...
+def test_determine_stale_empty_list():
+    assert determine_stale("[]", 0, 1) == {"all_open": set(), "open_stale": set()}
+
+
+def test_determine_stale_active_pr():
+    api_response = """[{
+      "number": 1,
+      "updatedAt": "1970-01-01T01:00:00"
+    }]"""
+    assert determine_stale(api_response, 0, 1) == {
```
Eric-Arellano: Nit that it'd be helpful to clarify what the two variables are, by either:

1. Using kwargs
2. Or setting the variables `current_time` and `expiration_period`

scripts/pr-previews/cleanup.py
```diff
@@ -61,36 +64,136 @@ def main() -> None:
...
+
+def test_determine_stale_stale_pr():
+    # unix_epoch = "1970-01-01T01:00:00"
+    # (checked wih `datetime.fromtimestamp(0).isoformat()`)
+    api_response = """[{
+      "number": 1,
+      "updatedAt": "1970-01-01T01:00:00"
+    }]"""
+    current_time = 100
+    expiration_period = 10
```
Eric-Arellano: In relation to the prior test's comment, FYI these can be kwargs. The values are so simple that there isn't much value to having them be dedicated variables, other than the naming benefit which you can get with kwargs. Nbd either way

Eric-Arellano (dismissed):
Thanks for writing tests!

PR #4494, opened 2026-01-08
Incorporate new MidCircuitMeasure method

### LINKED ISSUE #4401: Incorporate new `MidCircuitMeasure` method in Repetition codes tutorial

https://quantum.cloud.ibm.com/docs/en/tutorials/repetition-codes

### Select all that apply

- [x] new content request
- [ ] typo
- [ ] code bug
- [x] out-of-date content
- [ ] broken link
- [ ] other

Use the [`MidCircuitMeasure` method](https://quantum.cloud.ibm.com/docs/en/guides/measure-qubits#midcircuitmeasure-method) for the mid-circuit measurements in the circuits.

abbycross:
Closes #4401.

I'm expecting it to fail at first because I need to figure out how to adjust the code choosing a backend without breaking the execute-nb check.


miamico (dismissed):
Seems good! Just a comment on the `least_busy` change. Won't the current change break the patching for the testing? I'm also not sure whether there is a way around though since we need to filter the backends by the instruction. What do you think @abbycross ?

miamico:
Checking the [docs](https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/qiskit-runtime-service#least_busy), it looks like `least_busy` can use filters as `backends` so it should be directly portable

docs/tutorials/repetition-codes.ipynb
```diff
@@ -92,18 +94,15 @@
...
    "source": [
-    "num_qubits = 5\n",
-    "backend = service.least_busy(\n",
-    "    operational=True,\n",
-    "    simulator=False,\n",
-    "    min_num_qubits=num_qubits,\n",
-    "    dynamic_circuits=True,\n",
-    ")"
+    "# Choose a backend that supports `measure_2`.\n",
+    "service.backends(filters=lambda b: \"measure_2\" in b.supported_instructions)\n",
```
abbycross:
```suggestion
    "# Choose the least busy backend that supports `measure_2`.\n",
    "service.least_busy(filters=lambda b: \"measure_2\" in b.supported_instructions,\n",
        "    operational=True,\n",
        "    simulator=False,\n",
        "    dynamic_circuits=True,\n",
        ")"
```
@miamico something like this?

miamico: That is what I was thinking about! If it runs then it should be good as is

abbycross:
> @abbycross check out this [draft tutorial](https://github.com/haimeng-zhang/documentation/blob/7f811211ed777b302d7117a7cee1b922a233e8c8/docs/tutorials/dc_hex_ising.ipynb) which uses measure_2 on `ibm_boston`

Thank you so much @miamico - I will try this out!

kaelynj (approved):

PR #4496, opened 2026-01-08
Generate Qiskit 2.3

Eric-Arellano:


docs/api/qiskit/dev/_toc.json
```diff
@@ -2154,6 +2186,10 @@
         {
           "title": "v2",
           "children": [
+            {
+              "title": "v2.4",
+              "url": "/docs/api/qiskit/release-notes/2.4"
+            },
```
arnaucasau: We shouldn't have this entry here. I wonder why the API generation script added it 🤔. Did you just run `npm run gen-api` for it?

arnaucasau: Fixed in https://github.com/Qiskit/documentation/pull/4496/commits/1f30c76d95ef3078b46c50d854272ce52f0089aa, but we should create an issue to investigate it once we know what command generated the issue

Issue #4336, opened 2025-12-01
Deal with invalid external links


Eric-Arellano:
https://github.com/Qiskit/documentation/blob/69cb850fe7c5758d3c640c11fd2a2882a77dfb99/scripts/js/lib/links/ignores.ts#L158-L177

The infra team can help with updating API docs once we know what should happen to them, such as removing the link entirely vs using a new link. We should only use a new link if it's actually a valid replacement for the original content.

beckykd:
- [ ] https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_configuration_schema.json 
- Found a link in the API docs.  I opened an issue to update it. 
- The only other occurrence is being updated in https://github.com/Qiskit/documentation/pull/4313
- **Will we still get broken link errors from old versions of the API docs?**  

- [ ] https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_properties_schema.json
- The only links are in API docs.  I opened an issue to update it. 

- [ ] Link to https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_status_schema.json will be fixed in  https://github.com/Qiskit/documentation/pull/4313

beckykd:
The decision is to leave the broken links in old versions of API docs; adding the links to a new section in the ignore.ts file.  Ping appropriate infra team members with instructions to update the current API docs and they will perform their magic.

kaelynj:
@beckykd are we safe to close this since we'll leave broken links in old versions of API docs and it looks like what you've listed out is done?

PR #4506, opened 2026-01-09
Fix outdated links

beckykd:
Related to #4336

### DIFF
--- a/scripts/config/external-redirects.ts
+++ b/scripts/config/external-redirects.ts
@@ -37,8 +37,6 @@ export const externalRedirects = {
     "https://pubs.aip.org/aip/jcp/article/148/16/164104/196134/Bravyi-Kitaev-Superfast-simulation-of-electronic",
   "https://aip.scitation.org/doi/10.1063/1.5141835":
     "https://pubs.aip.org/aip/jcp/article/152/12/124107/953761/Quantum-orbital-optimized-unitary-coupled-cluster",
-  "https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_properties_schema.json":
-    "https://quantum.cloud.ibm.com/docs/api/qiskit-runtime-rest/tags/backends#tags__backends__operations__get_backend_configuration",
   ...Object.fromEntries(
     QISKIT_REMOVED_PAGES.map((link) => [
       `https://docs.quantum.ibm.com/api/qiskit/${link}`,
--- a/docs/guides/get-qpu-information.ipynb
+++ b/docs/guides/get-qpu-information.ipynb
@@ -186,7 +186,7 @@
    "id": "37fbdbc1-9f24-491b-8352-e56d767a5acd",
    "metadata": {},
    "source": [
-    "Use these keyword arguments to filter by any attribute in backend configuration ([JSON schema](https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_configuration_schema.json)) or status ([JSON schema](https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_status_schema.json)). A similar method is [`QiskitRuntimeService.least_busy()`](../api/qiskit-ibm-runtime/qiskit-runtime-service#least_busy), which takes the same filters as `backends()` but returns the backend that matches the filters and has the least number of jobs pending in the queue:"
+    "Use these keyword arguments to filter by any attribute in backend configuration or status. A similar method is [`QiskitRuntimeService.least_busy()`](../api/qiskit-ibm-runtime/qiskit-runtime-service#least_busy), which takes the same filters as `backends()` but returns the backend that matches the filters and has the least number of jobs pending in the queue:"
    ]
   },
   {


scripts/js/lib/links/ignores.ts
```diff
@@ -170,16 +170,10 @@ const ALWAYS_IGNORED_URLS__SHOULD_FIX: string[] = [
   "https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/default_pulse_configuration_schema.json",
 
   // These links are from old IQP and do not work any more
-  "https://api.quantum-computing.ibm.com/runtime",
   "https://auth.quantum-computing.ibm.com/api",
   "https://auth.quantum.ibm.com/api",
   "https://quantum-computing.cloud.ibm.com",
-  "https://us-east.quantum-computing.cloud.ibm.com/",
```
frankharkins, line 175:
```suggestion
```

Might fix the lint error

scripts/js/lib/links/ignores.ts
```diff
@@ -168,10 +168,14 @@ const ALWAYS_IGNORED_URLS__SHOULD_FIX: string[] = [
...
   // These links are from old IQP and do not work any more
-  "https://api.quantum-computing.ibm.com/runtime",
   "https://auth.quantum-computing.ibm.com/api",
   "https://auth.quantum.ibm.com/api",
   "https://quantum-computing.cloud.ibm.com",
+
+  // The pulse guide was removed in https://github.com/Qiskit/documentation/pull/4495 and should be fixed
+  // in the Qiskit latest, dev, and some historical versions. In the meantime, the app will redirect the
+  // page to pulse-migration so the links are not broken.
+  "/docs/guides/pulse",
```
arnaucasau: This part is not necessary anymore :). It was removed by https://github.com/Qiskit/documentation/pull/4608 and fixed by Abby

docs/api/qiskit-ibm-runtime/release-notes.mdx
```diff
@@ -654,7 +654,7 @@ The qiskit-ibm-runtime `v0.41.0` release focuses on the removal of IBM Quantum P
 *   The `channel_strategy` parameter in `QiskitRuntimeService` has been removed. To continue using Q-CTRL in your workflow, please explore the following options:
 
     > *   If your organization has an existing IBM Quantum Premium Plan instance: migrate to the Q-CTRL Performance Management Function, found in the [Qiskit Functions Catalog](https://quantum.ibm.com/functions).
-    > *   To continue using Qiskit Runtime with IBM Cloud: migrate to Q-CTRL Fire Opal, the same performance management product accessible directly through Q-CTRL. You can [connect your IBM Cloud API key and Qiskit Runtime CRN](https://docs.q-ctrl.com/fire-opal/discover/hardware-providers/how-to-authenticate-with-ibm-credentials) to Fire Opal. ([1966](https://github.com/Qiskit/qiskit-ibm-runtime/pull/1966))
```
arnaucasau, line 657: This link needs to be fixed in the runtime repo (`main` branch and `stable/0.45`):

- https://github.com/Qiskit/qiskit-ibm-runtime/blob/ce9f35518e35f6ca18ed85471b993ee6326b1f9c/release-notes/0.32.0.rst?plain=1#L47
- https://github.com/Qiskit/qiskit-ibm-runtime/blob/ce9f35518e35f6ca18ed85471b993ee6326b1f9c/release-notes/0.30.0.rst?plain=1#L18

docs/api/qiskit/0.46/qiskit.algorithms.optimizers.ISRES.mdx
```diff
@@ -13,7 +13,7 @@ python_api_name: qiskit.algorithms.optimizers.ISRES
...
+  Improved Stochastic Ranking Evolution Strategy (ISRES) is an algorithm for non-linearly constrained global optimization. It has heuristics to escape local optima, even though convergence to a global optima is not guaranteed. The evolution strategy is based on a combination of a mutation rule and differential variation. The fitness ranking is simply via the objective function for problems without nonlinear constraints. When nonlinear constraints are included, the [stochastic ranking proposed by Runarsson and Yao](https://www.semanticscholar.org/paper/Stochastic-ranking-for-constrained-evolutionary-Runarsson-Yao/827e3d01e52438e29aac1f6c4c8a2ed972a87c75) is employed. This method supports arbitrary nonlinear inequality and equality constraints, in addition to the bound constraints.
```
arnaucasau, line 16: This link needs to be added as a redirect to avoid being reverted when we generate the API docs in the future. Given that this is a historical API file, we don't need to fix it on the qiskit repo.

See Frank's comment for an example: https://github.com/Qiskit/documentation/pull/4506#pullrequestreview-3651167205

frankharkins:
> @frankharkins what should we do about links that are broken in historical API documents but don't exist in more recent documents? For example, qaiskit-ibm-runtime/0.40/qiskit-runtime-service links to https://auth.quantum.ibm.com/api. That API no longer exists and the link was just deleted from the later versions.

I think if a link has no replacement, it's OK for it to be broken in a historical version. These are archives after all and it might even be helpful for a user to know that the API isn't available any more.

Maybe just update the comment to explain we're not going to fix them.

docs/api/qiskit-ibm-runtime/release-notes.mdx
beckykd: > This link needs to be fixed in the runtime repo

Jessie said to just ignore it, since it's in a really old release note.  I'll add to the Ignore file.

arnaucasau: Unfortunately, we can't ignore the issue because it's in the release notes. Our script regenerates the whole release note file with each version, so we would revert the change in the future if it's not fixed.

We can only ignore issues with historical versions if it's in a page different than the release notes :(

arnaucasau: In my opinion we should fix it on the runtime repo and remove the redirect in `external-redirects.ts` given that it won't be necessary anymore

scripts/config/external-redirects.ts
```diff
@@ -37,6 +37,14 @@ export const externalRedirects = {
     "https://pubs.aip.org/aip/jcp/article/148/16/164104/196134/Bravyi-Kitaev-Superfast-simulation-of-electronic",
   "https://aip.scitation.org/doi/10.1063/1.5141835":
     "https://pubs.aip.org/aip/jcp/article/152/12/124107/953761/Quantum-orbital-optimized-unitary-coupled-cluster",
+  "https://www.cs.bham.ac.uk/~xin/papers/published_tec_sep00_constraint.pdf":
+    "https://www.semanticscholar.org/paper/Stochastic-ranking-for-constrained-evolutionary-Runarsson-Yao/827e3d01e52438e29aac1f6c4c8a2ed972a87c75",
+  "https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_properties_schema.json":
```
arnaucasau: This link shouldn't be added as a redirect. It was already fixed in the latest version and we have ignores for historicals

scripts/config/external-redirects.ts
```diff
@@ -37,6 +37,14 @@ export const externalRedirects = {
...
+    "https://quantum.cloud.ibm.com/docs/api/qiskit-runtime-rest/tags/backends#tags__backends__operations__get_backend_configuration",
+  "https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_configuration_schema.json":
```
arnaucasau: Ditto this link

scripts/config/external-redirects.ts
```diff
@@ -37,6 +37,14 @@ export const externalRedirects = {
...
+    "https://quantum.cloud.ibm.com/docs/api/qiskit-runtime-rest/tags/backends#tags__backends__operations__get_backend_configuration",
+  "https://docs.q-ctrl.com/fire-opal/discover/hardware-providers/how-to-authenticate-with-ibm-credentials":
+    "https://docs.q-ctrl.com/fire-opal/discover/hardware-providers/how-to-migrate-from-performance-management-on-ibm-cloud",
```
arnaucasau: We can remove it if we do https://github.com/Qiskit/documentation/pull/4506/changes#r2758072445

scripts/js/lib/links/ignores.ts
```diff
@@ -163,16 +163,16 @@ const ALWAYS_IGNORED_URLS__EXPECTED = [
 
 // These external URLs cause actual 404s and should probably be fixed.
 const ALWAYS_IGNORED_URLS__SHOULD_FIX: string[] = [
-  // These schemas are not available any more. They might have moved, but we're not sure where.
+  // These schemas are not available any more. References to these in the new API docs have been fixed,
+  // but they will never be fixed in the old API docs.  We decided to leave these links as-is for archival purposes.
   "https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_status_schema.json",
   "https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/default_pulse_configuration_schema.json",
+  "https://docs.q-ctrl.com/fire-opal/discover/hardware-providers/how-to-authenticate-with-ibm-credentials",
```
arnaucasau: Do we need this link to be ignored now? I think we can remove it if it only appears in the qiskit-ibm-runtime release notes

scripts/config/external-redirects.ts
frankharkins: I think, if the link points to the same resource (just at a different URL), then it's helpful to update the historical docs.

arnaucasau: Agreed! We originally decided to ignore historical versions because not only the link changed (See https://github.com/Qiskit/documentation/pull/4573/changes). If we are okay with keeping the old text, but the new link, we could indeed add the redirect and regenerate every runtime historical version to pull it

docs/api/qiskit-ibm-runtime/release-notes.mdx
beckykd: OK

beckykd: > In my opinion we should fix it on the runtime repo and remove the redirect in `external-redirects.ts` given that it won't be necessary anymore

This is done now - in Main and Stable .45.  They are in the merge queue.

arnaucasau: Nice, thank you Becky! I also fixed it in our artifact stored in Box 👍

docs/api/qiskit/0.46/qiskit.algorithms.optimizers.ISRES.mdx
```diff
@@ -13,7 +13,7 @@ python_api_name: qiskit.algorithms.optimizers.ISRES
 
   Improved Stochastic Ranking Evolution Strategy optimizer.
 
-  Improved Stochastic Ranking Evolution Strategy (ISRES) is an algorithm for non-linearly constrained global optimization. It has heuristics to escape local optima, even though convergence to a global optima is not guaranteed. The evolution strategy is based on a combination of a mutation rule and differential variation. The fitness ranking is simply via the objective function for problems without nonlinear constraints. When nonlinear constraints are included, the [stochastic ranking proposed by Runarsson and Yao](https://www.cs.bham.ac.uk/~xin/papers/published_tec_sep00_constraint.pdf) is employed. This method supports arbitrary nonlinear inequality and equality constraints, in addition to the bound constraints.
```
arnaucasau, line 16: This link would normally need an external redirect, but given that it's only in one historical version and it's a simple change, I replaced it in the Box artifact. So, this change is okay now and it doesn't need anything else 👍

scripts/config/external-redirects.ts
arnaucasau: Should the link point to [get_backend_properties](https://quantum.cloud.ibm.com/docs/en/api/qiskit-runtime-rest/tags/backends#tags__backends__operations__get_backend_properties) instead?

We also have `https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_configuration_schema.json`. Should we fix both of them then?

I was looking at the original [PR](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2569/changes) were both links were fixed and I realized that the properties one points to [get_backend_configuration](https://quantum.cloud.ibm.com/docs/api/qiskit-runtime-rest/tags/backends#tags__backends__operations__get_backend_configuration) instead of [get_backend_properties](https://quantum.cloud.ibm.com/docs/en/api/qiskit-runtime-rest/tags/backends#tags__backends__operations__get_backend_properties) too, so maybe we should fix it again.

I recommend removing the redirect for now to unblock the rest of the changes of this PR. We can address it in a follow-up. Adding this redirect implies having to regenerate 20 runtime versions

scripts/js/lib/links/ignores.ts
```diff
@@ -163,20 +163,18 @@ const ALWAYS_IGNORED_URLS__EXPECTED = [
 
 // These external URLs cause actual 404s and should probably be fixed.
 const ALWAYS_IGNORED_URLS__SHOULD_FIX: string[] = [
-  // These schemas are not available any more. They might have moved, but we're not sure where.
+  // These schemas are not available any more. References to these in the new API docs have been fixed,
+  // but they will never be fixed in the old API docs.  We decided to leave these links as-is for archival purposes.
   "https://github.com/Qiskit/ibm-quantum-schemas/blob/main/schemas/backend_status_schema.json",
```
arnaucasau, line 168: The comment is not true for the first schema. I checked and it only appears in a guide:

https://github.com/Qiskit/documentation/blob/c1ff226bec11b262ca80b45f4ec26a84c8bc442a/docs/guides/get-qpu-information.ipynb#L189

Thoughts on fix it too?

scripts/config/external-redirects.ts
frankharkins: Sorry! I was looking at the link above it (sematicscholar). For the schema, I'm not sure, maybe it's best to leave it broken? If I was trying to debug something using an old version of runtime, it might be confusing if I was linked to a different schema.

However, I agree we should do whatever we can to unblock this PR so it doesn't end up in review purgatory 😂

beckykd:
@arnaucasau on this comment: "Should the link point to [get_backend_properties](https://quantum.cloud.ibm.com/docs/en/api/qiskit-runtime-rest/tags/backends#tags__backends__operations__get_backend_properties) instead?"  
Jessie said to use that address, so let's keep it as-is.  I doubt those links get enough traffic to worry about updating them again :)

PR #4515, opened 2026-01-12
Clarify the need for `lift`

beckykd:
This change was made in response to a support issue.

fyi @LenaPer


docs/guides/classical-feedforward-and-control-flow.ipynb
```diff
@@ -208,12 +208,12 @@
...
     "\n",
     "\n",
-    "In the parity calculation, the first element of the constructed expression involves lifting the Python object `mr[0]`to a [`Value`](/docs/api/qiskit/circuit_classical#value)  node. This is not necessary for `mr[1]` and the possible following classical register, as they are inputs to `expr.bit_xor`, and any necessary lifting is done automatically in these cases. Such expressions can be built up in loops and other constructs."
+    "In the parity calculation, the first element of the constructed expression involves lifting the Python object `mr[0]` to a [`Value`](/docs/api/qiskit/circuit_classical#value) node (`lift` is used to turn arbitrary objects into classical expressions). This is not necessary for `mr[1]` and the possible following classical register, as they are inputs to `expr.bit_xor`, and any necessary lifting is done automatically in these cases. Such expressions can be built up in loops and other constructs. "
```
abbycross:
```suggestion
    "In the parity calculation, the first element of the constructed expression involves lifting the Python object `mr[0]` to a [`Value`](/docs/api/qiskit/circuit_classical#value) node (`lift` is used to turn arbitrary objects into classical expressions). This is not necessary for `mr[1]` and the possible following classical register, as they are inputs to `expr.bit_xor`, and any necessary lifting is done automatically in these cases. Such expressions can be built up in loops and other constructs."
```

docs/guides/classical-feedforward-and-control-flow.ipynb
```diff
@@ -257,7 +257,7 @@
     "        continue\n",
     "\n",
     "    # build XOR-parity expression\n",
-    "    parity = expr.lift(mr[left_bits[0]])  # lift the first bit to Value\n",
+    "    parity = expr.lift(mr[left_bits[0]])  # lift the first bit to Value so it will be treated like a Boolean.\n",
```
abbycross: Based on IBM Style, which says "In API programming information, use an initial lowercase b when
boolean refers to a primitive return type," would this use of Boolean need to be lower case?

beckykd:
```suggestion
    "    parity = expr.lift(mr[left_bits[0]])  # lift the first bit to Value so it will be treated like a boolean.\n",
```

beckykd: Seems like it!

PR #4521, opened 2026-01-13
SQD tutorial: add section on passing initial occupancies

kevinsung:
Adds a section showing how to pass initial occupancies, which allows SQD to run even if there are no valid configurations. It may also improve convergence.


nathanearnestnoble:
Thank you for adding this component @kevinsung , I think it will help some issues which a few users have raised.   

- Naive question: is there a scenario in which a user would not have the ability to guess the initial occupancies?  If not, perhaps it can just be the default workflow rather than an "advanced section"
- Perhaps just adding a sentence which briefly explains the initial occupancy you selected?

nathanearnestnoble:
Gave a read through of this and I like the updates!   Thanks @kevinsung for implementing.

PR #4535, opened 2026-01-14
put commuting observables in the same PUB

beckykd:
Related to a question received by support.

FYI @sumit-kale

### DIFF
--- a/docs/guides/primitive-input-output.ipynb
+++ b/docs/guides/primitive-input-output.ipynb
@@ -63,7 +63,7 @@
     "- A single `QuantumCircuit`, which may contain one or more [`Parameter`](/docs/api/qiskit/qiskit.circuit.Parameter) objects\n",
     "- A list of one or more observables, which specify the expectation values to estimate, arranged into an array (for example, a single observable represented as a 0-d array, a list of observables as a 1-d array, and so on). The data can be in any one of the `ObservablesArrayLike` format such as `Pauli`, `SparsePauliOp`, `PauliList`, or `str`.\n",
     "   <Admonition type=\"note\">\n",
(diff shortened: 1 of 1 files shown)


docs/guides/primitive-input-output.ipynb
```diff
@@ -62,6 +62,9 @@
     "For the Estimator primitive, the format of the PUB should contain at most four values:\n",
     "- A single `QuantumCircuit`, which may contain one or more [`Parameter`](/docs/api/qiskit/qiskit.circuit.Parameter) objects\n",
     "- A list of one or more observables, which specify the expectation values to estimate, arranged into an array (for example, a single observable represented as a 0-d array, a list of observables as a 1-d array, and so on). The data can be in any one of the `ObservablesArrayLike` format such as `Pauli`, `SparsePauliOp`, `PauliList`, or `str`.\n",
+    "   <Admonition type=\"note\">\n",
+    "   If you have two commuting observables in different PUBs but with the same circuit, they will not be estimated using the same measurement. Each PUB represents a different basis for measurement, therefore, separate measurements are required for each PUB. To ensure that commuting observables are estimated using the same measurement, they must be grouped within the same PUB.\n",
```
abbycross:
```suggestion
    "   If you have two commuting observables in different PUBs but with the same circuit, they will not be estimated using the same measurement. Each PUB represents a different basis for measurement, and therefore, separate measurements are required for each PUB. To ensure that commuting observables are estimated using the same measurement, they must be grouped within the same PUB.\n",
```

abbycross: I'll just commit this so you don't have to ask for a re-review :D

PR #4543, opened 2026-01-15
Fix CI

frankharkins:
Quite a few steps in our main PR checks were not running (see [before](https://github.com/Qiskit/documentation/actions/runs/21042653301/job/60508385246) vs [after](https://github.com/Qiskit/documentation/actions/runs/21042820693/job/60509012936)).


.github/workflows/main.yml
frankharkins, line 136: This line was not working because this job did not rely on `decide-jobs`

arnaucasau: Good catch!

PR #4545, opened 2026-01-15
Remove plotly figures from notebook

frankharkins:
Plotly's interactive figures contain a _lot_ of code: Some of it JSON (see this PR's diff), some of it HTML and JS (see https://github.com/Qiskit/documentation/pull/4539#discussion_r2695493368). Since we can't display the figures on our website, I've commented out the two `fig.show` lines. Users can uncomment them to render the figures in their local notebook client.


docs/guides/visualize-circuit-timing.ipynb
```diff
@@ -2649,7 +439,7 @@
     ")\n",
     "\n",
     "# Display the figure\n",
-    "fig.show(renderer=\"notebook\")\n",
+    "# fig.show(renderer=\"notebook\")\n",
```
Eric-Arellano: I'm curious what the content team thinks, but it might be worth explaining what's going on. Like "Uncomment the below line to display the figure"

abbycross: Yes, I think it would help to be more specific in the way Eric suggests.

PR #4548, opened 2026-01-16
Replace link extraction code with Rust

frankharkins:
Improving check time is important to our team (see https://github.com/Qiskit/documentation/issues/2854). A large portion of the time is spent in the internal link checker.

We discovered that parsing markdown and extracting links is a bottleneck in the link checker, so we rewrote that specific function in Rust (https://github.com/qiskit-community/mdx-link-extract/). From eyeballing the logs, this brings roughly a 10x speedup.


scripts/js/lib/links/extractLinks.test.ts
frankharkins, line 1: These tests are now in https://github.com/qiskit-community/mdx-link-extract/

docs/guides/quick-start.ipynb
```diff
@@ -62,7 +62,7 @@
     "\n",
     "- To install Python, first check the \"Programming Language\" section on the [Qiskit PyPI project page](https://pypi.org/project/qiskit/) to determine which Python versions are supported by the most recent release. For download instructions, see the [Python Beginners Guide.](https://wiki.python.org/moin/BeginnersGuide/Download)\n",
     "\n",
-    "<Admonition type = \"note\">\n",
+    "<Admonition type=\"note\">\n",
```
frankharkins, line 65: The only discrepancy I found between the JS and Rust MDX libraries is that the Rust library doesn't allow spaces here. This only occurs in a handful of places in our repo.

PR #4606, opened 2026-01-29
Explain stretch output

### LINKED ISSUE #4520: Add info on returned stretch values

https://quantum.cloud.ibm.com/docs/en/guides/stretch#view-stretch-values-in-qiskit-runtime

### Select all that apply

- [x] new content request
- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

The current example only shows how to request resolved stretch values. It doesn't say anything about how to interpret the returned values, and they are not exactly intuitive, e.g. 

```
'stretch_values': [{'name': 'x_0',
   'value': 5,
   'remainder': 1,
   'expanded_values': [[803, 6], [810, 5]]}]
```

I can help the team by providing enough information or by referring them to a specific Subject Matter Expert to write the material

beckykd:
closes #4520 

preview is here: https://qiskit.github.io/documentation/pr-4606/docs/guides/stretch#understand-the-metadata-output

### DIFF
--- a/docs/guides/stretch.mdx
+++ b/docs/guides/stretch.mdx
@@ -176,9 +176,9 @@ circuit.measure_all()
 
 The values returned for duration depend on the goal value and the calculated remainder.  For example, these are the durations returned for foo:
 
-- foo value + remainder (8+2 = 10) 
-- foo value * 3 (8 x 3 = 24)
-- foo value * 2 (8 x 2 = 16)
+- `foo value` + `remainder` (8+2 = 10) 
+- `foo value` * 3 (8 x 3 = 24)
+- `foo value` * 2 (8 x 2 = 16)
 
 You can use a visualization to help understand and verify the timing. 
 


docs/guides/stretch.mdx
```diff
@@ -100,21 +100,51 @@ qc.barrier()
...
+# Visualize the timing.
+# Use the sliders at the bottom, the controls at the top, and the legend on the side 
+# of the output to customize the view.
+draw_circuit_schedule_timing(vulqan_job.result()[0].metadata['compilation']['scheduler_timing']['timing'])
 ```
 <Admonition type="note">
 Although the total circuit time is returned in the "compilation" metadata, this is NOT the time used for billing (quantum time).
 </Admonition>
 
+## Understand the metadata output
```
jyu00: This should be a sub-header to `View stretch values in Qiskit Runtime`, since it's part of that topic

docs/guides/stretch.mdx
```diff
@@ -100,21 +100,51 @@ qc.barrier()
...
+
+The `stretch_values` metadata returns the following information:
+
+- **Name:** The name of the applied stretch
+- **Value:** The requested goal value
+- **Remainder:** The remainder from resolving the stretch
```
jyu00:
```suggestion
- **Remainder:** The remainder from resolving the stretch, which is added to the first delay that uses the stretch.
```

I think it's worth pointing out what we do with the remainder, given we had a `Note` on this earlier. It also makes it more obvious why the first value is `6` and the second is `5` in the example below.

docs/guides/stretch.mdx
```diff
@@ -100,21 +100,51 @@ qc.barrier()
...
+``` python
+'stretch_values': 
+ [{'name': 'x_0',
+   'value': 5,
+   'remainder': 1,
+   'expanded_values': [[803, 6], [810, 5]]}
+   ]
+```   
+
+You can also use a visualization (requires plotly) to help understand the timing. Use the sliders at the bottom, the controls at the top (hover over the image to reveal these), and the legend on the side of the output to customize the view.
```
jyu00: Does this figure actually have anything to do with `stretch`? If not, then we should just add a `Next steps` section that points to the [circuit schedule visualization](https://quantum.cloud.ibm.com/docs/guides/visualize-circuit-timing#qiskit-runtime-support) page, and/or mention that page there (but not the figure). 

If this figure has something to do with `stretch`, then we need to explain because it's not very obvious...

beckykd: It does - I'll ask Kavya to help me explain it.

docs/guides/stretch.mdx
```diff
@@ -100,21 +100,51 @@ qc.barrier()
 The concrete value of a stretch duration is resolved at compile time, after the circuit is scheduled. When running a Sampler job in Qiskit Runtime, you can view the resolved stretch values in the job result metadata.  Support for `stretch` in Qiskit Runtime is currently experimental, so you need to first set an experimental option to enable its retrieval, then access the data directly from the metadata as follows:
 
 ```python
-# enable stretch value retrieval
+# Enable stretch value retrieval.
 sampler.options.experimental = {
     "execution": {
         "stretch_values": True,
```
KavyaVarma, line 107:
```suggestion
        "stretch_values": True,
        "scheduler_timing": True,
```
This experimental option is required for the circuit visualization below to work.

docs/guides/stretch.mdx
KavyaVarma: The visualization is useful to verify that the starts and ends of delay regions match the returned `Stretch` values (if you hover over the right elements and calculate the duration of white spaces in the plot). 
Although - I'm not sure the image here is conveying this. Maybe something that specifically corresponds to the stretch values above would be helpful?

docs/guides/stretch.mdx
```diff
@@ -97,24 +98,66 @@ qc.barrier()
...
+``` python
+ [{'name': 'bar',
+  'value': 29,
+  'remainder': 1,
+  'expanded_values': [[1365, 30], [1404, 29]]},
+ {'name': 'foo',
+  'value': 8,
+  'remainder': 2,
+  'expanded_values': [[1365, 10], [1384, 24], [1417, 16]]}
+  ]
```
jyu00, line 174: I don't understand the values returned for `foo`. I interpret `Value` as the _ideal_ stretch value. For `bar`, the first one is `30` because it gets the remainder of 1 added. But why are the second and third durations for `foo` 24 and 16? We should explain that to avoid confusion.

docs/guides/stretch.mdx
```diff
@@ -97,24 +98,66 @@ qc.barrier()
...
+```   
+
+You can also use a visualization (which requires plotly) to help understand and verify the timing. 
+
+```python
+draw_circuit_schedule_timing(vulqan_job.result()[0].metadata['compilation']['scheduler_timing']['timing'])
+```
+
+In the following image, based on the example output, `bar` corresponds to the stretches on Qubit 3. The first stretch delay that uses `bar` starts at the end of the `init_play` (1394). The stretch duration is 30, so that delay ends when the `x` gate starts (1394+30 = 1424).
```
jyu00: Is this supposed to say `1364` and not `1394`? In the image, the `init_play` for qubit 3 ends soon after 1360, before 1380. Plus the example says the first foo starts at 1365.

docs/guides/stretch.mdx
```diff
@@ -97,24 +98,66 @@ qc.barrier()
...
+
+![The output from the `draw_circuit_schedule_timing` command is shown.](/docs/images/guides/stretch/stretch-visualization.avif)
```
jyu00, line 191: I think it'd help to add annotation to the image to indicate where the stretch value applies, and/or add the code that specified those values. Code is preferable since that's what users have.

Image

docs/guides/stretch.mdx
```diff
@@ -97,24 +98,66 @@ qc.barrier()
...
+  'remainder': 1,
+  'expanded_values': [[1365, 30], [1404, 29]]},
+ {'name': 'foo',
+  'value': 8,
+  'remainder': 2,
+  'expanded_values': [[1365, 10], [1384, 24], [1417, 16]]}
+  ]
+```   
+
+You can also use a visualization (which requires plotly) to help understand and verify the timing.
```
jyu00: We should still link to the visualization page, which has more information on how to interpret timing data

docs/guides/stretch.mdx
```diff
@@ -97,24 +98,94 @@ qc.barrier()
...
+  ]
+```   
+
+You can use a visualization to help understand and verify the timing. 
+
+```python
+draw_circuit_schedule_timing(job.result()[0].metadata['compilation']['scheduler_timing']['timing'])
+```
+
+In the following image, based on the example output, `bar` corresponds to the stretches on Qubit 3. The first stretch delay that uses `bar` starts at the end of the `init_play` (1394). The stretch duration is 30, so that delay ends when the `x` gate starts (1394+30 = 1424).
```
KavyaVarma: This explanation needs an update after the change to the stretch value output
```suggestion
In the following image, based on the example output, `bar` corresponds to the stretches on Qubit 3. The first stretch delay that uses `bar` starts at the end of the `init_play` (1365). The stretch duration is 30, so that delay ends when the `x` gate starts (1365+30 = 1395).
```

jyu00 (dismissed):
This is now very clear, thanks for making the updates!

docs/guides/stretch.mdx
```diff
@@ -97,24 +98,102 @@ qc.barrier()
...
+  'remainder': 2,
+  'expanded_values': [[1365, 10], [1384, 24], [1417, 16]]}
+  ]
+```   
+
+The values returned for duration depend on the goal value and the calculated remainder.  For example, these are the durations returned for foo:
+
+- foo value + remainder (8+2 = 10) 
+- foo value * 3 (8 x 3 = 24)
+- foo value * 2 (8 x 2 = 16)
```
abbycross: wondering about how foo will be auto-translated 🤣 maybe we need to include it in code format? `foo`

abbycross (approved):
👍

PR #4622, opened 2026-02-02
Update preview hash to support the `LeftColoredBar` component

arnaucasau:
This PR updates the preview image hash to add support to use the `LeftColoredBar` component. It also updates the MDX guide to add an example of how to use it.

This is a screenshot with an example of how different components would render inside it:

Screenshot 2026-02-02 at 13 10 01

### DIFF
--- a/mdx-guide.md
+++ b/mdx-guide.md
@@ -377,7 +377,7 @@ Here's an example of what this might look like in your notebook source.
 
 ### LeftColoredBar
 
-The `LeftColoredBar` components adds a colored bar that wraps all the content inside it. Currently we support two colors (`blue` and `purple`) that can be picked by using the `color` property:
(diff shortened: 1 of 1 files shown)


mdx-guide.md
```diff
@@ -374,3 +374,13 @@ Here's an example of what this might look like in your notebook source.
  ]
 }
 ```
+
+### LeftColoredBar
+
+The `LeftColoredBar` components adds a colored bar that wraps all the content inside it. Currently we support two colors (`blue` and `purple`) that can be picked by using the `color` property:
```
arnaucasau:
```suggestion
The `LeftColoredBar` component adds a colored bar that wraps all the content inside it. Currently we support two colors (`blue` and `purple`) that can be picked by using the `color` property:
```

PR #4628, opened 2026-02-03
Migrate generated markdown from `details` to `Accordion`

frankharkins:
We're switching to use the Carbon's `Accordion` component. This PR updates markdown-generating code that uses `details`. I've split this into separate commits to make review easier:

* 72a108082c015fbd797c6f5f7bc93e232b365224 requires the closest review: This updates the logic that writes the generated markdown
* c790c3867ffc70e64253a84017b60d01f5c7917c used find-and-replace to switch the "package versions" in each notebook to accordions. I executed a notebook to make sure this was equivalent to the generated markdown
* 2a60194b31b08b02452ee9d525ada52d0f589537 is the result of `npm run regen-apis -- -p qiskit`, but I reset all the image changes to avoid noise.

### DIFF
--- a/scripts/js/lib/api/releaseNotes.test.ts
+++ b/scripts/js/lib/api/releaseNotes.test.ts
@@ -40,8 +40,7 @@ test.describe("generateReleaseNotesIndex", () => {
     });
     const result = generateReleaseNotesIndex(pkg);
     expect(result).toEqual(
-      `
----
+      `---
 title: My Quantum Project release notes
 description: New features, bug fixes, and other changes in previous versions of My Quantum Project.
 ---
@@ -57,12 +56,11 @@ New features, bug fixes, and other changes in previous versions of My Quantum Pr
 - [v2.0](./2.0)
 </AccordionItem>
 
-<AccordionItem  title="v1">
+<AccordionItem title="v1">
 - [v1.3](./1.3)
 - [v1.2](./1.2)
 </AccordionItem>
-</Accordion>
-`.trim(),
+</Accordion>`,
     );
   });
 });


scripts/js/lib/api/releaseNotes.ts
```diff
@@ -169,18 +179,11 @@ New features, bug fixes, and other changes in previous versions of ${pkg.title}.
...
-    pkg.releaseNotesConfig.separatePagesVersions,
-  );
-  [...grouped.entries()].forEach(([majorVersion, versionList], idx) => {
-    markdown +=
-      renderVersionGroup(majorVersion, versionList, {
-        isLatestVersion: idx === 0,
-      }) + "\n\n";
-  });
-  return markdown.trim();
+`.trim();
```
arnaucasau: Nit: You can remove the `trim()` by moving the backtick next to the Accordion closing tag

```
</Accordion>`;
```

Ditto other places

scripts/js/lib/api/releaseNotes.ts
```diff
@@ -287,11 +290,10 @@ function renderVersionGroup(
   const items = versions
     .map((version) => `- [v${version}](./${version})`)
     .join("\n");
-  const openAttr = kwargs.isLatestVersion ? " open" : "";
+  const maybeOpenProp = kwargs.isLatestVersion ? "open" : "";
```
arnaucasau: If you recover the space before `open` you will remove it from Accordions without that prop. Right now they have a double space:

E.g.
```
<AccordionItem  title="v1">
```

This doesn't affect to the result, so as you prefer

docs/guides/add-job-tags.ipynb
```diff
@@ -30,8 +30,8 @@
     "  a new cell before or after this one.\n",
     "*/}\n",
     "\n",
-    "<details>\n",
-    "<summary><b>Package versions</b></summary>\n",
+    "<Accordion>\n",
+    "<AccordionItem title=\"Package versions\">\n",
```
arnaucasau, line 34: We don't want the title to be bold anymore?

frankharkins: I don't think it's necessary now we have better affordance of the accordion, but should probably defer to @Qiskit/docs-content-team

abbycross: I'm ok with it not bold!

PR #4635, opened 2026-02-05
corrected function output types to list[tuple[str, list[int], float]]:

### LINKED ISSUE #4632: Function's return anotation doesn't match the actual return type

https://quantum.cloud.ibm.com/docs/en/tutorials/quantum-approximate-optimization-algorithm
Step 1: Map classical inputs to a quantum problem
Optimization problem → Hamiltonian

### Select all that apply

- [ ] new content request
- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [x] other

The function is defined as 
> def build_max_cut_paulis(graph: rx.PyGraph) -> list[tuple[str, float]]:

but it actually returns list[tuple[str,list[int],float]

(No answer)

christopherporter1:
Fixes #4632

### DIFF
--- a/docs/tutorials/quantum-approximate-optimization-algorithm.ipynb
+++ b/docs/tutorials/quantum-approximate-optimization-algorithm.ipynb
@@ -295,7 +295,9 @@
     }
    ],
    "source": [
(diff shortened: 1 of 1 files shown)


abbycross (approved):
Thank you!

PR #4641, opened 2026-02-06
Fix typo of classical and quantum bounds in chsh-inequality.ipynb

ayush268:
Quantum bounds are 2*sqrt(2) and classical bounds are 2.


jlapeyre (approved):
Thanks @ayush268 !

The existing text is clearly a typo. The bounds are reversed (in the text only).

I have no write access. @miamico or someone with permission, can you please approve?

PR #4474, opened 2026-01-05
Corrected random range in chsh

christopherporter1:
fixes #4438

Issue #4438, opened 2025-12-18
CHSH game example error: random.randint(0, 2) generates invalid inputs


neeco1991:
### URL to the relevant course

https://quantum.cloud.ibm.com/learning/en/courses/basics-of-quantum-information/entanglement-in-action/qiskit-implementation#qiskit-implementation-1

### Select all that apply

- [ ] new content request
- [ ] typo
- [x] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

### Describe the fix or the content request.

I encountered an issue in the CHSH game code example where the input bits $x$ and $y$ are generated using an incorrect range. The code currently uses:

```python
x, y = random.randint(0, 2), random.randint(0, 2)
```

But random.randint(0, 2) ranges from 0 to 2, which doesn't make sense for a bit. 
in fact, if you run the whole code, you get:

```
Fraction of games won: 0.617
```

Which is wrong for my understanding. I believe the correct code sould be:

```python
x, y = random.randint(0, 1), random.randint(0, 1)
```

With this fix, the code correctly outputs a win fraction of ~0.85.

### For new content requests - if the request is accepted, do you want to write the content?

I will write (or already have written) a draft of the proposed content

PR #4645, opened 2026-02-09
Fix confusion between stdlib random and np.random

### LINKED ISSUE #4643: `random` is imported from an inconsistent source

In [Entanglement in action](https://quantum.cloud.ibm.com/learning/en/courses/basics-of-quantum-information/entanglement-in-action/qiskit-implementation), the `chsh_game` function uses `random.randint(0, 1)`, but that's only make sense for the `random` module that comes from python. But actually `random` is imported from NumPy in that notebook, and this function always returns 0 for this input, since the high limit is not inclusive. _@kalde in [#4438](https://github.com/Qiskit/documentation/issues/4438#issuecomment-3768709508)_ correctly points out this problem, but somehow that issue gets closed.

### Select all that apply

- [ ] new content request
- [ ] typo
- [x] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

Maybe changing `from numpy import pi, random` to
```py
from numpy import pi
import random
```
is a better solution since in my opinion, one should use NumPy's random as `np.random`, rather than polluting the global naming space.

(No answer)

jlapeyre:
The existing code that generates

https://quantum.cloud.ibm.com/learning/en/courses/basics-of-quantum-information/entanglement-in-action/qiskit-implementation

imports both `pi` and `random` from numpy.  The notebook uses nothing else from numpy. These two objects are used only in a scalar context. In some cases, using numpy can degrade performance. In qiskit there has been an effort, reflected in PRs, to avoid using numpy habitually, when not necessary. This is done even in non-performance critical code.

This PR changes the code to import `pi` from `math`. And it imports `random` from the stdlib.  When producing single samples, standard library `random.randint` is about 5 times faster than `numpy.random.randint`.

Furthermore, the existing code calls numpy's `random.randint(0, 1)`, which always returns zero. This was mistakenly introduced. Since this PR uses `randint` from the standard library, the range is 0 to 1, inclusive.  So it may return either 0 or 1. My best understanding is that this is the original intent of the code.

The following issues and PRs are on exactly the same code touched here
* #4438
* #4439
* #4474 
* #4643 

Closes #4643

### DIFF
--- a/learning/courses/basics-of-quantum-information/entanglement-in-action/qiskit-implementation.ipynb
+++ b/learning/courses/basics-of-quantum-information/entanglement-in-action/qiskit-implementation.ipynb
@@ -56,7 +56,8 @@
     "from qiskit.visualization import plot_histogram, array_to_latex\n",
     "from qiskit.result import marginal_distribution\n",
     "from qiskit.circuit.library import UGate\n",
-    "from math import pi, random"
+    "from math import pi\n",
+    "import random"
    ]
   },
   {


learning/courses/basics-of-quantum-information/entanglement-in-action/qiskit-implementation.ipynb
```diff
@@ -56,7 +56,7 @@
     "from qiskit.visualization import plot_histogram, array_to_latex\n",
     "from qiskit.result import marginal_distribution\n",
     "from qiskit.circuit.library import UGate\n",
-    "from numpy import pi, random"
+    "from math import pi, random"
```
abbycross: @jlapeyre the json was invalid - this fix now passes the linter but pls let me know if it doesn't match what is needed here

abbycross: I've now put `import random` on its own line in the notebook (vs. directly in the json), and it healed the lint problem - that looks more like what you wanted to accomplish

jlapeyre: @abbycross Thankyou. This is correct. Next time, I'll run `jq`.

PR #4648, opened 2026-02-09
Add admonition about verifying rep delay value

abbycross:
Add clarification that rep_delay can vary across QPUs, and to always verify it, as it has impact on execution time/capacity consumption.

### DIFF
--- a/docs/guides/repetition-rate-execution.ipynb
+++ b/docs/guides/repetition-rate-execution.ipynb
@@ -49,7 +49,7 @@
    "source": [
     "When a circuit is executed on an IBM&reg; quantum processing unit (QPU), an implicit reset is typically inserted at the beginning of the circuit to ensure the qubits are initialized to zero. This is controlled by the `init_qubits` flag, set as a [primitive execution option](/docs/api/qiskit-ibm-runtime/options-execution-options-v2).\n",
     "\n",
-    "However, the reset process is not perfect and can create state preparation errors. To alleviate the error, the QPU also inserts repetition delay time (or `rep_delay`) between circuits. Each backend has a different default `rep_delay`, but it's usually set to balance reset fidelity against total execution time. Run `backend.default_rep_delay` to find the default `rep_delay` for a specific QPU.\n",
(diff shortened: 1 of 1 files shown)


docs/guides/repetition-rate-execution.ipynb
```diff
@@ -49,22 +49,29 @@
    "source": [
     "When a circuit is executed on an IBM&reg; quantum processing unit (QPU), an implicit reset is typically inserted at the beginning of the circuit to ensure the qubits are initialized to zero. This is controlled by the `init_qubits` flag, set as a [primitive execution option](/docs/api/qiskit-ibm-runtime/options-execution-options-v2).\n",
     "\n",
-    "However, the reset process is not perfect, leading to state preparation errors. To alleviate the error, the system also inserts repetition delay time (or `rep_delay`) between circuits. Each backend has a different default `rep_delay`, but it's usually longer than T1 to allow the environment to reset the qubits. The default `rep_delay` can be queried by running `backend.default_rep_delay`.\n",
+    "However, the reset process is not perfect and can create state preparation errors. To alleviate the error, the QPU also inserts repetition delay time (or `rep_delay`) between circuits. Each backend has a different default `rep_delay`, but it's usually set to balance reset fidelity against total execution time. Run `backend.default_rep_delay` to find the default `rep_delay` for a specific QPU.\n",
```
jlapeyre: > the reset process is not perfect and can create state preparation errors

"and" doesn't mean that the first clause implies or contributes to the second. Colloquially, we understand it this way. So it is kind of ok. But I'm trying to think of something better.... maybe something like

"However, imperfections in the reset process introduce state-preparation errors." ?

jlapeyre (commented):
As usual, this is a big improvement. The language is simpler, more direct, has a uniform mode of address. I have one comment. Not a blocker, but worth commenting on.

docs/guides/repetition-rate-execution.ipynb
abbycross: How about `However, imperfections in the reset process can introduce state-preparation errors`?

jlapeyre: Yes, that's better.

abbycross:
```suggestion
    "However, imperfections in the reset process can introduce state-preparation errors. To alleviate the error, the QPU also inserts repetition delay time (or `rep_delay`) between circuits. Each backend has a different default `rep_delay`, but it's usually set to balance reset fidelity against total execution time. Run `backend.default_rep_delay` to find the default `rep_delay` for a specific QPU.\n",
```

PR #4690, opened 2026-02-18
Second error detection tutorial

MeltemTolunay:
Error detection tutorial for generalized Clifford circuits


MeltemTolunay:
Review/help about the following would be highly appreciated:

- The metric class and the gamma parameter are not very clearly documented in the Paulice repository and the paper. I'll need help to understand and explain how this is utilized in the check picking algorithm.
- Similarly, get_post_selection_method is not very clearly documented. I need guidance on how to submit this to hardware or use this function to post-select hardware results in the tutorial.
- An internal discussion on windowed, genetic, and windowed-genetic search would be helpful for me to explain the inner workings of the check picking algorithm better.
- A discussion on how exactly the noise model is used during check picking would also be beneficial.

PR #4713, opened 2026-02-24
Support ibm-quantum-schemas docs in qiskit-ibm-runtime

arnaucasau:
This PR tweaks our script to ensure we can generate the qiskit-ibm-runtime docs correctly after https://github.com/Qiskit/qiskit-ibm-runtime/pull/2623. It also adds a new TOC order for runtime.

This PR builds the new artifact on top of the current dev docs in https://github.com/Qiskit/documentation/pull/4713/commits/1c4b453a23667426382dd3c948f575ed62a374b8 to have an example for the review


scripts/js/lib/links/ignores.ts
```diff
@@ -290,6 +290,148 @@ function _runtimeObjectsInv(): FilesToIgnores {
         `/docs/api/qiskit-ibm-runtime/${vers}guides/executor-basic#the-outputs-of-the-executor`,
         `/docs/api/qiskit-ibm-runtime/${vers}guides/index`,
         `/docs/api/qiskit-ibm-runtime/${vers}guides/index#guides`,
+        // qiskit-ibm-runtime includes the docs from the ibm-quantum-schemas package
```
frankharkins: Are these temporary? If not, is there a way to generalize the rule? E.g. can we ignore all `ibm-quantum-schemas` pages?

arnaucasau: We should be able to fix them somehow, so they should be temporary. They are failing because our anchors include the word field on it.

Expected: `#circuit_b64`
Real one: `#field-circuit_b64`

Actually, let me see if we can fix that 🤔. If it's not easy, we can create an issue and fix them as a follow-up imo

arnaucasau: https://github.com/Qiskit/documentation/pull/4713/commits/59a5fd09b2269bca7fe86d6ef1299046f6600b36 should fix it

scripts/js/lib/api/generateApiComponents.ts
```diff
@@ -251,14 +258,15 @@ function prepareAttributeOrPropertyProps(
...
-  // between an attribute and property. However, we preserve the full string if it's `abstract property`.
+  // between an attribute and property. However, we preserve the full string if it's `abstract property` or `field`.
   //
   // Meanwhile, we preserve the non-modifier `em.property` elements to be processed below.
-  const rawModifiers = $child
-    .find("em.property")
-    .filter((i, el) => $(el).text().includes("property"));
+  const rawModifiers = $child.find("em.property").filter((i, el) => {
+    const text = $(el).text();
+    return text.includes("property") || text.includes("field");
```
frankharkins, line 268: Question: Does it make sense to be more specific here? Not sure what kind of assumptions we want to make.

E.g. would something like this be better?

```suggestion
    return ['property', 'abstract property', 'field'].includes(text);
```

frankharkins (approved):
Awesome work fixing the links so quickly

scripts/js/lib/api/generateApiComponents.ts
arnaucasau: Not sure. I think I'm leaning towards leaving it as-is because technically `property` is included in `abstract property`, so the second entry won't be used.

In my opinion we should find a sprint to try to refactor and simplify code. After having removed very old historical versions I'm sure we can also simplify the pipeline and make improvements like the one you said 🤔

frankharkins (approved):

scripts/js/lib/api/generateApiComponents.ts
frankharkins: > In my opinion we should find a sprint to try to refactor and simplify code

That would be good! Let's leave it as-is for now to keep the momentum

PR #4718, opened 2026-02-24
Ajc/fix admonitions

### LINKED ISSUE #4696: Add a new line after admonition tag and before text
To make sure there is adequate space between the box and the text, look for instances of
`<Admonition>My text</Admonition> ` and change to 
```
<Admonition>
  My text
</Admonition>
```

abbycross:
Closes #4696. Add carriage return both after <Admonition> and before </Admonition> so that the right spacing renders in the published page.


arnaucasau (approved):
Thanks!

PR #4725, opened 2026-02-26
Add two missing runtime entries for the new release

arnaucasau:
It adds two new TOC entries that I found by generating the latest docs from https://github.com/Qiskit/qiskit-ibm-runtime/pull/2623

### DIFF
--- a/scripts/js/lib/api/TocGrouping.ts
+++ b/scripts/js/lib/api/TocGrouping.ts
@@ -253,8 +253,18 @@ export const QISKIT_RUNTIME_GROUPING: TocGrouping = {
       kind: "module",
     },
     {
+      moduleId: "qiskit_ibm_runtime.noise_learner_v3",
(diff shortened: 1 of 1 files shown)


scripts/js/lib/api/TocGrouping.ts
```diff
@@ -252,11 +252,21 @@ export const QISKIT_RUNTIME_GROUPING: TocGrouping = {
       title: "qiskit_ibm_runtime.noise_learner",
       kind: "module",
     },
+    {
```
arnaucasau, line 255: @frankharkins I don't love that we have to add everything manual now, so I'll investigate how to make it automatic and only set the order (it's not possible at the moment AFAIK).

frankharkins: Thanks, let's make an issue and we can prioritize it today

EDIT: https://github.com/Qiskit/documentation/issues/4729

PR #4732, opened 2026-02-27
Move composer to getting started section

kaelynj:
I've gotten feedback stating that the composer is a helpful "getting started" tool and I would like us to prioritize its placement for better visibility to new users. This PR moves the `IBM Quantum Composer` documentation to the getting started section.


abbycross (approved):
👍

PR #4772, opened 2026-03-09
Remove outdated new pills

frankharkins:
Removes new pills that were added > 3 weeks ago


Eric-Arellano:
> I thought it would fail whenever we build a PR and prevent us from merging something new.

We could do that if you'd like! But we figured it would be really annoying because it would break unrelated PRs randomly. But we can do whatever process you'd like. If we made the change you suggest, we would definitely have less risk of not realizing that we need to remove new pills.

PR #4779, opened 2026-03-09
Add script to remind of Qiskit deprecations

frankharkins:
The version selector requires some manual intervention to move releases from "Stable" to "Deprecated" after a new major version has been released for over six months. This PR adds an action that will check PyPI to see if the version should be deprecated and makes a new issue if one does not already exist.

See https://github.com/Qiskit/documentation/issues/4778 for an example issue

### DIFF
--- a/scripts/js/commands/api/checkQiskitDeprecation.ts
+++ b/scripts/js/commands/api/checkQiskitDeprecation.ts
@@ -10,7 +10,6 @@
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
 
-
 // The version selector has the following rule for grouping Qiskit versions:
 //
 // > When a new major series like v2.0 comes out, the latest minor version


.github/workflows/check-qiskit-deprecation.yml
```diff
@@ -0,0 +1,32 @@
...
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+name: Check if Qiskit deprecation is needed in version selector
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 1 * * 1,2,3,4,5" # At 01:00 on Monday-Friday.
```
arnaucasau: Given how infrequent this is to find a version to deprecate, what do you think about running it maybe twice or three times a week instead? Even if we deprecate the version, it will still need some time until it gets deployed, so maybe a day or two of delay isn't a big deal?

My motivation was to save CI resources, but If we think it's super important to act on it asap, then every day is fine probably. (feel free to leave it as-is)

scripts/js/commands/api/checkQiskitDeprecation.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (!parsed) {
+      console.log(`Ignoring unparsable version '${s}'`);
+      return [];
+    }
+    return [parsed];
+  });
+
+  const versionsByMajor = groupBy(versions, (v) => v.major);
+  const [latestMajorNumber, previousMajorNumber] = Object.keys(versionsByMajor)
+    .map((n) => Number(n))
```
arnaucasau, line 53: Nit:
```suggestion
    .map(Number)
```

scripts/js/commands/api/checkQiskitDeprecation.ts
```diff
@@ -0,0 +1,93 @@
...
+  30 * // ~ 1 month
+  6;
+
+/** Extract required information from a list of version strings */
+export function processVersions(rawVersions: string[]): {
+  latestMajorNumber: number;
+  previousMajorNumber: number;
+  firstReleaseOfLatestMajor: string;
+} {
+  const versions = rawVersions.flatMap((s) => {
```
Eric-Arellano, line 42: Nice use of flatmap for nullish values

scripts/js/commands/api/checkQiskitDeprecation.ts
```diff
@@ -0,0 +1,93 @@
+// This code is a Qiskit project.
```
Eric-Arellano, line 1: It would probably be worth adding a comment to the top of this file with an explanation of the business logic

scripts/js/commands/api/checkQiskitDeprecation.ts
```diff
@@ -0,0 +1,93 @@
...
+      console.log(`Ignoring unparsable version '${s}'`);
+      return [];
+    }
+    return [parsed];
+  });
+
+  const versionsByMajor = groupBy(versions, (v) => v.major);
+  const [latestMajorNumber, previousMajorNumber] = Object.keys(versionsByMajor)
+    .map((n) => Number(n))
+    .sort()
```
arnaucasau: I think we need to use a different compare function for the sort. The default one won't sort well double digit versions. In the past we used the `compare` function from `semver`.

Example: if we have the versions `['0', '1', '2', '10']`, the current function will tell you that the latest version is `2` and the previous one is `10`

arnaucasau: Although if you are sorting just numbers, you could do:

```ts
.sort((a, b) => a - b)
```

🤔

scripts/js/commands/api/checkQiskitDeprecation.ts
frankharkins: While I think it's fine in this case, I have been tripped up a bit by point-free in JS recently so would prefer to leave as-is. You can sometimes get unexpected behaviour because `.map` tries to pass more arguments to the function (see [example](https://stackoverflow.com/a/262511)).

scripts/js/commands/api/checkQiskitDeprecation.ts
frankharkins: Great catch, can't believe I fell for this one again. Thanks!

.github/workflows/check-qiskit-deprecation.yml
frankharkins: Sure, daily is definitely unnecessary. Is once a week ok?

scripts/js/commands/api/checkQiskitDeprecation.ts
arnaucasau: Sure, I think it's fine as it is right now too 👍

PR #4803, opened 2026-03-12
Update backends for nb-tester patch and change its CLI behavior

kaelynj:
We're changing the behavior of the notebook tester to allow both a config path and patch to be specified, but for a patch that is explicitly specified in the command line to take priority over the config path. This PR also updates the available open plan backends.

This makes it easier to run something like `tox -- /path/to/notebook.ipynb --patch {some-patch}` in a local environment when you only want to run a handful of notebooks.

### DIFF
--- a/scripts/nb-tester/test/test_get_notebook_jobs.py
+++ b/scripts/nb-tester/test/test_get_notebook_jobs.py
@@ -30,7 +30,6 @@ QiskitRuntimeService.least_busy = patched_least_busy
 
 
 def test_no_config_file():
-    print("Testing with no config file:")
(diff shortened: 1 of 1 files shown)


scripts/nb-tester/test/test_get_notebook_jobs.py
```diff
@@ -28,25 +30,17 @@ def patched_least_busy(self, *args, **kwargs):
 
 
 def test_no_config_file():
+    print("Testing with no config file:")
```
frankharkins:
```suggestion
```

Issue #4757, opened 2026-03-05
Observability cron job is failing


Eric-Arellano:
https://github.com/Qiskit/documentation/actions/runs/22697952695/job/65808474034

PR #4827, opened 2026-03-16
Fix observability sync workflow

arnaucasau:
Part of https://github.com/Qiskit/documentation/issues/4757#event-23612009934

This PR loads the `GITHUB_TOKEN` in the observability sync workflow to fix the cron job is failing (https://github.com/Qiskit/documentation/actions/runs/22697952695/job/65808474034)

We were getting the following error:

```
 * [new branch]      actions/cron-22697952695 -> actions/cron-22697952695
gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:
  env:
    GH_TOKEN: ${{ github.token }}
```

To verify that it works I ran the workflow from this branch. See https://github.com/Qiskit/documentation/actions/runs/23155373200 & https://github.com/Qiskit/documentation/pull/4828

### DIFF
--- a/qiskit_bot.yaml
+++ b/qiskit_bot.yaml
@@ -153,7 +153,6 @@ notifications:
     - "`@abbycross`"
     - "`@beckykd`"
   "docs/guides/observability":
-    - "@GantaRoja"
     - "`@abbycross`"
     - "`@beckykd`"
   "docs/guides/responsibilities":


.github/workflows/observability-sync.yml
```diff
@@ -41,7 +41,8 @@ jobs:
 
       - name: Make pull request
         if: steps.changed-docs-files.outputs.CHANGED_FILES
-
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
arnaucasau, line 45: We are doing the same for the dev docs:

https://github.com/Qiskit/documentation/blob/0d0ba5604875ac135c8981ba2f3dd30faeef8b27/.github/workflows/dev-docs-sync.yml#L54-L55

Issue #4804, opened 2026-03-12
Generate a guide with audit events for Quantum System API


Eric-Arellano:
Extend https://github.com/Qiskit/documentation/blob/main/scripts/js/commands/generateObservabilityPage.ts to generate the actions for the Quantum System REST API. 

* The OpenAPI file's URL is TBD
* Should be a new URL: `/docs/guides/observability-quantum-system`
* The intro needs some tweaking, like using the correct product name & removing the Location region information

PR #4829, opened 2026-03-16
Add support to create the QS observability file

arnaucasau:
Closes https://github.com/Qiskit/documentation/issues/4804

This PR adds support to create the Quantum System observability page in our script. It also renames the old `observability.mdx` page to `observability-runtime-rest.mdx`

### DIFF
--- a/docs/guides/observability-quantum-system.mdx
+++ b/docs/guides/observability-quantum-system.mdx
@@ -5,7 +5,7 @@ description: Learn about activity tracking events for the IBM Quantum System ser
 
 # Activity tracking events for the IBM Quantum System service
 
-IBM Cloud&reg; services, such as the IBM Quantum System service, generate activity tracking events.
+IBM Cloud&reg; services, such as the IBM&reg; Quantum System service, generate activity tracking events.
 
 Activity tracking events report on activities that change the state of a service in IBM Cloud. You can use the events to investigate abnormal activity and critical actions and to comply with regulatory audit requirements.
 
--- a/scripts/js/commands/generateObservabilityPage.ts
+++ b/scripts/js/commands/generateObservabilityPage.ts
@@ -40,6 +40,10 @@ async function main() {
   );
 }
 
+function maybeAddPkgTitleRegisteredIcon(pkgTitle: string) {
+  return pkgTitle.replace("IBM", "IBM&reg;");
+}
+
 async function writeObservabilityFile(
   pkgTitle: OpenapiPkgTitle,
   destPath: string,
@@ -70,7 +74,7 @@ description: Learn about activity tracking events for ${pkgTitle}.
 
 # Activity tracking events for ${pkgTitle}
 
-IBM Cloud&reg; services, such as ${pkgTitle}, generate activity tracking events.
+IBM Cloud&reg; services, such as ${maybeAddPkgTitleRegisteredIcon(pkgTitle)}, generate activity tracking events.
 
 Activity tracking events report on activities that change the state of a service in IBM Cloud. You can use the events to investigate abnormal activity and critical actions and to comply with regulatory audit requirements.
 


docs/guides/observability-quantum-system.mdx
arnaucasau, line 1: Same file as the Runtime one, but with the correct package title, the QS endpoints, and without the region section.

Is the rest of the content correct?

Eric-Arellano: The only thing I'm not sure about is if it should be "Quantum System" or something like "the Quantum System API". I'll ask Raul

raulotaolea: I’m not 100% sure about this, but I think using _Quantum System_ makes more sense. My reasoning is that, although the quantum system is a combination of hardware and software - and calling it a service might not seem entirely clear at first - we are actually commercializing it under a _Hardware as a Service_ (HaaS) model, so it does make sense to consider it a service. And, as is typical for a service, it provides an API to use it. So referring to it as a _Quantum System_ when we mean the quantum system as a whole - or the service it represents - makes sense. And using _Quantum System API_ to refer to the API it provides is also clear and appropriate.

abbycross: For the title and first instances, let's follow the pattern in the rest of docs and use the full title with the word service, allowing for proper trademarking. And per Christine Vu, when we use "service" we should start with "the"; once we drop "service" we then also drop the "the".

- the IBM Quantum System service
- IBM Quantum System

That being said, I am concerned about using just IBM Quantum System or Quantum System on its own since the term doesn't clearly disambiguate from the QPUs that are not on-premises. Thoughts on how to make sure the distinction is very clear?

arnaucasau: Sounds good! Updated in https://github.com/Qiskit/documentation/pull/4829/commits/d17d33718b486d76aae24a43ea6d9d60b114975d 👍

docs/guides/_toc.json
```diff
@@ -937,7 +937,11 @@
             },
             {
               "title": "Activity tracking events for Qiskit Runtime",
-              "url": "/docs/guides/observability"
+              "url": "/docs/guides/observability-runtime-rest"
+            },
+            {
+              "title": "Activity tracking events for Quantum System",
```
Eric-Arellano: @abbycross should this be the full title Activity tracking events for the IBM Quantum System service?

abbycross: Sorry I missed this - yes, I would say "for the IBM Quantum System service".

arnaucasau: Cool! Done in https://github.com/Qiskit/documentation/pull/4829/commits/da473d62d306027ef0ffd2b4403519e187444828

docs/guides/observability-quantum-system.mdx
```diff
@@ -0,0 +1,34 @@
+---
+title: Activity tracking events for the IBM Quantum System service
+description: Learn about activity tracking events for the IBM Quantum System service.
+---
+
+# Activity tracking events for the IBM Quantum System service
+
+IBM Cloud&reg; services, such as the IBM Quantum System service, generate activity tracking events.
```
abbycross:
```suggestion
IBM Cloud&reg; services, such as the IBM&reg; Quantum System service, generate activity tracking events.
```

abbycross: I don't know if this will get overwritten since it was generated text?

arnaucasau: Yeah, it will get overwritten. Is it okay to add the `&reg;` to every instance or it should only be the first one?

abbycross: It should only be the first one. If that isn't going to work, maybe we can find another workaround...

arnaucasau: Yeah, I think that works! @Eric-Arellano I added a small function to add the icon when IBM is used in the first title instance https://github.com/Qiskit/documentation/pull/4829/commits/3e3d9066f45b88abc1bb4ca80cbc3f5e975130a3

Issue #4857, opened 2026-03-20
Add CI check that qiskit and qiskit-c are the same version


Eric-Arellano:
For every version, check that the `_pkg.json` is identical between the two. If not, give a helpful error message explaining how you need to run the same command with both `-p qiskit` and `-p qiskit-c`.

PR #4863, opened 2026-03-23
Add new check to ensure Qiskit and Qiskit C are in sync

arnaucasau:
Closes https://github.com/Qiskit/documentation/issues/4857

This PR adds a small check to ensure every version of Qiskit is also a version of Qiskit C, and vice versa.

Example of output:

```diff
❌ Out of sync versions found:

- qiskit
+ qiskit-c

@@ -6,1 +6,1 @@
-   "2.2.3",
+   "2.2.2",

The qiskit and qiskit-c packages share the same artifact, but you need to run the API geneartion script for each of them in every version. 

For example, to generate the v2.0.3 docs as the latest ones, you need to run two commands: 
	npm run gen-api -- -p qiskit -v 2.0.3 
	npm run gen-api -- -p qiskit-c -v 2.0.3 

Make sure you generated both qiskit and qiskit-c when adding or updating the API docs. Check that every version is in sync in their respective `_package.json` files!
```

### DIFF
--- a/scripts/js/commands/checkQiskitApiVersions.ts
+++ b/scripts/js/commands/checkQiskitApiVersions.ts
@@ -4,25 +4,25 @@ import { diff } from "jest-diff";
 import { getDevVersion, getReleasedVersions } from "../lib/apiVersions";
 
 async function main() {
-  const [QiskitHistoricalVersions, QiskitCurrentVersion] =
+  const [qiskitHistoricalVersions, qiskitCurrentVersion] =
     await getReleasedVersions("qiskit", false);
-  const QiskitDevVersion = await getDevVersion("qiskit");
+  const qiskitDevVersion = await getDevVersion("qiskit");
   const allQiskitVersions = [
-    QiskitDevVersion,
-    QiskitCurrentVersion,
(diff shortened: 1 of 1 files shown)


scripts/js/commands/checkQiskitApiVersions.ts
```diff
@@ -0,0 +1,52 @@
+import { isEqual } from "lodash-es";
+import { diff } from "jest-diff";
+
+import { getDevVersion, getReleasedVersions } from "../lib/apiVersions";
+
+async function main() {
+  const [QiskitHistoricalVersions, QiskitCurrentVersion] =
```
Eric-Arellano: nit, why is this UpperCase rather than camelCase?

arnaucasau: Ugh, good catch! I'm so use to write Qiskit with the Q upper-case that I accidentally used it like that for the variables. Fixed in https://github.com/Qiskit/documentation/pull/4863/commits/bc5caaba875b08c6e2f0ad7ceb75464b78c62c21

Issue #3100, opened 2025-05-08
Set up link checker in private repositories


Eric-Arellano:
We already have a script for checking links in other repositories: https://github.com/Qiskit/documentation/blob/main/scripts/js/commands/checkInternalLinksFromOtherRepo.ts. However, we haven't updated it since changing how links work that we now expect a `/docs` and `/learning` prefix. We should get this script working. 

That also means regenerating our saved links:

https://github.com/Qiskit/documentation/blob/f03f97347cea0c6b06c6cdc8817b11df86f1ada4/package.json#L34

When working on this, use our branch `functions-template`, which is how we run our scripts with private repositories. First, run `npm run check:internal-links` and ensure it's breaking; if not, add some test links to the files. Then, fix the code on `main`, and open a PR to update `functions-template` to use the new commit & to add `npm run check:internal-links` to the `./check` script.

PR #4886, opened 2026-03-25
Update function template

Eric-Arellano:
With this PR, `./fix`, `./check`, and `./start` work again. (Pending a PR I will make to `main` to fix some import paths.)

Note that `check:markdown` is a new script that merges the prior `images` and `metadata`, and adds a new check for ensuring the h1 and page SEO metadata title are identical. We cannot use `:` in the SEO metadata, so I use `-` for both the h1 and page metadata.

The internal link checker is still not working, per https://github.com/Qiskit/documentation/issues/3100.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.1.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "qiskit-documentation": "qiskit/documentation#5debd41828c1c26721e5378bca1b1a3d85774878"
+        "qiskit-documentation": "qiskit/documentation#7c4975c83b4812563fff721cb0d081b358a20947"
       },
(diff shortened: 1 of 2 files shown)


scripts/tool_runner/setup/installation_instructions.py
```diff
@@ -0,0 +1,24 @@
+# This code is a Qiskit project.
```
Eric-Arellano, line 1: Copied from `main`

scripts/tool_runner/__init__.py
```diff
@@ -0,0 +1,94 @@
+# This code is a Qiskit project.
```
Eric-Arellano, line 1: Copied from `main`

PR #4925, opened 2026-04-01
Remove remaining tables in Functions API docs and reformat

abbycross:


docs/api/functions/q-ctrl-performance-management.mdx
```diff
@@ -45,46 +53,46 @@ The name of the backend
 - Example: `"ibm_fez"`
 </Attribute>
 
-### `options`
+### Options
 <Attribute attributeTypeHint="`dict`" attributeValue="See Options section for more details">
-Input options; see [Options](#est-options) section for more details
+Input options; see [Options list](#est-options) for more details
```
kaelynj: I would add in a similar explanation from the IBM Circuit function to explain that the input options are a nested dictionary:

Maybe something like
> Similar to Qiskit Runtime primitives, options for this function are specified as a nested dictionary. The full list of options and their default values are below.

docs/api/functions/q-ctrl-performance-management.mdx
```diff
@@ -105,39 +113,36 @@ The name of the backend
 - Example: `"ibm_fez"`
 </Attribute>
 
-### `options`
+### Options
 <Attribute attributeTypeHint="`dict`" attributeValue="See the Options section for details">
-Input options; see [Options](#sam-options) section for more details
+Input options; see [Options list](#sam-options) section for more details
```
kaelynj: Same feedback as above from the estimator inputs section

PR #4934, opened 2026-04-02
Keep Functions API toc open

abbycross:
Table of contents for the Qiskit Functions API ref docs should always display all Functions.


abbycross:
Boo! this did not work as expected in production. Will need to try again

PR #4946, opened 2026-04-06
Update OBP tutorial to conform to new template

ritajitmajumdar1:
Updated the tutorial to conform to the new template. Need to try a few more times with different settings to obtain better result.


abbycross:
Hi @ritajitmajumdar1! Ok for me to move your new file to the `/tutorials` directory and replace the old operator_backpropagation file with your new one? Then I can run checks on it, once it's in that directory.

ritajitmajumdar1:
Yes, sure.

abbycross (dismissed):
I did a round of copyediting and ran the checks. Ready from my side

henryzou50 (changes requested):
Thanks @ritajitmajumdar1 for the changes and @abbycross for the updates and review. Overall, nice work converting this to the template format! Here are some requested suggestions.

**1. Add a job tag to the large-scale estimator**

So it should be something like:

```python
estimator.options.environment.job_tags = ["TUT_OBP"]
```

**2. Restore live execution in the large-scale block**

Uncomment `# job = estimator.run(pubs)` and remove `job = service.job("d79ukpjc6das739jds5g")`, the hardcoded job ID won't work for any other reader.


**3. Step 4 no longer prints value of the results**

The previous version printed `Expectation value without backpropagation: 0.788...` etc. The new version only assigns variables before plotting. Worth restoring the `print(...)` lines so readers see the numbers especially if they are close, not just bars.


### Some minor grammar fixes

**Learning outcomes & Background:**
- "at the cost of increased number of circuit **execution**" → "**executions**"
- "construct XYZ Hamiltonians and their time-evolution **circuit**" → "**circuits**"

**Step 3:**
- "we create two [Primitive Unified Blocs](...) (**PUB**)" → "(**PUBs**)" 
- "execute the **pubs**" → "**PUBs**" 

**Large-scale example intro:**
- "we use **tensor network** and obtain" → "we use **a tensor network** and obtain"

abbycross:
I tried implementing those changes suggested by @henryzou50 - can you verify that I updated the code properly (i.e., the first three items in Henry's list)?

henryzou50 (dismissed):
@abbycross Thanks for the updates! LGTM! I can rerun the notebook so that we can see the print statements, and then push that update once it is ready.

henryzou50:
After rerunning the notebook with a few backends, the results look off for the large hardware results. Investigating this at the moment. Here are the current ibm_boston results at the moment for reference: 
image

When the previous results before this PR was like: 
image

henryzou50:
I managed to improved the hardware results for all cases by reverting `noise_factors` back to `[1, 1.2, 1.4]`. The narrower `[1, 1.1, 1.2]` window made the linear extrapolation too sensitive and pushed the no-backpropagation result past the physical bound. Note that while the results are better, "No backpropagation" actually outperforms "backpropgation", but "backprogation with truncation" still has the best results. Two important notes here: (1) these are single-run results, and the error bars on each bar are not negligible, all three values are statistically consistent with the exact 0.89 line, and the ordering of the three bars can change run to run. (2) The qualitative takeaway from OBP is still the depth reduction (24 → 20 → 18 two-qubit layers), which is what enables the technique to scale to deeper Trotter circuits where unmitigated noise would otherwise wash out the signal entirely.
image

Here are some other minor edits I made along the way: 

- Dropped the as Estimator alias so the small-scale section uses `StatevectorEstimator` directly, which makes it explicit that Part 1 is a noiseless simulation, not hardware.
- Added a short markdown cell after the small-scale plot explaining that the original and backpropagated bars agree because the simulator is noiseless, and pointing forward to the hardware example as the place where OBP's actual benefit shows up.
- Reintroduced the 2-qubit depth prints for the original, backpropagated, and truncated circuits so the depth savings are visible inline before the hardware run.

Otherwise, I would say this PR is good to go. @nathanearnestnoble Can you take a quick review on it before it gets merged?

ritajitmajumdar1:
image
The consistency is a bit problematic. This is after running on ibm_kingston. Now it shows backpropagation w/ and w/o truncation are similar and both better than no backpropagation.

henryzou50:
@ritajitmajumdar1 Nice catch! I reran with `ibm_kingston` and got better results where it shows backpropagation w/ and w/o truncation are similar and both better than no backpropagation. And yes, I agree, the inconsistency is a bit of a problem.

nathanearnestnoble:
On consistency in results: I think the consistency issue can hopefully be characterized a bit more systematically with @emilkovacev.  

Aside from this, I would only suggest we add reference to the pauli prop package: https://github.com/Qiskit/pauli-prop/tree/main/docs/tutorials. I think it fits best 'next steps'....could maybe go in 'prereqs'. 

Thank you @ritajitmajumdar1 and @henryzou50 for pushing this forward!

docs/tutorials/operator-back-propagation.ipynb
```diff
@@ -764,6 +764,7 @@
     "<Admonition type=\"tip\" title=\"Recommendations\">\n",
     "- [Approximate quantum compilation for time evolution circuits](/docs/tutorials/approximate-quantum-compilation-for-time-evolution)\n",
     "- [Multi-product formulas to reduce Trotter error](/docs/tutorials/multi-product-formula)\n",
+    "- [`pauli-prop`](https://github.com/Qiskit/pauli-prop), a Rust-accelerated package for Pauli propagation, with [tutorials](https://github.com/Qiskit/pauli-prop/tree/main/docs/tutorials) covering OBP, classical expectation-value estimation, and noisy simulation.\n",
```
abbycross:
```suggestion
    "- [`pauli-prop`](https://github.com/Qiskit/pauli-prop), a Rust-accelerated package for Pauli propagation, with [tutorials](https://github.com/Qiskit/pauli-prop/tree/main/docs/tutorials) covering OBP, classical expectation-value estimation, and noisy simulation\n",
```

nathanearnestnoble (approved):

PR #4947, opened 2026-04-06
Update error detection tutorial to conform to new template

MeltemTolunay:


nathanearnestnoble:
Thank you @MeltemTolunay .   I am noticing there is no 'prerequisite' section -- could you add appropriate material (or flag to me if there are none).

nathanearnestnoble:
I am also noticing we import qiskit-device-benchmarking, for purpose of using the `plot_gate_map` function.   Is it not possible to instead use the same [qiskit visualization function](https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.visualization.plot_gate_map)?

nathanearnestnoble:
please also be sure to include steps for small simulator example.  Otherwise, I think this looks good!  (Thank you!)

henryzou50 (changes requested):
Nice work on the changes! The tutorial looks great and is mostly formatted to the Tutorial template. I just have a few change requests so that it more exactly matches the template and is consistent with the other updated tutorials. 

## Template conformance

1. **Step 1–4 subheadings are in the wrong section.** The template (and recent conversions like the SKQD and SQD tutorials) put `### Step 1` … `### Step 4` under **Small-scale simulator example**, and compress the steps into a single block in **Large-scale hardware example**. Right now this PR does the opposite — the small-scale section has no step breakdown, and the large-scale section has all four `### Step` headings. Can we restructure to match? For the large-scale example, we can still call out the steps using inline comments inside the code block, e.g.:
   ```python
   # -------------------------Step 1-------------------------
   ...
   # -------------------------Step 2-------------------------
   ...
   ```
2. **Missing `## Prerequisites` section.** The template places this between Learning outcomes and Background. As Nate mentioned, could you add appropriate material (or flag to him if there are none).
3. **`## Large-scale hardware example: Prepare a GHZ state on real hardware`** the template expects bare `## Large-scale hardware example`. Consider dropping the suffix to match.

## Requirements section

These two entries can be removed, as installing Qiskit with visualization support already pulls in both, and the specific versions don't matter here:

```
- NumPy v2.3.2 (`pip install numpy`)
- Matplotlib v3.10.7 (`pip install matplotlib`)
```

## Job tags on hardware runs

Per the template, hardware jobs should have a unique tag set via `sampler.options.environment.job_tags = ["TUT_XYZ"]` before the Sampler `run()` call in Step 3. We should add one here. Suggested abbreviations for the title "Low-overhead error detection with spacetime codes":

- `TUT_LOEDSC` — strict: first letter of every word (skipping "with")
- `TUT_EDSC` — error detection / spacetime codes (shorter, still distinctive)

I'd lean toward `TUT_EDSC`, but choose which ever you prefer.

## Grammar / typos

- Markdown cell 8:
  - Existing: "This ancilla qubit, or the check qubit, starts the in the $\ket{+}$ state."
  - Fix: "This ancilla qubit, or the check qubit, starts in the $\ket{+}$ state." (drop the extra "the")
- Markdown cell 9:
  - Existing: "If it's measured in '1', then this means that that an error occurred in the payload circuit, and we discard that measurement."
  - Fix: "…then this means that an error occurred…" (duplicated "that")
- Markdown cell 20:
  - Existing: "We remove edges/nodes whose CZ errors, measurement errors, and the T2 values are worse than the thresholds below."
  - Fix: "…measurement errors, and T2 values are worse than…" (drop "the" for parallelism)

henryzou50:
> I am also noticing we import qiskit-device-benchmarking, for purpose of using the `plot_gate_map` function. Is it not possible to instead use the same [qiskit visualization function](https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.visualization.plot_gate_map)?

Yes, we can drop the `qiskit-device-benchmarking` dependency and use `qiskit.visualization.plot_gate_map` instead. The one caveat is that for a given backend, the built-in function falls back to a spring layout (hexagonal honeycomb) since it doesn't have hardcoded coordinates for that qubit count, so it doesn't look as clean as the grid layout we had before. We can still use it without passing coordinates, but to match the original look we'd want to supply `qubit_coordinates`. Something like:

```python
# Build heavy-hex grid coordinates for the 156-qubit backend
qubit_coordinates = []
for r in range(8):
    y = 14 - 2 * r
    for col in range(16):
        qubit_coordinates.append([y, col])
    if r < 7:
        bridge_cols = [3, 7, 11, 15] if r % 2 == 0 else [1, 5, 9, 13]
        for col in bridge_cols:
            qubit_coordinates.append([y - 1, col])

# Plot the layout of GHZ and check qubits on the device
plot_gate_map(
    backend,
    label_qubits=True,
    qubit_coordinates=qubit_coordinates,
    line_color=[
        "black"
        if edge[0] in ghz_qubits + checks and edge[1] in ghz_qubits + checks
        else "lightgrey"
        for edge in backend.coupling_map.graph.edge_list()
    ],
    qubit_color=[
        "blue"
        if i in ghz_qubits
        else "salmon"
        if i in checks
        else "lightgrey"
        for i in range(0, backend.num_qubits)
    ],
)
```

The coords are system-size specific (this one is hardcoded for 156-qubit Heron), so if the backend changes we'd need to adjust them. Here how it looks like: 
image

With no `qubit_coordinates`, it looks like: 
image

Let me know if any of these are what you prefer, and happy to push that change if you like.

MeltemTolunay:
> Thank you @MeltemTolunay . I am noticing there is no 'prerequisite' section -- could you add appropriate material (or flag to me if there are none).

I have added this section. I think the tutorial is pretty self-contained, so I could only identify one relevant tutorial. I'm happy to add more if you can think of another relevant tutorial to refer to.

MeltemTolunay:
> I am also noticing we import qiskit-device-benchmarking, for purpose of using the `plot_gate_map` function. Is it not possible to instead use the same [qiskit visualization function](https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.visualization.plot_gate_map)?

After reading Henry's detailed comment, I'm leaving it as it is for now. Do you have a preference for this? I can make the change.

MeltemTolunay:
> please also be sure to include steps for small simulator example. Otherwise, I think this looks good! (Thank you!)

Qiskit patterns steps are now included for the small simulator example.

MeltemTolunay:
@henryzou50 Thank you for the review! Here are my updates/explanations:

Template conformance:
1. I have added the steps to the small-scale simulator example as well. I would like to specifically leave the large-scale hardware example section as it is, without merging the individual steps. The reason for this is that, in this particular tutorial, the small example is quite different from the hardware example. The first section only explains how and why error detection works with a random Clifford circuit. However, the hardware section builds a GHZ state. This task has quite complex subsections with helper functions. I think consolidating the hardware section would make it worse.
2. I added the prerequisites section.
3. I modified the title.

Requirements section:
I removed the mentioned requirements.

Job tags on hardware runs:
I added the second job tag.

Grammar / typos
All are fixed. Thank you for catching these!

henryzou50 (dismissed):
Thanks for the additional updates and changes. LGTM!

henryzou50:
@nathanearnestnoble I would say this is ready to be merged into main, but I will let you take a final look first.

abbycross (dismissed):
Thanks!

nathanearnestnoble:
The notebook has dependency on the qiskit_device_benchmarking because of the associated "plot_gate_map" -- however, this exists in qiskit visualization: https://quantum.cloud.ibm.com/docs/en/api/qiskit/qiskit.visualization.plot_gate_map.  I think we should remove dependency on the benchmarking repository as it doesnt have unique functionality that is needed and is more complicated to install.

abbycross (approved):

PR #4944, opened 2026-04-06
Update SKQD tutorial to conform to new template

kevinsung:

PR #4884, opened 2026-03-25
Update SQD tutorial to conform to new template

kevinsung:
- Removed a significant chunk of the background on the LUCJ ansatz. The tutorial already links to the paper as well as the ffsim documentation page on it.
- Removed the snippets that ran transpilation without `ffsim.PRE_INIT` for the sake of comparison. I thought it was unnecessary and potentially confusing
- Added small-scale simulator example using a (10e, 8o) active space derived from the STO-6G basis set. It uses noiseless simulation and trivially converges in the first iteration of SQD.

PR #4988, opened 2026-04-13
Some cleanup from recent merges

abbycross:
Making some fixes to work updated in #4944 and #4884


nathanearnestnoble:
Thanks for bringing together @abbycross !!!   Realizing we still need to add job tags

- For Sample based krylov quantum diagonalization could we add `sampler.options.environment.job_tags = ["TUT_SKQD"]` right before sampler.run
- For sample based quantum diagonalization could we add `sampler.options.environment.job_tags = ["TUT_SQD"]` right before sampler.run

nathanearnestnoble (approved):

PR #5003, opened 2026-04-17
Implicit solvent calculations tutorial

abbycross:
By @blannix


docs/tutorials/implicit-solvent-calculations.ipynb
```diff
@@ -0,0 +1,1756 @@
...
...describe how a solute compound interacts with a solvent, without modeling the solvent system directly. Instead, an approximation is made wherein the solute system model is wrapped in a mathematical representation of a dielectric medium characterized empirically. This dielectric approximation then interacts with the solvent, which is itself modeled directly. The dielectric medium influences characteristics of the solute system, such as its ground state energy, by interacting with its electron field. This is important to biophysical models used, say, in drug discovery, because compounds behave differently in various dielectric environments. Modeling a compound in the air (in vacuo) will describe different behavior than modeling it in water. Since pharmaceutical compounds must enter into the human body, which itself is composed mostly of water, it is helpful to model a compound in a solution like water instead of in vacuo. With implicit solvent models, we can achieve this behavior cheaply, although the end result will generally be more approximate than the counterpart: more computationally costly explicit solvent models that create direct representations of the solution molecules.\n",
```
abbycross:
```suggestion
    "Implicit solvent calculations are frequently used in computational biophysics. These models describe how a solute compound interacts with a solvent, without modeling the solvent system directly. Instead, an approximation is made wherein the solute system model is wrapped in a mathematical representation of a dielectric medium characterized empirically. This dielectric approximation then interacts with the solute, which is itself modeled directly. The dielectric medium influences characteristics of the solute system, such as its ground state energy, by interacting with its electron field. This is important to biophysical models used, say, in drug discovery, because compounds behave differently in various dielectric environments. Modeling a compound in the air (in vacuo) will describe different behavior than modeling it in water. Since pharmaceutical compounds must enter into the human body, which itself is composed mostly of water, it is helpful to model a compound in a solution like water instead of in vacuo. With implicit solvent models, we can achieve this behavior cheaply, although the end result will generally be more approximate than the counterpart: more computationally costly explicit solvent models that create direct representations of both solute and solvent molecules.\n",
```

beckykd (approved):

PR #4813, opened 2026-03-13
Updates to two tutorials: QAOA, and Advanced techniques for QAOA

abbycross:
Continuation of #4805 by @ibrahim-shehzad : Making updates to match the updated tutorial template

Moving to a write-access branch to run the notebook tester

[Here is the PR preview](https://qiskit.github.io/documentation/pr-4813/docs/tutorials/advanced-techniques-for-qaoa)

Also adding in updates from @mrvee-qC-bee to [QAOA](https://qiskit.github.io/documentation/pr-4813/docs/tutorials/quantum-approximate-optimization-algorithm) tutorial (finishing up work started in https://github.com/Qiskit/documentation/pull/5029)

PR #5029, opened 2026-04-23
Update to Ajc/qaoa pr 4805

mrvee-qC-bee:
PR to update QAOA notebook 

- best_solution was broken. It sets min_cost = 1000 but never updates it inside the loop, so basically every sample passes the <= check and the function returns whatever happens to come last. Made it toupdate min_cost when we find something better.

- The docstring on build_max_cut_paulis refers to a build_max_cut_graph function that doesn't exist anywhere — looks like a leftover from wherever this code originally lived. Rewrote it to describe what the function actually returns.

- The HCH_C HC​ formula was written two different ways in two different places — Background had ∑QijZiZj\sum Q_{ij} Z_iZ_j ∑Qij​Zi​Zj​, Step 1 had ∑ZiZj\sum Z_iZ_j ∑Zi​Zj​. Made them match, with Step 1 explicitly saying "for unweighted Max-Cut this reduces to…".

- Step 1 had a dangling "Start by defining a 5-node graph:" sentence that led into nothing (the graph was already defined in a prior cell, and the next thing was building the Hamiltonian). Deleted.

- Put back the general Hamiltonian HC=∑ijQijZiZj+∑ibiZiH_C = \sum_{ij} Q_{ij} Z_iZ_j + \sum_i b_i Z_i HC​=∑ij​Qij​Zi​Zj​+∑i​bi​Zi​ in Background with a note that bi=0b_i = 0 bi​=0 for unweighted Max-Cut. The rewrite had trimmed it to the simplified form only, which left readers with no hook if they wanted to adapt this to weighted graphs or other QUBO problems.

- Brought back the single-job / batch / session illustration in Step 3 — the rewrite had dropped it when compressing the explanation.

- Renamed the "Small-scale simulator example" heading to just "Small-scale example." The code in that section runs on real hardware (simulator=False), so "simulator example" is incorrect. Added a sentence making this explicit.

- Split the 85-line utility-scale cell into four cells, one per step, with short markdown transitions between them. Matches the style of the small-scale walkthrough and makes the block less intimidating to scroll through.

- Fixed the grammar on "The steps to build the workflow are followed the same as above" — now reads "The workflow follows the same steps as above, applied to a much larger graph."

- Added a classical brute-force baseline after the 5-node example. At n=5 you're enumerating 32 bitstrings, which is trivial, and it gives readers a way to see how close QAOA actually got to the real optimum.

- Added a comment explaining the evaluate_sparse_pauli bit-trick. The np.packbits / np.bitwise_xor.reduce stuff is dense, and without context it's not obvious why you'd do that instead of just calling SparsePauliOp.expectation_value. The answer is performance — it's dramatically faster across 10,000 samples — and the comment now says so.

- Added interpretation for the convergence plot. A reader looking at a jagged curve shouldn't have to guess whether that's normal or a sign something's wrong. Walks through what a healthy curve looks like, what a bad one means, and the specific way COBYLA fails when the noise exceeds the per-step cost improvement (COBYLA is derivative-free, so its linear-approximation model just can't tell descent from jitter).

- Added a short comment on the initial γ, β values. The original just set them with no explanation. Made it clear that these aren't principled defaults, just bounded midpoints of the periodic parameter ranges.

- Added a one-liner on what optimization_level=3 means — slower to transpile, shorter circuits, more robust on hardware. Otherwise a newcomer sees the number and has to guess what the range is.

- Replaced the hardcoded "Heron device" label in the CDF plot with backend.name. That string was going to read wrong the moment anyone ran this on a different backend.

- Renamed the module-scope n = 5 and n = 100 to n_small and n_large. Notebooks get executed out of order all the time, and having both bound to n is asking for a confusing bug.

- Dropped the redundant bare import matplotlib (was only used for one matplotlib.rcParams.update call that works identically as plt.rcParams.update).


mrvee-qC-bee:
@abbycross fyi!

abbycross:
I've moved the changes to #4813; we can close this PR.

PR #5046, opened 2026-04-29
Adds Qiskit addon prose content pipeline  

eharvey328:
Extracts prose HTML and Jupyter notebooks for Qiskit addons and converts them to MDX to be consumed by our docs app. **The goal is to extend the existing API docs pipeline for minimal changes.**

- Adds a new addons command `npm run gen-addon -- -p {package}`
  - function like the existing `npm run gen-api -- -p {package} -v {version}`

### Why a separate pipeline instead of reusing the API pipeline

- The API pipeline (`runApiDocsPipeline`) targets `apidocs/**` stubs and generates `docs/api/{pkg}/`.
- Addons ship _guides_ (HTML) + _notebooks_ (`.ipynb`) under the Sphinx root — no stubs, no
historical versioning, no release notes per-page. 
- The addon pipeline reuses the same shared stages
(`pipelineStages.ts`, `notebookStages.ts`) but diverges in:

| Concern | API pipeline | Addon pipeline |
|---|---|---|
| Input glob | `apidocs/**`, `stubs/**` | Everything except apidocs/tutorials/release-notes |
| Output path | `docs/api/{pkg}/` | `docs/addons/{pkg}/` |
| Images | `public/docs/images/api/` | `public/docs/images/addons/` |
| TOC generator | `generateToc.ts` | `generateAddonToc.ts` |
| Frontmatter | auto-generated from URL | extracted from Sphinx `<h1>` |
| Notebooks | not processed | processed (link rewriting, frontmatter cell injected) |
| Release notes | yes (own pipeline stage) | excluded from glob |
| Versioning | `latest`/`historical`/`dev` | `latest` only |

---

### New files

| File | Purpose |
|---|---|
| `scripts/js/lib/api/addonDocsPipeline.ts` | Orchestrator: HTML→MD, notebooks, images, TOC, tutorials manifest |
| `scripts/js/lib/api/pipelineStages.ts` | Shared HTML stages extracted from the old `conversionPipeline.ts`; used by both pipelines |
| `scripts/js/lib/api/notebookStages.ts` | Notebook-specific stages (read, process, write, collect images) |
| `scripts/js/lib/api/generateAddonToc.ts` | Builds `_toc.json` with back-link, API reference |
| `scripts/js/commands/api/updateAddonDocs.ts` | CLI entry point: `--package`, `--version` |
| `scripts/js/commands/api/updateDocsShared.ts` | Shared CLI helpers (yargs options, artifact download, output wipe) extracted from `updateApiDocs.ts` |
| `scripts/js/lib/api/apiDocsPipeline.ts` | Renamed/extracted from `conversionPipeline.ts`; API pipeline now lives here |

## Design decisions

1.  **`extractfrontMatter=true` in addon pipeline**: Addons don't follow the same URL-slug naming convention as API stubs, so URL-derived titles were wrong. Extracting from `<h1>` is more reliable.
2. **parentLabel** added to `_toc` to provide text for the parent back arrow (link) in the toc. Added as a new property for backwards compatibility with learning content. 
3. **Loading all `objects.inv` files for cross-package link resolution**: Addon notebooks link to symbols in _other_ packages (`qiskit.github.io/{pkg}/stubs/...`). The old pipeline only resolved links against the _current_ package's inventory, which isn't enough for addons.
   - **New pattern:** Both pipelines now call `ObjectsInv.loadPublishedApis(publicBaseFolder)` first, which scans `public/docs/api/` on disk and returns a `Map<pkgName, ObjectsInv>` for every package that has a published inventory. That map flows down through `postProcess` → `updateLinks` and `processNotebooks` → `rewriteNotebookLinks`, where `resolveStubUrl(url, allInvs)` uses it to look up the right package's inventory by name.
   - **`resolveStubUrl` handles two URL forms:**
     - `stubs/{Symbol}` — entry lookup by name; the entry's `uri` already carries the correct anchor.
     - `apidocs/{page}` — page-level lookup; the source `#anchor` is preserved and appended.
   - `std:` domain guard in `shouldIncludeEntry` — excludes non-API entries (guide pages, index, install) from published inventories so they can't produce false cross-package matches.


scripts/js/lib/links/ignores.ts
```diff
@@ -587,8 +587,57 @@ function _qiskitCRegexes(): FilesToIgnores {
   };
 }
 
+function _addonStaleTutorialLinks(): FilesToIgnores {
```
axelhzf: When this is merged, we should open an issue so we don't forget about this.

scripts/js/lib/api/pipelineStages.ts
```diff
@@ -0,0 +1,208 @@
...
+    .clone()
+    .find("a.headerlink")
+    .remove()
+    .end()
+    .text()
+    .trim();
+  const isRootIndex = url.endsWith(`/${pkg.name}/index`);
+  const description = isRootIndex
+    ? `Documentation for the latest version of ${pkg.title}`
+    : `${h1} for the lastest version of ${pkg.title}`;
```
axelhzf:
```suggestion
    : `${h1} for the latest version of ${pkg.title}`;
```

scripts/js/lib/api/htmlToMd.ts
```diff
@@ -261,7 +262,7 @@ function buildAdmonition(
       {
         type: "mdxJsxAttribute",
         name: "title",
-        value: toText(titleNode),
+        value: titleNode ? toText(titleNode) : "",
```
axelhzf: The Admonition component has

```
<div className="bg-layer-02 text-text-primary font-600 p-12 py-8 flex items-center gap-8">
  <AdmonitionIcon type={type} />
  {title ?? capitalize(type)}
</div>
```

so maybe we should default to `undefined` instead of `""`

scripts/js/lib/api/generateAddonToc.ts
```diff
@@ -0,0 +1,266 @@
...
+// obtain a copy of this license in the LICENSE file in the root directory
+// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+//
+// Any modifications or derivative works of this code must retain this
+// copyright notice, and modified files need to carry a notice indicating
+// that they have been altered from the originals.
+
+// Generates the _toc.json for an addon's content pages.
+//
+// Addon TOCs have a fixed three-part shape:
```
axelhzf: How hard would be to process the html to extract the toc preserving the order, and ignore the entries that we are not including?

scripts/config/cspell/cSpell.json
```diff
@@ -43,7 +43,18 @@
...
     "ebitsa",
     "ebitsb",
-    "pmzero"
+    "pmzero",
+    "textrm",
+    "Frontmatter",
+    "leftarrow",
+    "forall",
+    "arxiv",
+    "subpace", // fix in addon-sqd
```
axelhzf: Can we add an issue for this?

scripts/js/lib/api/htmlToMd.ts
```diff
@@ -338,6 +340,16 @@ function buildMathExpression(node: any, type: "math" | "inlineMath"): any {
   return { type: type, value };
 }
 
+function remarkEscapeMathPipesInTables() {
```
eharvey328: Some table have math with `|` in them which causes the markdown to add columns.

This fixes it by escaping pipes to `\\vert`

|  Source | Issue	| 
|----------|:-------------:|
| Image | Image

README.md
```diff
@@ -321,6 +328,43 @@ Additionally, If you are regenerating a dev version, then you can add `--dev` as
...
+   npm run gen-addon -- -p <pkg-name>
+   ```
+
+   To regenerate all addon packages at once:
+
+   ```sh
+   npm run gen-addon -- --all
+   ```
+
+   Or if you don't need to download but need to re-run the pipeline you can pass "--ship-download" to speed things up
```
Alienah:
```suggestion
   Or if you don't need to download but need to re-run the pipeline you can pass "--skip-download" to speed things up
```

scripts/js/lib/api/Notebooks.ts
```diff
@@ -0,0 +1,17 @@
+export type NotebookCell = {
```
Alienah, line 16: Do we need to add this header here?

```suggestion
// This code is a Qiskit project.
//
// (C) Copyright IBM 2026.
//
// This code is licensed under the Apache License, Version 2.0. You may
// obtain a copy of this license in the LICENSE file in the root directory
// of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
//
// Any modifications or derivative works of this code must retain this
// copyright notice, and modified files need to carry a notice indicating
// that they have been altered from the originals.

// Jupyter notebook stages used by the addon docs pipeline. The API pipeline
// does not process notebooks today, but these stages live alongside the other
// shared stages so that any future pipeline needing notebook handling can
// reuse them without duplication.

export type NotebookCell = {
```

README.md
```diff
@@ -321,6 +328,43 @@ Additionally, If you are regenerating a dev version, then you can add `--dev` as
 
 In this case, no commit will be automatically created.
 
+## Generate or update addon docs
+
+Addon docs for packages like are generated by a separate pipeline from the API pipeline. Content is published to `docs/addons/{pkg}/` rather than `docs/api/{pkg}/`.
+
+### Generate addon docs for a new release
+
+3. Run:
```
Alienah:
```suggestion
1. Run:
```

README.md
```diff
@@ -321,6 +328,43 @@ Additionally, If you are regenerating a dev version, then you can add `--dev` as
...
+   npm run gen-addon -- --all
+   ```
+
+   Or if you don't need to download but need to re-run the pipeline you can pass "--ship-download" to speed things up
+
+   ```sh
+   npm run gen-addon -- --all --skip-download
+   ```
+
+4. Open a pull request with the generated changes.
```
Alienah:
```suggestion
2. Open a pull request with the generated changes.
```

scripts/js/lib/api/updateLinks.ts
```diff
@@ -177,6 +178,8 @@ export function normalizeUrl(
 }
 
 export function relativizeLink(link: Link): Link | undefined {
+  rewriteQiskitAddonLinks(link);
```
brownjedi: hmm, are we not using the return from this function?

scripts/js/lib/api/updateLinks.ts
```diff
@@ -254,7 +277,13 @@ export async function updateLinks(
...
           }
-
+          const resolvedStub = maybeObjectsInv?.resolveStubUrl(
+            node.url,
+            allObjectInvs,
+          );
+          if (resolvedStub) {
+            node.url = resolvedStub;
+          }
           node.url = normalizeUrl(node.url, resultsByName, itemNames, kwargs);
```
brownjedi, line 303: I think we are overwriting node.url immediately right? is this supposed to be inside an else condition?

brownjedi: nevermind 🤦🏾. I feel dumb haha

Alienah (approved):
Nice work! 🚀

PR #5088, opened 2026-05-11
Update spin chain vqe tutorial

mrvee-qC-bee:
Change log just to keep track. Updates made:

- Conform to new template, addition of learning outcomes and prerequisites

- Additional background text and context for spin chain problems

- Renamed "Small-scale simulator example" → "Small-scale example" in cell 5. The code under it uses simulator=False, so the old heading was wrong. Also tweaked the body text since it was promising a "hardware run" later that doesn't exist.

- Fixed the barren plateau link in cell 21 — it was pointing at the SQD add-on page for some reason. Now points to the variational-algorithm-design course section on barren plateaus.
 
- Fixed a rendering bug in the SPSA section (cell 11) where a paragraph started with a comma after a display-math block.
 
- Reverted URLs from absolute back to relative in cells 1 and 23 to match our usual convention (/docs/..., /learning/courses/...). Left the ibm.com blog link alone since it's external.
 
- Added a sentence in the Background clarifying that "nearest-neighbor" in the code means the hardware's native coupling on the first N qubits, not a strict 1D chain — the Hamiltonian is written as if it's 1D but the code uses backend.target.build_coupling_map().
 
- Executed cell till iteration convergence. Result plots available and uploaded

- fix stray text in Step 1: First, we import some generic packages and create some helper functions." This is leftover from a prior version, as imports/helpers were already moved into the Setup section.

- The background says "we use the EfficientSU2 ansatz", but the code in cell uses `efficient_su2`. Note that `EfficientSU2` was depreciated in qiskit, so we can just mention it as `efficient_su2`. 

Contributed and reviewed with @henryzou50 @nathanearnestnoble


mrvee-qC-bee:
@abbycross fyi!

henryzou50 (dismissed):
LGTM and thanks for the updates @mrvee-qC-bee! I will let @abbycross take a look as well to confirm that things also look good on her side.

abbycross:
Thanks for flagging - I will get to this asap! 😎

abbycross:
I think this looks great, thank you! I made a couple small copyediting tweaks. What do you think about making the Challenge section an h2 header rather than an h3, and moving it before the References section, rather than after the Next steps section? I think that would be my preference, unless you have another reason to keep it where it is. @mrvee-qC-bee @henryzou50

henryzou50:
> I think this looks great, thank you! I made a couple small copyediting tweaks. What do you think about making the Challenge section an h2 header rather than an h3, and moving it before the References section, rather than after the Next steps section? I think that would be my preference, unless you have another reason to keep it where it is. @mrvee-qC-bee @henryzou50

Yes, I agree and I like that organization for it. I just pushed that change.

nathanearnestnoble (approved):
LGTM!

PR #4787, opened 2026-03-10
Update pages around usage

abbycross:
cc: @kevin-stevens-ibm @jyu00

PR #5091, opened 2026-05-12
Convention for using "usage"

raulotaolea:
Guidelines for explaining Qiskit Runtime service consumption:

- the term "usage" refers to the consumption of the service.
- service consumption is measured through metrics.
- currently, we support a single metric called `qpu_charge_time_seconds`, which includes both the execution time of circuits on the QPU and additional fixed charges per job and per circuit.
- when we refer to the usage of a job, this is shorthand for the value of the `qpu_charge_time_seconds` metric consumed by that job within the Qiskit Runtime service.
- in the case of batches, usage is the sum of the `qpu_charge_time_seconds` values of all jobs included in the batch.
- in the case of sessions, because the QPU is reserved for the exclusive use of the session, the entire time during which the session remains active is considered usage, regardless of whether the QPU is actively executing workloads. Conceptually, this is equivalent to continuously consuming QPU capacity from the moment the session starts until it ends. This is also equivalent to measuring the session wall-clock time.


raulotaolea:
@abbycross @beckykd I created this PR proposing updates in the `ajc/pricing` branch.

abbycross:
Thanks @raulotaolea! Something has gone wrong - maybe this needs an update from main? I see over 2000 files in this PR :)

raulotaolea:
> Thanks @raulotaolea! Something has gone wrong - maybe this needs an update from main? I see over 2000 files in this PR :)

@abbycross This PR will merge with your branch [ajc/pricing](https://github.com/Qiskit/documentation/pull/4787). I updated it with `main` before creating my own. I think if you sync your branch with `main`, you should only see my changes.

PR #5092, opened 2026-05-12
Fix dependency requirements

kaelynj:
A dependency chain conflict is preventing us from using the most recent version of runtime. This PR will temporarily remove the `qiskit-ibm-transpiler` requirement from the notebook tester so that we can continue checking the majority of our notebooks against `qiskit-ibm-runtime==0.46.1`


abbycross (dismissed):
approving to unblock :)

abbycross (approved):

PR #5132, opened 2026-05-19
Fix misplaced images on Composer guide 

### LINKED ISSUE #4058: Circuit figures do not match text on IBM Quantum Composer guide

Sections: Probabilities view, Q-sphere view and Statevector view
link: https://quantum.cloud.ibm.com/docs/en/guides/composer#probabilities-view

### Select all that apply

- [ ] new content request
- [ ] typo
- [ ] code bug
- [x] out-of-date content
- [ ] broken link
- [x] other

In the 3 sections mentioned above, there is a circuit figure that does not match the other figures and the discussion just below. In the screenshot below, one can see a simple circuit with one qubit:

Image

The 1st figure caption does not match its content. The 2nd figure matches the probabilities of the shown circuit, but with a different style than the current composer. Moreover, the figure caption is also a bit misleading, as it shows only the probabilities, not the quantum circuit.

Finally, the discussion just below these figures presents a completely different state vector, not resulting for the circuits shown before. The correct circuit that obtains the referred state is shown in the figure below:

Image

abbycross:
Closes #4058


abbycross:
I have replaced three erroneous circuit images and added dark-mode versions of them as well.

docs/guides/composer.mdx
kaelynj: Unrelated to what you've done so far, but I've also found an issue with this image. The last circle here is mislabeled and should read "Maximally mixed" rather than "Maximally entangled"

docs/guides/composer.mdx
```diff
@@ -400,9 +400,9 @@ This view visualizes the probabilities of the quantum state as a bar graph. The
 
 Consider the following quantum circuit and its probabilities view:
 
-![a quantum circuit](/docs/images/guides/composer/58f83286-9707-4e43-893e-2bd161c744cd.avif "A circuit consisting of a column of Hadamards that creates an equal superposition of the computational basis states, followed by a two-qubit controlled-Z (CZ) gate.")
+![a quantum circuit](/docs/images/guides/composer/2H-circuit.svg "A circuit consisting of a column of Hadamards that creates an equal superposition of the computational basis states, followed by a two-qubit controlled-Z (CZ) gate.")
```
kaelynj, line 401: Looking at the Composer, I don't actually have the ability to add a CZ gate to my circuit. Symbolically the closest thing is an RZZ gate, but that will produce a different state. You could either replace these with an example using the CX gate or with `RZZ(pi/2)` (the Composer uses and angle of pi/2 by default).

abbycross: I was able to make the CZ using the Z gate and adding a control to it (the Z icon, then the dot icon, which you can edit to say which qubit is the control and which is the target) - is that ok?

docs/guides/composer.mdx
abbycross: Not sure what the source of this image is - these aren't circles generated directly by Composer, correct? If so, we need design help with the image. Can we take the whole "Reduced purity" section down in the meantime, so this PR isn't blocked by it @kaelynj ?

cc: @mackblank @jarusso (see https://quantum.cloud.ibm.com/docs/en/guides/composer#phase-disk)

abbycross: Followup work is being tracked [here](https://github.com/Qiskit/documentation/issues/5154)

beckykd (approved):

PR #5143, opened 2026-05-20
Repetition codes experiment extension

rdputteribm:
Changed the tutorial into a quantum memory experiment with multiple cycles of error correction, as discussed with @nathanearnestnoble .

Also brought the tutorial (more) in line with the tutorial template.


docs/tutorials/repetition-codes.ipynb
```diff
@@ -17,15 +17,54 @@
...
+    "\n",
+    "We will compare the error rate vs. time to the error rates of individual, unencoded qubits, and also to use of the repetition code where we only detect and correct errors after final measurement, but not dynamically during the circuit.\n",
+    "\n",
+    "Note: the repetition code only allows for correction of bit-flip errors and therefore is not a complete error correction code. However, because of its simplicity, it is a good starting point to illustrate how to implement error correction on a quantum computer. The memory experiment below only tests for a single type of error (a $|\\bar{1}\\rangle$ decaying to $|\\bar{0}\\rangle$) and technically only demonstrates the protection of classical information."
```
abbycross:
```suggestion
    "In this tutorial, we demonstrate the bit-flip code using a simple quantum memory experiment. We will prepare the encoded qubit in the logical state $|\\bar{1}\\rangle \\equiv |111\\rangle$ and then implement multiple cycles of idle time (to accrue errors) followed by error detection and correction. We then quantify the probability of a logical error as a function of the number of cycles (time), where the logical error probability is the probability of finding the qubits in a state that doesn't recover the $|\\bar{1}\\rangle$ after final measurement (concretely, states corresponding to an error are $|000\\rangle$, $|001\\rangle$, $|010\\rangle$, and $|100\\rangle$).\n",
    "\n",
    "We will compare the error rate versus time to the error rates of individual, unencoded qubits, and also to use of the repetition code where we only detect and correct errors after final measurement, but not dynamically during the circuit.\n",
    "\n",
    "Note: the repetition code only allows for correction of bit-flip errors and therefore is not a complete error correction code. However, because of its simplicity, it is a good starting point to illustrate how to implement error correction on a quantum computer. The memory experiment below only tests for a single type of error (a $|\\bar{1}\\rangle$ decaying to $|\\bar{0}\\rangle$) and technically only demonstrates the protection of classical information."
```

docs/tutorials/repetition-codes.ipynb
```diff
@@ -69,83 +131,129 @@
...
     "\n",
+    "Below you will build circuits that (1) prepare the data qubits in the logical $|\\bar{1} \\rangle$ state, then (2) run multiple cycles of a $5 \\mu s$ delay followed by error correction (including reset of the syndrome qubits), and (3) read out the state of the data qubits.\n",
     "\n",
-    "def initialize_qubits(circuit: QuantumCircuit):\n",
-    "    \"\"\"Initialize qubit to |1>\"\"\"\n",
-    "    circuit.x(qreg_data[0])\n",
-    "    circuit.barrier(qreg_data)\n",
-    "    return circuit\n",
-    "\n",
+    "We will also test the lifetime of the $|1\\rangle$ state without using error correction by including 3 reference qubits that we prepare in the $|1\\rangle$ state, leave idle, and then read out."
```
abbycross:
```suggestion
    "We will also test the lifetime of the $|1\\rangle$ state without using error correction by including three reference qubits that we prepare in the $|1\\rangle$ state, leave idle, and then read out."
```

docs/tutorials/repetition-codes.ipynb
```diff
@@ -256,251 +383,352 @@
...
-    "### Generate ISA circuits"
+    "#### Generate ISA circuits"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "cc8f818f",
+   "metadata": {},
+   "source": [
+    "We will start by finding an initial layout (i.e. selection of physical qubits to map our circuits too) by transpiling the longest of our circuits using the preset pass manager with optimization level 3."
```
abbycross:
```suggestion
    "We will start by finding an initial layout (as in, a selection of physical qubits to map our circuits to) by transpiling the longest of our circuits using the preset pass manager with optimization level 3."
```

henryzou50 (commented):
Overall this tutorial update looks great, and thanks for the changes! A few notes:

**Structure vs. template:** Having only a large-scale hardware example here is fine, but per the template, I believe should still include both the `## Small-scale simulator example` and `## Large-scale hardware example` headers, with a brief explanation under the small-scale section about why we're skipping the simulator example in this case, similar to how we did it for other tutorials. 

**Tutorial Survey:** The link is correct, but most of the other updated tutorials don't include this section. Probably fine to keep, just flagging for consistency.

I pushed some minor changes for grammar, typos, and consistency, but otherwise great job!

rdputteribm:
Hi @henryzou50 ,
Thanks a lot for your review! 

> Structure vs. template: Having only a large-scale hardware example here is fine, but per the template, I believe should still include both the ## Small-scale simulator example and ## Large-scale hardware example headers, with a brief explanation under the small-scale section about why we're skipping the simulator example in this case, similar to how we did it for other tutorials.

I am happy to add an empty simulator header + explanation if that is preferred (I realize it's inthe template), but would have a slight preference for keeping it simple and just keep leaving it out. What is an example of a tutorial that does not do a small-scale simulator run, but still has a header for it?

> Tutorial Survey: The link is correct, but most of the other updated tutorials don't include this section. Probably fine to keep, just flagging for consistency.

Good catch. Since the survey is not in the template, let me just remove it.

henryzou50:
> What is an example of a tutorial that does not do a small-scale simulator run, but still has a header for it?

@rdputteribm  Good question, there's a clean example for this. [`combine-error-mitigation-techniques`](https://github.com/Qiskit/documentation/blob/main/docs/tutorials/combine-error-mitigation-techniques.ipynb) keeps the `## Small-scale simulator example` header but skips the actual run, with just a one-line explanation of why:

> ## Small-scale simulator example
>
> We will forgo this step since runtime error mitigation is not supported on simulators.

It then continues straight into the `## Large-scale hardware example` section.

Given that, I'd lean toward following the template here: keep both the `## Small-scale simulator example` and `## Large-scale hardware example` headers, and add a brief note under the small-scale one explaining why we skip the simulator in this case. It keeps us consistent with the template and the other tutorials while staying minimal.

henryzou50:
I added the minor changes for the structure so that it is more consistent with the other tutorials. This PR is good for review now, and thanks for the changes @rdputteribm! It looks good to me.

henryzou50:
@abbycross Can you take a look at this again when you have the chance?

abbycross (dismissed):

abbycross (dismissed):

abbycross (approved):

PR #5147, opened 2026-05-20
Minor clarity improvements

Eric-Arellano:


docs/guides/considerations-set-up-runtime.mdx
```diff
@@ -41,7 +41,7 @@ This guide uses the following terms:
 
 Before you set up IBM Quantum Platform for your organization, you need to make these decisions:
 
-* How are user identities defined? You can set up IBM Cloud users, users from another IDP, or both.
+* How are user identities defined? You can set up IBM Cloud users, users from another identity provider (IDP), or both.
```
Eric-Arellano, line 44: We already defined the acronym twice in earlier sections, but docs aren't read linearly—a user might not have read those earlier sections.

abbycross (approved):
Thanks!

PR #5153, opened 2026-05-21
Update content on responsible quantum pg

### LINKED ISSUE #5145: Update responsible computing file
I have some new updates from Ryan M.

abbycross:
Closes #5145


abbycross:
@abdonrd I tried making an edit (I removed standalone `)` characters that were being used to enumerate a list) in case that is what is breaking it - but the PR preview build is still failing :(

scripts/pr-previews/builder.py
abdonrd, line 1: The PR preview build was failing because `scripts/pr-previews/builder.py` drops `_toc.json` and `index.*` files from any section that has no changed files in this PR.

The copytree ignore filter adds the entire subdirectory (e.g. `docs/guides`, `learning/courses/*`) to the ignore list when nothing inside it changed, so copytree never descends into it, and the rules meant to always preserve `_toc.json` and `index.*` never get a chance to run. With no `_toc.json` files copied, the Next.js app's `/api/toc/[[...tocPath]]` route can't generate static params during `output: export`, producing the misleading error:

```
Page "/api/toc/[[...tocPath]]" is missing "generateStaticParams()"
```

This only surfaces when a PR's only changes are in files at the root of `docs/` (like `responsible-quantum-computing.mdx`), which is why other PRs build fine.

Fix: copy all `_toc.json` and `index.*` files explicitly after the copytree pass, alongside the existing hardcoded list. The ignore filter is simplified since those special cases are no longer needed there.

abdonrd (approved):

beckykd (approved):

PR #5170, opened 2026-05-26
Update multi-product-formula tutorial to conform to new template

henryzou50:
## Summary

Brings `multi-product-formula.ipynb` in line with new tutorial template and the conventions established by recent tutorial updates.

Both the small-scale simulator example and the large-scale hardware example now follow the standard four-step layout. Labels match content: Step 2 = transpile/optimize, Step 3 = execute, Step 4 = post-process (compute MPF coefficients and combine). Previously, Step 2 was doing execution and Step 3 was doing post-processing.

Other changes:
- **Background section** rewritten to focus on the Trotter-step (k) vs. circuit-depth trade-off rather than the PF order. r and k_j are introduced on first appearance, eta_n / chi exponents are made explicit for symmetric vs. non-symmetric PFs, and the "stability" paragraph is replaced with a derivation-based explanation of why t/k_min <~ 1 matters.
- **Prerequisites** point to the compilation-methods tutorial and SuzukiTrotter / LieTrotter API pages (the previous links were mismatched).
- **Requirements** trimmed (removed unused quimb / qiskit-quimb).
- **Per-cell narrative** restored for the LSE inspection cells, identity_factory / Néel MPS setup, and the dynamic-coefficients loop.
- **Small-scale post-plot commentary** corrected: no longer reads as if MPFs work better at t/k > 1, and the sampling-error formula sigma_MPF^2 = sum_j x_j^2 sigma_{k_j}^2 is given explicitly.
- **Large-scale section** corrected: the misleading "Steps 1-4 combined" header is dropped (that cell is only Step 1), the inline comment on the k=6 baseline is fixed (depth-comparable single-circuit run, not one that matches the MPF's effective Trotter error), and the results discussion is rewritten against the latest hardware run (dynamic MPF beats every individual product formula; exact-static MPF is worst because ||x||_1 = 5.63 amplifies hardware noise).
- **TeNPy `unit_cell_width` UserWarning** suppressed in the Setup cell. The default is correct for Chain lattices, which is what `CouplingMap.from_line(...)` produces here.
- **cspell:ignore** extended with the new technical / LaTeX terms.

## Acknowledgments

Thanks to Boseong for the detailed review of the prior draft. The descriptions, Step-label fixes, and several of the Background corrections above incorporate that feedback directly.


docs/tutorials/multi-product-formula.ipynb
```diff
@@ -2,153 +2,262 @@
...
+    "$$\n",
     "\n",
-    "where $\\chi$ is the order of the symmetric PF and $s_p = \\left( 4 - 4^{1/(2p-1)} \\right)^{-1}$. For long time-evolutions, one can split the time interval $t$ into $k$ intervals, called Trotter steps, of duration $t/k$ and approximate the time-evolution in each interval with a $\\chi$ order product formula $S_{\\chi}$. Thus, the PF of order $\\chi$ for time-evolution operator over $k$ Trotter steps is:\n",
+    "where $\\langle A \\rangle_{k_j}(t)$ is the expectation value of an observable $A$ at time $t$ estimated from a Trotter circuit with $k_j$ steps, and the coefficients $\\{x_j\\}_{j=1}^r$ are chosen so that the leading Trotter-error terms in the combination cancel. We will revisit this expression in [Step 4](#step-4-post-process-and-return-result-in-desired-classical-format), where we evaluate it explicitly to combine our Trotter results. The key practical point is that the deepest circuit in the MPF only needs $k_{\\max}$ steps, which is much smaller than the single $k$ that would be required to reach the same effective Trotter error directly. The shallower circuits make the MPF approach better suited to noisy hardware.\n",
```
abbycross: Which step 4 should this link to, since we have two headers with this same title?

Also - should the step names ever change, I suggest future-proofing it with this syntax on the line right before the relevant step 4:
`<span id="step-4-or-whatever-anchor-name-you-want"></span>`
and then you can use `[Step 4](#step-4-or-whatever-anchor-name-you-want)` to call it.

henryzou50: Oh this should link to step 4 of the small scale example. I just pushed a change for that and did your suggestion! That is good to know and thanks!

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -2,153 +2,262 @@
...
-   "id": "d08a5e51",
+   "id": "c4d0b2f2",
+   "metadata": {},
+   "source": [
+    "## Learning outcomes\n",
+    "After completing this tutorial, you can expect to understand the following information:\n",
+    "- How multi-product formulas (MPFs) reduce Trotter error in Hamiltonian simulation by combining expectation values from multiple shallow circuits\n",
+    "- When MPFs are beneficial over standard product formulas and when they are not the right tool\n",
+    "- How to compute static and dynamic MPF coefficients using the `qiskit_addon_mpf` package\n",
+    "- How to execute an MPF workflow end-to-end on IBM Quantum hardware, including transpilation, error mitigation, and post-processing"
```
abbycross:
```suggestion
    "- How to execute an MPF workflow end-to-end on IBM Quantum&reg; hardware, including transpilation, error mitigation, and post-processing"
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -2,153 +2,262 @@
...
     "$$\n",
     "\n",
-    "$\\rho_k(t)$ is an approximation for $\\rho(t)$ with the Trotter approximation error ||$\\rho_k(t)-\\rho(t) ||$. If we consider a linear combination of Trotter approximations of $\\rho(t)$:\n",
+    "where the integer exponents $\\{\\eta_n\\}$ are the orders of the successive Trotter-error terms for the chosen product formula. For a *symmetric* $2\\chi$-order PF, the leading error in $\\left[S_{2\\chi}(t/k)\\right]^k$ scales as $1/k^{2\\chi}$, with subsequent corrections at $1/k^{2\\chi+2}, 1/k^{2\\chi+4}, \\ldots$ &mdash; so the exponents are $\\eta_n = 2\\chi + 2n$. For non-symmetric PFs, both odd and even powers contribute and $\\eta_n = 2\\chi + n$. See Ref. [\\[1\\]](#references) for the full derivation. The first equation in the system above ensures unbiasedness (the MPF reproduces the exact expectation value in the $k_j \\to \\infty$ limit), and the remaining $r-1$ equations successively cancel the first $r-1$ Trotter-error terms. When the resulting $L_1$-norm $\\|x\\|_1$ is too large (which amplifies sampling noise), one can instead solve an approximate optimization that caps $\\|x\\|_1$ while minimizing $\\|Ax - b\\|$.\n",
```
abbycross:
```suggestion
    "where the integer exponents $\\{\\eta_n\\}$ are the orders of the successive Trotter-error terms for the chosen product formula. For a *symmetric* $2\\chi$-order PF, the leading error in $\\left[S_{2\\chi}(t/k)\\right]^k$ scales as $1/k^{2\\chi}$, with subsequent corrections at $1/k^{2\\chi+2}, 1/k^{2\\chi+4}, \\ldots$ &mdash; so the exponents are $\\eta_n = 2\\chi + 2n$. For non-symmetric PFs, both odd and even powers contribute and $\\eta_n = 2\\chi + n$. See Ref. [\\[1\\]](#references) for the full derivation. The first equation in the system above ensures unbiasedness (the MPF reproduces the exact expectation value in the $k_j \\to \\infty$ limit), and the remaining $r-1$ equations successively cancel the first $r-1$ Trotter-error terms. When the resulting $L_1$-norm $\\|x\\|_1$ is too large (which amplifies sampling noise), you can instead solve an approximate optimization that caps $\\|x\\|_1$ while minimizing $\\|Ax - b\\|$.\n",
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -2,153 +2,262 @@
...
-    "* $k_j$ determines both the circuit depth and level of Trotter approximation. Smaller values of $k_j$ lead to shorter circuits, which incur fewer circuit errors but will be a less accurate approximation to the desired state.\n",
+    "### When to use MPFs\n",
     "\n",
+    "MPFs are most beneficial when:\n",
     "\n",
-    "The key here is that the remaining Trotter error given by $\\mu(t)$ is smaller than the Trotter error that one would obtain by simply using the largest $k_j$ value.\n",
+    "- **Circuit depth is the bottleneck.** If hardware noise limits the depth you can run, MPFs let you achieve higher effective Trotter accuracy from shallower circuits.\n",
+    "- **You need accurate expectation values, not full state preparation.** MPFs operate at the level of expectation values &mdash; they combine classical numbers, not quantum states. This makes them ideal for observable estimation via the Estimator primitive.\n",
+    "- **You combine a modest number of Trotter-step counts.** Typically combining $r = 3$&ndash;$5$ different step counts $k_j$ is sufficient to cancel several leading Trotter-error terms while keeping $\\|x\\|_1$ manageable.\n",
```
abbycross:
```suggestion
    "- **Circuit depth is the bottleneck.** If hardware noise limits the depth you can run, use MPFs to achieve higher effective Trotter accuracy from shallower circuits.\n",
    "- **You need accurate expectation values, not full state preparation.** MPFs operate at the level of expectation values &mdash; they combine classical numbers, not quantum states. They are therefore ideal for observable estimation when using the Estimator primitive.\n",
    "- **You combine a modest number of Trotter-step counts.** Typically combining $r = 3$&ndash;$5$ different step counts $k_j$ is sufficient to cancel several leading Trotter-error terms while keeping $\\|x\\|_1$ manageable.\n",
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -2,153 +2,262 @@
...
     "\n",
-    "You can view the usefulness of this from two perspectives:\n",
+    "### When MPFs may not help\n",
```
abbycross:
```suggestion
    "### When MPFs might not help\n",
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -2,153 +2,262 @@
...
     "\n",
-    "1. For a fixed budget of Trotter steps that you can execute, you can obtain results with a Trotter error that is smaller in total.\n",
-    "2. Given some target number of Trotter steps that is too large to execute, you can use the MPF to find a collection of lower-depth circuits to run that results in a similar Trotter error."
+    "- **Very short evolution times.** When $t$ is small enough that a single low-order Trotter formula is already accurate, the overhead of running multiple circuits is unnecessary.\n",
+    "- **State-preparation tasks.** MPFs produce a corrected *expectation value*, not a corrected quantum state. If you need the actual time-evolved state (e.g., as input to another quantum subroutine), MPFs do not apply.\n",
```
abbycross:
```suggestion
    "- **State-preparation tasks.** MPFs produce a corrected *expectation value*, not a corrected quantum state. If you need the actual time-evolved state (for example, as input to another quantum subroutine), MPFs do not apply.\n",
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -2,153 +2,262 @@
...
+    "- **Trotter-step counts that violate the convergence regime.** The static-coefficient derivation expands each individual $\\left[S_{2\\chi}(t/k_j)\\right]^{k_j}$ as a series in $t/k_j$; this expansion only converges well when $t/k_{\\min} \\lesssim 1$. If $k_{\\min}$ is chosen too small for the given $t$, the shallowest circuit is far outside the perturbative regime, the higher-order error terms that the MPF leaves uncanceled become large, and the cancellation can require large coefficients. The $L_1$-norm $\\|x\\|_1$ is the practical diagnostic: when $\\|x\\|_1 \\gg 1$, the sampling overhead $\\propto \\|x\\|_1^2$ may outweigh the Trotter-error reduction. See the [guide on choosing Trotter steps](https://qiskit.github.io/qiskit-addon-mpf/how_tos/choose_trotter_steps.html) for details.\n",
```
abbycross:
```suggestion
    "- **Trotter-step counts that violate the convergence regime.** The static-coefficient derivation expands each individual $\\left[S_{2\\chi}(t/k_j)\\right]^{k_j}$ as a series in $t/k_j$; this expansion only converges well when $t/k_{\\min} \\lesssim 1$. If $k_{\\min}$ is chosen too small for the given $t$, the shallowest circuit is far outside the perturbative regime, the higher-order error terms that the MPF leaves uncanceled become large, and the cancellation can require large coefficients. The $L_1$-norm $\\|x\\|_1$ is the practical diagnostic: when $\\|x\\|_1 \\gg 1$, the sampling overhead $\\propto \\|x\\|_1^2$ might outweigh the Trotter-error reduction. See the [guide on choosing Trotter steps](https://qiskit.github.io/qiskit-addon-mpf/how_tos/choose_trotter_steps.html) for details.\n",
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -210,89 +301,46 @@
...
   {
    "cell_type": "markdown",
-   "id": "46470a22",
+   "id": "1512ba0e",
    "metadata": {},
    "source": [
-    "Then we create the circuits implementing the approximate Trotter time-evolutions."
+    "#### Build Trotter circuits\n",
+    "\n",
+    "We create the circuits implementing the approximate Trotter time-evolutions for each time point and each Trotter step count. The `CollectAndCollapse` pass defined in Setup collects XX and YY rotations into single XX+YY gates, enabling more efficient tensor-network simulation later."
```
abbycross:
```suggestion
    "We create the circuits implementing the approximate Trotter time-evolutions for each time point and each Trotter step count. The `CollectAndCollapse` pass defined in the Setup section collects XX and YY rotations into single XX+YY gates, to prepare for more efficient tensor-network simulation later."
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -342,66 +390,84 @@
...
    "cell_type": "markdown",
-   "id": "b84e4000",
+   "id": "a384f017",
    "metadata": {},
    "source": [
-    "We also calculate the exact expectation values for comparison."
+    "<span id=\"small-scale-step-4\"></span>\n",
+    "### Step 4: Post-process and return result in desired classical format\n",
+    "\n",
+    "Step 4 is where the MPF is actually constructed. Even though the coefficients $x_j$ are *computed* here (and, for the dynamic variant, this computation can be intensive), conceptually they are a classical recipe for combining the quantum measurements from Step 3 into a single corrected expectation value &mdash; so we treat the whole coefficient + combination workflow as post-processing.\n",
```
abbycross:
```suggestion
    "Step 4 is where the MPF is actually constructed. Even though the coefficients $x_j$ are *computed* here (and, for the dynamic variant, this computation can be intensive), conceptually they are a classical recipe for combining the quantum measurements from Step 3 into a single corrected expectation value &mdash; so we treat the whole coefficient and combination workflow as post-processing.\n",
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -416,132 +482,67 @@
...
     "$$\n",
     "\n",
-    "\n",
-    "\n",
-    "For more details, refer to the documentation of the Linear System of Equations ([LSE](https://qiskit.github.io/qiskit-addon-mpf/stubs/qiskit_addon_mpf.static.LSE.html)).\n",
-    "\n",
-    "We can find a solution for $x$ analytically as $x = A^{-1}b$; see for example Refs. [\\[1\\]](#references) or [\\[2\\]](#references).\n",
-    "However, this exact solution can be \"ill-conditioned\", resulting in very large L1-norms of our coefficients, $x$, which can lead to bad performance of the MPF.\n",
-    "Instead, one can also obtain an approximate solution that minimizes the L1-norm of $x$ in order to attempt to optimize the MPF behavior."
+    "The first row enforces unbiasedness ($\\sum_j x_j = 1$); the second and third rows cancel the leading $1/k^2$ and next-order $1/k^3$ Trotter-error terms respectively."
```
abbycross:
```suggestion
    "The first row enforces unbiasedness ($\\sum_j x_j = 1$); the second and third rows cancel the leading $1/k^2$ and next-order $1/k^3$ Trotter-error terms, respectively."
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -416,132 +482,67 @@
...
    "source": [
     "##### Set up the LSE\n",
     "\n",
-    "Now that we have chosen our $k_j$ values, we must first construct the LSE, $Ax=b$ as explained above.\n",
-    "The matrix $A$ depends not only on $k_j$ but also our choice of PF, in particular its _order_.\n",
-    "Additionally, you might take into account whether the PF is symmetric or not (see [\\[1\\]](#references)) by setting `symmetric=True/False`.\n",
-    "However, this is not required, as shown by Ref. [\\[2\\]](#references)."
+    "We use `setup_static_lse` from `qiskit_addon_mpf.static` to assemble the matrix $A$ and right-hand-side vector $b$ described above. The matrix $A$ depends not only on $k_j$ but also on our choice of product formula &mdash; in particular its *order* $\\chi$ and whether it is *symmetric*. The `symmetric` flag controls the exponent pattern $\\eta_n$ (symmetric formulas only produce even-power Trotter-error terms; see Ref. [\\[1\\]](#references)). Note that, as shown in Ref. [\\[2\\]](#references), setting `symmetric=True` is not strictly necessary even when the underlying PF is symmetric &mdash; the non-symmetric LSE remains valid (it just enforces extra unneeded constraints).\n",
```
abbycross:
```suggestion
    "We use `setup_static_lse` from `qiskit_addon_mpf.static` to assemble the matrix $A$ and right-hand-side vector $b$ described above. The matrix $A$ depends not only on $k_j$ but also on our choice of product formula &mdash; in particular its *order* $\\chi$ and whether it is *symmetric*. The `symmetric` flag controls the exponent pattern $\\eta_n$ (symmetric formulas only produce even-power Trotter-error terms; see Ref. [\\[1\\]](#references)). Note that, as shown in Ref. [\\[2\\]](#references), setting `symmetric=True` is not strictly necessary even when the underlying PF is symmetric &mdash; the non-symmetric LSE remains valid (it enforces extra unneeded constraints).\n",
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -416,132 +482,67 @@
...
+    "\n",
+    "For our example we already set `order = 2` and `symmetric = False` in Step 1."
```
abbycross:
```suggestion
    "For our example we have already set `order = 2` and `symmetric = False` in Step 1."
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -1706,164 +1041,69 @@
...
-      "relative error with MPF dynamic coeffs 0.06645896853467514\n"
-     ]
-    }
-   ],
-   "source": [
-    "def relative_error(ev, exact_ev):\n",
-    "    return abs(ev - exact_ev)\n",
+    "The plot above illustrates the interplay between Trotter error and sampling error.\n",
     "\n",
+    "- **Trotter error.** The individual product formulas (grey markers) deviate from the exact curve more and more as time grows, with the largest deviation for $k=1$ &mdash; that circuit is the shallowest, but it is also already in the regime where $t/k \\gtrsim 1$, so the leading $1/k^{2}$ error term is large. The MPF combinations (colored markers) cancel several of these leading Trotter-error terms, so they track the exact curve much more closely than any single $k_j$ circuit. The remaining gap reflects the higher-order Trotter terms that the MPF does *not* cancel: an order-$2$, $r=3$ static MPF only kills the first two error orders, and at large $t/k_{\\min}$ the uncancelled tail eventually dominates &mdash; so MPF is not a free pass to use very shallow circuits at arbitrary times.\n",
```
abbycross:
```suggestion
    "- **Trotter error.** The individual product formulas (grey markers) deviate from the exact curve more and more as time grows. The $k=1$ circuit has the largest deviation and is the shallowest, but it is also already in the regime where $t/k \\gtrsim 1$, so the leading $1/k^{2}$ error term is large. The MPF combinations (colored markers) cancel several of these leading Trotter-error terms, so they track the exact curve much more closely than any single $k_j$ circuit. The remaining gap reflects the higher-order Trotter terms that the MPF does *not* cancel: an order-$2$, $r=3$ static MPF only kills the first two error orders, and at large $t/k_{\\min}$ the uncanceled tail eventually dominates &mdash; so MPF does not necessarily invite using very shallow circuits at arbitrary times.\n",
```

abbycross: With translations in mind, I tried rewriting the last sentence to avoid colloquialisms, but perhaps you can come up with something better?

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -1706,164 +1041,69 @@
...
     "\n",
-    "relative_error_k = [relative_error(ev, exact_obs.real) for ev in evs_exp]\n",
-    "relative_error_mpf = relative_error(evs_exp @ mpf_coeffs, exact_obs.real)\n",
-    "relative_error_approx_mpf = relative_error(\n",
-    "    evs_exp @ coeffs_approx.value, exact_obs.real\n",
-    ")\n",
-    "relative_error_dynamic_mpf = relative_error(\n",
-    "    evs_exp @ mpf_dynamic_coeffs_list[7], exact_obs.real\n",
-    ")\n",
+    "- **Sampling error.** The wider error bars on the MPF curves are a direct consequence of the linear combination: propagating independent per-circuit standard errors $\\sigma_{k_j}$ gives a total variance $\\sigma_{\\text{MPF}}^2 = \\sum_j x_j^2 \\, \\sigma_{k_j}^2$. So the larger the $\\|x\\|_2$ (and in practice the $\\|x\\|_1$, which is what we control), the more shots are required to reach a given target uncertainty. This is the trade-off behind the approximate-solver option in the Background: we cap $\\|x\\|_1$ to keep this overhead manageable. Crucially, unlike Trotter error, sampling error shrinks with $1/\\sqrt{N_{\\text{shots}}}$, so it can always be reduced by spending more shots.\n",
```
abbycross:
```suggestion
    "- **Sampling error.** The wider error bars on the MPF curves are a direct consequence of the linear combination: propagating independent per-circuit standard errors $\\sigma_{k_j}$ gives a total variance $\\sigma_{\\text{MPF}}^2 = \\sum_j x_j^2 \\, \\sigma_{k_j}^2$. Therefore, the larger the $\\|x\\|_2$ (and in practice the $\\|x\\|_1$, which is what we control), the more shots are required to reach a given target uncertainty. This is the trade-off behind the approximate-solver option in the Background: we cap $\\|x\\|_1$ to keep this overhead manageable. Crucially, unlike Trotter error, sampling error shrinks with $1/\\sqrt{N_{\\text{shots}}}$, so it can always be reduced by spending more shots.\n",
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -1706,164 +1041,69 @@
...
    "cell_type": "markdown",
-   "id": "0130299b",
+   "id": "6fa763ff",
    "metadata": {},
    "source": [
-    "## Part II: scale it up\n",
+    "## Large-scale hardware example\n",
     "\n",
-    "Let's scale the problem up beyond what is possible to simulate exactly. In this section we will focus on reproducing some of the results shown in Ref. [\\[3\\]](#references)."
+    "Let's scale the problem up beyond what is possible to simulate exactly. In this section we reproduce some of the results shown in Ref. [\\[3\\]](#references), using a 50-qubit XXZ chain at time $t = 3$. We use the same four-step workflow as above, but pull out the hardware-specific details (qubit filtering, transpilation, and error mitigation) into their own subsections after Step 1."
```
abbycross:
```suggestion
    "In this section we scale the problem up beyond what is possible to simulate exactly. We reproduce some of the results shown in Ref. [\\[3\\]](#references), using a 50-qubit XXZ chain at time $t = 3$. We use the same four-step workflow as above, but pull out the hardware-specific details (qubit filtering, transpilation, and error mitigation) into their own subsections after Step 1."
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -1706,164 +1041,69 @@
...
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
-   "source": [
-    "L = 50\n",
-    "# Generate some coupling map to use for this example\n",
-    "coupling_map = CouplingMap.from_line(L, bidirectional=False)\n",
-    "graphviz_draw(coupling_map.graph, method=\"circo\")"
+    "Note that even though we are still in Step 1 here (mapping + circuit construction), we also pre-compute the dynamic coefficients alongside the static ones in this cell. Dynamic coefficients depend on $H$ and $t$ but not on the quantum measurements, so they can be computed any time before Step 4. We do it now to keep all the MPF-specific setup in one place."
```
abbycross:
```suggestion
    "Note that even though we are still in Step 1 here (mapping and circuit construction), we also pre-compute the dynamic coefficients alongside the static ones in this cell. Dynamic coefficients depend on $H$ and $t$ but not on the quantum measurements, so they can be computed any time before Step 4. We do it now to keep all the MPF-specific setup in one place."
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -2516,28 +1610,49 @@
...
+    "A few observations about the hardware results above:\n",
+    "\n",
+    "- **Coefficient norm matters more than mathematical optimality.** The exact-static MPF has $\\|x\\|_1 = 5.63$, and its final estimate ends up far from the true value &mdash; that large coefficient norm amplifies the residual gate noise, decoherence, and ZNE error on each $\\langle A \\rangle_{k_j}$ by roughly the same factor, swamping the Trotter-error cancellation we gained. The approximate-static MPF (capped at $\\|x\\|_1 = 2$) and the dynamic MPF (whose coefficients here are similarly modest in magnitude) avoid this blow-up and land much closer to the reference.\n",
```
abbycross:
```suggestion
    "- **Coefficient norm matters more than mathematical optimality.** The exact-static MPF has $\\|x\\|_1 = 5.63$, and its final estimate ends up far from the true value &mdash; the large coefficient norm amplifies the residual gate noise, decoherence, and ZNE error on each $\\langle A \\rangle_{k_j}$ by roughly the same factor, overwhelming the Trotter-error cancellation we gained. The approximate-static MPF (capped at $\\|x\\|_1 = 2$) and the dynamic MPF (whose coefficients here are similarly modest in magnitude) avoid this overwhelm and land much closer to the reference.\n",
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -2516,28 +1610,49 @@
...
+    "\n",
+    "- **The dynamic MPF is the best estimator in this run.** Tailoring the coefficients to this specific Hamiltonian, initial state, and time gives an answer closer to the exact reference than any single Trotter circuit &mdash; including the deeper $k = 6$ baseline &mdash; and substantially closer than either static MPF. This is the regime MPFs are designed for: shallow constituent circuits combined with well-chosen, small-norm coefficients.\n",
```
abbycross:
```suggestion
    "- **The dynamic MPF is the best estimator in this run.** Tailoring the coefficients to this specific Hamiltonian, initial state, and time gives an answer closer to the exact reference than any single Trotter circuit &mdash; including the deeper $k = 6$ baseline &mdash; and is substantially closer than either static MPF. This is the regime MPFs are designed for: shallow constituent circuits combined with well-chosen, small-norm coefficients.\n",
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -2516,28 +1610,49 @@
...
     "\n",
-    "Overall, the approximate static coefficients still give a more accurate solution than the product formula with higher number of Trotter steps with the same amount of Trotter error in the noiseless setting.\n",
+    "The practical takeaway is that on hardware, MPFs should be paired with strong error mitigation on each individual $\\langle A \\rangle_{k_j}$, the coefficient $L_1$-norm should be kept modest (use the approximate solver, or prefer the dynamic MPF when its tensor-network setup is tractable), and the Trotter steps $k_j$ should be chosen so that $t/k_{\\min} \\lesssim 1$. With those choices, as we see here for the dynamic MPF, the MPF can recover the depth-vs-accuracy advantage shown in Ref. [\\[3\\]](#references). Note also that individual runs are noisy &mdash; on a different submission of the same job (or a different backend), the relative ordering of these estimators can shift; the qualitative trend that small-$\\|x\\|_1$ MPFs do well and large-$\\|x\\|_1$ MPFs are amplified by hardware noise is the robust one."
```
abbycross:
```suggestion
    "The practical takeaway is that on hardware, MPFs should be paired with strong error mitigation on each individual $\\langle A \\rangle_{k_j}$, the coefficient $L_1$-norm should be kept modest (use the approximate solver, or prefer the dynamic MPF when its tensor-network setup is tractable), and the Trotter steps $k_j$ should be chosen so that $t/k_{\\min} \\lesssim 1$. With those choices, as we see here for the dynamic MPF, the MPF can recover the depth-versus-accuracy advantage shown in Ref. [\\[3\\]](#references). Note also that individual runs are noisy &mdash; on a different submission of the same job (or a different backend), the relative ordering of these estimators can shift; the qualitative trend that small-$\\|x\\|_1$ MPFs do well and large-$\\|x\\|_1$ MPFs are amplified by hardware noise is the robust one."
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -2516,28 +1610,49 @@
...
+   "metadata": {},
+   "source": [
+    "## Next steps\n",
+    "<Admonition type=\"tip\" title=\"Recommendations\">\n",
     "\n",
-    "It is also important to note that in the example that reproduces the experiment in Ref. [\\[3\\]](#references), the time point $t=3$ is beyond the limit at which it is expected that the PF with $k=2$ behaves well, which is $t/k>1$ as discussed in this [guide](https://qiskit.github.io/qiskit-addon-mpf/how_tos/choose_trotter_steps.html)."
+    "If you found this work interesting, you might be interested in the following material:\n",
+    "- [How to choose the Trotter steps for an MPF](https://qiskit.github.io/qiskit-addon-mpf/how_tos/choose_trotter_steps.html) — practical guidance on selecting $k_j$ values to avoid instabilities\n",
+    "- [How to use the approximate model](https://qiskit.github.io/qiskit-addon-mpf/how_tos/using_approximate_model.html) — tuning the $L_1$-norm constraint and solver options for the approximate static MPF\n",
+    "- [qiskit-addon-mpf API reference](https://qiskit.github.io/qiskit-addon-mpf/) — full documentation for static, dynamic, and backend modules\n",
```
abbycross:
```suggestion
    "- [`qiskit-addon-mpf` API reference](https://qiskit.github.io/qiskit-addon-mpf/) — full documentation for static, dynamic, and backend modules\n",
```

abbycross:
I've finished a copyedit sweep ✅

docs/tutorials/multi-product-formula.ipynb
henryzou50: Good point, I went with: "so MPF does not guarantee that very shallow circuits remain accurate at arbitrary times." This should read a bit more naturally than "invite using" and should translate cleanly. Also fixed a stray double-backslash in \gtrsim that snuck in. Let me know if you'd prefer different phrasing!

abbycross: Sounds much better, thanks @henryzou50!

abbycross (dismissed):
Thanks!

nathanearnestnoble (dismissed):
I think this tutorial looks good to publish!   I am curious for us to evaluate how necessary the backend filtering is for performance and whether we should apply this principle to all of the large scale hardware demos (+ updating backend characterization).

henryzou50:
In the latest commit, I made a few revisions to the large-scale hardware example:

- **Removed the backend coupling-map filtering.** I tested running without the `filter_coupling_map` step (and the adaptive threshold loop) and got similar results across multiple runs, so it doesn't appear to be necessary here. For that reason I think it's better to remove it for now and keep the workflow simplified, we just use the default preset pass manager instead.
- **Aligned the section with the tutorial template** — replaced the repeated `### Step N:` subheaders with inline `# ---- Step N ----` markers in the code cells, matching the other recently updated tutorials.
- **Adjusted the Trotter steps** to `k_j = [3, 4, 6]` (so `t/k_min = 1`) with a deeper `k = 10` baseline, and updated the results commentary to match the reran outputs.

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -342,66 +390,84 @@
...
    "metadata": {},
    "source": [
-    "Next, we calculate the time-evolved expectation values from the Trotter circuits."
+    "### Step 2: Optimize problem for quantum hardware execution\n",
+    "\n",
+    "For the small-scale example we target the Aer simulator. Two transformations happen before the circuits are ready to execute:\n",
+    "\n",
+    "1. **Gate collection at the Hamiltonian-simulation level.** In the Setup cell we built a `CollectAndCollapse` pass that fuses adjacent `rxx` and `ryy` rotations into a single `XXPlusYYGate`. We already applied this pass when we built the Trotter circuits in Step 1 (the `pm.run(...)` call). This both reduces the two-qubit gate count and produces a structure that is more amenable to tensor-network simulation for the dynamic-coefficient computation later.\n",
+    "\n",
+    "2. **Lowering to the simulator's ISA.** Below we run Qiskit's preset pass manager at `optimization_level=3` to lower each Trotter circuit to the simulator's instruction-set architecture (ISA)."
```
abbycross:
```suggestion
    "2. **Lowering to the simulator's ISA.** Below we run the Qiskit preset pass manager at `optimization_level=3` to lower each Trotter circuit to the simulator's instruction-set architecture (ISA)."
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -2080,239 +1188,113 @@
...
    ]
   },
   {
    "cell_type": "markdown",
-   "id": "25f03e39",
+   "id": "b5d388b1",
    "metadata": {},
    "source": [
-    "### Step 2: Optimize problem for quantum hardware execution"
+    "We now optimize the circuits for the chosen backend. We use Qiskit's preset pass manager at `optimization_level=3`, which automatically selects a good set of physical qubits and routes each circuit onto the device topology."
```
abbycross:
```suggestion
    "We now optimize the circuits for the chosen backend. We use the Qiskit preset pass manager at `optimization_level=3`, which automatically selects a good set of physical qubits and routes each circuit onto the device topology."
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -2516,28 +1495,49 @@
...
+    "A few observations about the hardware results above:\n",
+    "\n",
+    "- **Going deeper is not free on hardware.** The single-circuit baselines tell the story directly: the $k = 6$ circuit is essentially exact ($-0.256$ vs. the reference $-0.244$), yet the deeper $k = 10$ baseline is *worse* ($-0.061$, off by $\\sim 0.18$), not better. Once the Trotter error is already small, adding steps mostly deepens the circuit and accumulates more gate noise and decoherence. This is precisely the regime MPFs are built for: reach the accuracy of a deep circuit using only shallow constituents.\n",
```
abbycross:
```suggestion
    "- **Going deeper is not free on hardware.** The single-circuit baselines tell the story directly: the $k = 6$ circuit is essentially exact ($-0.256$ versus the reference $-0.244$), yet the deeper $k = 10$ baseline is *worse* ($-0.061$, off by $\\sim 0.18$), not better. Once the Trotter error is already small, adding steps mostly deepens the circuit and accumulates more gate noise and decoherence. This is precisely the regime MPFs are built for: reach the accuracy of a deep circuit using only shallow constituents.\n",
```

docs/tutorials/multi-product-formula.ipynb
```diff
@@ -1706,164 +1041,67 @@
...
-    "$$\n",
-    "\\hat{\\mathcal{H}}_{XXZ} = \\sum_{i=1}^{L-1} J_{i,(i+1)}\\left(X_i X_{(i+1)}+Y_i Y_{(i+1)}+ 2\\cdot Z_i Z_{(i+1)} \\right) \\, ,\n",
-    "$$\n",
+    "- An **XXZ Hamiltonian** on 50 sites with random couplings drawn from $\\mathcal{U}(0.5, 1.5)$ (Ref. [\\[3\\]](#references)).\n",
+    "- A **symmetric** second-order Trotter formula with $k_j = [3, 4, 6]$ (so $\\chi=1$, `symmetric=True`).\n",
+    "- A single fixed evolution time $t = 3$. With $k_{\\min}=3$ this gives $t/k_{\\min}=1$, keeping the shallow constituents inside the Trotter-convergence regime where the leading-error model the MPF relies on is valid.\n",
+    "- An additional **single-circuit comparison run with $k = 10$ Trotter steps**, used as a baseline. We chose $k = 10$ because its two-qubit depth on hardware is deeper than the deepest MPF constituent ($k_{\\max}=6$) plus the overhead of running multiple MPF circuits &mdash; deep enough to be noise-limited, which is the regime in which the MPF combination is meant to win. It is a \"single deep circuit\" comparison against the MPF combination, not a circuit that targets the MPF's effective Trotter error (which would require many more steps).\n",
```
abbycross: Curious about the language "the regime in which it is meant to win" - is that standard to use "win" instead of something like "succeed"? I'm imagining it's a technical field-specific turn of phrase?

henryzou50: People in the field do say a method "wins" to mean it beats the baseline, but can be seen as not standard terminology and could read as too casual here. I've changed it to "expected to outperform the single-circuit baseline," which says exactly what we mean. Updated the matching code comment too.

henryzou50:
Thanks for the changes and review @abbycross, I push a change to address your latest question. Let me know if there is anything else to change!

abbycross (approved):
Thanks!!

PR #5218, opened 2026-06-04
Add PEC with SLC tutorial

henryzou50:
## Summary

Adds a new tutorial, **"Probabilistic error cancellation with shaded lightcones."** It is an updated version of the SLC-PEC tutorial from the QDC Challenges 2025 repo: https://github.com/qiskit-community/qdc-challenges-2025/blob/main/day3_tutorials/Track_A/slc_pec/qdc_tutorial_slc.ipynb

Probabilistic error cancellation (PEC) produces unbiased expectation values but at a sampling overhead (γ²) that grows exponentially with circuit noise. Shaded lightcones (SLC) bound each noise term's contribution to the observable so the mitigation budget is spent only where it matters, trading a small bounded residual bias for a large reduction in overhead. The tutorial implements the full PEC+SLC workflow on a 20-qubit mirrored Ising circuit (observable ⟨X₆Z₁₃⟩) using `qiskit-addon-slc`, `samplomatic`, `NoiseLearnerV3`, and the `QuantumProgram`/`Executor` classes in Qiskit Runtime, following the four-step Qiskit pattern, and compares full PEC, PEC+SLC, and the unmitigated baseline.

## Major changes from the original

- **Updated to released Qiskit Runtime (0.47.0).** Migrated off the `executor_preview` branch. `Executor`, `QuantumProgram`, and `NoiseLearnerV3` are imported from the top-level `qiskit_ibm_runtime`; noise-learner options are passed as a plain dict (no `NoiseLearnerV3Options`). `qiskit-addon-slc` 0.1.0 is already the latest release, so its pin is kept.
- **Removed hardcoded credentials.** Replaced the `shared_service` (token + `instance` CRN) and the named `QiskitRuntimeService(name="qdc-2025")` with a plain `QiskitRuntimeService()` that reads saved credentials.
- **Runs live, no saved jobs.** Removed all saved-job loading (`shared_service.job(...)` IDs and the `load_cached_*` branches); the noise learner and `Executor` execute live, with the noise-learning job tagged `TUT_SLC`.
- **Noise-learning depth tuned.** The deepest NLv3 `layer_pair_depths` value is set to half the circuit's two-qubit-layer depth, so the total learned depth matches the depth of the circuit being mitigated (computed programmatically from the boxed circuit). This noticeably improves the learned model.
- **Dropped the cluster-resources demonstration.** Removed the committed `merged_bounds_cluster.pickle` load; `local_scales` and the execution now use the live, laptop-computed `merged_bounds`, and the sampling-overhead plot compares full PEC vs PEC+SLC only (with autoscaling y-limits).
- **Cleanup.** Removed `logging.basicConfig` INFO logging (kept the multiprocessing `set_start_method` / `OMP_NUM_THREADS` setup SLC requires), stripped INFO-log and local-path stream outputs, and converted in-page anchor links to plain references.
- **Reformatted to the tutorial template.** Added title/description frontmatter, learning outcomes, prerequisites, adapted background (relative doc links, committed AVIF figures), requirements, setup, and next steps. The small-scale simulator section is kept but explains why simulation is skipped (SLC mitigates *learned hardware* noise).

## Files

- New: `docs/tutorials/pec-with-shaded-lightcones.ipynb` + images under `public/docs/images/tutorials/pec-with-shaded-lightcones/` (3 background schematics and the extracted cell outputs).
- Registered in `docs/tutorials/_toc.json` and `docs/tutorials/index.mdx` (Error mitigation section), added to the notebook-CI exclude list in `scripts/config/notebook-testing.toml`, and given an owner in `qiskit_bot.yaml`.

## Results / testing

Ran on IBM Quantum hardware (`ibm_boston`), `tox -e lint`, the notebook normalizer, spelling, and internal-link checks all pass.


docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+    "---\n",
+    "title: Probabilistic error cancellation with shaded lightcones\n",
+    "description: Use probabilistic error cancellation (PEC) with shaded lightcones to cut PEC sampling overhead and mitigate errors in expectation value estimation on hardware\n",
+    "---\n",
+    "\n",
+    "{/* cspell:ignore samplomatic samplex Trotterized Trotter Lindblad lightcone lightcones Eddins reverser broadcastable postselection postselect antinoise eigval atol starmap unmitigated NoiseLearner PauliLindblad QubitSparsePauli Kingston kicked qubit canonical isa plm qspl Rzz pauli TREX levelname xticks expvals markerfacecolor ncols */}\n",
+    "\n",
+    "# Probabilistic error cancellation with shaded lightcones\n",
+    "\n",
+    "*Usage estimate: 10 minutes on an IBM Heron processor (NOTE: This is an estimate only. Your runtime might vary.)*"
```
abbycross:
```suggestion
    "*Usage estimate: 10 minutes on a Heron processor (NOTE: This is an estimate only. Your runtime might vary.)*"
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "slc-lout-0000-4000-8000-000000000001",
+   "metadata": {},
+   "source": [
+    "## Learning outcomes\n",
+    "After going through this tutorial, users should understand:\n",
+    "- What probabilistic error cancellation (PEC) is, and why its sampling overhead $\\gamma^2$ grows exponentially with the total noise acting on the circuit\n",
+    "- How shaded lightcones (SLC) bound each noise term's contribution to the target observable, letting you spend the mitigation budget where it matters and trade a bounded residual bias for lower sampling overhead\n",
```
abbycross:
```suggestion
    "- How shaded lightcones (SLC) bound each noise term's contribution to the target observable, so that you can spend the mitigation budget where it matters and trade a bounded residual bias for lower sampling overhead\n",
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+    "- How to learn layer noise with `NoiseLearnerV3` and inject anti-noise through `samplomatic` and the `Executor`\n",
+    "- How to combine PEC and PEC+SLC with TREX and post-selection to estimate an expectation value on hardware\n",
```
abbycross:
```suggestion
    "- How to learn layer noise with `NoiseLearnerV3` and inject anti-noise through `samplomatic` and the Executor primitive\n",
    "- How to combine PEC and PEC+SLC with TREX and postselection to estimate an expectation value on hardware\n",
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+    "\n",
+    "## Prerequisites\n",
+    "We suggest that users are familiar with the following topics before going through this tutorial:\n",
+    "- The [Qiskit patterns](/docs/guides/intro-to-patterns) workflow\n",
+    "- Using the [Estimator](/docs/api/qiskit-ibm-runtime/estimator-v2) primitive to calculate expectation values of an observable\n",
```
abbycross:
```suggestion
    "- Using the [Estimator](/docs/guides/get-started-with-estimator) primitive to calculate expectation values of an observable\n",
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+    "\n",
+    "If we can identify errors that do not need to be mitigated, we can decrease this sampling cost exponentially. A first step in this direction is implementing locally aware error mitigation, which uses a quickly computable conventional \"lightcone\" to reduce the PEC overhead by bounding an observable's sensitivity to errors throughout the circuit, extending the feasibility of PEC to larger scales for some problems. Errors outside of this lightcone cannot affect the measured outcome and can therefore be excluded from the error cancellation procedure. This exclusion decreases the sampling overhead, in some cases substantially, without introducing additional bias. In particular, for measuring a local observable $O$ of a fixed-depth circuit, the required sampling overhead eventually plateaus when scaling the number of qubits in the circuit (see Fig. 2b in [Locality and error mitigation of quantum circuits](https://arxiv.org/abs/2303.06496)).\n",
```
abbycross:
```suggestion
    "If we can identify errors that do not need to be mitigated, we can decrease this sampling cost exponentially. A first step in this direction is implementing locally-aware error mitigation, which uses a quickly computable conventional \"lightcone\" to reduce the PEC overhead by bounding an observable's sensitivity to errors throughout the circuit, extending the feasibility of PEC to larger scales for some problems. Errors outside of this lightcone cannot affect the measured outcome and can therefore be excluded from the error cancellation procedure. This exclusion decreases the sampling overhead, in some cases substantially, without introducing additional bias. In particular, for measuring a local observable $O$ of a fixed-depth circuit, the required sampling overhead eventually plateaus when scaling the number of qubits in the circuit (see Fig. 2b in [Locality and error mitigation of quantum circuits](https://arxiv.org/abs/2303.06496)).\n",
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+    "\n",
+    "Shaded lightcones (SLC) go further, using classical simulations to more tightly bound the sensitivity to errors throughout the circuit. This trades some QPU time for CPU time and reduces the sampling overhead needed to renormalize the bias. Instead of a hard cutoff, each potential error in the circuit is assigned a graded \"shade\" that upper-bounds the susceptibility of the observable to that error. This refined characterization allows for more efficient, targeted applications of PEC with reduced variance, while giving the user the ability to controllably tune the bias in the observable estimation. See [Lightcone shading for classically accelerated quantum error mitigation](https://arxiv.org/abs/2409.04401) for more details.\n",
+    "\n",
+    "Our workflow for the SLC addon leverages the [`samplomatic`](https://github.com/Qiskit/samplomatic) library together with the `QuantumProgram` and `Executor` classes added to Qiskit Runtime in `qiskit-ibm-runtime` 0.47.0, allowing users to have more modular control of execution settings for error suppression and mitigation while retaining ease of use.\n",
```
abbycross:
```suggestion
    "Our workflow for the SLC addon leverages the [`samplomatic`](https://qiskit.github.io/samplomatic/guides/index.html) library together with the `QuantumProgram` and `Executor` classes added to Qiskit Runtime in `qiskit-ibm-runtime` v0.47.0, allowing users to have more modular control of execution settings for error suppression and mitigation while retaining ease of use. Read more in the [directed execution model](/docs/guides/directed-execution-model) guide.\n",
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+    "\n",
+    "e. CPU — Estimate observable. Compute the expectation value, applying measurement-based post-selection to reduce non-Markovian noise impact.\n",
+    "\n",
+    "### Noise learning overview\n",
+    "Noise learning is a common step in several error-mitigation methods, carried out by the [noise learner](/docs/guides/noise-learning); it also appears in the [probabilistic error amplification tutorial](/docs/tutorials/probabilistic-error-amplification). In `NoiseLearnerV3`, a user can specifically identify the to-be-learned noise layers as [`CircuitInstruction`](/docs/api/qiskit/qiskit.circuit.CircuitInstruction) objects, which allows users to compute the desired SLC noise bounds for each layer in the manner described above. The learned Pauli-Lindblad model provides coefficients to be used in the PEC+SLC prioritization. The way in which the gates are collected into layers can be determined by using the `generate_boxing_pass_manager` and `find_unique_box_instructions` convenience functions, and then fed into the SLC utility function `generate_noise_model_paulis`, as described in Step 2 below.\n",
```
abbycross:
```suggestion
    "Noise learning is a common step in several error-mitigation methods, carried out by the [noise learner](/docs/guides/noise-learning); it also appears in the [probabilistic error amplification tutorial](/docs/tutorials/probabilistic-error-amplification). In `NoiseLearnerV3`, you can specifically identify the to-be-learned noise layers as [`CircuitInstruction`](/docs/api/qiskit/qiskit.circuit.CircuitInstruction) objects, so that you can compute the desired SLC noise bounds for each layer in the manner described above. The learned Pauli-Lindblad model provides coefficients to be used in the PEC+SLC prioritization. The way in which the gates are collected into layers can be determined by using the `generate_boxing_pass_manager` and `find_unique_box_instructions` convenience functions, and then fed into the SLC utility function `generate_noise_model_paulis`, as described in Step 2 below.\n",
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+   "id": "slc-reqs-0000-4000-8000-000000000003",
+   "metadata": {},
+   "source": [
+    "## Requirements\n",
+    "Before starting this tutorial, be sure you have the following installed:\n",
+    "\n",
+    "- Qiskit SDK v2.2 or later, with [visualization](/docs/api/qiskit/visualization) support (`pip install 'qiskit[visualization]'`)\n",
+    "- Qiskit Runtime v0.47 or later (`pip install qiskit-ibm-runtime`)\n",
+    "- Shaded lightcone Qiskit addon 0.1 or later (`pip install qiskit-addon-slc`)\n",
+    "- Qiskit addon utils 0.3 or later (`pip install 'qiskit-addon-utils'`)\n",
```
abbycross:
```suggestion
    "- Qiskit SDK v2.2 or later, with [visualization](/docs/api/qiskit/visualization) support\n",
    "- Qiskit Runtime v0.47 or later (`pip install qiskit-ibm-runtime`)\n",
    "- Shaded lightcones Qiskit addon v0.1 or later (`pip install qiskit-addon-slc`)\n",
    "- Qiskit addon utils v0.3 or later (`pip install 'qiskit-addon-utils'`)\n",
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+    "\n",
+    "### Step 1: Map the problem\n",
+    "For ease of demonstration, we select a 1D mirror Ising chain. The 1D Ising chain gives a nicely dense circuit structure, which is convenient for showcasing PEC implementations. A mirror circuit makes it straightforward to know the expected outcome (namely, we should measure an observable of 1).\n",
+    "\n",
+    "Because we run a mirror circuit, for every gate in the second half of the circuit there is an inverse gate in the first half. As the measured observable $\\langle X_6 Z_{13}\\rangle$ has non-Z-basis measurements, and the executor accounts for the desired basis at the end of the circuit, we provide a `prepare_basis` function that inserts the appropriate gates at the start of the mirror circuit. This detail is specific to our mirror-circuit demonstration. The `get_measurement_bases` function lets us identify which gates are needed and where to append them, while keeping track of qubit-index subtleties arising from `box` annotation conventions, as discussed in the section on preparing canonical basis measurements."
```
abbycross:
```suggestion
    "Because we run a mirror circuit, for every gate in the second half of the circuit there is an inverse gate in the first half. As the measured observable $\\langle X_6 Z_{13}\\rangle$ has non-Z-basis measurements, and Executor accounts for the desired basis at the end of the circuit, we provide a `prepare_basis` function that inserts the appropriate gates at the start of the mirror circuit. This detail is specific to our mirror-circuit demonstration. We use the `get_measurement_bases` function to identify which gates are needed and where to append them, while keeping track of qubit-index subtleties arising from `box` annotation conventions, as discussed in the section on preparing canonical basis measurements."
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "slc-step2-000-4000-8000-000000000013",
+   "metadata": {},
+   "source": [
+    "### Step 2: Optimize\n",
+    "We will optimize details associated with the circuit to be run, the observable to be measured, and the noise-learning parameters. As a starting point, we instantiate a backend with fractional gates turned on. These fractional gates allow for greater sensitivity in some of our post-selection filtering."
```
abbycross:
```suggestion
    "We optimize details associated with the circuit to be run, the observable to be measured, and the noise-learning parameters. As a starting point, we instantiate a backend with fractional gates turned on. These fractional gates allow for greater sensitivity in some of our post-selection filtering."
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+    "backend = service.backend(backend.name, use_fractional_gates=True)\n",
+    "print(f\"Selected backend: {backend.name}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "81dd2580-d4a8-45bb-bd05-707c64aa4ac7",
+   "metadata": {},
+   "source": [
+    "First, we will transpile our circuit to ISA instructions, [as required for execution on our QPUs](https://www.ibm.com/quantum/blog/isa-circuits). For the data collected in this experiment, we hand select our qubits based on evaluation of the highest quality chain."
```
abbycross:
```suggestion
    "First, we will transpile our circuit to ISA instructions, [as required for execution on our QPUs](/docs/guides/transpile#instruction-set-architecture). For the data collected in this experiment, we hand-select our qubits based on evaluation of the highest quality chain."
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7e7135e7-1e86-4d16-b806-30d4a9225e88",
+   "metadata": {},
+   "source": [
+    "### Box the circuit\n",
+    "For ease of implementation, we will utilize the `generate_boxing_pass_manager` transpilation pass, which places the circuit instructions into annotated boxes. These boxes clearly indicate where, in the case of PEC, antinoise should be injected into the circuit. For details on settings, refer to the [Samplomatic documentation.](https://qiskit.github.io/samplomatic/)\n",
```
abbycross:
```suggestion
    "For ease of implementation, we utilize the `generate_boxing_pass_manager` transpilation pass, which places the circuit instructions into annotated boxes. These boxes clearly indicate where, in the case of PEC, anti-noise should be injected into the circuit. For details on settings, refer to the [Samplomatic documentation](https://qiskit.github.io/samplomatic/).\n",
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+    "\n",
+    "Note that the SLC workflow requires the use of `inject_noise_strategy=\"individual_modification\"` later in the process because this allows us to uniquely identify each `BoxOp` in the circuit.\n",
+    "\n",
+    "The `find_unique_box_instructions` function iterates through the provided boxed circuit and identifies those that have unique 2Q layers or measurements, for the purpose of noise learning and noise injection."
```
abbycross:
```suggestion
    "The `find_unique_box_instructions` function iterates through the provided boxed circuit and identifies those that have unique two-qubit (2Q) layers or measurements, for the purpose of noise learning and noise injection."
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "aa8794ef-7906-4ee3-a9b0-e0399b3e9129",
+   "metadata": {},
+   "source": [
+    "### Prepare canonical bases measurements\n",
+    "Due to how the qubits are labeled when identifying unique 2Q layers, one must take special care to keep track of qubit ordering. Below, we introduce the notion of `canonical_qubits` as a means to appropriately update the qubit ordering when providing it to the executor, as a result of how qubit order is captured when boxing circuits and finding unique instructions. See the [Qubit ordering convention](https://qiskit.github.io/samplomatic/guides/samplex_io.html#qubit-ordering-convention) documentation for details."
```
abbycross:
```suggestion
    "Note that we must take special care to keep track of qubit ordering. Below, we introduce the notion of `canonical_qubits` as a means to appropriately update the qubit ordering when providing it to Executor, as a result of how qubit order is captured when boxing circuits and finding unique instructions. See the [Qubit ordering convention](https://qiskit.github.io/samplomatic/guides/samplex_io.html#qubit-ordering-convention) documentation for details."
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+    "  2. Backward propagation (bound effect on initial state). Propagate each error to the start of the circuit and compute its commutator with the initial state.\n",
+    "\n",
+    "b. QPU — Learn noise rates. Use `NoiseLearnerV3` to estimate rates of the Pauli-Lindblad noise model.\n",
+    "\n",
+    "c. CPU — Prioritize mitigation\n",
+    "\n",
+    "  1. Update merged bounds with learned noise rates. Combine forward and backward bounds that were previously computed and update them with learned noise rates.\n",
+    "  2. Rank noise components to mitigate by using the computed bounds and learned rates. Prioritize each possible noise error based on its estimated impact on bias and the associated expense to correct.\n",
+    "\n",
+    "d. QPU — Insert antinoise and run. Execute the circuit of interest with antinoise (inverse noise) specified by using `Box` annotations.\n",
```
abbycross:
```suggestion
    "d. QPU — Insert anti-noise and run. Execute the circuit of interest with anti-noise (inverse noise) specified by using `Box` annotations.\n",
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "slc-wfnote-00-4000-8000-000000000025",
+   "metadata": {},
+   "source": [
+    "### Workflow for lightcone shading, noise learning, and anti-noise injection\n",
+    "\n",
+    "> **Note**: In this tutorial, we run the SLC bound computations *before* the noise learning completes, so the to-be-mitigated circuit is run as close in time as possible to the learned noise model. In principle this workflow can be parallelized further: a noise-learning job can run while, in parallel, the noise bounds are estimated. For an arbitrary quantum circuit the noise-bound computation can scale with a weakly exponential dependence, so parallelizing the bound computation (for example, across many CPU cores) yields tighter bounds for a given compute-time budget, and the QPU executions and bound computations can themselves be parallelized for the most efficient workflow."
```
abbycross: Rather than use the markdown `>` convention to indicate a note, we should use the Admonition component here:
```
<Admonition type="note" title="Title is optional">
text text text
</Admonition>
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ce0f9cf4-565a-4f18-baa5-bec5c7b354a2",
+   "metadata": {},
+   "source": [
+    "#### c. Compute backward bounds\n",
+    "\n",
+    "This part of noise prediction evaluates how an error at a particular layer can affect the input state $\\rho$. The `compute_backward_bounds` function first inverts the circuit, removes measurement gates, and then proceeds with a similar analysis as was done for the forward-bound computations."
```
abbycross:
```suggestion
    "This part of noise prediction evaluates how an error at a particular layer can affect the input state $\\rho$. The `compute_backward_bounds` function first inverts the circuit and removes measurement gates, and then proceeds with a similar analysis as was done for the forward-bound computations."
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -0,0 +1,1801 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "c8960e51-733e-4458-b77c-287ba3abdb8d",
+   "metadata": {},
+   "source": [
+    "#### Preview merged bounds without learned noise rates\n",
+    "\n",
+    "The `merged_bounds` function determines the point in the circuit where switching from backward bounds to forward bounds minimizes the total estimated bias on the desired observable. This bias is computed as the sum of the backward-bound contributions for all noise locations before that point, plus the forward-bound contributions for all noise locations after it. Currently, this is done uniformly for all qubits.\n",
+    "\n",
+    "> **Important Note**: The point to switch from forward to backward bounds depends on the learned noise rates."
```
abbycross: Use an Admonition component here too.

docs/tutorials/pec-with-shaded-lightcones.ipynb
henryzou50: Oh that a good way to put it, will do and keep that in mind for the future.

docs/tutorials/pec-with-shaded-lightcones.ipynb
henryzou50: Done!

henryzou50:
Appreciate the review and suggestions @abbycross. I just finished applying them!

abbycross (dismissed):
Thank you!

abbycross:
@nathanearnestnoble do you want to do a review before we merge?

docs/tutorials/pec-with-shaded-lightcones.ipynb
mrossinek, line 1: > trade a bounded residual bias for lower sampling overhead

Technically speaking, it is possible to retain a zero-bias estimate and still reduce the sampling overhead (geometric lightcone). But it is true that the shading is what allows you a further trading. Might be worth pointing out though.

mrossinek, line 1: > How to learn layer noise with `NoiseLearnerV3` and inject anti-noise through `samplomatic` and the Executor primitive

I suggest typesetting `Executor` as code to be consistent with the typesetting of NLv3 and samplomatic

mrossinek: I see that this gets mentioned further down, but am wondering if this single line needs some adjusting. Up to you

mrossinek, line 1: > that upper-bounds the susceptibility of the observable to that error.

Very much a nit-pick, but: we are bouding the susceptibility of the observable's expectation value to that error. In the backward bounds case, that also depends on the chosen initial state of the circuit (all-zero in typical workflows and sufficient for the purposes of this tutorial).

mrossinek, line 1: > In the case of our mirrored Ising circuit

I find it a bit odd, that the first time this gets mentioned is during the `Post-processing overview`. The phrasing suggests that this has already been established before but it has not.

mrossinek, line 1: > `pip install 'qiskit-addon-utils'`

Nit-pick: this is the only dependency that is typeset with surrounding quotes although that is not strictly necessary

mrossinek, line 1: > namely, we should measure an observable of 1

Similar nit-pick as before: the phrasing is misleading since not the observable is 1, but its expectation value.

mrossinek, line 1: > Because we run a mirror circuit, for every gate in the second half of the circuit there is an inverse gate in the first half. As the measured observable 
 has non-Z-basis measurements, and Executor accounts for the desired basis at the end of the circuit, we provide a prepare_basis function that inserts the appropriate gates at the start of the mirror circuit. This detail is specific to our mirror-circuit demonstration. We use the get_measurement_bases function to identify which gates are needed and where to append them, while keeping track of qubit-index subtleties arising from box annotation conventions, as discussed in the section on preparing canonical basis measurements.

I think this whole paragraph can be improved. We motivate that the mirror circuit is what gives us the known exp val of `1`, but truly that depends on the observable. This paragraph is trying to argue that the mirror circuit requires us to insert the `prepare_basis` gates for the observable, but in reality _we_ require that because _we_ want the exp val to be 1 for this particular combination of observable and mirror circuit.

All that is to say: I feel like the argument in this paragraph is a bit reversed. Can this be rewritten somewhat to more clearly motivate this:
1. _we_ choose to perform time evolution under Ising
2. _we_ want an observable which has one `X` and one `Z` component (because this is nice for highlighting the SLC bounds on different errors)
3. _we_ want a known exp val of `1`
4. _that_ requires us to prepend some gates to the circuit to ensure that `XZ` gets measured in the `+1` eigenstate at the end of the mirror circuit

mrossinek, line 1: The code blocks `In[3] - In[9]` can probably cleaned up somewhat.

- we should comment on free parameters (like `num-trotter_steps` and `rx_angle`)
- we can remove unused code paths from the functions defined in `In[7]` to make this less "daunting"

mrossinek, line 1: > `backend = service.least_busy(operational=True, simulator=False)`

The intro claims to run on a Heron device, but this does not get enforced. Is this of concern to you?

mrossinek, line 1: > we hand-select our qubits based on evaluation of the highest quality chain.

This is unlikely to be accurate given that the backend gets fetched based on `least_busy` as well as this hand-picked set being outdated quickly. I would remove this from the notebook since it sets a false precedent for users.

Instead, we should comment on the importance of layouting and suggest hand-picking (ideally point to a resource/tutorial where users can learn on how to run some quick experiments to inform their selection) but for the purposes of this tutorial it is probably best to rely on the Qiskit transpiler to select a layout, given the chosen backend's coupling map and current calibration data.

mrossinek, line 1: > `alt="Output of the previous code cell"`

These meta-texts for the images are probably useless. Is there no way of adjusting these?

mrossinek, line 1: For `In[14]` (the `boxes_pm` setup) it is important to know, that SLC currently assumes that the `InjectNoise` annotations use `site="after"`: https://github.com/Qiskit/qiskit-addon-slc/issues/33

Also be aware of `samplomatic`'s default value changing for this: https://github.com/Qiskit/samplomatic/pull/372

mrossinek, line 1: > `*before*`

I think this is unintentional formatting in the `Workflow for lightcone shading...` section

mrossinek, line 1: > Predict to-be-learned noise-model Paulis

Is the explanation of `generate_noise_model_paulis` really clear? To me it makes sense, but I have gotten feedback offline that this can be fairly confusing for someone new to the concept of noise learning. Would be good to get a sanity check on this section being clear.

mrossinek, line 1:
```
slc_atol = 1e-8
slc_eigval_max_qubits = 18
slc_evolution_max_terms = 1000
slc_num_processes = 8
slc_timeout = 60
```

I think it is improtant that these settings be explained in more detail.

Also:
- `slc_atol` has been split into two settings (https://github.com/Qiskit/qiskit-addon-slc/pull/69) which will get updated with `0.2` (release date TBD; happy to sync on that)
- `slc_timeout` is triggering in this notebook. That is done _by design_ which should be explained to the user to avoid confusion.

mrossinek, line 1: > which tracks how the observable spreads backward through the circuit and uses that spread to place upper bounds on the effect of each noise operator

I think it may be useful to be more concrete and refer to the bound exploiting the limited speed at which information can spread.

mrossinek, line 1: > We can further tighten the bounds by accounting for lightcone limitation. In principle, this gives us a smoother transition from the computed bounds to the trivial bounds set forth after the timeout was reached. Here, the effect is not as visible because the lightcones have already reached the edge of the circuit.

This feels a bit redundant to the previous paragraph.

mrossinek, line 1: > The point to switch from forward to backward bounds depends on the learned noise rates.

Yes, but at this point our `noise_model_rates` are initialized to `None` which assumes even rates and is meant just as a preview. This should be clarified.

mrossinek (commented):
Thanks for your work on this! This already looks great, below are mostly nit-picks and a few more general comments.

docs/tutorials/pec-with-shaded-lightcones.ipynb
henryzou50: Thanks and good point, I adjusted the learning outcome to note that a purely geometric lightcone already reduces the overhead with zero bias, and that the shading is what enables the further bias/overhead trade-off.

henryzou50: Done, thanks!

henryzou50: Fixed, it now refers to the observable's expectation value, and I added a note that the backward bounds also depend on the initial state (all-zero here).

henryzou50: Good catch, I just rephrased so the mirrored Ising circuit is introduced as forward-looking ("constructed in Step 1 below")

henryzou50: Removed, thanks!

henryzou50: Fixed here and in the Background section.

henryzou50: Nice suggestion, I restructured Step 1 as a numbered list following exactly this chain of choices, so the basis-preparation gates are now motivated by our choice of observable and known expectation value rather than by the mirror circuit itself.

henryzou50: Done! I added comments on the free parameters and removed the unused code paths (barrier flag, inverse_first flag, and a redundant branch in prepare_basis).

henryzou50: Oh yes good point and nice catch, I removed the hard-coded layout and now let the preset transpiler choose it from the coupling map and current calibration data, deriving layout from final_index_layout() for the downstream bookkeeping. The text now discusses why layout selection matters and points to the real-time benchmarking tutorial for readers who want to hand-pick a chain.

henryzou50: These alt texts are hardcoded by the docs repo's notebook-output extraction tooling (scripts/notebook-normalizer) and are the same across all tutorials, they can't be customized per image without an infrastructure change.

henryzou50: Good catch, the newer samplomatic has made the default flipped to "before". I now pass inject_noise_site="after" explicitly and added a caution admonition linking both the SLC issue and the samplomatic PR.

henryzou50: Fixed!

henryzou50: Fair point, I rewrote this section to first recall what the sparse Pauli-Lindblad model contains and then frame this step as predicting the model's structure (which NoiseLearnerV3 later fills in with rates). I bleieve that should land better for readers new to noise learning.

henryzou50: Done and good points! I added an explanation for each setting and an admonition making explicit that the timeout triggers by design and why that's safe.

henryzou50: Thanks, I reworded to be concrete about the bound exploiting the limited speed at which information spreads (the observable's backward lightcone growing at most one qubit per two-qubit-gate layer).

henryzou50: Good point, I trimmed the redundancy, the visualization note now only covers what the previous paragraph didn't.

henryzou50: Done! I clarified in the admonition that at this point the rates are all None (treated as uniform) and the merged bounds are only a preview, recomputed with the learned rates in Step 3.

henryzou50:
Thanks for the review and feedback @mrossinek! I applied them in my latest commit and will shorty follow up with another commit to rerun the experiments and have it be defaulted to a heron device. Currently the backends are in maintenance, but will do the rerun once that is complete.

mrossinek (commented):
Thanks for addressing all of my comments! I gave this another read and it's very nice now! I left a few final comments on ReviewNB but it looks like they did not sync to here :shrug: Let me know if you cannot see them..

henryzou50:
@mrossinek Thanks you for another read! I just updated the notebook to include the rerun results, which did unfortunately took a while due to the maintenance. The results looks good and I added your suggestion of having it default to a heron device. 

I can't see the final comments on ReviewNB, can you send them here?

docs/tutorials/pec-with-shaded-lightcones.ipynb
henryzou50: Changed to default to a Heron device!

mrossinek:
On the `Box the circuit` cell:

Tiny nit-pick: we uniquely identify the InjectNoise annotations of the BoxOp instructions. The reason to highlight this, is that non-noisy boxes cannot be identified.

mrossinek:
On the "bounds tightening" step:

This makes it sound like this tightening function only takes the speed-limit of the observable moving through the circuit is taken into account. But that should already be taken fully into account by the SLC computation. The key here is that we leverage the speed limit of information flowing through the circuit to tighten the previously computed bounds with respect to themselves. The way the internal code is commenting on this is two-fold:

1. Let the information flow from the commutator-bounds to the bounds on the backpropagated observable (if the observable commutes with a Z error, then it cannot have much X or Y component there).
2. Also do the reverse: information flow from the bounds on the backpropagated observable to the commutator-bounds (if the observable has low X and Y component on a qubit, then it must almost commute with Z there).


@aeddins-ibm can probably comment on this much more concisely than I can...

mrossinek:
And a quick note on that new discussion:
> On this device, full PEC carries a sampling overhead of ... After large early swings, the plain-PEC estimate recovers the signal, but settles at the upper edge of the 10% bias band (around 1.1), slightly overshooting the exact value.

The 10% bias bound does NOT apply to PEC. For vanilla PEC we would expect it to recover an unbiased estimate which it does not (although this one looks fairly reasonable, which is okay at this size of problem).

henryzou50:
Thanks for the quick review @mrossinek! I applied your suggested changes:

- Box the circuit: Reworded so that it's the `InjectNoise` annotations of the `BoxOp` instructions that are uniquely identified, and added a note that boxes without an `InjectNoise` annotation cannot be identified this way.
- Bounds tightening: Rewrote the paragraph so it's clear the geometric spread of the lightcone is already fully captured by the SLC computation, and that `tighten_with_speed_limit` instead uses the speed limit of information flow to tighten the previously computed bounds with respect to each other, in both directions: from the commutator bounds to the bounds on the backpropagated observable (commuting with a Z error -> little X/Y component there), and the reverse (little X/Y component -> almost commuting with Z there). Happy to further refine the wording if it's still not quite right.
- 10% bias band: Removed the framing of the plain-PEC result against the 10% bias band. The bullet now states that the band applies only to the PEC+SLC estimator, that plain PEC is in principle unbiased, and that the residual overshoot around 1.1 comes from statistical fluctuations and imperfections in the learned noise model, which is reasonable at this problem size.

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -371,7 +368,7 @@
    "metadata": {},
    "source": [
     "### Step 2: Optimize\n",
-    "We optimize details associated with the circuit to be run, the observable to be measured, and the noise-learning parameters. As a starting point, we instantiate a backend with fractional gates turned on. These fractional gates allow for greater sensitivity in some of our post-selection filtering."
+    "We optimize details associated with the circuit to be run, the observable to be measured, and the noise-learning parameters. As a starting point, we select the least busy Heron processor and instantiate it with fractional gates turned on. These fractional gates allow for greater sensitivity in some of our post-selection filtering."
```
abbycross:
```suggestion
    "We optimize details associated with the circuit to be run, the observable to be measured, and the noise-learning parameters. As a starting point, we select the least-busy Heron processor and instantiate it with fractional gates turned on. These fractional gates allow for greater sensitivity in some of our post-selection filtering."
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -212,9 +212,12 @@
...
+    "3. We want the ideal expectation value of this observable to be exactly 1, so that the quality of the mitigated estimates is easy to judge. Mirroring the circuit gets us most of the way there: for every gate in the second half of the circuit there is an inverse gate in the first half, so that the ideal circuit acts as the identity.\n",
+    "4. Finally, obtaining the ideal expectation value of 1 requires the circuit to start — and therefore, thanks to the mirroring, ideally also end — in a +1 eigenstate of the observable. Because the observable contains a non-$Z$ component, and the `Executor` primitive accounts for the desired measurement basis at the end of the circuit, we prepend the corresponding basis-preparation gates at the start of the mirror circuit with a `prepare_basis` function. We use the `get_measurement_bases` function to identify which gates are needed and where to place them, while keeping track of qubit-index subtleties arising from `box` annotation conventions, as discussed in the section on preparing canonical basis measurements."
```
abbycross:
```suggestion
    "4. Finally, obtaining the ideal expectation value of 1 requires the circuit to start &mdash; and therefore, thanks to the mirroring, ideally also end &mdash; in a +1 eigenstate of the observable. Because the observable contains a non-$Z$ component, and the `Executor` primitive accounts for the desired measurement basis at the end of the circuit, we prepend the corresponding basis-preparation gates at the start of the mirror circuit with a `prepare_basis` function. We use the `get_measurement_bases` function to identify which gates are needed and where to place them, while keeping track of qubit-index subtleties arising from `box` annotation conventions, as discussed in the section on preparing canonical basis measurements."
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -404,55 +408,33 @@
    "id": "81dd2580-d4a8-45bb-bd05-707c64aa4ac7",
    "metadata": {},
    "source": [
-    "First, we will transpile our circuit to ISA instructions, [as required for execution on our QPUs](/docs/guides/transpile#instruction-set-architecture). For the data collected in this experiment, we hand-select our qubits based on evaluation of the highest quality chain."
+    "First, we will transpile our circuit to ISA instructions, [as required for execution on our QPUs](/docs/guides/transpile#instruction-set-architecture). The choice of physical qubits to run on has a large impact on the quality of the results, because coherence times and gate fidelities vary across a device and drift over time. For a careful experiment, it can pay off to hand-select a chain of qubits informed by quick benchmarking experiments — see, for example, the [Real-time benchmarking for qubit selection](/docs/tutorials/real-time-benchmarking-for-qubit-selection) tutorial. For the purposes of this tutorial, we rely on the Qiskit transpiler to select a layout based on the backend's coupling map and its most recent calibration data."
```
abbycross:
```suggestion
    "First, we will transpile our circuit to ISA instructions, [as required for execution on our QPUs](/docs/guides/transpile#instruction-set-architecture). The choice of physical qubits to run on has a large impact on the quality of the results, because coherence times and gate fidelities vary across a device and drift over time. For a careful experiment, it can pay off to hand-select a chain of qubits informed by quick benchmarking experiments (see, for example, the [Real-time benchmarking for qubit selection](/docs/tutorials/real-time-benchmarking-for-qubit-selection) tutorial). For the purposes of this tutorial, we rely on the Qiskit transpiler to select a layout based on the backend's coupling map and its most recent calibration data."
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -653,12 +642,24 @@
...
+    "- `slc_evolution_max_terms`: the maximum number of Pauli terms retained while propagating operators through the circuit. Truncating the smallest terms keeps the computation tractable at the cost of slightly looser bounds.\n",
+    "- `slc_num_processes`: the number of CPU processes across which the bound computation is parallelized.\n",
+    "- `slc_timeout`: the time budget (in seconds) for the bound computation. Any layers not reached within this budget are assigned the trivial bound of 2.\n",
+    "\n",
+    "<Admonition type=\"note\">\n",
+    "In this tutorial, the timeout is reached *by design*: we keep `slc_timeout` deliberately small so that the notebook runs quickly, and to demonstrate that reaching the timeout is safe. The computed shades are upper bounds at every stage, so hitting the timeout only makes the bounds looser (trivial) for the layers that were not reached — it never underestimates the effect of an error. This appears in the visualizations below as the leftmost layers being filled entirely with the trivial bound. Increasing the timeout (or the number of processes) tightens the bounds and further reduces the sampling overhead.\n",
```
abbycross:
```suggestion
    "In this tutorial, the timeout is reached *by design*: we keep `slc_timeout` deliberately small so that the notebook runs quickly, and to demonstrate that reaching the timeout is safe. The computed shades are upper bounds at every stage, so hitting the timeout only makes the bounds looser (trivial) for the layers that were not reached; it never underestimates the effect of an error. This appears in the visualizations below as the leftmost layers being filled entirely with the trivial bound. Increasing the timeout (or the number of processes) tightens the bounds and further reduces the sampling overhead.\n",
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -770,13 +771,16 @@
    "id": "1aee6b29-8077-4019-9e37-3e80292aadf6",
    "metadata": {},
    "source": [
-    "#### b. Compute and tighten forward bounds\n",
-    "We next tighten the bounds by using the `tighten_with_speed_limit` function, which tracks how the observable spreads backward through the circuit and uses that spread to place upper bounds on the effect of each noise operator, taking the smaller of the forward bound just computed, and the backward-propagation bound."
+    "#### b. Tighten forward bounds with quantum speed limits\n",
+    "We next tighten the bounds by using the `tighten_with_speed_limit` function. The geometric spread of the observable's lightcone is already fully accounted for in the forward bounds computed above. What this function adds is a self-consistency pass that leverages the limited speed at which information can flow through the circuit — the lightcone grows by at most one qubit per layer of two-qubit gates — to tighten the previously computed bounds with respect to each other. The information flows in both directions:\n",
```
abbycross:
```suggestion
    "We next tighten the bounds by using the `tighten_with_speed_limit` function. The geometric spread of the observable's lightcone is already fully accounted for in the forward bounds computed above. What this function adds is a self-consistency pass that leverages the limited speed at which information can flow through the circuit &mdash; the lightcone grows by at most one qubit per layer of two-qubit gates &mdash; to tighten the previously computed bounds with respect to each other. The information flows in both directions:\n",
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -951,13 +955,13 @@
     "The `merged_bounds` function determines the point in the circuit where switching from backward bounds to forward bounds minimizes the total estimated bias on the desired observable. This bias is computed as the sum of the backward-bound contributions for all noise locations before that point, plus the forward-bound contributions for all noise locations after it. Currently, this is done uniformly for all qubits.\n",
     "\n",
     "<Admonition type=\"note\" title=\"Important note\">\n",
-    "The point to switch from forward to backward bounds depends on the learned noise rates.\n",
+    "The optimal point at which to switch from backward to forward bounds depends on the learned noise rates. At this stage we have not yet learned any rates: `noise_model_rates` was initialized to `None` for every layer, which `merge_bounds` treats as uniform noise rates. The merged bounds shown here are therefore only a preview — we recompute them with the actual learned rates in Step 3.\n",
```
abbycross:
```suggestion
    "The optimal point at which to switch from backward to forward bounds depends on the learned noise rates. At this stage we have not yet learned any rates: `noise_model_rates` was initialized to `None` for every layer, which `merge_bounds` treats as uniform noise rates. The merged bounds shown here are therefore only a preview; we recompute them with the actual learned rates in Step 3.\n",
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -1700,15 +1704,17 @@
...
+    "- The unmitigated result sits far outside the 10% bias band (near 0.35) and is unaffected by the number of randomizations.\n",
+    "- On this device, full PEC carries a sampling overhead of $\\gamma^2 \\approx 1.6\\times10^2$. After large early swings, the plain-PEC estimate recovers the signal and settles around 1.1, slightly overshooting the exact value. Note that the 10% bias band applies only to the PEC+SLC estimator: plain PEC is in principle unbiased, so this residual deviation is not covered by a bias guarantee but instead reflects statistical fluctuations and imperfections in the learned noise model, which is reasonable at this problem size.\n",
+    "- SLC reduces the overhead a further ~3.4-fold (to $\\gamma^2 \\approx 48$, for a residual-bias bound of about 10%). PEC+SLC also fluctuates strongly at few randomizations — initially even reaching into the unphysical region — but its estimate converges steadily onto the exact value, settling near 1.0 well inside the band. It matches or exceeds the accuracy of plain PEC at roughly a third of the sampling cost, demonstrating the benefit of lightcone shading.\n",
```
abbycross:
```suggestion
    "- SLC reduces the overhead a further ~3.4-fold (to $\\gamma^2 \\approx 48$, for a residual-bias bound of about 10%). PEC+SLC also fluctuates strongly at few randomizations, initially even reaching into the unphysical region, but its estimate converges steadily onto the exact value, settling near 1.0 well inside the band. It matches or exceeds the accuracy of plain PEC at roughly a third of the sampling cost, demonstrating the benefit of lightcone shading.\n",
```

docs/tutorials/pec-with-shaded-lightcones.ipynb
```diff
@@ -1700,15 +1704,17 @@
...
+    "- The error bars of both mitigated estimates contract as randomizations accumulate; at a fixed number of randomizations, the PEC+SLC estimator has the smaller statistical variance, consistent with its smaller $\\gamma^2$.\n",
+    "\n",
+    "Note that the learned noise rates — and therefore the overheads and mitigated estimates — depend on the backend and its calibration at the time of execution, so rerunning this notebook can produce quantitatively different values."
```
abbycross:
```suggestion
    "Note that the learned noise rates (and therefore the overheads and mitigated estimates) depend on the backend and its calibration at the time of execution, so rerunning this notebook can produce quantitatively different values."
```

abbycross (approved):

PR #5228, opened 2026-06-06
Update query-model-of-computation.ipynb

mgsein:
Added "of" between both us to in the second paragraph of query model of computation part.


abbycross:
Thanks for opening a fix - it was done in https://github.com/Qiskit/documentation/pull/5231 so closing this one.

PR #5249, opened 2026-06-11
Add some notes about shots and precision

### LINKED ISSUE #5082: Add a note about target precision

https://quantum.cloud.ibm.com/docs/en/guides/estimator-options

### Select all that apply

- [x] new content request
- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

We should add a note saying that the `precision` parameter specifies the _target_ precision, meaning we will try but it's not guaranteed that we'll reach that precision. 

jyu00:
Per user feedback, this PR adds a note saying an error is raised if the primitive doesn't know how to divide shots among twirls when twirling is enabled. 

Also closes #5082


docs/guides/estimator-options.ipynb
```diff
@@ -263,6 +267,10 @@
     "For example, if precision is specified in all four places, the one with highest precedence (precision specified in the PUB) is used.\n",
     "\n",
     "<Admonition type=\"note\">\n",
+    "Although precision specified in the PUB and in `run` have higher precedence, the job would fail if `twirling` is enabled and the product of `num_randomizations` and `shots_per_randomization` is smaller than the shots needed to achieve the precision. In this scenario `EstimatorV2` does not know how to allocate the shots among the specified `num_randomizations`.\n",
+    "</Admonition>\n",
+    "\n",
+    "<Admonition type=\"note\">\n",
```
abbycross, line 273: Just confirming that you want two admonition boxes in a row - it's fine if intentional

Image

docs/guides/estimator-options.ipynb
```diff
@@ -263,6 +267,10 @@
     "For example, if precision is specified in all four places, the one with highest precedence (precision specified in the PUB) is used.\n",
     "\n",
     "<Admonition type=\"note\">\n",
+    "Although precision specified in the PUB and in `run` have higher precedence, the job would fail if `twirling` is enabled and the product of `num_randomizations` and `shots_per_randomization` is smaller than the shots needed to achieve the precision. In this scenario `EstimatorV2` does not know how to allocate the shots among the specified `num_randomizations`.\n",
```
abbycross:
```suggestion
    "Although precision specified in the PUB and in `run` have higher precedence, the job fails if `twirling` is enabled and the product of `num_randomizations` and `shots_per_randomization` is smaller than the shots needed to achieve the precision. In this scenario, `EstimatorV2` is unable to allocate the shots among the specified `num_randomizations`.\n",
```

docs/guides/sampler-options.ipynb
```diff
@@ -144,12 +144,16 @@
     "\n",
     "Thus, if shots are specified in all possible places, the one with highest precedence (shots specified in the PUB) is used.\n",
     "\n",
+    "<Admonition type=\"note\">\n",
+    "Although shots specified in the PUB and in `run` have higher precedence, the job would fail if `twirling` is enabled and the product of `num_randomizations` and `shots_per_randomization` is smaller than the `shots` value. In this scenario `SamplerV2` does not know how to allocate the shots among the specified `num_randomizations`.\n",
```
abbycross:
```suggestion
    "Although shots specified in the PUB and in `run` have higher precedence, the job fails if `twirling` is enabled and the product of `num_randomizations` and `shots_per_randomization` is smaller than the `shots` value. In this scenario, `SamplerV2` is unable to allocate the shots among the specified `num_randomizations`.\n",
```

docs/guides/estimator-options.ipynb
jyu00: I kept them separate only because they refer to different things, but I don't feel too strongly either way. Do we usually use 1 box to group these together?

abbycross: If they are related enough, we can list them as bullet points in the same admonition - but I think it's fine to keep these as two admonitions, even one after the other like this, since it adds readability and emphasis, which I think you want here. I just wanted to be sure that was the intent.

abbycross (approved):
Thanks!

PR #5254, opened 2026-06-11
Add Julia tutorial for simulating time evolution of the Ising model

haimeng-zhang:
### Summary
Julia is a dynamic language designed for high-performance numerical and scientific computing, making it a natural fit for quantum simulation workflows. The tutorial shows how Julia is used for both classical pre- and post-processing (e.g., building Hamiltonians, running ODE solvers, computing expectation values) and for orchestrating quantum hardware jobs, eliminating the need to switch between languages or environments. The example used in this tutorial is simulating time evolution of the transverse-field Ising model. 

To interface with IBM Quantum hardware from Julia, this tutorial uses two packages from the Qiskit ecosystem: `Qiskit.jl` wraps the Qiskit C API and provides circuit construction and transpilation functionality in Julia; `QiskitIBMRuntime.jl` connects to IBM Quantum hardware through the Qiskit IBM Runtime service, enabling job submission and result retrieval directly from Julia.

### Files
The PR adds the following files:

- the notebook `docs/tutorials/time-evolution/time-evolution.ipynb` 
- the files `docs/tutorials/time-evolution/Project.toml` and `docs/tutorials/time-evolution/Manifest.toml`, which are needed to set up the Julia environment and install the dependencies to run the notebook.


garrison:
Before I forget ... the runtime client has a pending rename, which will be complete once https://github.com/JuliaRegistries/General/pull/158556 merges.

docs/tutorials/time-evolution/Project.toml
```diff
@@ -0,0 +1,10 @@
+[deps]
```
kaelynj, line 1: Is it possible to remove both the toml files? These files won't render on the site and readers won't be able to see their contents. I would say if it's not critical to have the _exact_ package versions installed you could just list these as packages the user should install. If it is, I would add the contents of `Project.toml` into the notebook through an accordion object.

docs/tutorials/time-evolution/time-evolution.ipynb
kaelynj, line 1: Another option that would make this easier to pass through our CI checks would be to convert this file to markdown. This would bypass our notebook/lint checker which, for this particular case, I think is fine to do.

abbycross: I am doing the conversion locally and will upload as soon as I've compared the two versions to make sure they are identical. I will also move the file outside of the /time-evolution/ directory (right now the URL would be /docs/tutorials/time-evolution/time-evolution)

frankharkins: Haimeng messaged me yesterday about the linter. Bumping squeaky from `0.7.0` to `0.7.1` in these two places should fix it:

* [tox.ini](https://github.com/Qiskit/documentation/blob/c8724d5d55770f9bdacb9cd53a1f5df8b6c6dcc5/tox.ini#L15)
* [scripts/nb-tester/pyproject.toml](https://github.com/Qiskit/documentation/blob/c8724d5d55770f9bdacb9cd53a1f5df8b6c6dcc5/scripts/nb-tester/pyproject.toml#L16)

abbycross: Hm, I updated squeaky (one was already done) but I'm still getting errors. Should we convert to markdown after all, @kaelynj?

kaelynj: I think we should still. My understanding is that `ruff` doesn't support Julia and we would have to set up a new linting pipeline in CI for this notebook. 

Eventually we should find a solution for if we start adding more content in different languages, but for now I think it's fine to convert this to markdown. (Or have `ruff` ignore the file).

garrison: > Should we convert to markdown after all [...] ?

Is the idea to convert it to markdown just to satisfy the linter?  Might it make more sense to add this notebook to the linters' ignore list?  Keeping it in notebook form has a few advantages that I can think of, namely that users could download it and execute, and also that we could then ourselves download it and manually execute it at a later date, to ensure that the instructions still work with the latest version of Qiskit.jl.  Sure, manual testing is not as reliable as automated testing, but keeping it as a notebook still provides a path to testing it easily.

frankharkins: @abbycross you still have the bad metadata from before the squeaky upgrade, you need to revert 9eba3cf7f11c7a2c2971ab29f30465c32888f2b0 then re-run `./fix`

kaelynj: > > Should we convert to markdown after all [...] ?
> 
> Is the idea to convert it to markdown just to satisfy the linter? Might it make more sense to add this notebook to the linters' ignore list? Keeping it in notebook form has a few advantages that I can think of, namely that users could download it and execute, and also that we could then ourselves download it and manually execute it at a later date, to ensure that the instructions still work with the latest version of Qiskit.jl. Sure, manual testing is not as reliable as automated testing, but keeping it as a notebook still provides a path to testing it easily.

Yeah it's mostly to satisfy the linter. I'd be fine with just adding it to the ignore list and keeping it as a notebook.

@haimeng-zhang you should be able to do this by adding this line to the `docs/ruff.toml` file
```
extend-exclude = ['tutorials/time-evolution/time-evolution.ipynb
```

docs/tutorials/time-evolution/time-evolution.ipynb
```diff
@@ -0,0 +1,1124 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "id": "4ea0efac",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "using Qiskit\n",
```
garrison, line 172: After some discussion with other Qiskit developers, I think it will be better to use the idiomatic Julia API to construct a circuit, described at https://github.com/Qiskit/Qiskit.jl#idiomatic-julia-interface-qiskitoperations

```suggestion
    "using Qiskit\n",
    "using Qiskit.Operations\n",
```

docs/tutorials/time-evolution/time-evolution.ipynb
```diff
@@ -0,0 +1,1124 @@
...
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "function make_trotter_circuit(h::Vector, J::Vector, n::Int, δt::Float64, n_trotter_steps::Int)\n",
+    "    qc = QuantumCircuit(n, n)\n",
+    "\n",
+    "    # Neel state initialization\n",
+    "    for i in 1:2:n\n",
+    "        qc.x(i)\n",
```
garrison, line 576: Here's an example of what a corresponding change would look like.  Throughout, all mutating operations starting with `qc.` should be replaced with the corresponding method ending with a `!`.

```suggestion
    "        x!(qc, i)\n",
```

haimeng-zhang: Sorry for coming to this late. I've made this change to the tutorial. @garrison, please take a look.

docs/tutorials/time-evolution/time-evolution.ipynb
haimeng-zhang: Updated accordingly.

docs/tutorials/time-evolution/time-evolution.ipynb
haimeng-zhang: Added this exclude line to the `docs/ruff.toml` file.

docs/tutorials/time-evolution/Project.toml
haimeng-zhang: Hi @kaelynj, Sorry for the late reply on this! I edited the `Requirements` section of the notebook. It now has a code cell that directly installs the tutorial dependencies; this way the installation experience is self-contained in the notebook. It also provide an alternative approach to install from the `.toml` files, by linking to the Qiskit documentation github repo for those files, for users who want exact reproducibility. Let me know if this looks good to you :)

docs/tutorials/time-evolution/time-evolution.ipynb
```diff
@@ -0,0 +1,1647 @@
...
+    "Pkg.activate(mktempdir())          # fresh temporary environment\n",
+    "Pkg.add([\n",
+    "    \"Qiskit\",\n",
+    "    \"QiskitIBMRuntime\",\n",
+    "    \"OrdinaryDiffEq\",\n",
+    "    \"TensorNetworkQuantumSimulator\",\n",
+    "    \"JSON\",\n",
+    "    \"LinearAlgebra\",\n",
+    "    \"Plots\",\n",
+    "    \"SparseArrays\",\n",
```
garrison: It's not necessary to include the standard library packages here.

```suggestion
    "    \"Plots\",\n",
```

haimeng-zhang: Applied the change manually

henryzou50 (commented):
Thanks for putting this together @haimeng-zhang! Overall the tutorial looks great and was easy to setup for me. I did note one potential bug and a few small changes below.

## 1. Bug: Rx angles are a factor of 2 too small 

`Rx(θ) = exp(−iθX/2)`, so the half-step `exp(−i·h·X·δt/2)` needs **θ = h·δt**, but both circuits use `h[i] * δt / 2` -- the `/2` *double-counts* the ½ already in the gate definition, so the circuits evolve under an effective field of **h/2**. (The Rzz angle is correct.) I confirmed numerically with the notebook's own exact-solution code at N=12: the TN circuit as written differs from exact(h=1) by up to **0.31** in site magnetization, but matches exact(h=0.5) to 3×10⁻⁴. This also contradicts the text before the final figure ("classical and tensor network results are now in close agreement").

The fix is dropping `/ 2` from the four Rx angle expressions (two per function — the Rx layer appears before and after the Rzz layer in each Trotter step). Rzz lines are untouched.

In `make_trotter_circuit_tn`:

```diff
         # first half X rotation
-        append!(circuit, [("Rx", [(i,)], h[i]* δt / 2) for i in 1:n])
+        append!(circuit, [("Rx", [(i,)], h[i] * δt) for i in 1:n])
         # ZZ interactions
         append!(circuit, [("Rzz", [(i,), (i+1,)], 2 * J[i] * δt) for i in 1:n-1])
         # second half X rotation
-        append!(circuit, [("Rx", [(i,)], h[i] * δt / 2) for i in 1:n])
+        append!(circuit, [("Rx", [(i,)], h[i] * δt) for i in 1:n])
```

In `make_trotter_circuit` (both `rx!` loops):

```diff
         for i in 1:n
-            rx!(qc, h[i] * δt / 2, i)
+            rx!(qc, h[i] * δt, i)
         end
```

After the fix, the notebook needs a full re-run, including the hardware jobs and all saved outputs.

## 2. Template alignment: label the simulator vs. hardware sections

Right now the hardware Steps 1–4 are H3 headers nested under `## Small-scale simulator example`, so readers can't tell where simulation ends and hardware begins. Suggest restructuring the headers to follow the tutorial template:

- Keep `## Small-scale simulator example` — covering the exact solution and the TN simulation.
- Add `## Large-scale hardware example` immediately before `### Step 1: Map classical inputs to a quantum problem`, with a one-line intro (and a brief note on the N=20 scale).

Two small additions while in there:

- Add the usage estimate after the title (template format): *Usage estimate: X minutes on ibm_y. (NOTE: This is an estimate only. Your runtime may vary.)*
- Add a short `## Next steps` section at the end linking related material (e.g., Qiskit.jl repo, a related time-evolution tutorial).

## 3. Reproducibility: state the Julia version

The Manifest was generated with Julia 1.11.5, but Requirements doesn't mention a version. note that Option 2's "exact environment" depends on it. Add one line to Requirements, e.g. "This tutorial was developed with Julia 1.11; install via `juliaup add 1.11`."

## 4. Minor code cleanups

- Delete `s = siteinds("S=1/2", g)`, as it is unused.
- Delete the `fidelities` accumulation (or mention the final fidelity in the text) -- currently unused after the loop.
- Delete the bare `job_list` display cell -- redundant with the status loop that follows..

public/docs/images/tutorials/time-evolution/time-evolution/extracted-outputs/cc5496cc-0.svg
garrison, line 1: I wonder if adding an exception here could resolve the lint failure:

https://github.com/Qiskit/documentation/blob/96ba18bdbd57bdc6b6df672254bfee09d4793bb7/scripts/js/commands/checkStaleImages.ts#L27-L29

abbycross:
The Execute-notebooks check failure is probably because it can only properly run on a write-access branch. I will merge this branch to a new one and re-run the checks. I'll also clean up the flagged images that are breaking the linter.

abbycross:
Work is continuing here: https://github.com/Qiskit/documentation/pull/5557

PR #5291, opened 2026-06-22
Fix wording in AI transpiler passes

kaelynj:
Leftover wording which reference the qiskit transpiler service was still left on the page. This PR removes that language since the package is meant to only be run locally now


abbycross (approved):
Unless we're adding any clarifications to the changelog, this is all set!

kaelynj:
> Unless we're adding any clarifications to the changelog, this is all set!

Yeah I'm waiting to hear back from the team if there's anything that needs to go out. In the meantime these changes should be sufficient!

PR #5300, opened 2026-06-24
Change package name to qiskit-paulice

abbycross:


jenglick:
Just discussed with @kaelynj -- we will leave URLs as-is. Sorry for the runaround!

Issue #5265, opened 2026-06-12
Improve Resource Controller API discussion of extensions vs parameters


Eric-Arellano:
### URL to the relevant guide or resource

https://quantum.cloud.ibm.com/docs/en/guides/access-instances-platform-apis

### Select all that apply

- [x] new content request
- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

### Describe the fix or the content request.

Per https://github.com/IBM/qauvern/issues/153, `extensions` vs `parameters` has been confusing with using the REST API to manage instances. For example, you can run into problems if you _consume_ values from `parameters` rather than `extensions`. It would also probably help to have some examples or generally more docs.

### For new content requests - if the request is accepted, do you want to write the content?

I will write (or already have written) a draft of the proposed content

Eric-Arellano:
https://github.com/Qiskit/documentation/pull/5320 rewrites it all. We need to also update the REST API index page, which is managed in inner source.

kaelynj:
Closing this and following it up in inner source.

PR #5320, opened 2026-07-01
Rewrite Resource Controller docs

Eric-Arellano:
Part of https://github.com/Qiskit/documentation/issues/5265.

This rewrite is mostly based on findings from https://github.com/IBM/qauvern using the Resource Controller API extensively.

We also add two warnings about quirks with the API.


docs/guides/access-instances-platform-apis.mdx
```diff
@@ -1,44 +1,65 @@
...
-- [Update a resource instance](https://cloud.ibm.com/apidocs/resource-controller/resource-controller#update-resource-instance)
-- [Get a resource instance](https://cloud.ibm.com/apidocs/resource-controller/resource-controller#get-resource-instance)
-- [Get a list of all resource instances](https://cloud.ibm.com/apidocs/resource-controller/resource-controller#list-resource-instances)  
-    
-    When getting a list of all resource instances, note that you can filter to only Qiskit Runtime instances with `resource_id = "b6049020-80f4-11eb-a0f7-e35ec9b4054f"`
+Use the [`GET /v2/resource_instances/{crn}`](https://cloud.ibm.com/apidocs/resource-controller/resource-controller#get-resource-instance) endpoint to get information about a particular instance. The [CRN](/docs/guides/instances) must be URL-encoded in the path.
 
-</Admonition>
+In addition to the standard Resource Controller fields, the response includes Quantum-specific fields in `extensions`. Use `extensions` rather than `parameters` because `parameters` may include a subset of the full fields.
```
beckykd:
```suggestion
In addition to the standard Resource Controller fields, the response includes quantum-specific fields in `extensions`. Use `extensions` rather than `parameters` because `parameters` might include a subset of the full fields.
```

docs/guides/access-instances-platform-apis.mdx
```diff
@@ -1,44 +1,65 @@
...
+- **`backends`** — Array of strings. The allow-list of backend names available to this instance. `["ANY"]` means all backends on the plan are available (the default). `[]` means no backends are available.
 
+## Get a list of all instances
 
-- **instance_limit_seconds** (string or null, `^[0-9]+$`): The limit to set on the instance. For Open Plan: if omitted, this defaults to 10 minutes.
-.
-- **usage_allocation_seconds** (string or null, `^[0-9]+$`): Allotted time for the instance, used by the fair-share scheduler to determine the priority in the queue, based on usage of all QPUs. (This does not apply to Pay-As-You-Go instances.)
-- **backends** (array of lists of backends (strings or null) or array with "ANY"): the allow-list of names of backends that are available to this instance. Default: `["ANY"]`, means all backends available on this plan. `[]` means no backends.
+Use the [`GET /v2/resource_instances`](https://cloud.ibm.com/apidocs/resource-controller/resource-controller#list-resource-instances) endpoint to get a list of all your instances. Set the `resource_id` query parameter to `b6049020-80f4-11eb-a0f7-e35ec9b4054f` to filter to IBM Quantum instances.
```
beckykd:
```suggestion
Use the [`GET /v2/resource_instances`](https://cloud.ibm.com/apidocs/resource-controller/resource-controller#list-resource-instances) endpoint to get a list of all your instances. Set the `resource_id` query parameter to `b6049020-80f4-11eb-a0f7-e35ec9b4054f` to filter to IBM Quantum&reg; instances.
```

Eric-Arellano:
You can see a preview at https://qiskit.github.io/documentation/pr-5320/docs/guides/access-instances-platform-apis

docs/guides/access-instances-platform-apis.mdx
```diff
@@ -1,44 +1,65 @@
...
+Use the [`POST /v2/resource_instances`](https://cloud.ibm.com/apidocs/resource-controller/resource-controller#create-resource-instance) endpoint to create (provision) a new instance. Pass a JSON body with the header `"Content-Type: application/json"`.
+
+Required fields:
+
+- **`name`** — A human-readable name for the instance.
+- **`target`** — The region, such as `us-east` or `eu-de`.
+- **`resource_plan_id`** — The plan for this instance. See the [plan ID table](#get-a-list-of-all-instances).
+- **`resource_group`** — The [resource group](https://cloud.ibm.com/docs/account?topic=account-rgs) to use.
+
+You can also include a `parameters` object to set Quantum-specific values:
```
beckykd:
```suggestion
You can also include a `parameters` object to set quantum-specific values:
```

docs/guides/access-instances-platform-apis.mdx
```diff
@@ -1,44 +1,65 @@
...
+
+- **`instance_limit_seconds`** — Integer, or `null`. The usage time limit for the instance. See [Set instance allocation limits](/docs/guides/allocation-limits).
+- **`usage_allocation_seconds`** — Integer, or `null`. The time allocated to this instance, used by the [fair-share scheduler](/docs/guides/fair-share-scheduler) to determine queue priority. See [Set instance allocation limits](/docs/guides/allocation-limits). Not applicable to Pay-As-You-Go instances.
+- **`backends`** — Array of strings. The allow-list of backend names available to this instance. `["ANY"]` means all backends on the plan are available. `[]` means no backends are available.
```
beckykd:
```suggestion
- **`backends`** — Array of strings. The allowlist of backend names available to this instance. `["ANY"]` means all backends on the plan are available. `[]` means no backends are available.
```

docs/guides/access-instances-platform-apis.mdx
```diff
@@ -1,44 +1,65 @@
...
\ No newline at end of file
+## Update an instance
+
+Use the [`PATCH /v2/resource_instances/{crn}`](https://cloud.ibm.com/apidocs/resource-controller/resource-controller#update-resource-instance) endpoint to update the limit, allocation, and permitted backends for an instance. The [CRN](/docs/guides/instances) must be URL-encoded in the path.
+
+Pass a `parameters` JSON object in the request body with the fields you want to change, along with the header `"Content-Type: application/json"`. Omitted fields are left unchanged.
+
+- **`instance_limit_seconds`** — Integer, or `null`. The usage time limit for the instance. See [Set instance allocation limits](/docs/guides/allocation-limits).
+- **`usage_allocation_seconds`** — Integer, or `null`. The time allocated to this instance, used by the [fair-share scheduler](/docs/guides/fair-share-scheduler) to determine queue priority. See [Set instance allocation limits](/docs/guides/allocation-limits). Not applicable to Pay-As-You-Go instances.
+- **`backends`** — Array of strings. The allow-list of backend names available to this instance. `["ANY"]` means all backends on the plan are available. `[]` means no backends are available.
```
beckykd:
```suggestion
- **`backends`** — Array of strings. The allowlist of backend names available to this instance. `["ANY"]` means all backends on the plan are available. `[]` means no backends are available.
```

docs/guides/access-instances-platform-apis.mdx
```diff
@@ -1,44 +1,65 @@
...
-</Admonition>
+In addition to the standard Resource Controller fields, the response includes Quantum-specific fields in `extensions`. Use `extensions` rather than `parameters` because `parameters` may include a subset of the full fields.
 
-## `parameters`
+The `extensions` object includes these fields:
 
-In the body when provisioning or updating an instance:
+- **`instance_limit_seconds`** — Integer, or `null`. The usage time limit for the instance. See [Set instance allocation limits](/docs/guides/allocation-limits).
+- **`usage_allocation_seconds`** — Integer, or `null`. The time allocated to this instance, used by the [fair-share scheduler](/docs/guides/fair-share-scheduler) to determine queue priority. See [Set instance allocation limits](/docs/guides/allocation-limits).
+- **`backends`** — Array of strings. The allow-list of backend names available to this instance. `["ANY"]` means all backends on the plan are available (the default). `[]` means no backends are available.
```
beckykd:
```suggestion
- **`backends`** — Array of strings. The allowlist of backend names available to this instance. `["ANY"]` means all backends on the plan are available (the default). `[]` means no backends are available.
```

docs/guides/access-instances-platform-apis.mdx
```diff
@@ -1,44 +1,245 @@
...
+
+The `extensions` object includes these fields:
+
+- `instance_limit_seconds` — Integer, or `null`. The usage time limit for the instance. See [Set instance allocation limits](/docs/guides/allocation-limits).
+- `usage_allocation_seconds` — Integer, or `null`. The time allocated to this instance, used by the [fair-share scheduler](/docs/guides/fair-share-scheduler) to determine queue priority. See [Set instance allocation limits](/docs/guides/allocation-limits).
+- `backends` — Array of strings. The allowlist of backend names available to this instance. `["ANY"]` means all backends on the plan are available (the default). `[]` means no backends are available.
+
+<Admonition type="caution" title="Some extensions fields may be outdated">
+  The following fields in the `extensions` object may be outdated. This can happen when IBM Quantum Support changes your instance's configuration. The current workaround is to instead use the [Qiskit Runtime REST API](/docs/api/qiskit-runtime-rest) to get this metadata.
```
riley2012: I wonder if we can update this wording: the IBM Quantum Support team doesn't directly change an instance's configuration — it might seem like they have permission to do that, which they don't. Can we say something like (this may be too wordy, but its here for the idea): This can happen when an account update made by IBM Quantum Support results in a change to your instance, for example when a backend is removed from an account, it is automatically removed from an instance of the account.

riley2012: Minor nit:
> The following fields
We plan to add two more fields in the near future (functions, and custom_functions), but for now, should this be singular?

Eric-Arellano: Yeah, it probably should be singular for now. I indeed was trying to future-proof it for your upcoming docs, but that's confusing to users

docs/guides/access-instances-platform-apis.mdx
```diff
@@ -1,44 +1,245 @@
...
+
+Pass a `parameters` JSON object in the request body with the fields you want to change, along with the header `"Content-Type: application/json"`. Omitted fields are left unchanged.
+
+- `instance_limit_seconds` — Integer, or `null`. The usage time limit for the instance. See [Set instance allocation limits](/docs/guides/allocation-limits).
+- `usage_allocation_seconds` — Integer, or `null`. The time allocated to this instance, used by the [fair-share scheduler](/docs/guides/fair-share-scheduler) to determine queue priority. See [Set instance allocation limits](/docs/guides/allocation-limits). Not applicable to Pay-As-You-Go instances.
+- `backends` — Array of strings. The allowlist of backend names available to this instance. `["ANY"]` means all backends on the plan are available. `[]` means no backends are available.
+
+<Admonition type="caution" title="Always include a unique timestamp">
+  The API silently ignores the request if `parameters` is identical to the previous request. In the `parameters` object, always include a `timestamp` field set to the current time so each request is treated as unique.
+</Admonition>
```
riley2012, line 120: Nice! and thank you! 💯

docs/guides/access-instances-platform-apis.mdx
```diff
@@ -1,44 +1,245 @@
...
+The `extensions` object includes these fields:
+
+- `instance_limit_seconds` — Integer, or `null`. The usage time limit for the instance. See [Set instance allocation limits](/docs/guides/allocation-limits).
+- `usage_allocation_seconds` — Integer, or `null`. The time allocated to this instance, used by the [fair-share scheduler](/docs/guides/fair-share-scheduler) to determine queue priority. See [Set instance allocation limits](/docs/guides/allocation-limits).
+- `backends` — Array of strings. The allowlist of backend names available to this instance. `["ANY"]` means all backends on the plan are available (the default). `[]` means no backends are available.
+
+<Admonition type="caution" title="Some extensions fields may be outdated">
+  The following fields in the `extensions` object may be outdated. This can happen when IBM Quantum Support changes your instance's configuration. The current workaround is to instead use the [Qiskit Runtime REST API](/docs/api/qiskit-runtime-rest) to get this metadata.
+
+  - `backends`. Instead, use [`GET /v1/instance`](/docs/api/qiskit-runtime-rest/tags/instances#tags__instances__operations__get_instance).
```
riley2012: The endpoint to use for backends is `GET /v1/backends`

docs/guides/access-instances-platform-apis.mdx
```diff
@@ -1,44 +1,245 @@
...
-- [Update a resource instance](https://cloud.ibm.com/apidocs/resource-controller/resource-controller#update-resource-instance)
-- [Get a resource instance](https://cloud.ibm.com/apidocs/resource-controller/resource-controller#get-resource-instance)
-- [Get a list of all resource instances](https://cloud.ibm.com/apidocs/resource-controller/resource-controller#list-resource-instances)  
-    
-    When getting a list of all resource instances, note that you can filter to only Qiskit Runtime instances with `resource_id = "b6049020-80f4-11eb-a0f7-e35ec9b4054f"`
+## Get an instance
 
+Use the [`GET /v2/resource_instances/{crn}`](https://cloud.ibm.com/apidocs/resource-controller/resource-controller#get-resource-instance) endpoint to get information about a particular instance. The [CRN](/docs/guides/instances) must be URL-encoded in the path.
+
+In addition to the standard Resource Controller fields, the response includes quantum-specific fields in `extensions`. Use `extensions` rather than `parameters` because `parameters` might include a subset of the full fields.
```
riley2012: I feel like it would be good to mention `parameters` here in the response, to explain what they are, but then be done with them and only refer to `extensions` later in this doc. Also, the issue is not only that `parameters` are a subset, but the actual data that gets written (by the Service Broker) can be different than what it was passed, and extensions should have the current configuration that was stored by the service.

Maybe something like this?

The response includes quantum-specific fields in both `parameters` and `extensions`. Read configuration from `extensions` rather than `parameters`: `parameters` echoes your most recent request, while `extensions` reflect the configuration that was stored by the service.

docs/guides/access-instances-platform-apis.mdx
```diff
@@ -1,44 +1,245 @@
...
+If your account has multiple plans and you want to filter by plan, set the `resource_plan_id` query parameter to one of the following values:
+
+| Plan | `resource_plan_id` |
+|------|---------------------|
+| Premium | `7f666d17-7893-47d8-bf9d-2b2389fc4dfc` |
+| Flex | `53bde9d3-cdbb-46f5-a98f-60ebcadf7260` |
+| Pay-As-You-Go | `5304b575-3cff-4455-90dc-ae4367762093` |
+| Open | `850b21a7-71de-4e53-9441-1abdd202f35d` |
+
+Each result includes `extensions` and `parameters` fields as described in [Get an instance](#get-an-instance).
```
riley2012: Since `parameters` is explained in [Get an instance](#get-an-instance), can this be simplified by pointing users only to the field that matters to them?
Something like - Each result includes the same `extensions` fields as [Get an instance](#get-an-instance).

riley2012 (commented):
@Eric-Arellano these are great improvements. Thanks for working on it! I left some comments in case you find them helpful.

docs/guides/access-instances-platform-apis.mdx
Eric-Arellano: Nack. The `/v1/backends` endpoint returns account-level info. The instances endpoint returns the backends for the specific instance

```
{
  "backends": [
    "ibm_backend"
  ],
  "instance_limit": 60.7,
  "instance_limit_seconds": "61",
  "plan_id": "flex-plan-id-1",
  "usage_allocation_seconds": "200"
}
```

See https://quantum.cloud.ibm.com/docs/en/api/qiskit-runtime-rest/tags/instances#tags__instances__operations__get_instance

riley2012: `GET /v1/backends` requires a `Service-CRN` header. It returns the backends that are configured for the instance that is passed in the header.

The endpoints that returns backends for an account is `GET /accounts/:id`.

Eric-Arellano: Oh, hm. Thanks for pointing that out. That being said, is there any issue with using `GET /v1/instance`? I like it because it includes the other metadata that the Resource Controller would return: limit and allocation

docs/guides/access-instances-platform-apis.mdx
```diff
@@ -1,44 +1,245 @@
...
+- `usage_allocation_seconds` — Integer, or `null`. The time allocated to this instance, used by the [fair-share scheduler](/docs/guides/fair-share-scheduler) to determine queue priority. See [Set instance allocation limits](/docs/guides/allocation-limits).
+- `backends` — Array of strings. The allowlist of backend names available to this instance. `["ANY"]` means all backends on the plan are available (the default). `[]` means no backends are available.
+
+<Admonition type="caution" title="The backends field may be outdated">
+  The `backends` field in the `extensions` object may be outdated. This can happen when IBM Quantum Support changes your account in a way that impacts instances. For example, when a backend is removed from an account, it will update the `backends` for the instance, but that change is currently not yet reflected in the Resource Controller API.
+
+  Instead, the current workaround is to use the [Qiskit Runtime REST API](/docs/api/qiskit-runtime-rest) with the [`GET /v1/backends`](/docs/api/qiskit-runtime-rest/backends#tags__backends__operations__list_backends) endpoint. (Make sure that you set the `Service-CRN` header to your instance's CRN.)
```
abbycross:
```suggestion
  Instead, the current workaround is to use the [Qiskit Runtime REST API](/docs/api/qiskit-runtime-rest) with the [`GET /v1/backends`](/docs/api/qiskit-runtime-rest/tags/backends) endpoint. (Make sure that you set the `Service-CRN` header to your instance's CRN.)
```
Something funky is going on with our anchors on that page and needs further investigation. This URL should work in the meantime, I think

Eric-Arellano: The issue isn't the anchor, but the bad URL path

abbycross (approved):
👍

PR #5345, opened 2026-07-06
Fix composer svgs

eharvey328:
- saw these svgs had our entire platform css defs in them (tailwind, monaco, fonts)...  over 30MB per file 😬


brownjedi (approved):
🚀

PR #5348, opened 2026-07-09
Syncs course metadata with Platform's displayed values

eharvey328:
The course titles and descriptions are hard-coded in the platform docs repo. They can be seen here: https://quantum.cloud.ibm.com/learning/courses.

this syncs those hard-coded values back to the source so we can pull the values instead of maintaining a second copy.

### DIFF
--- a/learning/courses/basics-of-quantum-information/_toc.json
+++ b/learning/courses/basics-of-quantum-information/_toc.json
@@ -1,6 +1,6 @@
 {
   "parentUrl": "/learning/courses",
-  "title": "Basics of quantum information",
+  "title": "Basics of Quantum Information",
   "collapsed": true,
   "children": [
     {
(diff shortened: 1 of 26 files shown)


abbycross:
Hm, the titles seen on [that page](https://quantum.cloud.ibm.com/learning/courses) are not consistent with IBM style when they use Camel Case - all titles should be Sentence case. Can you flag for me in the internal repo where we can update the capitalization? cc: @kcmccormibm

eharvey328:
> Hm, the titles seen on [that page](https://quantum.cloud.ibm.com/learning/courses) are not consistent with IBM style when they use Camel Case - all titles should be Sentence case. Can you flag for me in the internal repo where we can update the capitalization? cc: @kcmccormibm

Updated! Looks like all the titles are good then. Only descriptions need updating.

abbycross (approved):

PR #5416, opened 2026-07-23
add metadata to content

eharvey328:
Moving content metadata out from our docs app to the content here. This way the docs app does not need to keep separate data mappings and can simply pull and display the data.


axelhzf (approved):
🚀

Alienah (approved):

abbycross (approved):

PR #5419, opened 2026-07-24
Updated changelog with QR rename

raulotaolea:
Added a new July 2026 section under IQP section with QR rename.


docs/guides/latest-updates.mdx
```diff
@@ -10,17 +10,22 @@ in_page_toc_max_heading_level: 2
...
 
 {/* remember to update the date in the previous line each time this page is updated!!! */}
 
 Keep up with the latest and greatest from Qiskit and IBM Quantum&reg;!
 
 <Admonition type="note" title="What's new (last 60 days)">
 
+27 July 2026
+
+- **Qiskit Runtime Service is now IBM Quantum Compute Service** &mdash; This is a naming change only and does not require any code changes or updates to existing applications. You may notice the new name throughout the user interface and documentation. [*Read more*](#qr-rename)
```
abbycross:
```suggestion
- **Qiskit Runtime service is now IBM Quantum Compute Service** &mdash; This is a naming change only and does not require any code changes or updates to existing applications. You might notice the new name throughout the user interface and documentation. [*Read more*](#qr-rename)
```

docs/guides/latest-updates.mdx
```diff
@@ -10,17 +10,22 @@ in_page_toc_max_heading_level: 2
...
+---
+
 16 July 2026
 
-- **New pre-flight checks and new `backend` methods in Qiskit Functions v0.17.0** &mdash; Use the new methods in `qiskit-ibm-catalog` to check runtime capacity and backend access before you run, and `run()` now validates both before it submits a job. [*Read more*](#new-methods)
+- **New pre-flight checks and new `backend` methods in Qiskit Functions v0.17.0** &mdash; Use the new methods in `qiskit-ibm-catalog` to check runtime capacity and backend access before you run, and `run()` now validates both before it submits a job. [*Read more*](new-methods)
```
abbycross:
```suggestion
- **New pre-flight checks and new `backend` methods in Qiskit Functions v0.17.0** &mdash; Use the new methods in `qiskit-ibm-catalog` to check runtime capacity and backend access before you run, and `run()` now validates both before it submits a job. [*Read more*](#new-methods)
```

docs/guides/latest-updates.mdx
```diff
@@ -50,6 +55,11 @@ Jump to [IBM Quantum Platform](#latest-platform) | [Qiskit SDK](#latest-qiskit-s
 
 This section summarizes the recent enhancements and new features for the new [IBM Quantum Platform](https://quantum.cloud.ibm.com/).
 
+### July 2026
+
+<span id="qr-rename"></span>
+- **Qiskit Runtime Service is now IBM Quantum Compute Service** - Qiskit Runtime previously referred to both an internal runtime mechanism and IBM's managed execution service, creating confusion between open-source Qiskit and IBM's proprietary platform. The new naming clearly separates Qiskit, the open-source SDK for building portable quantum applications, from IBM Quantum Compute, the managed service for executing them on IBM quantum hardware. This provides a clearer architecture without changing workflows, APIs, access plans, or the customer experience.
```
abbycross:
```suggestion
- **Qiskit Runtime service is now IBM Quantum Compute Service** &mdash; Qiskit Runtime previously referred to both an internal runtime mechanism and IBM's managed execution service. The new naming clearly separates Qiskit, the open-source SDK for building portable quantum applications, from IBM Quantum Compute, the managed service for executing on IBM Quantum hardware. This provides a clearer architecture without changing workflows, APIs, access plans, or the user experience.
```

docs/guides/latest-updates.mdx
```diff
@@ -448,9 +453,9 @@ Wondering what has changed? Read the [changelog](https://github.com/Qiskit/qiski
 *To see the release notes for all versions, visit the [Qiskit Runtime client release notes](/docs/api/qiskit-ibm-runtime/release-notes).*
 
 <span id="latest-runtime-service"></span>
-## Qiskit Runtime service
+## IBM Quantum Compute Service
 
-This summary of the latest changes to the Qiskit Runtime service applies to anyone using Qiskit Runtime, regardless of how you communicate with the service (by using `qiskit-ibm-runtime` or otherwise), or which version of the client SDK you use.
+This summary of the latest changes to the IBM Quantum Compute Service applies to anyone executing on IBM Quantum hardware, regardless of how you communicate with the service (by using `qiskit-ibm-runtime` or otherwise), or which version of the client SDK you use.
```
abbycross: Please check my rewording here @raulotaolea

raulotaolea: @abbycross In this case, this section refers to the Qiskit Runtime execution environment that runs on the quantum computer itself. It does not refer to the service hosted on IBM Cloud. Therefore, Qiskit Runtime should be kept here.

That said, we should avoid using the term "service" in this context to prevent confusion. Although it is technically a service—internally exposed through the Quantum System API (QSA), which both IQP and HPC integrations interact with at a low level—we want to clearly distinguish it from the IBM Cloud offering, now called IBM Quantum Compute Service.

To make that distinction explicit, I suggest referring to it as IBM Qiskit Runtime. This makes it clear that we are talking about IBM's proprietary implementation of the Qiskit Runtime Execution Environment, rather than the IBM Cloud offering, now called IBM Quantum Compute Service.

abbycross: We also have a section in the changelog called [Qiskit Runtime client](https://quantum.cloud.ibm.com/docs/en/guides/latest-updates#qiskit-runtime-client-0460) - are we distinguishing that with enough clarity in this case?

abbycross: Also @jyu00 do you have a take on this?

jyu00: I think leaving it as `Qiskit Runtime` will make things more confusing 🙂. 

Internally we often refer to the non-cloud part of the software stack as Qiskit Runtime, but I don't think that's true externally. Our product is listed as Qiskit Runtime _Service_, and the first thing you initialize in the client SDK is `QiskitRuntimeService`. 

My proposal is to take this opportunity and combine the existing Qiskit Runtime and IQP sections under a single header - `IBM Quantum Compute Service`. Especially since the division is a bit arbitrary right now (e.g. `store` is listed under IQP but `measure_2` under QRT, even though they're both part of EDC support).

raulotaolea: Overall, I agree. However, I think Qiskit Runtime, understood as the execution environment for Qiskit's primitives on the quantum computer, is a legitimate concept that we do need to refer to from time to time. The problem is that, until now, we've also used the same name for the cloud service, which leads to situations like the example you just mentioned with `Store`.

The product rationale behind the name `qiskit-ibm-runtime` is that any hardware provider that wants Qiskit users to run workloads on its quantum computers must provide two provider-specific components:

- A Qiskit extension (formerly provider) that enables users to submit workloads to the provider's cloud or execution service. In IBM's case, this is the `qiskit-ibm-runtime` library, which communicates with IBM Quantum Compute Service.
- A runtime (execution environment) that processes Qiskit primitives, translates them into hardware-specific operations, and executes them on the provider's quantum hardware. In IBM's case, this is the IBM Qiskit Runtime Execution Environment, which runs on the selected quantum computer.

Historically, both the cloud service and the execution environment were referred to as Qiskit Runtime, which is why the client class was named `QiskitRuntimeService`. With the introduction of IBM Quantum Compute Service, the architecture becomes much clearer: the client library connects to the cloud service, which dispatches the workload to the selected quantum computer, where the runtime execution environment processes and executes the primitives.

I think we should reflect this architecture more explicitly in this changelog page. We should keep IBM Quantum Platform (we're working on this in parallel), create an IBM Quantum Compute section that clearly establishes it as the cloud service, and keep Qiskit Runtime as a separate concept -- perhaps renaming it to something like IBM Qiskit Runtime Execution Environment -- to refer specifically to the runtime that executes Qiskit primitives on the quantum computer.

The remaining inconsistency is that the client library is still called `qiskit-ibm-runtime` and exposes the `QiskitRuntimeService` class rather than `IBMQuantumComputeService`. @jyu00, I think we can address that by introducing the alias you proposed. This would preserve backward compatibility while making the naming consistent with the architecture and giving us a much clearer overall framework.

jyu00: > Historically, both the cloud service and the execution environment were referred to as Qiskit Runtime

Which is why I'm suggesting avoid the term `Qiskit Runtime` altogether. It'd be hard for people to realize it means two things, and we are only renaming one of them. 

> I think we should reflect this architecture more explicitly in this changelog page. We should keep IBM Quantum Platform (we're working on this in parallel), create an IBM Quantum Compute section that clearly establishes it as the cloud service, and keep Qiskit Runtime as a separate concept -- perhaps renaming it to something like IBM Qiskit Runtime Execution Environment -- to refer specifically to the runtime that executes Qiskit primitives on the quantum computer.

I don't think most users can tell you the difference between IQP, IBM Quantum Compute, and Qiskit Runtime Execution Environment. In fact I'm not sure I can. Take the new `Executor` primitive as an example. My Executor job runs in the QRT environment, but I submit the job to IBM Quantum Compute and view the result on IQP. So where should this announcement go?  More importantly, do users care about the differences? They just know they use IBM Quantum Compute service, so why not just have a single section for the product?

francabrera: > I don't think most users can tell you the difference between IQP, IBM Quantum Compute, and Qiskit Runtime Execution Environment.

I agree with that. There is a similar question being asked to some internal users by our UX Research team because a similar topic/question was discussed on how some questions in the annual survey were formulated splitting our product in components/areas that our external users may not know how to differentiate.

Personally, I would not differentiate them at that granular level. It may make sense to differentiate "python packages", "web interfaces", "rest API"... but probably not as a way of groping but to describe the update itself, as part of the content.

kaelynj: > I don't think most users can tell you the difference between IQP, IBM Quantum Compute, and Qiskit Runtime Execution Environment.

I agree with this as well, the granular distinction between the IBM Quantum Compute service and the Qiskit Runtime Execution environment is not a helpful one _at the user facing level_.

> This makes it clear that we are talking about IBM's proprietary implementation of the Qiskit Runtime Execution Environment, rather than the IBM Cloud offering, now called IBM Quantum Compute Service.

Users interface with these two systems through either the `qiskit-ibm-runtime` package or through the REST API _as if they were the same thing._ To my knowledge there is no ability to interact with them separately and so to identify them as such is not helpful information (and will likely just lead to confusion) as this level of granularity between these systems isn't documented anywhere else. (Primarily because it isn't relevant to the workflow of submitting jobs).

raulotaolea: I know users don't distinguish between these concepts today. My understanding is that the goal of this effort is precisely to make the architecture clearer and reduce that ambiguity over time.

That's why I think it's useful to distinguish between IBM Quantum Platform, IBM Quantum Compute, and Qiskit Runtime in the places where we explicitly refer to those concepts.

@jyu00 , if you're saying that going forward we'll no longer explicitly refer to the runtime, then I'm perfectly happy to remove it and update those references to IBM Quantum Compute instead.

As long as we have a section where we can document each new feature or change we want to include in the changelog, I'm happy with that approach. 

Right now, we have the following sections:
- IBM Quantum Platform
- Qiskit SDK v2.3
- Qiskit Runtime client 0.46.0
- Qiskit Runtime service
- Qiskit Transpiler Service client v0.15.2
- Qiskit addons
- Qiskit Functions
- Qiskit Code Assistant

Then the new entries would be:
- IBM Quantum Platform
- IBM Quantum Compute Service
- Qiskit SDK v2.3
- Qiskit Runtime client 0.46.0
- Qiskit Transpiler Service client v0.15.2
- Qiskit addons
- Qiskit Functions
- Qiskit Code Assistant

If you agree, this would mean renaming [Qiskit Runtime service](https://quantum.cloud.ibm.com/docs/en/guides/latest-updates#qiskit-runtime-service) section to "IBM Quantum Compute Service". @jyu00 @francabrera are you still comfortable having items such Directed execution model, Dynamic circuits, and similar features under the new "IBM Quantum Compute" section title?

jyu00: My suggestion was actually to merge the existing `IBM Quantum Platform` section with the `Qiskit Runtime` section and name it `IBM Quantum Compute Service`. This merged section would then cover all server-side changes (besides Functions), so people don't have to try to understand the different component names. Functions are separated out because it's a separately licensed entity.

abbycross: I've taken a stab at combining the two as Jessie suggests, and interleaved the archival entries.

abbycross: Since this PR doesn't have a preview, here are screenshots from my local build
Screenshot 2026-07-28 at 5 23 24 PM
Screenshot 2026-07-28 at 5 23 30 PM

raulotaolea: I wrote the text in the screenshot assuming we would keep a dedicated section for Qiskit Runtime, understood as the execution environment, and I prefixed it with IBM to make it clear that it refers to IBM's implementation of Qiskit Runtime.

Now that we've decided to simplify the model and present IBM Quantum Compute Service as the main concept, I'm not sure whether that explanation still makes sense. My original intention was simply to explain the different components involved in the execution flow and how they relate to each other.

Do you think it's still worth describing those underlying components in this context, even if we don't expose them elsewhere or have dedicated sections for them? Or would you rather treat IBM Quantum Compute Service as a black box and avoid introducing those implementation details altogether?

If we go with the latter approach, we could write something along these lines:

> Qiskit Runtime Service is now IBM Quantum Compute Service — This is a naming change only and does not require any code changes or updates to existing applications. You might notice the new name throughout the user interface and documentation.
> 
> IBM Quantum Compute Service is the quantum execution service within IBM Quantum Platform for running workloads on IBM Quantum systems. It provides the cloud capabilities required to execute workloads reliably at scale, including authentication, scheduling, queue management, workload routing, usage tracking, limits, and analytics.
> 
> The open-source `qiskit-ibm-runtime` library continues to be the client used to submit workloads from Qiskit applications. Existing APIs, workflows, access plans, and the user experience remain unchanged.
> 
> This renaming clarifies the role of the service without changing how applications are developed or executed.

raulotaolea: > My suggestion was actually to merge the existing `IBM Quantum Platform` section with the `Qiskit Runtime` section and name it `IBM Quantum Compute Service`. This merged section would then cover all server-side changes (besides Functions), so people don't have to try to understand the different component names. Functions are separated out because it's a separately licensed entity.

I agree with merging the existing _Qiskit Runtime_ section into _IBM Quantum Compute Service_. That aligns well with the goal of simplifying the terminology presented to users.

However, I don't think the same applies to the _IBM Quantum Platform_ section. Regardless of the rename itself, from a product perspective I think it's important that we continue to distinguish between the platform and the services it provides.

_IBM Quantum Platform_ is the platform. _IBM Quantum Compute Service_ is one of the services within that platform. They represent different layers of the product hierarchy, and I think our changelog should reflect that distinction rather than hide it. Otherwise, platform-level changes -- such as documentation, learning content, Composer, and similar capabilities -- would gradually end up being described as changes to IBM Quantum Compute Service, even though they belong to IBM Quantum Platform.

abbycross (approved):

PR #5422, opened 2026-07-25
Fix missing links on API attributes

eharvey328:
see https://github.com/Qiskit/documentation/issues/5311

The Attribute href was not getting captured. Added a new href prop to be consumed by the Attribute component in docs app. 

- migrated `ibm-qiskit-runtime` api v0.48.0 (latest)


scripts/js/lib/api/generateApiComponents.ts
```diff
@@ -97,7 +103,11 @@ function prepareProps(
   apiType: ApiObjectName,
   id: string,
   headerLevel: number,
-  options: { isCApi: boolean },
+  options: {
+    isCApi: boolean;
+    kebabCaseAndShorten?: boolean;
```
axelhzf: nit: `kebabCaseAndShorten` seems to describe how to do something, rather than what it's used for

What do you think about doing something like

```
getPathHref: (path) => kebabCaseAndShortenPage(hrefPath, pkgName)
```

eharvey328: It comes from these pkg options. Idk if that makes it feel more clear

https://github.com/Qiskit/documentation/blob/attribute-hint-href/scripts/js/lib/api/processHtml.ts#L30-L40

eharvey328: I changed it to use a callback called `normalizeUrl` https://github.com/Qiskit/documentation/pull/5422/commits/aec2a2045c2aeacbf66e399305bf7034737d400a

axelhzf (approved):

PR #5437, opened 2026-07-28
Fix python check and check-tutorials-index on Windows

### LINKED ISSUE #2835: `./check` script fails on Windows
I'm not sure what's going on. Running `tox -e lint` passed, but then `npm run check` fails with this:

```
  File "C:\1docs\check", line 24, in <module>
    main()
  File "C:\1docs\check", line 19, in main
    run_cmd(["npm", "run", "check"])
  File "C:\1docs\scripts\tool_runner\__init__.py", line 22, in run_cmd
    subprocess.run(cmd, check=True)
  File "C:\Python312\Lib\subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python312\Lib\subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Python312\Lib\subprocess.py", line 1538, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

Even weirder, `npm run check` works when run directly in the command line. This seems like an issue with how we implement `subprocess.run()` maybe.

micpap25:
Fixes #2835

Since `npm` is actually a wrapper on Windows for `npm.cmd`, use `shutil` to determine the correct `npm` command.
This exposed as well that `check-tutorials-index.py` was broken on Windows. This was due to two issues.
1. The path of the `expected_link` was not formatted correctly for Windows (front-slash vs. back-slash). Use `as_posix()` to fix this.
2. There are characters not in Window's default character set in the notebooks (Emojis, accented letters, fancy quotes). I removed all the fancy quotes and then set the encoding as `utf-8` explicitly. (Perhaps other encodings can explicitly use `utf-8` for cleanliness?) 

EDIT: I noted down the notebooks that have special characters which require explicitly using `utf-8` for the future.
```
docs\tutorials\global-data-quantum-optimizer.ipynb
docs\tutorials\loop-string-hadron-dynamics.ipynb
docs\tutorials\readout-error-mitigation-sampler.ipynb
docs\tutorials\warm-start-qaoa.ipynb
```


scripts/tool_runner/qiskit_docs_contributor_tools/__init__.py
```diff
@@ -37,6 +37,7 @@ def run_cmd(name: str, cmd: list[str], *, progress: str | None) -> None:
         stderr=subprocess.STDOUT,
         text=True,
         encoding="utf-8",
+        check=False
```
Eric-Arellano, line 40: Why this change? Isn't it the default?

check
```diff
@@ -23,16 +24,18 @@ from scripts.tool_runner.qiskit_docs_contributor_tools import (
 
 logger = logging.getLogger(__name__)
 
+npm_cmd = shutil.which("npm")
```
Eric-Arellano, line 28: Could you please add a brief comment explaining why we do this?

Eric-Arellano (commented):
Nice job! Thanks for taking the time to fix this, and good write up in the issue. You're correct that shell=True is heavily discouraged, so this approach is better

scripts/tool_runner/qiskit_docs_contributor_tools/__init__.py
micpap25: So, it is the default, but pylint recommends that it be explicit: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/subprocess-run-check.html

Eric-Arellano: Hm, I find that noisy and we don't use Pylint in this repo, but that's fine.

Eric-Arellano (approved):
Thank you!

abbycross:
I think the Execute notebooks check failure means we need to move this into a write-access branch, so I'll merge it into a new one.

PR #5439, opened 2026-07-29
Update Attribute to include "Type" and "Defalut value" as children

eharvey328:
- removes `attributeTypeHint` and `attributeTypeHintHref` props on `Attribute` mdx component.
  - now the type and default value are added to the body as markdown to be parsed in docs
  -  allows multiple types with multiple links this way (e.g., `method: QiskitType | dict[Qiskititem] | None`)  
  - this also makes it so hrefs get automatically transformed by the pipeline since they are no longer a prop string  

- did not implement an ignore list for links to python standard types because these links are already display in other places in the docs. If we want we can handle in a follow up. 

### Packages migrated:

|  Package | Version	|  
|----------|:-------------:|
| qiskit | 2.5.2 (latest), 2.6.0-dev (dev) |
| qiskit-ibm-runtime | 0.49.0 (latest), 0.50.0-dev (dev) |
| pauli-props | 0.2.0 (latest) |
| qiskit-addon-aqc-tensor  | 0.3.1 (latest) |
| qiskit-addon-cutting | 0.10.0 (latest) |
| qiskit-addon-mpf | 0.3.0 (latest) |
| qiskit-addon-obp | 0.3.0 (latest) |
| qiskit-addon-paulice | 0.2.1 (latest) |
| qiskit-addon-pna | 0.2.0 (latest) |
| qiskit-addon-slc | 0.1.0 (latest) |
| qiskit-addon-sqd | 0.13.1 (latest) |
| qiskit-addon-utils | 0.4.0 (latest) |


axelhzf (dismissed):

eharvey328:
@axelhzf I got the inline-code styling to work

The challenge is the source does not output these as inline-code blocks so we have to add them somewhere. 
But we can't just wrap the entire thing because we lose the links 
```mdx
`[Value](href) | [Another](href)` // links becomes a string
```

We need to parse [each token individually](https://github.com/Qiskit/documentation/pull/5439/changes/7775115ef64cb9c44cdb984933aa8bbdf94dab03#diff-9a468d2cdf072516097df6c928349b3f9396c0a79c93b6b18b0b7ef46e4802c0R594-R604):
```mdx
[`Value`](href) | [`Another`](href)
```

That works for the type, but default value can have complex values with spaces and chars:
```mdx
Default value:  \<default case>
```

Since this represents a value, it is safe to assume there won't be any type def links. So we can [wrap the whole thing in a code block](https://github.com/Qiskit/documentation/pull/5439/changes/7775115ef64cb9c44cdb984933aa8bbdf94dab03#diff-9a468d2cdf072516097df6c928349b3f9396c0a79c93b6b18b0b7ef46e4802c0R322)

axelhzf (dismissed):

axelhzf (approved):

Alienah (approved):

PR #5482, opened 2026-08-06
Update changelogs

beckykd:


abbycross:
Thanks Becky! Excited for this change...

One minor thing - instead of `changes` in the new filenames, how about `changelog`

We'll also need to add a redirect and decide where it should point (probably the Compute Service one?...)

mackblank:
Looking good so far!! A couple design notes - 

The heading for the IBM Quantum Compute Service changelog is not the same as the others - it looks like it's a section heading and not a page title (which trickles down to the date headers- those are not the same size as what's on the other pages)
image
image

Also I'm not sure how I feel about older updates being kept in the accordions still - it feels odd to me for the older dates to be visible in the in-page nav but in a closed accordion by default

abbycross (approved):
Whoo hoo - thanks!

mackblank (approved):
Love it!

PR #5406, opened 2026-07-21
Configure functions access

abbycross:
Incorporate the docs provided by @riley2012 into the [resource controller page](https://quantum.cloud.ibm.com/docs/en/guides/access-instances-platform-apis).

File to review: https://qiskit.github.io/documentation/pr-5406/docs/guides/access-instances-platform-apis

PR #5488, opened 2026-08-10
Add Functions access configuration section to Get started guide

pandasa123:
## Background

Adds a **Configure Functions access on an instance** section to the Get started with Qiskit Functions guide, documenting how account administrators grant an instance access to catalog and custom Qiskit Functions through the IBM Cloud Resource Controller REST API. 

The section provides cURL and Python examples for looking up the account-level configuration, granting access, removing access, and cross-links the shared bearer-token and CRN setup to the existing REST guides. It also adds a pointer to this section from the instance-management guide.


pandasa123:
Ideally, this should be collapsed in an accordion or something to start @abbycross, but I wasn’t sure how to do that

Also, this guide is getting large, might be worth thinking about re-organising / subsite

pandasa123:
Closing since a [better set of docs are coming](https://github.com/Qiskit/documentation/pull/5406)

PR #5496, opened 2026-08-11
Update twirling default in Sampler options

jyu00:
I noticed it says twirling is enabled by default for Sampler - this is incorrect. It's only enabled by default for Estimator.


abbycross (approved):

abbycross:
Thanks!

PR #5499, opened 2026-08-11
add new QMOO tutorial

kcmccormibm:
New portfolio optimization tutorial using the QMOO method introduced in [Kotil, et al.](https://arxiv.org/pdf/2503.22797)

# Summary

Adds a new tutorial, Quantum approximate multi-objective optimization applied to a portfolio optimization problem. The tutorial shows how to trace the risk/return/diversification Pareto front of a cardinality-constrained portfolio with QAOA. The cardinality constraint is enforced with an XY mixer, the three objectives are scalarized with the weighted-sum method, and the weights are swept to map out the Pareto front.

# What's included

- Small-scale walkthrough (8-asset, exact simulation) that can be verified by brute force.
- Large-scale example (40 assets, 8 GICS sectors, K=6) sized for IBM Heron hardware, using QAP-truncation to sparsify the covariance and moocore for hypervolume and dominated-solution filtering.
- QAOA angles trained classically (exact statevector at small scale, MPS at 40 qubits); pre-trained angles ship in instances/ so the slow training step is skipped by default.
- Real market data and parameters file in `datasets/tutorials` and dowloaded at runtime

# Testing / notes
Hardware section selects the least-busy Heron device via least_busy(min_num_qubits=156) (the 156-qubit threshold restricts selection to Heron, excluding Eagle and Nighthawk).
Adapted from Kotil et al. (2025), Nature Computational Science (arXiv:2503.22797), which develops the method for max-cut.


henryzou50 (commented):
Thanks Katie, this is a nice tutorial and great work here!

**Housekeeping I pushed today** (so you don't have to): merged `main` into the branch to resolve the conflicts (kept both sides everywhere; restored the `implicit-solvent-calculations` entry in `qiskit_bot.yaml` that was commented out), ran `tox -e fix` on the notebook (import split + formatting), and ran prettier on `qaoa_params.json` (array layout only, the values are untouched). Lint should be green now. One heads-up for later: the retrain branch writes `qaoa_params.json` via `json.dump`, which produces formatting prettier rejects, we likely need to rerun `npx prettier --write` on it whenever the angles are regenerated.

I also ran the workflow end-to-end locally, and it nicely reproduces the committed outputs exactly. But while verifying I found two correctness bugs that affect the hardware results, plus a couple of potential blockers:

### 1. Trained γ is bound with the wrong sign (circuit anti-optimizes)

`qaoa_training_pipeline` maximizes energy, so we train on the negated op (`training_op = sum(-1/n_obj * H_k)`), but the trainer's evaluator internally builds its ansatz from that same negated op, i.e. it optimizes angles for exp(+iγH). The notebook then binds those angles into `qaoa_ansatz(combined_cost_op)` = exp(−iγH), which differs by γ -> −γ.

I verified this by replicating the small-scale pipeline exactly (same Hamiltonians, trainer, seed, and it reproduces the committed trained angles β=3.329, γ=3.445) and comparing the shot-weighted scalarized objective `Σ c_k f_k(x)` across 60 weight vectors:

| binding | scalarized objective (lower = better) |
|---|---|
| notebook binding | −1.834 -- worse than random on **60/60** weight vectors |
| negated-op fix | −2.017 -- better than random on **60/60** |
| random feasible baseline | −1.926 (best possible: −2.121) |

It's invisible in the committed small-scale outputs only because 200×500 shots over 8 qubits sample all 256 bitstrings, so the distinct-portfolio union enumerates the whole space regardless of what the circuit favors.

** Possible fix (either):** build the sampling ansatz from `sum(-c[k] * H_k)` so it matches the trained circuit (this is what I tested above), or bind `-opt_gammas`. The shipped `qaoa_params.json` was trained the same way (per its `note`), so would apply to the hardware section too.

### 2. Bitstrings are never reversed before scoring/reporting

`get_counts()` returns qubit 0 (= asset 0) at the *rightmost* position. The return Hamiltonian's Z on qubit 0 carries coefficient μ₀/2, but `evaluate_portfolio`/`evaluate_40` and the "tickers held" table read left-to-right, so every portfolio is scored and reported as its mirror image (asset i <-> asset N−1−i). Same small-scale masking as above, but at 40 assets the printed Pareto ticker lists are the wrong portfolios, and the hypervolume comparison scores scrambled candidates (the committed 14.145 vs 12.973 margin is likely an artifact of bugs 1+2 partially offsetting).

** Possible fix:** reverse once at collection, e.g. `bs = bs.replace(" ", "")[::-1]`, in both collection loops (the warm-start QAOA tutorial handles this explicitly). Post-selection should be unaffected.

### 3. Blockers for anyone running the notebook fresh

- **`qaoa-training-pipeline` isn't on PyPI**, and its GitHub `main` just removed `ScipyTrainer.train()` (replaced by `provide_params()`), so a fresh install breaks the notebook. Only the `v0.1.0` tag works. The Requirements cell should give the exact command: `pip install "git+https://github.com/qiskit-community/qaoa_training_pipeline.git@v0.1.0"` (and `pip install qiskit-addon-opt-mapper`). Might be worth asking whether a PyPI release is planned, since git-main drift will silently break this tutorial again.
- The `load_params_file = False` branch crashes after training: `os.makedirs(os.path.dirname(params_path))` with a bare filename -> `makedirs("")` raises `FileNotFoundError`. We can just drop that line.

### 4. Consistency with other tutorials (smaller items)

- [ ] Add a `## References` section for Kotil et al. (see warm-start-qaoa for the `[\[1\]](#Reference1)` pattern)
- [ ] Wrap Next steps in `<Admonition type="tip" title="Recommendations">`, the "companion unconstrained notebook" bullet points to something that doesn't exist in this repo. Either remove or link it. 
- [ ] Add a job tag: `sampler_hw.options.environment.job_tags = ["TUT_QAMOO"]`
- [ ] Remove the `version-info` cell (this is a guides-only convention; no tutorial has one and it'll never get populated since this notebook is CI-excluded) and the `hardware-unrun` tags (no tooling reads them; `[groups.exclude]` already controls testing)
- [ ] The market-data cell (first hardware cell) has no committed output, after the fixes above, please re-run the whole notebook top-to-bottom so outputs, figures, and execution counts are regenerated cleanly

### 5. Wording nits 

- [ ] "The constraint is built into the circuit's structure, **for free**" is slightly oversold given post-selection is needed two sentences later. I suggest noting the product-state init trades depth for discarded shots
- [ ] "a penalty term … **does not fit on hardware**" -> "leads to much deeper circuits on hardware"

kcmccormibm:
Thanks @henryzou50 and sorry for the delay! All your comments have been addressed, but I'm now concerned that at the 40-asset scale the QAOA hypervolume no longer clearly beats random. I'll continue looking at this next week to see if I can improve the performance, and welcome any ideas you might have!

henryzou50 (commented):
Thanks Katie! I reran the small example successfully and confirmed that the sign and bitstring-ordering fixes are in place.

On beating random: I think an honest comparison might be enough. One helpful finding: choosing 6 of 40 assets gives 3,838,380 feasible portfolios. I enumerated them and computed the 61-point exact Pareto front in about two seconds locally. Comparing against that front, plus several random seeds with a fixed hypervolume reference point, could help clarify the performance.

A few things still need attention before merging:
- Finish regenerating the outputs. The committed market-data cell still shows a 404 error, and the subsequent hardware cells have no outputs. We’ll need the corrected results and figures to assess the hardware example.
- Restore stable dataset URLs. Both downloads now point to `km/add-qmoo-portfolio`. Those should use main or an immutable commit for publication so deleting the PR branch won’t break the tutorial.
- Fix the second reference. arXiv:2101.05742 is Sack and Serbyn’s Quantum annealing initialization of the quantum approximate optimization algorithm, but the bibliography and initialization comment attribute it to Zhou et al.
- Run the formatting/normalization checks again after regenerating the notebook.

My suggestion would be to establish the corrected comparison first, then decide whether further tuning is needed. If the hardware result is comparable to random, we can say that directly and explain the limitations without overstating what the example demonstrates.

kcmccormibm:
Thanks @henryzou50! All your remaining points should be addressed now. I went ahead and added a few sentences explaining that the QAOA result was comparable to random and what we would need to change to improve it.

henryzou50 (commented):
Thanks for the updates, @kcmccormibm! The sign fix (checked against `ScipyTrainer` in `v0.1.0`), the bit-order fix, and the exact-front benchmark all look right.

I pushed two commits: a merge of `main` to fix the conflict, and a copyedit pass to make this more align with other tutorials. Please pull before making more changes.

**Before merge:**
1. **Lint:** `ruff format` and squeaky both fail. Could you move the long inline comments onto their own lines, then run `tox -e fix`?
2. **Brute force:** the large-scale intro says 40 assets is "too large to verify by brute force," but the new benchmark enumerates all 3.8M portfolios. Maybe say the 2^40 statevector is what's too large.
3. **Benchmark wording:** QAOA gets 81.4% vs. random's 83.8% ± 1.3%, which is below random, not "comparable." Could we word it so it holds after a re-run, e.g. "QAOA does not beat uniform random sampling at this size"?
4. **Front comment:** "the front IS the set of answers" overclaims (15 sampled vs. 61 exact). Maybe "the best trade-offs found by the sampler"?
5. **Framing:** since hardware doesn't beat random, could the intro say early that this teaches the workflow, not a quantum advantage?

**Optional (benchmark cell cleanup):**
- **One scoring helper.** The same risk/return/diversification math appears three times: in `evaluate_40`, in the exact enumeration, and in `random_feasible_hv`. A single vectorized helper that takes an array of selected-asset indices could replace all three, so a later change to an objective can't make them disagree:
  ```python
  def score_40(P):
      """Score portfolios given as an (n, K) array of asset indices."""
      risk = sigma_40[P[:, :, None], P[:, None, :]].sum((1, 2))
      div = D_40[P[:, :, None], P[:, None, :]].sum((1, 2))
      return np.column_stack([-risk, mu_40[P].sum(1), div])
  ```
  `evaluate_40` could then convert a bitstring with `np.flatnonzero` and call `score_40`.
- **Compute the exact front once.** Right now `hypervolume` runs on all 3.8M portfolios, and `filter_dominated` runs on them again inside the `print`. Computing `front_exact = filter_dominated(fis_exact, maximise=True)` once and taking the hypervolume of those 61 points gives the same number and removes the second filter pass.
- **Memory.** `np.array(list(itertools.combinations(...)))` builds about 3.8M Python tuples first, a few hundred MB briefly. `np.fromiter(itertools.chain.from_iterable(...), dtype=np.int16).reshape(-1, K_40)` skips the tuples.
- **Small example.** 200 × 500 shots covers all 256 bitstrings (all 70 valid portfolios), so the 8-asset run is effectively brute force. That's fine as a sanity check, but the text shouldn't suggest QAOA found the front there.

PR #5511, opened 2026-08-13
Fix DD documentation inconsistency for reset qubits

t-imamichi:
The DD option and the figure shown below are inconsistent.

Because `dynamical_decoupling.skip_reset_qubits` defaults to False, the XX sequence is also applied to q2.
To match the figure, `dynamical_decoupling.skip_reset_qubits=True` should be specified.

In general, setting `dynamical_decoupling.skip_reset_qubits=True` provides better accuracy because it avoids applying DD sequences to reset (clean) qubits, as illustrated in the figure. However, the default value remains False for backward compatibility.

Reference: https://quantum.cloud.ibm.com/docs/en/guides/estimator-options#available-options

image

### DIFF
--- a/docs/guides/error-mitigation-and-suppression-techniques.ipynb
+++ b/docs/guides/error-mitigation-and-suppression-techniques.ipynb
@@ -115,7 +115,8 @@
    "source": [
     "estimator = Estimator(mode=backend)\n",
     "estimator.options.dynamical_decoupling.enable = True\n",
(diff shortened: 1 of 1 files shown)


t-imamichi:
The CI error does not appear to be related to my fix.

> AccountNotFoundError: "Unable to find account. Please make sure an account with the channel name 'ibm_quantum_platform' is saved."

PR #5533, opened 2026-08-18
Small changes to changelogs

abbycross:
We don't want the filenames to include the version number, unless we want to add new files with every new version.


mackblank:
ooh seeing the updates in the in-page nav is neat! ibm_torino is quite a bit larger than that label size text though, any way we could get that to be the same text size? 
image

kaelynj (approved):

abbycross:
@mackblank I remember long ago we used to have that trouble with code font in headers, and it might have been resolved by adjusting the css?? Would need assistance from @Qiskit/frontend for it - let's do so in a followup!

PR #5559, opened 2026-08-24
Update nbtester deps

emilkovacev:
nb-tester dependencies do not include dependencies required by tutorials. This PR attempts to sync dependencies to ensure they are up-to-date.

~There are two dependencies missing from requirements.txt file that need to be added eventually:~
* ~qctrl-visualizer~
* ~imblearn~

^Update: It turns out both dependencies had different import names and pip install targets, so the imports are actually valid.

These dependencies require additional fixes applied to tutorial notebooks, so I am saving them for a follow-up PR.


kaelynj (approved):
LGTM thanks! 🚀

PR #5564, opened 2026-08-24
Remove old channel code

jyu00:
Remove the old `channel="ibm_quantum"`, which is no longer supported (even though it's commented out).


abbycross (approved):
Good catch, thanks @jyu00!

PR #5588, opened 2026-08-31
Changed index to j

### LINKED ISSUE #5581: Simultaneous use of "i" as subscript and imaginary number in Module "Get started with Qiskit in the classroom"

URL: https://quantum.cloud.ibm.com/learning/en/modules/quantum-mechanics/get-started-with-qiskit
Section: Quantum Computers
Sub-Section: The quantum phase
First through third paragraphs

### Select all that apply

- [ ] new content request
- [x] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

The paragraph explains that each amplitudes $c_0$ and $c_1$ can be written in a general form $c_i$ = | $c_i$ |e^{i$φ_i$}, where $φ_i$ is the phase. The second paragraph mentions "factoring out" e^{i$φ_0$}. The third paragraph introduces a different way to express the quantum state as |v> = | $c_0$ ||0> + | $c_1$ |e^{iφ}|1>.

Focusing only on notation, the "i" is being used simultaneously as: (a) a subscript to denote an index for the values 0 and 1 that are used with the amplitudes for the qubit states |0> and |1>; and (b) a superscript to denote the imaginary number i (i.e., square root of negative 1).

For beginners in a classroom who may want to do the math and derive the rewritten equation by hand, they may get confused and treat "i" only as a subscript and miss the fact that the imaginary number i plays an important role in the relative phase.

I suggest replacing the subscript with a different letter such as j. The first paragraph would then present the general form of the amplitudes as $c_j$ = | $c_j$ |e^{i$φ_j$}, where $φ_j$ is the phase. No changes would be necessary to the second and third paragraphs as they are currently written.

Please note in this request I tried to use notation to properly denote subscripts and superscripts. Content in between {} are meant to be superscripts (exponents). For example, in e^{i$φ_j$} I'm trying to say that e is raised to an exponent, where this exponent is the product of the imaginary number i and $φ_j$

(No answer)

christopherporter1:
Fixes #5581 
Changed index from i to j to avoid confusion with imaginary unit.


abbycross (approved):
Approving to unblock - thanks!

PR #5590, opened 2026-08-31
move punctuation outside math and link expressions in learning

abbycross:
Because the machine translation does not read the content inside math or hyperlink expressions, moving punctuation outside of those expressions so that translation does not break. 

Also removing some expired "isNew" pills from the tocs


learning/courses/quantum-machine-learning/classical-ml-review.ipynb
```diff
@@ -106,7 +106,7 @@
...
+    "For some feature maps, it may be possible to write the inner product of feature-mapped vectors $\\vec{\\Phi}^T(\\vec{x}_i)\\cdot \\vec{\\Phi}(\\vec{x}_j)$ as a simple function $k(\\vec{x}_i, \\vec{x}_j)$ of the original (lower-dimensional) variables $\\vec{x}_i$ and $\\vec{x}_j$. For some choices of $\\Phi$, we may even be able to write $\\vec{\\Phi}^T(\\vec{x}_i)\\cdot \\vec{\\Phi}(\\vec{x}_j)$ as a simple function of the lower-dimensional inner product $\\vec{x}_i^T\\vec{x}_j$ This is computationally very beneficial because we can access the space in which data are linearly separable, but without the cost of manipulations in higher dimensions. In fact, since the feature-mapped vectors only appear in the $f_2$ in inner products, we may not even need to explicitly carry out the feature mapping to compute the inner products.  We call the function $k(\\vec{x}_i, \\vec{x}_j)$ that calculates the inner products the “kernel function”, and this way of avoiding the feature map computation is called the “kernel trick”. In fact, the feature mapped vectors could even be infinite dimensional, but the kernel might still be very efficiently computable.\n",
```
abbycross:
```suggestion
    "For some feature maps, it may be possible to write the inner product of feature-mapped vectors $\\vec{\\Phi}^T(\\vec{x}_i)\\cdot \\vec{\\Phi}(\\vec{x}_j)$ as a simple function $k(\\vec{x}_i, \\vec{x}_j)$ of the original (lower-dimensional) variables $\\vec{x}_i$ and $\\vec{x}_j$. For some choices of $\\Phi$, we may even be able to write $\\vec{\\Phi}^T(\\vec{x}_i)\\cdot \\vec{\\Phi}(\\vec{x}_j)$ as a simple function of the lower-dimensional inner product $\\vec{x}_i^T\\vec{x}_j$. This is computationally very beneficial because we can access the space in which data are linearly separable, but without the cost of manipulations in higher dimensions. In fact, since the feature-mapped vectors only appear in the $f_2$ in inner products, we may not even need to explicitly carry out the feature mapping to compute the inner products.  We call the function $k(\\vec{x}_i, \\vec{x}_j)$ that calculates the inner products the “kernel function”, and this way of avoiding the feature map computation is called the “kernel trick”. In fact, the feature mapped vectors could even be infinite dimensional, but the kernel might still be very efficiently computable.\n",
```

francabrera (approved):

axelhzf (approved):

kaelynj (approved):

PR #5593, opened 2026-09-01
Update rustworkx attributes for version used in qiskit~=2.5.2

emilkovacev:
qiskit~=2.5.2 is pinned [here](https://github.com/Qiskit/documentation/blob/51d4c0bc421fe3b2eecc5b4a4420c38991e2eaff/scripts/nb-tester/requirements.txt#L4), which correlates to rustworkx 0.18.1. Rustworkx types and function calls were modified such that `rx.Graph` -> `rx.PyGraph`. This PR updates advanced-techniques-for-qaoa to match that format.

### DIFF
--- a/docs/tutorials/advanced-techniques-for-qaoa.ipynb
+++ b/docs/tutorials/advanced-techniques-for-qaoa.ipynb
@@ -408,7 +408,7 @@
     "\n",
     "    def find_initial_mappings(\n",
     "        self,\n",
-    "        program_graph: rx.Graph,\n",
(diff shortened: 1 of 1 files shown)


mtreinish (approved):
FWIW, `Graph` was never a class rustworkx. The two main graph classes in the library have always been `PyGraph` and `PyDiGraph` (there is also a `PyDAG` class which dates back to the earliest releases of retworkx and is still supported as a backwards compatibility alias of `PyDiGraph`).

abbycross (approved):

PR #5595, opened 2026-09-01
Fix unmatched fstrings in tutorials

emilkovacev:
Revises quotes within fstring to be different than surrounding quotes within two tutorials.

Original:
```
var = {"a": 1}
f"{var["a"]}" # inner quotes conflict with outer quotes
```

Fix:
```
var = {"a": 1}
f"{var['a']}" # inner quotes use '', outer quotes use ""
```


abbycross (approved):
Thanks!

PR #5596, opened 2026-09-01
Update var type in tutorial

emilkovacev:
Change type of 'shots' from float to int in real-time-benchmarking tutorial. This was previously causing the tutorial to crash with `TypeError: shots must be an integer` because of `shots = 1000.0` (Python `1e3` is evaluated to float `1000.0`, not `1000`).


abbycross (approved):
Thanks!

PR #5603, opened 2026-09-03
test the qiskit-noise-learning subsite

abbycross:
A test to understand how the pipeline works in preparation to migrating API docs/guides


abbycross:
cc: @kaelynj Some things to resolve:

- How to make the citations render properly (see formalism/index for examples)
- How to make toc render properly (what do we need in the source repo for this to happen? Do we need a guides/ dir and some other expected files, like index.rst and install.rst for the toc to generate properly?) 
- right now we don't even see the API docs in my local preview left nav (using --apis), though those files have imported, and I can get to them with a direct URL
- I don't see a package file and so am using 0.1.0 as a version placeholder. I assume this is also why the release notes are broken
- Check all the broken links in the CI link checker - is this a pipeline problem?

abbycross:
@eharvey328 @kaelynj I don't know why this would affect generation of the toc, but is part of the problem that some of the docs are markdown files? I see that the other packages use only rst and ipynb. Worth converting all the md to one of those?

eharvey328 (dismissed):
LGTM! Thank you

kaelynj (approved):

PR #5615, opened 2026-09-08
Update to Functions guides/tutorials/api reference

abbycross:
Two new application function docs.


HuangJunye (commented):
Looks great to me. I can't see the API reference preview (perhaps it's expected?) but otherwise everything looks good. We are just waiting for the updated figure

docs/guides/aqarios-constrained-quantum-optimizer.ipynb
```diff
@@ -0,0 +1,799 @@
...
+    "\n",
+    "| Problem               | Variables | Constraints | Qubits | Accuracy | Feasibility | Total runtime (s) | QPU usage (s) | Total shots |\n",
+    "|-----------------------|-----------|-------------|--------|----------|-------------|-------------------|---------------|-------------|\n",
+    "| MIS es60fst02         | 186       | 280         | 124    | 100.0%   | True        | 940.2             | 147.0         | 25.0k       |\n",
+    "| MIS sloane\\_2dc\\_128  | 128       | 5173        | 74     | 100.0%   | True        | 347.6             | 106.0         | 12.5k       |\n",
+    "| Max-3-Cut Karate Club | 99        | 33          | 99     | 98.7%    | True        | 279.5             | 166.0         | 25.0k       |\n",
+    "| Max-4-Cut Karate Club | 132       | 33          | 132    | 100.0%   | True        | 258.3             | 200.0         | 25.0k       |"
+   ]
+  },
+  {
```
HuangJunye, line 83: Need to add the following disclaimer for all optimization functions in the benchmarking section. I am not sure where is the best location:
1. right below the "Benchmarks" section title
2. right below the benchmarks table.
@abbycross what do you think?

This is the text that needs to be added

```
Disclaimer: Performance may depend on both the problem instance and subsequent processing steps. In some cases, classical samples and quantum-generated samples may achieve similar final solution quality after equivalent post-processing. Evaluation should therefore consider the complete optimization workflow.
```

abbycross: My vote, for visibility, is to put it below the Benchmarks section title. 

Also note - we avoid using the word "may" because its dual meaning is confusing to machine translation. I suggest updating to 

Disclaimer: Performance can depend on both the problem instance and subsequent processing steps. In some cases, classical samples and quantum-generated samples might achieve similar final solution quality after equivalent post-processing. Evaluation should therefore consider the complete optimization workflow.

abbycross:
> Looks great to me. I can't see the API reference preview (perhaps it's expected?) but otherwise everything looks good. We are just waiting for the updated figure

Correct - API refs don't appear in these PR previews, but you can preview them locally.

docs/guides/aqarios-constrained-quantum-optimizer.ipynb
HuangJunye: I agree with the location and the change to the text. Thank you!

docs/tutorials/parity-twine-optimizer-ms.ipynb
```diff
@@ -0,0 +1,594 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "805c80a1180e79fe",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "catalog = QiskitFunctionsCatalog(\n",
+    "    channel=\"ibm_quantum_platform\",\n",
+    ")\n",
```
HuangJunye:
```suggestion
    "catalog = QiskitFunctionsCatalog(channel="ibm_quantum_platform")\n",
```

docs/tutorials/parity-twine-optimizer-sk.ipynb
```diff
@@ -0,0 +1,542 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "c4a7dc1cbb3d35ff",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "catalog = QiskitFunctionsCatalog(\n",
+    "    channel=\"ibm_quantum_platform\",\n",
+    ")\n",
```
HuangJunye:
```suggestion
    "catalog = QiskitFunctionsCatalog(channel="ibm_quantum_platform")\n",
```

docs/tutorials/parity-twine-optimizer-ms.ipynb
HuangJunye, line 1: The two tutorials notebooks do not have cell outputs. Should I run them and generate the outputs?

abbycross: Yes please!

beckykd (dismissed):

PR #5627, opened 2026-09-10
Some improvements to the SQD tutorial

### LINKED ISSUE #5626: SQD tutorial small-scale data no longer matches text

https://quantum.cloud.ibm.com/docs/en/tutorials/sample-based-quantum-diagonalization

### Select all that apply

- [ ] new content request
- [ ] typo
- [ ] code bug
- [x] out-of-date content
- [ ] broken link
- [ ] other

The text says
>The first plot shows that in this simulation we are already within 1 mH of the exact answer after the first iteration (chemical accuracy is typically accepted to be 1 kcal/mol ≈≈ 1.6 mH).

But the plot no longer matches.

Image

This was introduced when we switched to the optimized LUCJ pass manager, which automatically drops interactions, making the ansatz less expressive. To fix it, we can use a square lattice topology and/or increase the number of repetitions.

kevinsung:
Fixes #5626 .

- demonstrate usage of `max_dim`
- use square lattice connectivity and higher number of reps in small-scale example, to improve energies
- in hardware example, determine connectivity from backend processor family
- don't limit optimization iterations for compressed factorization for LUCJ, since it's much faster in `ffsim >= 0.0.81`.
- put nitrogen atoms symmetrically around Z axis to eliminate warning in newer PySCF versions
- set "chemical accuracy" to 1.6 mHa and simplify y axis plotting logic


abbycross:
@kevinsung is it still ok that the text talks about heavy-hex still?

kevinsung:
> Kevin J. Sung (@kevinsung) is it still ok that the text talks about heavy-hex still?

Good catch, I will update that.

nathanearnestnoble (dismissed):
LGTM!

abbycross (dismissed):

henryzou50 (changes requested):
Thanks Kevin! This mostly looks good, but two small things to address:

- Could we bump the SQD requirement to v0.12? The new `max_dim` argument isn’t available in v0.11.
- Could we lower `min_num_qubits=133` in the hardware example? That filter excludes Nighthawk’s 120 qubits, so the new Nighthawk connectivity branch won’t be reached.

henryzou50 (approved):
Thanks, LGTM!

abbycross (approved):

PR #5655, opened 2026-09-15
added notes on state indexing

### LINKED ISSUE #5526: Ambiguity in index convention in Amplitude Encoding formula (ordinal vs. binary indexing of basis states)

I am refering to URL: https://quantum.cloud.ibm.com/learning/en/courses/quantum-machine-learning/data-encoding 
Section "Methods of encoding", subsection "Amplitude encoding".

### Select all that apply

- [ ] new content request
- [ ] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [x] other

In this subsection, the general formula is written as:

$$|\psi_x^{(j)}\rangle=\frac{1}{\alpha}\sum_{i=1}^{N}x_i^{(j)}|i\rangle $$

with the sum running from $i=1$ to $i=N$.

In the example right after, $x^{(1)}=(4,8,5)$ with $N=3$ is encoded into a 2-qubit state:

$$|\psi(\vec{x}^{(1)})\rangle=\frac{1}{\sqrt{105}}(4|00\rangle+8|01\rangle+5|10\rangle+0|11\rangle) $$

This result is correct but it relies on reading $|i\rangle$  in the general formula as the i-th basis state in a fixed ordering (i.e. $x_1\rightarrow |00\rangle, x_2\rightarrow |01\rangle, x_3 \rightarrow |10\rangle$) rather than **the basis state whose label is the binary representation of integer $i$** (which would give $x_1\rightarrow |01\rangle, x_2 \rightarrow |10\rangle, x_3 \rightarrow |11\rangle$, leaving $|00\rangle$ unassigned, i.e. effectively carrying coefficient 0). Since the sum starts at $i=1$ the two conventions don't line up unless this is made explicit.

It would help to add a short note clarifying that $|i\rangle$ refers to the positional indexing of the basis states rather than the binary value of $i$.

Thanks for putting this course together, it's been a great resource!

(No answer)

christopherporter1:
Fixes #5526


learning/courses/quantum-machine-learning/data-encoding.ipynb
```diff
@@ -23,7 +23,7 @@
...
+    "A dataset is a set of $M$ data vectors: $\\text{X} = \\{\\vec{x}^{(j)}\\,|\\,j\\in [M]\\}$, where each vector is $N$ dimensional, that is, $\\vec{x}^{(j)}=(\\vec{x}^{(j)}_1,\\ldots,\\vec{x}^{(j)}_N)\\in\\mathbb{R}^N$. You may also see vector components numbered from $0$ to $N-1$ in some contexts. This could be extended to complex data features. In this lesson, we may occasionally use these notations for the full set $(\\text{X}),$ and its specific elements like $\\vec{x}^{(j)}$. But we will mostly refer to the loading of a single vector from our dataset at a time, and will often simply refer to a single vector of $N$ features as $\\vec{x}$.\n",
```
```suggestion
    "A dataset is a set of $M$ data vectors: $\\text{X} = \\{\\vec{x}^{(j)}\\,|\\,j\\in [M]\\}$, where each vector is $N$ dimensional, that is, $\\vec{x}^{(j)}=(\\vec{x}^{(j)}_1,\\ldots,\\vec{x}^{(j)}_N)\\in\\mathbb{R}^N$. You might also see vector components numbered from $0$ to $N-1$ in some contexts. This could be extended to complex data features. In this lesson, we may occasionally use these notations for the full set $(\\text{X})$, and its specific elements like $\\vec{x}^{(j)}$. But we will mostly refer to the loading of a single vector from our dataset at a time, and will often simply refer to a single vector of $N$ features as $\\vec{x}$.\n",
```

learning/courses/quantum-machine-learning/data-encoding.ipynb
```diff
@@ -247,7 +247,7 @@
     "|\\psi^{(j)}_x\\rangle = \\frac{1}{\\alpha}\\sum_{i=1}^N x^{(j)}_i |i\\rangle\n",
     "$$\n",
     "\n",
-    "where $N$ is the same dimension of the data vectors as before, $\\vec{x}^{(j)}_i$ is the $i^{th}$ element of $\\vec{x}^{(j)}$ and $|i\\rangle$ is the $i^{th}$ computational basis state. Here, $\\alpha$ is a normalization constant to be determined from the data being encoded. This is the normalization condition imposed by quantum mechanics:\n",
+    "where $N$ is the same dimension of the data vectors as before, $\\vec{x}^{(j)}_i$ is the $i^{th}$ element of $\\vec{x}^{(j)}$. The computational basis states are assumed to be in a fixed ordering, which is referenced here by $i$. Equivalently, one may write the encoding as $\\sum_{i=0}^{N-1} x_i |i\\rangle$, where $|i\\rangle$ denotes the basis state labeled by the binary representation of integer $i$. Either works as long as we are consistent. Here, $\\alpha$ is a normalization constant to be determined from the data being encoded. This is the normalization condition imposed by quantum mechanics:\n",
```
```suggestion
    "where $N$ is the same dimension of the data vectors as before, $\\vec{x}^{(j)}_i$ is the $i^{th}$ element of $\\vec{x}^{(j)}$. The computational basis states are assumed to be in a fixed ordering, which is referenced here by $i$. Equivalently, one can write the encoding as $\\sum_{i=0}^{N-1} x_i |i\\rangle$, where $|i\\rangle$ denotes the basis state labeled by the binary representation of integer $i$. Either works as long as we are consistent. Here, $\\alpha$ is a normalization constant to be determined from the data being encoded. This is the normalization condition imposed by quantum mechanics:\n",
```

abbycross (approved):

PR #5660, opened 2026-09-15
swapped transposition of vectors and added explanation

### LINKED ISSUE #5448: Projected Hamiltonian python definitions for Sample-based Quantum Diagonalization (SQD)

https://quantum.cloud.ibm.com/learning/en/courses/quantum-diagonalization-algorithms/sqd-overview

### Select all that apply

- [ ] new content request
- [x] typo
- [ ] code bug
- [ ] out-of-date content
- [ ] broken link
- [ ] other

The subspace Hs as defined in the snippets should be Hs = np.array([[x1.T @ H @ x1, x1.T @ H @ x2], [x2.T @ H @ x1, x2.T @ H @ x2]]) and not Hs = np.array([[x1 @ H @ x1.T, x1 @ H @ x2.T], [x2 @ H @ x1.T, x2 @ H @ x2.T]]). The projected Hamiltonian doesn't change, however the order might be confusing to someone not very familiar with Dirac notation: since x is our column vector (ket) the transpose .T should be on the bra (row linear functional) side. Hence <x.T | H | x>. 

(No answer)

christopherporter1:
Fixes #5448


learning/courses/quantum-diagonalization-algorithms/sqd-overview.ipynb
```diff
@@ -200,10 +200,18 @@
...
+    "Hs = np.array([[x1.T @ H @ x1, x1.T @ H @ x2], [x2.T @ H @ x1, x2.T @ H @ x2]])\n",
     "print(Hs)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "4b67c42b",
+   "metadata": {},
+   "source": [
+    "Note that in writing the projected Hamiltonian elements as `x1.T @ H @ x2`, we are implicitly adopting the common convention that the state vectors `x1` and `x2` correspond to kets (column vectors), consistent with expressions such as $x_1 = |011\\rangle$. However, NumPy's 1D arrays do not intrinsically distinguish between row and column vectors, and for the vectors used here the transpose operation has no numerical effect. Consequently, writing `x1 @ H @ x2.T` would produce the same numerical result, although `x1.T @ H @ x2` more closely matches the usual Dirac-notation correspondence $\\langle x_1|H|x_2\\rangle$⟩."
```
abbycross: I had to reload the PR preview so I haven't yet seen how this renders, but it seems like this would produce a double set of angles - one with LaTeX and the other from the ⟩ symbol?: `$\\langle x_1|H|x_2\\rangle$⟩`

abbycross: Just confirmed that it is a double set:
Screenshot 2026-09-16 at 11 55 14 AM

abbycross:
```suggestion
    "Note that in writing the projected Hamiltonian elements as `x1.T @ H @ x2`, we are implicitly adopting the common convention that the state vectors `x1` and `x2` correspond to kets (column vectors), consistent with expressions such as $x_1 = |011\\rangle$. However, NumPy's 1D arrays do not intrinsically distinguish between row and column vectors, and for the vectors used here the transpose operation has no numerical effect. Consequently, writing `x1 @ H @ x2.T` would produce the same numerical result, although `x1.T @ H @ x2` more closely matches the usual Dirac-notation correspondence $\\langle x_1|H|x_2\\rangle$."
```

abbycross (approved):
Thanks!

PR #5671, opened 2026-09-18
Cp/tfqa 2

christopherporter1:
Adding lesson 02 to Tools for Quantum Advantage


learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "Consider what happens to the state $|+\\rangle$ as time passes:\n",
+    "\n",
+    "$$\n",
+    "|\\psi(t)\\rangle = e^{-iHt/\\hbar}|\\psi(0)\\rangle \\rightarrow e^{-iZ\\omega_B t/2}|\\psi(0)\\rangle\n",
+    "$$\n",
+    "\n",
+    "Here we have used the fact that a magnetic field in the Z direction causes precession about the Z axis with a frequency $\\omega_B$ that depends on the effective magnetic moment and magnetic field strength. The details are less important than the fact that this interaction causes evolution about the Z axis, resulting in opposite phase accumulation for the two computational basis states. Applying this operator to each term in $|\\psi\\rangle$ we find\n",
```
abbycross:
```suggestion
    "Here we have used the fact that a magnetic field in the Z direction causes precession about the Z axis with a frequency $\\omega_B$ that depends on the effective magnetic moment and magnetic field strength. The details are less important than the fact that this interaction causes evolution about the Z axis, resulting in opposite phase accumulation for the two computational basis states. Applying this operator to each term in $|\\psi\\rangle$, we find the following:\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "=\\frac{1}{\\sqrt{2}}\\left( e^{-i\\omega_B t/2}|0\\rangle +e^{i\\omega_B t/2}|1\\rangle \\right)\n",
+    "$$\n",
+    "\n",
+    "This time-dependent phase corresponds to precession about the Z axis in the Bloch sphere picture. This is shown in the right half of the figure above.\n",
+    "\n",
+    "If we knew that this interaction were occurring in a controlled way, we could predict it. To model this type of noise, we instead consider a distribution of possible magnetic field strengths, each occurring with some classical probability (other sources of noise would need o be modeled differently). We need to consider what happens when there is a non-zero classical probability that the state has not rotated at all, and also some probability that it has rotated by a small amount, or even by a large amount. This distribution of possible rotations is the reason for the spreading of the state in the Bloch sphere picture below. If there is more such dephasing noise, the phase will be less well-defined. In the limit of strong dephasing, the qubit becomes completely dephased, corresponding to the loss of the quantum coherence stored in the state.\n",
```
abbycross:
```suggestion
    "If we knew that this interaction were occurring in a controlled way, we could predict it. To model this type of noise, we instead consider a distribution of possible magnetic field strengths, each occurring with some classical probability (other sources of noise would need to be modeled differently). We need to consider what happens when there is a non-zero classical probability that the state has not rotated at all, and also some probability that it has rotated by a small amount, or even by a large amount. This distribution of possible rotations is the reason for the spreading of the state in the Bloch sphere picture below. If there is more such dephasing noise, the phase will be less well-defined. In the limit of strong dephasing, the qubit becomes completely dephased, corresponding to the loss of the quantum coherence stored in the state.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "X|\\psi(t_0)\\rangle = \\frac{1}{\\sqrt{2}}\\left( e^{-i\\omega_B t_0/2}|1\\rangle +e^{i\\omega_B t_0/2}|0\\rangle \\right)\n",
+    "$$\n",
+    "\n",
+    "Now when the second interval of $t_0$ passes, the same interaction takes place. But now the amplitudes are associated with the opposite Z eigenstates, meaning the sign of the rotation about the Z axis has switched. After allowing the system to evolve in time for another $t_0$ we have the state:\n",
+    "\n",
+    "$$\n",
+    "e^{-iZ\\omega_B (t-t_0)/2}\\frac{1}{\\sqrt{2}}\\left( e^{-i\\omega_B t_0/2}|1\\rangle +e^{i\\omega_B t_0/2}|0\\rangle \\right)\n",
+    "=\\frac{1}{\\sqrt{2}}\\left( e^{-i\\omega_B (t-t_0)/2}e^{i\\omega_B t_0/2}|1\\rangle +e^{i\\omega_B (t-t_0)/2}e^{-i\\omega_B t_0/2}|0\\rangle \\right)\n",
+    "$$\n",
+    "And inserting $t = 2t_0$ we have\n",
```
abbycross:
```suggestion
    "And inserting $t = 2t_0$ we have:\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "Applying the final X gate does nothing in this case, but it is generally necessary:\n",
+    "\n",
+    "$$\n",
+    "|\\psi(2t_0)\\rangle = \\frac{1}{\\sqrt{2}}\\left(|0\\rangle +|1\\rangle \\right) = |\\psi(t=0)\\rangle\n",
+    "$$\n",
+    "\n",
+    "We have recovered the original quantum state, including its relative phase. This process is an especially simple example of dynamical decoupling (DD).\n",
+    "\n",
+    "More generally, __dynamical decoupling__ involves inserting some single-qubit gates to reduce the effect of interactions with systems outside the qubit (decoupling it from the environment). The complete refocusing of all possible phase histories back to the original state, as shown above, is somewhat idealized; but it is still a possible scenario. Let's discuss when DD is useful and what caveats exist.\n",
```
abbycross:
```suggestion
    "We have recovered the original quantum state, including its relative phase. This process is an especially simple example of dynamical decoupling.\n",
    "\n",
    "More generally, __dynamical decoupling__ (DD) involves inserting some single-qubit gates to reduce the effect of interactions with systems outside the qubit (decoupling it from the environment). The complete refocusing of all possible phase histories back to the original state, as shown above, is somewhat idealized, but it is still a possible scenario. Let's discuss when DD is useful and what caveats exist.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "$$\n",
+    "And inserting $t = 2t_0$ we have\n",
+    "$$\n",
+    "|\\psi(2t_0)\\rangle=\\frac{1}{\\sqrt{2}}\\left(|1\\rangle +i |0\\rangle \\right)\n",
+    "$$\n",
+    "\n",
+    "Applying the final X gate, we have:\n",
+    "\n",
+    "$$\n",
+    "|\\psi(2t_0)\\rangle = \\frac{1}{\\sqrt{2}}\\left(|0\\rangle +i |1\\rangle \\right) = |\\psi(t=0)\\rangle = |+i\\rangle\n",
```
abbycross, line 129: Unclear to me whether this is intentionally a duplicate of material that showed up earlier on the page??

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "The first and most obvious caveat is that we assumed a purely dephasing channel in our treatment. Real-world interactions typically produce a mixture of dephasing and other error mechanisms. In the treatment above, we chose to apply X gates specifically. This is referred to as an `XX` sequence in dynamical decoupling. This particular sequence is appropriate for purely dephasing errors. But there are other sequences that might be more broadly applicable, like `XY4` (shown in the circuit diagram below) and the more complex `XY8`.\n",
+    "\n",
+    "![A quantum circuit showing a sequence of four quantum gates used in dynamical decoupling: an X gate, a Y gate, a second X gate, and a second Y gate.](/learning/images/courses/tools-for-quantum-advantage/error-suppression/xy4.avif)\n",
+    "\n",
+    "Another caveat is that dynamical decoupling adds single-qubit gates, which can add single-qubit errors due to gate imperfections or even crosstalk. These single-qubit error rates are typically much lower than multi-qubit gate errors, so this is often not a major concern. But it is something to keep in mind if many qubits are employing DD many times throughout your circuit.\n",
```
abbycross:
```suggestion
    "Another caveat is that dynamical decoupling adds single-qubit gates, which can add single-qubit errors due to gate imperfections or even crosstalk. These single-qubit error rates are typically much lower than multi-qubit gate errors, so this is usually not a major concern, but it is something to keep in mind if many qubits are employing DD many times throughout your circuit.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "DD is useful when the external coupling has time to affect the state of the qubit. Because qubits are painstakingly well isolated and external couplings should be weak, this type of noise is most noticeable when a qubit sits idle for long periods. For short idle times, the effect of DD could be to add single-qubit gate errors while suppressing very little noise; the fidelity of your circuit could actually be reduced!\n",
```
abbycross:
```suggestion
    "DD is useful when the external coupling has time to affect the state of the qubit. Because qubits are painstakingly well isolated and external couplings should be weak, this type of noise is most noticeable when a qubit sits idle for long periods. For short idle times, the effect of DD could be to add single-qubit gate errors while suppressing very little noise; the fidelity of your circuit could actually be reduced.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "402016ef",
+   "metadata": {},
+   "source": [
+    "## Dynamical decoupling using Qiskit\n",
+    "\n",
+    "Let us explore the use of DD by examining the case of several qubits prepared in the $|+\\rangle$ state, which then remain idle for a long time. In the absence of errors, preparing the multi-qubit state $|\\psi\\rangle = |+\\rangle^{\\otimes N}$ and then applying a Hadamard gate before measurement (thereby measuring in the X basis) should always yield 0. With noise, the measurement outcome will be 0 only a fraction of the time rather than with 100% probability. We calculate the average `X` expectation value across multiple qubits. That is, we are interested in\n",
```
abbycross:
```suggestion
    "Let us explore the use of DD by examining the case of several qubits prepared in the $|+\\rangle$ state, which then remain idle for a long time. In the absence of errors, preparing the multi-qubit state $|\\psi\\rangle = |+\\rangle^{\\otimes N}$ and then applying a Hadamard gate before measurement (thereby measuring in the X basis) should always yield 0. With noise, the measurement outcome will be 0 only a fraction of the time rather than with 100% probability. We calculate the average `X` expectation value across multiple qubits. That is, we are interested in the following:\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "service = QiskitRuntimeService()  # assumes credentials are saved\n",
+    "backend = service.backend(\"ibm_fez\")  # adjust if needed"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "63145ef8",
+   "metadata": {},
+   "source": [
+    "Now we define some helper functions. First, we want to address the point about idle times being long. What exactly does \"long\" mean in this context? One could simply report the idle time in microseconds. However, it is useful to know how many two-qubit gates could be executed during that same interval. This makes the idle times more directly indicative of the circuit depth that could have been executed during the same period. The first helper function gets the two-qubit gate native to the backend and determines the duration of that gate.\n",
```
abbycross:
```suggestion
    "Now we define some helper functions. First, we want to address the point about long idle times. What exactly does \"long\" mean in this context? We could simply report the idle time in microseconds. However, it is useful to know how many two-qubit gates could be executed during that same interval. This makes the idle times more directly indicative of the circuit depth that could have been executed during the same period. The first helper function gets the two-qubit gate native to the backend and determines the duration of that gate.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "    p0 /= total\n",
+    "    return 2.0 * p0 - 1.0"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "740f9f9c",
+   "metadata": {},
+   "source": [
+    "Now we specify the details of our test including the number of qubits and the gate sequence to be used in DD (in this case `XX`). Note especially that we set the optimization level to zero. In practice, you would often select a higher optimization level; here we use level 0 to ensure that the effects of the errors targeted by DD remain visible. Finally we determine the characteristic time for two-qubit gates on this backend and print some relevant times."
```
abbycross:
```suggestion
    "Now we specify the details of our test, including the number of qubits and the gate sequence to be used in DD (in this case `XX`). Note especially that we set the optimization level to zero. In practice, you would often select a higher optimization level; here we use level 0 to ensure that the effects of the errors targeted by DD remain visible. Finally, we determine the characteristic time for two-qubit gates on this backend and print some relevant times."
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "48793948",
+   "metadata": {},
+   "source": [
+    "#### Execute\n",
+    "\n",
+    "We're finally ready to execute on hardware. We use sampler to obtain many measurements of each qubit, and we will use it twice: once with DD explicitly turned off, and once with DD turned on and using the `XX` gate sequence."
```
abbycross:
```suggestion
    "We're finally ready to execute on hardware. We use Sampler to obtain many measurements of each qubit, and we will use it twice: once with DD explicitly turned off, and once with DD turned on and using the `XX` gate sequence."
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1ffed158",
+   "metadata": {},
+   "source": [
+    "As you can see, with no explicit delay the expectation values are indeed near 1, which we would expect if all phase information were preserved. If all phase information were lost, there would be no preference for the final rotation to produce $|0\\rangle$ rather than $|1\\rangle$, and the average expectation value would approach zero. In the data, we see that as delay times increase, the average expectation value of `X` decreases, starting to approach zero. Note that DD has been very effective here; the expectation values with DD are typically more than 20% better (closer to 1) than the values without DD. But also note that the first delayed data point corresponds to a delay roughly equivalent to 100 2-qubit gate operations. This reinforces the point that DD is most useful when qubits remain idle for relatively long periods.\n",
```
abbycross:
```suggestion
    "As you can see, with no explicit delay the expectation values are indeed near 1, which we would expect if all phase information were preserved. If all phase information were lost, there would be no preference for the final rotation to produce $|0\\rangle$ rather than $|1\\rangle$, and the average expectation value would approach zero. In the data, we see that as delay times increase, the average expectation value of `X` decreases, starting to approach zero. Note that DD has been very effective here; the expectation values with DD are typically more than 20% better (closer to 1) than the values without DD. But also note that the first delayed data point corresponds to a delay roughly equivalent to 100 two-qubit gate operations. This reinforces the point that DD is most useful when qubits remain idle for relatively long periods.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "#### Check your understanding\n",
+    "\n",
+    "If we apply DD using `XY4` to the same circuit as before, do you expect it to yield results that are much better, much worse, or about the same, as DD using `XX`? Explain.\n",
+    "\n",
+    "<Accordion>\n",
+    "<AccordionItem title=\"Answer\">\n",
+    "\n",
+    "About the same, perhaps slightly worse. The circuit we used had states rotated into the XY plane. This state stores information primarily in its phase, making it especially sensitive to dephasing errors rather than T1 relaxation. XY4 might help with a wider variety of errors, but XX is already optimized to help the circuit we are using. XY4 might be just as good, but not add anything substantial, or the fact that XY4 contains more gates might allow additional gate errors to make the results slightly worse.\n",
```
abbycross:
```suggestion
    "About the same, perhaps slightly worse. The circuit we used had states rotated into the XY plane. This state stores information primarily in its phase, making it especially sensitive to dephasing errors rather than T1 relaxation. XY4 might help with a wider variety of errors, but XX is already optimized to help the circuit we are using. XY4 might be just as good, but would not add anything substantial, or the fact that XY4 contains more gates might allow additional gate errors to make the results slightly worse.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "</AccordionItem>\n",
+    "</Accordion>\n",
+    "\n",
+    "\n",
+    "The last result used the simplest DD gate sequence `XX`. Let's see how to implement a more complex sequence, `XY4`. We define a sampler in the next section."
```
abbycross:
```suggestion
    "The last result used the simplest DD gate sequence `XX`. Let's see how to implement a more complex sequence, `XY4`. We define a Sampler in the next section."
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b9d9e170",
+   "metadata": {},
+   "source": [
+    "## Pauli twirling\n",
+    "\n",
+    "We should begin by noting that Pauli twirling is often used not as an error-suppression technique, but as an error-shaping technique: making noise/errors behave differently, sometimes more predictably, to enable other methods. Although Pauli twirling does not *prevent* errors, it might prevent their coherent accumulation.\n",
```
abbycross:
```suggestion
    "We should begin by noting that Pauli twirling is often used not as an error-suppression technique, but as an error-shaping technique: it makes noise/errors behave differently, sometimes more predictably, to enable other methods. Although Pauli twirling does not *prevent* errors, it might prevent their coherent accumulation.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "|\\psi_\\text{init}\\rangle = a|00\\rangle + b|01\\rangle + c|10\\rangle + d|11\\rangle\n",
+    "$$\n",
+    "\n",
+    "As always, we are using the qubit ordering convention $|q_1,q_0\\rangle$. Applying a CX gate with $q_1$ as the target yields\n",
+    "\n",
+    "$$\n",
+    "CX|\\psi_\\text{init}\\rangle = |\\psi_\\text{final}\\rangle = a|00\\rangle + b|11\\rangle + c|10\\rangle + d|01\\rangle\n",
+    "$$\n",
+    "\n",
+    "Now let us consider the third circuit shown, using X gates for Pauli twirling. One could simply multiply the matrices together and verify that they yield a CNOT matrix. Alternatively, we can track the operation on an arbitrary quantum state through the circuit, as we do below. The states at different points in the circuit have been labeled a-d.\n",
```
abbycross:
```suggestion
    "Now let us consider the third circuit shown, using X gates for Pauli twirling. We could simply multiply the matrices together and verify that they yield a CNOT matrix. Alternatively, we can track the operation on an arbitrary quantum state through the circuit, as we do below. The states at different points in the circuit have been labeled a-d.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "#### Check your understanding:\n",
+    "\n",
+    "Verify that the Pauli twirling in the fourth panel in the figure above also leaves the logical effect of the CNOT unchanged.\n",
+    "\n",
+    "<Accordion>\n",
+    "<AccordionItem title=\"Answer\">\n",
+    "\n",
+    "We follow the example above and show that the action on any arbitrary two-qubit state is equivalent to a CNOT operation. We will refer to the labeled points in this diagram.\n",
+    "\n",
+    "![A quantum circuit with two qubits. The initial state is labeled \"a\", then an X gate acts on qubit 0 and a Y gate acts on qubit 1. After that the state is labeled \"b\". Then a CNOT gate acts with qubit 0 as the control and qubit 1 as the target. After that CNOT gate the state is labeled \"c\". Finally a Y gate acts on qubit 0 and a Z gate acts on qubit 1. The final state is labeled \"d\".](/learning/images/courses/tools-for-quantum-advantage/error-suppression/twirl-check.avif)\n",
```
abbycross:
```suggestion
    "![A quantum circuit with two qubits. The initial state is labeled \"a\", then an X gate acts on qubit 0 and a Y gate acts on qubit 1. After that, the state is labeled \"b\". Then a CNOT gate acts with qubit 0 as the control and qubit 1 as the target. After that, CNOT gate the state is labeled \"c\". Finally, a Y gate acts on qubit 0 and a Z gate acts on qubit 1. The final state is labeled \"d\".](/learning/images/courses/tools-for-quantum-advantage/error-suppression/twirl-check.avif)\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5925601c",
+   "metadata": {},
+   "source": [
+    "### When to use Pauli twirling\n",
+    "\n",
+    "As presented here, Pauli twirling is applied only to multi-qubit gates. Applying a similar protocol to single-qubit gates would require different logic and is generally not useful in practice. Pauli twirling itself uses several single-qubit gates (the Pauli gates). The additional Pauli gates would likely introduce more error than would be gained by randomizing any coherent error accumulation. Error rates associated with two-qubit gates are much larger than those associated with single-qubit gates. Further, some single-qubit gates are non-Clifford, which cannot be fully twirled. This is why Qiskit includes Pauli twirling options which automatically twirl around two-qubit gates, and not around single-qubit gates.\n",
```
abbycross:
```suggestion
    "As presented here, Pauli twirling is applied only to multi-qubit gates. Applying a similar protocol to single-qubit gates would require different logic and is generally not useful in practice. Pauli twirling itself uses several single-qubit gates (the Pauli gates). The additional Pauli gates would likely introduce more error than would be gained by randomizing any coherent error accumulation. Error rates associated with two-qubit gates are much larger than those associated with single-qubit gates. Further, some single-qubit gates are non-Clifford, which cannot be fully twirled. This is why Qiskit includes Pauli twirling options that automatically twirl around two-qubit gates, and not around single-qubit gates.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "This was implicit in the figure above: Pauli twirling was implemented around the CX gates, but not around the Hadamard gate.\n",
+    "\n",
+    "Let's see two examples of Pauli twirling in action.\n",
+    "\n",
+    "### Pauli twirling to suppress coherent accumulation\n",
+    "\n",
+    "To observe how Pauli twirling can turn coherent error accumulation into slower incoherent accumulation, we want a circuit and observable that function as a coherent-error stress test. The sole purpose is to make coherently accumulating two‑qubit errors visible, and then show how Pauli twirling converts that coherent buildup into stochastic decay.\n",
+    "\n",
+    "CNOT (or CZ) gates are a common source of coherent errors. The simplest experiment we can do in this case is to initialize a state (say $|+\\rangle$), apply layers of paired CNOT gates (using the fact that two CNOTs yield an identity) and check how errors accumulate as the number of layers increases, both with and without Pauli twirling.\n",
```
abbycross:
```suggestion
    "CNOT (or CZ) gates are a common source of coherent errors. The simplest experiment we can do in this case is to initialize a state (say $|+\\rangle$), apply layers of paired CNOT gates (using the fact that two CNOTs yield an identity), and check how errors accumulate as the number of layers increases, both with and without Pauli twirling.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "    return qc"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c16a052f",
+   "metadata": {},
+   "source": [
+    "Because we have rotated our basis before measurements, a measurement of $|0\\rangle$ corresponds to the state having been in $|+\\rangle$ just before the final Hadamard gate, and similar for $|1\\rangle$ and $|-\\rangle$. Therefore, our expectation value $\\langle X \\rangle$ can be simply be calculated from the counts of $|0\\rangle$ minus the counts of $|1\\rangle$."
```
abbycross:
```suggestion
    "Because we have rotated our basis before measurements, a measurement of $|0\\rangle$ corresponds to the state having been in $|+\\rangle$ just before the final Hadamard gate, and similar for $|1\\rangle$ and $|-\\rangle$. Therefore, our expectation value $\\langle X \\rangle$ can be simply calculated from the counts of $|0\\rangle$ minus the counts of $|1\\rangle$."
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "pm = generate_preset_pass_manager(backend=backend, optimization_level=0)\n",
+    "isa_circuits = [pm.run(c) for c in circuits]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "796c2626",
+   "metadata": {},
+   "source": [
+    "Keep in mind that each layer consists of more than one two qubit gate. Monitor the transpiled, two-qubit depth using the function below."
```
abbycross:
```suggestion
    "Keep in mind that each layer consists of more than one two-qubit gate. Monitor the transpiled, two-qubit depth by using the function below."
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "fa118805",
+   "metadata": {},
+   "source": [
+    "To understand why, let's look at a map of the layout of our backend (in this image, `ibm_fez`, but you can do equivalent analyses on any backend).\n",
+    "\n",
+    "![A diagram of the layout of qubits on a quantum computer called ibm_fez. It shows how qubits are chosen to minimize swapping of information.](/learning/images/courses/tools-for-quantum-advantage/error-suppression/ghz-transpile.avif)\n",
+    "\n",
+    "We see that the qubits were selected in a chain to minimize swap gates and thus circuit depth. Further, each CZ gate implemented in the circuit is between adjacent qubits. Finally, all five qubits have relatively low error rates, including readout-assignment error rates. You can check these error rates on any backend [here](https://quantum.cloud.ibm.com/computers). Finding such a layout is not difficult for a simple linear chain. But as problems become more complex, the optimization of the circuit layout becomes more difficult and more valuable.\n",
```
abbycross:
```suggestion
    "We see that the qubits were selected in a chain to minimize swap gates and thus circuit depth. Further, each CZ gate implemented in the circuit is between adjacent qubits. Finally, all five qubits have relatively low error rates, including readout-assignment error rates. You can check these error rates on any backend on the [Compute resources](https://quantum.cloud.ibm.com/computers) page. Finding such a layout is not difficult for a simple linear chain, but as problems become more complex, the optimization of the circuit layout becomes more difficult and more valuable.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "Now we configure our sampler primitive. We turn off other suppression/mitigation tools to focus on Pauli twirling."
```
abbycross:
```suggestion
    "Now we configure our Sampler primitive. We turn off other suppression/mitigation tools to focus on Pauli twirling."
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "# print(asdict(sampler_no_twirl.options))\n",
+    "# print(asdict(sampler_twirl.options))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8411f367",
+   "metadata": {},
+   "source": [
+    "Finally we run our jobs. You can optionally print the job ID's for later retrieval."
```
abbycross:
```suggestion
    "Finally, we run our jobs. You can optionally print the job ID numbers for later retrieval."
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ab237dc5",
+   "metadata": {},
+   "source": [
+    "The results using Pauli twirling are no better (and even slightly worse) than without twirling! What happened?\n",
```
abbycross:
```suggestion
    "The results using Pauli twirling are no better (and even slightly worse) than without twirling. What happened?\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "Two things happened. First, Pauli twirling does not reduce the total amount of noise — it reshapes coherent, systematic errors into stochastic Pauli‑type errors, so that error growth becomes predictable and modelable. There was never any promise of error reduction, except in special cases.\n",
```
abbycross:
```suggestion
    "Two things happened. First, Pauli twirling does not reduce the total amount of noise &mdash; rather, it reshapes coherent, systematic errors into stochastic Pauli‑type errors, so that error growth becomes predictable and modelable. There was never any promise of error reduction, except in special cases.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "Secondly, in GHZ circuits, some coherent errors can partially cancel or act like benign phase shifts because of the symmetry of the GHZ construction; twirling removes this accidental protection and replaces it with uncorrelated stochastic Pauli noise, so GHZ fidelity becomes slightly worse under twirling.\n",
+    "\n",
+    "This second claim requires some explanation. The claim is not that GHZ circuits are protected from __all__ kinds of coherent error accumulation, only some kinds and that in those cases the protection is destroyed by twirling. Specifically, let's consider coherent over-rotation associated with the CX gates. Let us call the real CX gate with over-rotation $\\tilde{CX}$:\n",
```
abbycross:
```suggestion
    "This second claim requires some explanation. The claim is not that GHZ circuits are protected from __all__ kinds of coherent error accumulation, only some kinds &mdash; and that in those cases the protection is destroyed by twirling. Specifically, let's consider coherent over-rotation associated with the CX gates. Let us call the real CX gate with over-rotation $\\tilde{CX}$:\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "$$\n",
+    "\\tilde{CX}\\equiv e^{-i\\epsilon K} \\cdot CX\n",
+    "$$\n",
+    "\n",
+    "where $K$ is any product of Pauli operators, like $X$, $Z$, $Z \\otimes X$, $Z^{\\otimes N}$, etc. For general states, any of these over-rotations could affect measurement statistics (and thus measures of state fidelity). A subset of these, however, leave many standard GHZ observables unchanged, including operators such as $Z_i Z_j$ and $Z^{\\otimes N}$. In the context of preparing a GHZ state the relevant over-rotation errors of this type would be\n",
```
abbycross:
```suggestion
    "where $K$ is any product of Pauli operators, like $X$, $Z$, $Z \\otimes X$, $Z^{\\otimes N}$, etc. For general states, any of these over-rotations could affect measurement statistics (and thus measures of state fidelity). A subset of these, however, leaves many standard GHZ observables unchanged, including operators such as $Z_i Z_j$ and $Z^{\\otimes N}$. In the context of preparing a GHZ state, the relevant over-rotation errors of this type would be:\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\\tilde{CX}\\equiv e^{-i\\epsilon Z_c Z_t} \\cdot CX\n",
+    "$$\n",
+    "\n",
+    "The preparation of the entire $N$-qubit GHZ state would look like:\n",
+    "\n",
+    "$$\n",
+    "|\\psi_\\text{GHZ}\\rangle=\\prod_{j=0}^{N-2}{e^{-i\\epsilon Z_j Z_{j+1}} \\cdot CX_{j,j+1}}H^{\\otimes N}|0^{\\otimes N}\\rangle\n",
+    "$$\n",
+    "\n",
+    "After preparing the GHZ state, the state is ideally\n",
```
abbycross:
```suggestion
    "After preparing the GHZ state, the state is ideally:\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "$$\n",
+    "|\\text{GHZ}^{\\otimes N}\\rangle=\\frac{1}{\\sqrt{2}}\\left(|0\\dots 0\\rangle + |1\\dots 1\\rangle\\right).\n",
+    "$$\n",
+    "\n",
+    "This state is a simultaneous eigenstate of a large set of Pauli operators, including\n",
```
abbycross:
```suggestion
    "This state is a simultaneous eigenstate of a large set of Pauli operators, including the following:\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "Pauli twirling fundamentally changes this situation. Twirling does not preserve the coherent over‑rotation error as a deterministic $Z_jZ_{j+1}$​ process. Instead, it converts the coherent error channel into an effective stochastic Pauli channel. As a result, the error channel now includes terms such as $X$, $Y$, $X \\otimes Z$, and $Y \\otimes X$, which do not commute with the GHZ stabilizers.\n",
+    "\n",
+    "When these stochastic Pauli errors occur, they create real bit‑flip and phase‑flip faults on individual qubits or pairs of qubits. These errors take the state out of the GHZ stabilizer subspace, reduce interference between $|0\\dots 0\\rangle$ and $|1\\dots 1\\rangle$, and directly lower GHZ fidelity and multi‑qubit parity signals. In other words, Pauli twirling removes the coherent structure of the error but also removes the symmetry‑based cancellation that previously made those errors benign. The result is a slightly worse GHZ state—not because twirling adds noise, but because it converts a mostly harmless coherent error into genuinely damaging stochastic errors.\n",
```
abbycross:
```suggestion
    "When these stochastic Pauli errors occur, they create real bit‑flip and phase‑flip faults on individual qubits or pairs of qubits. These errors take the state out of the GHZ stabilizer subspace, reduce interference between $|0\\dots 0\\rangle$ and $|1\\dots 1\\rangle$, and directly lower GHZ fidelity and multi‑qubit parity signals. In other words, Pauli twirling removes the coherent structure of the error but also removes the symmetry‑based cancellation that previously made those errors benign. The result is a slightly worse GHZ state &mdash; not because twirling adds noise, but because it converts a mostly harmless coherent error into genuinely damaging stochastic errors.\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "\n",
+    "This example highlights an important lesson: Pauli twirling is not a universal improvement strategy. It is most beneficial when coherent errors accumulate across a circuit in a manner destructive to the required fidelity. In highly symmetric circuits like GHZ state preparation, some coherent errors are naturally aligned with the state’s stabilizers, and deliberately randomizing them can eliminate this accidental protection.\n",
+    "\n",
+    "#### Combining methods\n",
```
abbycross:
```suggestion
    "#### Combine methods\n",
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "qcs[0].draw(\"mpl\")\n",
+    "# circuits[0].draw(\"mpl\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8acf4927",
+   "metadata": {},
+   "source": [
+    "At the high end, some of these are quite deep. Let's use the AerSimulator with no noise model to obtain the ideal states at the end of each of these circuits. That way we can benchmark our results from real quantum computers using Hellinger fidelity."
```
abbycross:
```suggestion
    "At the high end, some of these are quite deep. Let's use the AerSimulator with no noise model to obtain the ideal states at the end of each of these circuits. We can then can benchmark our results from real quantum computers using the Hellinger fidelity."
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
abbycross:
```suggestion
    "Because we have rotated our basis before measurements, a measurement of $|0\\rangle$ corresponds to the state having been in $|+\\rangle$ just before the final Hadamard gate, and similar for $|1\\rangle$ and $|-\\rangle$. Therefore, our expectation value $\\langle X \\rangle$ can be simply calculated from the counts of $|0\\rangle$ minus the counts of $|1\\rangle$."
```

learning/courses/tools-for-quantum-advantage/error-suppression.ipynb
```diff
@@ -0,0 +1,1893 @@
...
+    "|\\psi_c\\rangle = a|11\\rangle + b|00\\rangle + c|01\\rangle + d|10\\rangle\\\\\n",
+    "|\\psi_d\\rangle = a|00\\rangle + b|11\\rangle + c|10\\rangle + d|01\\rangle\n",
+    "$$\n",
+    "\n",
+    "This is exactly $|\\psi_\\text{final}\\rangle$ that we obtained before with no twirling. Indeed, this twirled gate sequence leaves the logical action of the payload unchanged. However, if different valid twirling sequences are selected randomly from layer to layer, coherent error accumulation can be converted into effectively stochastic (incoherent) error accumulation. To be clear, one does not choose a single twirling pattern and use it throughout the circuit. Instead, different valid twirling sequences are selected for different layers. An example with many entangling layers might look like this.\n",
+    "\n",
+    "![A quantum circuit with four qubits and three CNOT gates in a ladder arrangement. In twirling, each of these CNOT gates is surrounded by a different set of Pauli gates.](/learning/images/courses/tools-for-quantum-advantage/error-suppression/twirl-mult.avif)\n",
+    "\n",
+    "\n",
+    "#### Check your understanding:\n",
```
abbycross:
```suggestion
    "#### Check your understanding\n",
```

abbycross (approved):
Approving so that it can publish tomorrow. @christopherporter1 when you are back, let's check that one spot I flagged with duplicate content.

Issue #26, opened 2023-09-07
Add an Overview page to the Run section


javabster:
The Run section should have an overview page explaining what the "run" phase is, similar to what we already have for the [build section](https://docs.quantum-computing.ibm.com/build)

PR #5680, opened 2026-09-21
add nuclear shell tutorial

aaryav3:
a tutorial to solving nuclear shell problems using a pooled SQD method (with tunable resource constraints) 
contains exactly [this](https://github.com/Qiskit/qiskit-fortran/pull/26) tutorial, with an added enhancement from qiskit-addon-sqd, which adds an additional configuration recovery step in post-processing


henryzou50 (changes requested):
Thanks, Aaryav, for adapting this into a Python notebook and adding configuration recovery! This looks great! I have a few changes I’d like to see before we merge:

1. Fix the equal-dimension classical comparison. Running classical_selection with the saved target of 3,990 returns only 1,258 determinants, because the early stopping condition counts products before symmetry filtering. The reported 1,252 keV improvement therefore compares different subspace sizes. Please compare at a common actual dimension, verify those dimensions match, and regenerate the outputs and plots.

2. Make the comparison plot reflect either outcome. The annotation currently uses abs(advantage) and hardcodes “lower,” so it would still claim the sampled result wins when the classical result is better. Please make the annotation and title reflect the signed difference.

3. Clearly link to the Fortran version near the top. Please add a short paragraph immediately below the title explaining that this is the Python implementation, linking to the Fortran version, and noting the added configuration-recovery step. Something like:
> This tutorial presents a Python implementation of the nuclear shell-model workflow. A [Fortran version](https://github.com/Qiskit/qiskit-fortran/blob/nuclear-dynamics/applications/nuclear_shell/tutorial.ipynb) is also available in the `qiskit-fortran` repository. This Python version extends the workflow with self-consistent configuration recovery using `qiskit-addon-sqd`.

We should also merge [qiskit-fortran PR #26](https://github.com/Qiskit/qiskit-fortran/pull/26) so we can link to its merged location.
   @nathanearnestnoble  and @kaelynj: should that link point to the tutorial in qiskit-fortran, where it currently lives, or should we also host the Fortran tutorial in the docs repo? Keeping it in qiskit-fortran gives us one source of truth. Hosting it in docs could make maintenance and testing more consistent, especially if we add other Fortran tutorials, but keeping both copies would create two sources of truth unless we establish a synchronization process or designate one canonical location.

4. Revise the rationale for skipping simulation. Noisy simulation can exercise post-selection and configuration recovery, so the claim that these require real hardware noise is too strong. Please describe hardware execution as a choice for this tutorial and rename “Small-scale simulator example” to reflect that it runs on hardware.

5. Align the stated depth budget with the code. The prose says 600, while DEPTH_BUDGET is 300.

henryzou50 (commented):
Thanks, Aaryav for the updates! The updated equal-dimension comparison, signed plot annotation, and hardware-example wording address the main points from my earlier review. A few items remain before publication:

1. Include the Fortran companion.
   As discussed, please add it under `docs/tutorials/assets/nuclear_sqd_pooled/fortran/`, including the source, required data, and build/run instructions. The notebook’s current link to `qiskit-fortran/main/applications/nuclear_shell` returns 404. Update the prominent note near the top to link to the new docs directory, retaining the explanation that Python adds configuration recovery.

2. Remove the remaining simulator claims.
   “How to tell a good run from a bad one” and the summary still say noise-assisted recovery is invisible on a simulator. Noisy simulation can demonstrate this, so please make those sections consistent with the corrected introduction. For example:
   Noise can introduce additional configurations that improve the selected subspace. This can also be studied with noisy simulation; this tutorial demonstrates the effect using hardware samples.

3. Qualify the classical-computing comparison.
   Please change “past the reach of exact diagonalization” and similar statements to “beyond this notebook’s dense exact solver.” The limitation is this implementation, rather than classical shell-model methods generally. Also describe the comparison as the PT2-based classical baseline used here, rather than the “best” classical subspace. Matching dimensions makes the comparison meaningful, but does not establish that the classical baseline is optimal.

4. Clarify the solver recommendation in “Next steps.”
   `qiskit_addon_sqd.fermion.solve_sci` wraps an electronic-structure solver. The shared proton/neutron product structure alone does not make it a drop-in replacement for this nuclear Hamiltonian. Please recommend a sparse or iterative solver supporting the nuclear interaction, or explain and validate the mapping needed to use `solve_sci`.

henryzou50 (approved):
Thanks, Aaryav! The Python tutorial looks good now. I made a few minor wording changes to clarify the summary, solver-scaling guidance, and excitation-energy explanation, and resolved the merge conflict with main.

Next, I’ll create a follow-up PR from a branch in the docs repo so CI can run with the required access. I’ll preserve your authorship, make sure you’re credited in the first commit, and identify you as the tutorial’s developer in the PR description.

I’ll also add the Fortran companion from your merged `qiskit-fortran` work under `docs/tutorials/assets/nuclear_sqd_pooled/fortran/`, adapt the build/run instructions, and update the notebook’s link. That should put us in a good position for final checks. Thanks again for all your work!

henryzou50:
Closing this as the work is continuing in #5694

